PR #6, opened 2015-09-24
Randhound decentralized randomness cothority prototype

bford:
This contains the bare-bones beginnings of an implementation of a scalable decentralized randomness-scavenging protocol I'm tentatively calling RandHound.  It's not yet complete or ready to merge, but it does basically work in single-node fashion and can be tested with go test.

It includes and depends on an extremely simple abstract networking-interface layer currently in rand/net.go, which represents my current (immature, still-evolving) thoughts on what the common abstract network interface we want to use might look like.  I already think it can/should be reworked in various ways, but I think it at least sort of illustrates the minimalistic ideal I'm shooting for here.


Daeinar:
I started an attempt to get the framework up and running again but was only half successful. Currently, I am running in a `protobuf` error and can't quite tell what's causing it. To be more precise: on executing `go test` you get 
```
panic: no constructor for interface abstract.Secret [recovered]
```
caused through https://github.com/DeDiS/protobuf/blob/master/decode.go#L304

The relevant function call in RandHound is: https://github.com/DeDiS/cothority/blob/randhound/app/rand/cli.go#L215

How can you decode `abstract.Secret` with `protobuf`? Is this related to your recent Suite marshalling discussion https://github.com/DeDiS/crypto/pull/56?

ineiti:
Le 13/12/2015 18:04, Philipp Jovanovic a écrit :
>
> I started an attempt to get the framework up and running again but was 
> only half successful. Currently, I am running in a |protobuf| error 
> and can't quite tell what's causing it. To be more precise: on 
> executing |go test| you get
>
> |panic: no constructor for interface abstract.Secret [recovered] |
>
> caused through 
> https://github.com/DeDiS/protobuf/blob/master/decode.go#L304
>
> The relevant function call in RandHound is: 
> https://github.com/DeDiS/cothority/blob/randhound/app/rand/cli.go#L215
>
> How can you decode |abstract.Secret| with |protobuf|? Is this related 
> to your recent Suite marshalling discussion DeDiS/crypto#56 
> <https://github.com/DeDiS/crypto/pull/56>?
>
> —
> <https://github.com/DeDiS/cothority/pull/6#issuecomment-164277567>.
>
https://github.com/DeDiS/cothority/blob/development/lib/sign/messagessign.go#L129

No comment on the rest of the issue ;) or just a small one:

The main issue is that `point` and `secret` are defined as an interface, 
and no decoder can know what implementation of that interface is to be 
used. For this reason I proposed to change `point` and `secret` to a 
structure, so that it can be automatically decoded. But for 
security-reasons it's not worth it.

Even though every time I see somebody stumble on that same problem, I 
feel like we should do something about it. Perhaps already write a 
better panic-message in the protobuf-decoder?

happy hacking,

Linus

Daeinar:
Thanks Linus for your explanations, so it is indeed related. Yeah we should definitely do something about it. I'll talk to the rest to discuss the best approach.

ineiti:
We have philipp's implementation now.

PR #7, opened 2015-09-30
Monitor

nikkolasg:
Tried to modularize **when possible** the differents statistics generated by the differents app and their collections by the monitor.go.
We have now different entries each time a server or a client sends something to the log servers.
Theses entries are added to a Stats interface which can be either a shamir stats or a collective stats or whatever.
The stats interface write itself to the report file.


deploy/stats.go
```diff
@@ -58,66 +291,154 @@ type RunStats struct {
...
+func AverageStats(stats ...Stats) (Stats, error) {
+	if len(stats) == 0 {
+		return nil, errors.New("No stats given to average")
+	}
+	switch stats[0].(type) {
+	case *CollStats:
+		return collStatsAverage(stats...)
+	case *ShamirStats:
+		return shamirStatsAverage(stats...)
+	}
```
ineiti: Can't you put that in the interface, like

```
type Stats interface{
...
  CollectStatsAverage([]Stats)(Stats, error)
}
...
func AverageStats(stats []Stats){
...
return stats[0].CollectStats(stats)
}
```

and then define ```CollectStats``` in each implementation of ```Stats```? That way, every new stats that is defined doesn't need his entry in this switch/case part. But then again, I don't know how many of these will exist anyway.

deploy/stats.go
```diff
@@ -58,66 +291,154 @@ type RunStats struct {
...
+			a := b.(*CollStats)
+			st = append(st, a.round)
+			s.SysTime += a.SysTime
+			s.UserTime += a.UserTime
+			s.Rate += a.Rate
+			s.Times = append(s.Times, a.Times...)
+		default:
+			return s, errors.New("Average did not receive a CollStats struct")
+		}
+	}
```
ineiti: Personally I prefer a ```if/then/else``` over a ```switch``` with only one branch.

deploy/stats.go
```diff
@@ -13,7 +15,112 @@ import (
...
+// Update will update the time struct with the min / max  change
+// + compute new avg + new dev
+// k is the number of times we've added something ("index" of the update)
+// needed to compute the avg + dev
+// streaming dev algo taken from http://www.johndcook.com/blog/standard_deviation/
+func (t *StreamStats) Update(newTime float64) {
+	t.n += 1
+	// nothings takes 0 ms to complete, so we know it's the first time
+	if t.min > newTime || t.min == 0.0 {
+		t.min = newTime
```
ineiti: You can also take away the ```t.min == 0.0``` and put it into the ```if t.n == 1``` condition below.

Or put the ```if t.n == 1``` branch at the start, initialise everything as needed and then do all other calculations unconditionally.

deploy/start.go
```diff
@@ -198,9 +194,33 @@ func RunTests(name string, ts []T) {
...
 
-		rs[i] = RunStatsAvg(runs)
+		s, err := AverageStats(runs...)
+		if err != nil {
+			dbg.Fatal("Could not average stats for test ", i)
+		}
+		rs[i] = s
+		rs[i].WriteTo(f)
+		// Write the header if you still havent done it
+		if !headerWritten {
```
ineiti: Writing the header just seems odd in the loop - can't you take it out and decide upon the ```t T``` which header you have to write? Then the file-open and defer come out of the loop.

deploy/stats.go
```diff
@@ -25,31 +132,157 @@ type StatsEntry struct {
...
+}
+
+// basic check to see if we got somme real data
+func (s *ShamirStats) Valid() bool {
+	return s.round.Avg() > 0.0 && s.setup.Avg() > 0.0
 }
 
-type RunStats struct {
+// Average all these stats
+func shamirStatsAverage(stats ...Stats) (Stats, error) {
```
bford: I bet it wouldn't be hard to create a completely generic version of this function, which takes a varargs-list of interface{}, which it expects to be a slice of structs of float values, and creates/returns a tuple of structs of the same type containing the averages, mins, maxes, std devs, etc, of each field across all the input structs.  Then that could be reused for whatever application- or context-specific stats we want to deal with in any situation.  For example, the signature might be:

func SummarizeStats(stats ...interface{}) (min, max, avg, stddev interface{}) { ...}

Not sure it's necessarily worth the trouble, but something to consider anyway. :)

deploy/stats.go
nikkolasg: Will do.

deploy/stats.go
nikkolasg:
```
func SummarizeStats(stats ...interface{}) (min, max, avg, stddev interface{}) { ...}
```
has the limitation of only computing the min/max/avg/dev of ONE value. Our stats interfaces holds multiple values right now. More over, there is some fields in our stats where we don't need the  Dev / min / max, but rather just the average. There is definitly some improvements to do on the average part, I will first start by integrating the average in the stats interface.

deploy/start.go
nikkolasg: Yes. I can take the first test in the array of tests, then get a stat to write the header.
We'll need to refactor these "tests arrays" anyway later as they horribly lack of usability and checkability features (for example, check that all tests are using the same "app")

ineiti: you saw that I did some work on my ‘restructure_main’-branch - I think it would be good to go on from there.

deploy/stats.go
bford: Not if it's implemented the way I have in mind.  The thing you pass in would be a struct containing multiple (arbitrarily many) float64 fields, and it would compute the min/max/avg/stddev across all them in parallel.  For example:

type Blah struct {
   foo float64
   bar float64
   baz float64
}

var inputs []Blah
...
var min, max, avg, stddev Blah
min, max, avg, stddev = SummarizeStats(inputs...)

This will leave min.foo containing the minimum of the foo fields in all the inputs structs, avg.bar containing the average of the bar fields in all the inputs structs, etc.  Clear?

deploy/stats.go
nikkolasg: It's kinda like the "official" use of the ".(type)" statement (https://golang.org/doc/effective_go.html#conversions). 
But if we want the ``ìf / else```, we could do it with a ```if a,ok := b.(*CollStats) ```, is that Ok ?

deploy/stats.go
nikkolasg: I agree that an generic method like this would have its utility but I don't think it applies so well here. Like I said, our stats contains many fields and some of them are just here for information (like the number of hosts) and we do not need to avg'd / ... them and some of them do not need to be min'd / max'd / dev'd at all, only avg (ex. SysTime).
Of course we could compute the the stats for each fields and then take only those of interests but it doesn't seem the right approach for solving this problem, at least for me !!

deploy/stats.go
```diff
@@ -25,31 +132,156 @@ type StatsEntry struct {
...
+
+	// incoporate the Entry into theses stats
+	// i is the "Index" of this entry (i.e. # times we have added entries)
+	AddEntry(e Entry) error
+	// Valid tells is the stats received some real data and is not
+	// empty or full of garbage
+	Valid() bool
+
+	// Average will let you average a bunch of stats from a already existing
+	Average(stats ...Stats) (Stats, error)
```
nikkolasg: Done the change.

ineiti:
Merge the development-branch into this one and then test if it still works for all of our apps. Once this is confirmed, I think it's a good idea to go on and merge it into the development-branch.

ineiti:
Closing and re-open against development

PR #8, opened 2015-09-30
Seperation of "app"s and "deploy"

ineiti:
The apps and the deploy-part were heavily intermingled. This is an effort to better separate the two, so that:

  * each app is independent of the others and of the deterlab-simulation
  * deterlab-simulation knows as little as possible about each app
  * testrun-configurations are separate files and don't reside in ```deploy2deter``` anymore

Instead of a ```main``` in the root-directory and one in the ```app```-directory, each app has it's own main. The ```main``` from the root-directory has been moved to ```deploy```.

The testrun-configurations are as follows:

```
Machines = 8
App = "shamir"

hpn, bf, rounds
1, 2, 10
2, 2, 10
```

Where the first lines define configuration-options for the deterlab-simulation and the second part is a list of configuration-fields that are read in by each ```app```.

Different optimisations are included in this pull request:

   * better error-reporting when something goes wrong in the deterlab-setup
      * when a server is down, it proposes a command-line
      * setting higher open-file-limits on all servers
   * ```debug_lvl``` additions
      * ```Lvlf*``` which behaves in the same way as ```ftm.Printf```
      * warning if a big string of ```Lvl*``` is not printed (see performance leaks)
   * ```app```-library for easy reading of the configuration
   * renamed ```deter``` module to ```users```, as it is running on the users-server of deterlab


ineiti:
The directory ```test_data``` is not created automatically when the tests are run.

lib/app/app.go
```diff
@@ -0,0 +1,184 @@
...
+	AmRoot      bool   // is the host root (i.e. special operations)
+	TestConnect bool   // Dylan-code to only test the connection and exit afterwards
+	Mode        string // ["server", "client"]
+	Name        string // Comes from deter.go:187 - "Name of the node"
+	Server      string // Timestamping servers to contact
+}
+
+// Initialize before 'init' so we can directly use the fields as parameters
+// to 'Flag'
+var Flags FlagConfig
```
nikkolasg: Since FlagConfig is holding all the flags we are passing to the executable, could we simply rename it "Flags" ?
There is too much "config" everywhere, we should restrict that as much as possible

lib/app/app.go
```diff
@@ -0,0 +1,184 @@
...
+				log.Fatal(Flags.Hostname, "improperly formatted hostname: should be host:port")
+			}
+			p, _ := strconv.Atoi(port)
+			// uncomment if more fine grained memory debuggin is needed
+			//runtime.MemProfileRate = 1
+			res := http.ListenAndServe(net.JoinHostPort(Flags.PhysAddr, strconv.Itoa(p + 2)), nil)
+			dbg.Lvl3("Memory-stats server:", res)
+		}()
+	}
+}
```
nikkolasg: I know there is a lot of restructuration left to do, so this is just a side node that these two methods don't belong here, but rather in deploy/ or maybe later, in a possibly "logging" package.

nikkolasg:
A few global comments:
    - We should run this new code for all apps (including shamir) to see if all works well
    - **Organization part**:
        - I think we should have a "/config" and a "/lib" folders separate, with the "/config" containing the config for each differents app instead of a "/lib/app/*_config.go". Also, between /lib/app/*_config & deply/runconfig, the difference is unclear. Maybe we could rename "deploy/runconfig" => simulation, as it contains the data to run multiple simulations. As I understood, you plan to remove the "lib/config" existing folder, we could then move the "/lib/app/*" to "/config", and going even further, we could put the "simulations" files inside "/config/simulations".
        - "/lib/deploy" => why not move it inside "/deploy" ? As it is only used by "deploy", and from what it looks, is more initialization / reading config than helpers or special library with specific purpose. I is tightly coupled with the "deploy" package and therefore, I think it belongs best inside "/deploy".
    - Good job with the spearation of each app with their respective main and the simple and magnificient "app.ReadConfig(conf)"   

As there is a full load of changes, I only noted what looked wrong from the surface. There may be others things to note, but we will maybe open a new branch each time we find something that needs to be changed again (and we surely will).At least, we are now much more closer to separating app & deploy than we were before ;)

ineiti:
* moved ```lib/deploy``` to ```deploy``` and ```deploy/interface```
   * Renamed ```runconfig``` to ```simulations```
   * Made sure all apps work (sign, stamp and shamir)

lib/app/app.go
ineiti: Renamed ```Flags``` to ```RunFlags``` and ```FlagConfig``` to ```Flags```

lib/app/app.go
ineiti: ```ConnectLogservers``` is only used for the graph and the monitoring-function. Once your ```monitor``` is merged, we can talk about how to take that out.

```ServeMemoryStats``` could probably be taken away - I'd like to keep it for the moment, just in case we need it for further statistics.

ineiti:
Closed #13

ineiti:
Any other comments? I would like to merge this, together with the 'Montior' pull-request of Nicolas. Personally I work based on this branch anyway.

nikkolasg:
Since this is a restructuration branch, we may merge it as it is. We are surely going to do a lot of other refactoring that may depend on this branch (and this branch is already used by many others) so I think we better merge it now and look for other optimizations to do in the next refactoring step.

ineiti:
Closing and making pull-request to development-branch.

PR #12, opened 2015-09-30
Plot matplotlib

ineiti:
A first throw at a python-program. Yay. I put it on top of my branch 'restructure_mains'.

   * do we really want that in the main branch, as it is python? Where else to put it?
   * what other options do we need to print?


ineiti:
sign_multi

ineiti:
sign_multi csv time

ineiti:
I'm still trying to get a reproducible run on the Deterlab-servers. Of course it's very dependant on:

   * branching-factor
   * type of machines reserved in deterlab
   * number of machines reserved (512 processes on 8 machines are slower than 128 processes on 32 machines)

So we'll have to define the outlines of the experiment. The faster machines that can hold more processes are only in limited number.

Perhaps we'll have to add a "Amazon EC2"-simulation support.

ineiti:
OK, here's a new plot. ATTENTION: Cothority-values for 4 and 8 are just copies of '16'

Another point is that the first round often takes very long with Cothority, thus the high 'max' value and low variance.

sign_huge csv round

ineiti:
Next graph. Most important changes:

- from 2 to 16 hosts each process was on his own host, both for JVSS and Cothority
- discarded time for first round, as it was always much higher

sign_huge_start_2 csv round


With the new graph, it looks like we could even dump the 'minmax'-part and only show the times:

sign_huge_start_2 csv round

ineiti:
And the corresponding System- and User-time for Cothority:

sign_huge_start_2 csv time

System- and User-time for JVSS:

shamir_multi_adapted_hosts csv time

This looks good for cothority, as we seem to be mostly bound by network, as the user-time doesn't really increase till 8192 - probably swapping and such.

JVSS is very computation-intensive, as the rising time of the user-part indicates.

If you prefer to have Cothority also loglog plotted:

sign_huge_start_2 csv time

ineiti:
Going to re-open against ```development```-branch

PR #11, opened 2015-10-02
Network-layer

nikkolasg:
Evaluate network layer solutions. First evaluate already present solutions such as coconet, the randhound and the schnorr_sign.
Then try to decide on a specific implementation of the network layer (either one of the solutions or a mix of the above or a totally new one)

By the way, a nice slide on how to build scalable distributed resilient systems : http://www.slideshare.net/rolandkuhn/reactive-streams?next_slideshow=1

We should definitly absorb some of theses ideas into our implementation.


nikkolasg:
Done. Great tool "hub", which enables a lot of transformation on your github, notably changing a issue into a pull request ;)

ineiti:
OK, this looks like a good start. We do need to have some of these:

   * tests - we're way behind when it comes to tests and we need to use tests at least for new code. If I understand correctly, the ```gob.go``` is some kind of test. Make it a real test
   * documentation - if I run ```godoc -server=:8080```, there is nearly no documentation at all. For the very least we should have some introduction. Also if you do ```_example```-files, they will get included here

PR #5, opened 2015-10-02
Localhost simulator

nikkolasg:
I'd like to have a new deployment-model for a 'local' deployment. Would it be a good idea to separate the code, so that

deploy/deterlab
and
deplo/localhost

are two directories holding each time it's own deployment-method?


nikkolasg:
Ok it's working. Please note that we've deliberately left out the monitoring / logging / stats part for now  as this branch will mostly be used to test quickly some new features and not to test any scalability.  So the exec does not terminate since it is waiting for stats that never come. 
( the logging part is a big refactoring step that will have to be done later ;)

ineiti:
I'm closing that one as I merged it into my 'restructure_mains'-tree.

PR #22, opened 2015-10-07
Plot matplotlib

ineiti:
We just need to figure out where to put that ```matplotlib```-directory


bford:
Where it currently is seems reasonable to me.  It's testing infrastructure for use with deployments on testbeds, which I believe is pretty much what the 'deploy' directory is about right?

nikkolasg:
We may later need to turn this python script into something a bit more general so people could make their own graphs independently of what application they have launched but for the moment it's good enough for me. 
On the other hand, I would find it more convenient to put it either:
 * into our /lib folder, since this is a plotting script and not a deployment one.
 * or into a "futur" logger / logging folder, where we might put the to-be-discussed logging infrastructure (which is currently the "monitor.go"  + "users.go" file for now, which is not something usable for the long term).This matplotlib script is used from results of the logging server so it may make sense to put it close to that !

But anyway, that's not something urgent / important **right now**, so for me it's fine ;)

ineiti:
I should at least put it into its own class (classes, yay!) and put different files for the different graphs.

PR #27, opened 2015-10-07
also copy global-config of simulation to app-configuration

ineiti:
This allows to use both configuration-places in the ```simulation```-files to be used for a configuration-option. For example:

```
App = naive
SkipChecks = true

Machines, rounds
1, 30
2, 30
```

Works as well as

```
App = naive

Machines, rounds, SkipChecks
1, 30, true
2, 30, true
```


bford:
Looks good!

nikkolasg:
Ok for me !

Issue #25, opened 2015-10-07
Localhost-wait


ineiti:
It doesn't wait for the end of the experiment before quitting to the command line...

Liamsi:
I wanted to try the localhost simulation and stumbled upon this issue, too. The monitor only starts (and returns a valid ResultSet) in  case the app is run in "deterlab mode", but RunTest() returns before the end of the experiments in case it got empty/invalid ResultSet.  
I'll open a PR. Feel free to merge or ignore it, if you have other or better plans.

ineiti:
The long-term fix is to enable monitoring localhost, too. But that is further down the path...  So I'll just merge this one. Thanks!

bford:
Thanks for jumping right in, Ismail, and welcome to the team!

Liamsi:
Thanks for the warm welcome, Bryan. Can't wait for December :-)
And thanks for merging, Linus!

PR #28, opened 2015-10-07
quick fix for issue #25

Liamsi:
see issue #25


ineiti:
For longer fixes you'll have to do your own branch off from ```development```. But I think for one-liners this is OK.

Issue #31, opened 2015-10-08
Localhost stop before test is finished


nikkolasg:
In localhost simulation, the Monitor() directly returns if the platform is not deterlab. When monitor returns, the "test" is finished, and a call to platform.Stop() is called. Depending on the situation, it can happens (it has), that apps do not have enough time to finish their test before being killed by the localhost platform.

We should think about that, and this issue may be part of a logging facility restructuration since this is clearly a logging issue...

PR #33, opened 2015-10-09
Fix localhost monitor return

ineiti:
OK, this is the correct fix for #31. I added a new method to the platform.Platform interface, ```Wait``` which will wait for any started processes to quit. ```Start``` is non-blocking, so we can start an eventual monitor.

There were some issues with ```proto/sign``` with regard to correctly quit the signing. Now there is a ```CloseAll``` message in the handling of the messages which starts from the root-node and propagates to all children nodes.


ineiti:
Closes #31

deploy/platform/localhost.go
```diff
@@ -190,6 +207,9 @@ func (d *Localhost) Stop() error {
 	if err != nil {
 		dbg.Lvl3("Error stopping localhost", err)
 	}
+
+	// Wait for eventual connections to clean up
+	time.Sleep(time.Second)
 	return nil
 }
```
nikkolasg: There is no need for both Wait() + Stop() since basically they do the same thing, one is blocking whereas the other is not.
 We could somehow use a Wait() for regular "stops" of the platform and a Clean / Cleanup () for killing everything the hardway (the way Stop() is done right now), when a process can not finish after a certain timeout.
Another possiblity is to transform Wait() to take a timeout. If the timeout is passed, then Wait() kills the processes and return back to the caller. That way, both functionalities are present and there is no one more method to the interface.

ineiti:
Looks good to me - comments?

nikkolasg:
Looks fine for me. There are still some issues regarding the "if platform == "deterlab" " but that kind of thing will have to be resolved in the future and is not the main concern here.
it's been two days. Shall we merge ?

bford:
Sounds good.  (And for platform/testbed-related stuff don't feel like you need to get my sign-off before merging; just the two of you [Linus & Nicolas])

PR #35, opened 2015-10-13
Conode

ineiti:
First shot at a README for the Conode-project - COthority NODE...


ineiti:
Please use the appropriate flag-params instead of doing your own switch/help thing.

ineiti:
Some things to do:
- verify that sn.Rounds are deleted after some time
- clean up StoreLocalMerkleProof

ineiti:
cliutils/key.go - decide where does the output and where does the suite come. Sometimes suite is first, sometimes the output-variable is first.

app/conode/stamp/stamp.go
```diff
@@ -0,0 +1,394 @@
...
+	Hash string
+	// the root of the merkle tree
+	Root string
+	// the inclusion-proof from root to the hash'd file
+	Proof []string
+	// The signature challenge
+	Challenge string
+	// The signature response
+	Response string
+}
```
nikkolasg: Instead of having everything as string, and our method "readSignatureFile" returns many arguments, we could just make this struct already with the rights fields, i.e. 
```Hash []byte
Root []byte (or hashid.HashId)
Proof []Proof
Challenge abstract.Secret
Response abstract.Secret
````
And then having a method on them 
```func (s *SignatureFile) ReadSigFile(fileName string)```which sets the rights fields. 
Not only we increase the readability (no more 4 x returns value), but it could also later take an second argument as a the format of the signature. Eventually we'll need maybe to comply to some standard regarding a signature file format.

Comment (ineiti) [inline app/conode/stamp/stamp.go]: > And then having a method on them 
> ```func (s *SignatureFile) ReadSigFile(fileName string)```which sets the rights fields. 
> Not only we increase the readability (no more 4 x returns value), but it could also later take an second argument as a the format of the signature. Eventually we'll need maybe to comply to some standard regarding a signature file format.
I prefer having ```NewSignatureFile(fileName string)``` and returning the signature. Else you have to create the ```SignatureFile``` first, then call ```ReadSigFile``` on it.

File: app/conode/stamp/stamp.go [hunk 2]
```diff
@@ -0,0 +1,448 @@
...
+	// the inclusion-proof from root to the hash'd file
+	Proof     []string
+	// The signature challenge
+	Challenge string
+	// The signature response
+	Response  string
+	// The aggregated commitment used for signing
+	Commitment string
+}
+
```
nikkolasg: Right, we can use both ideas:
* Have a 
```NewSignature(filename string)```
* Which basically do 
```new(Signature).ReadFile(file string)```

Is that ok for you?

ineiti: sounds good - code?

PR #38, opened 2015-10-22
added jsonmarshal to signingmessage and stampreply/request

ineiti:
Replaced gob with json encoder just above network layer


ineiti:
In order for external processes to be able to connect to the conode, we need JSON instead of gob (which is a go-only-implementation). This changes the standard-decoders in coconet/tcpnet.

Also changed the names for Get/Put to GetData/PutData, as other methods used the same name and searching for them got cumbersome.

app/conode/stamp/stamp.go
```diff
@@ -188,7 +189,9 @@ func StampFile(file, server string) {
 		ReqNo: 0,
 		Sreq:  &defs.StampRequest{Val: myHash}}
 
-	err = conn.Put(msg)
+	out, _ := json.Marshal(msg)
+	dbg.Print(string(out))
```
nikkolasg: debugging part still left!

proto/sign/signingMessages.go
```diff
@@ -109,58 +110,77 @@ func (sm *SigningMessage) UnmarshalBinary(data []byte) error {
 	return protobuf.DecodeWithConstructors(data, sm, cons)
 }
 
+type JSONdata struct{
+		Len int64
+		Data []byte
+}
+
```
nikkolasg: We should find the approriate place to put that struct instead of declaring it two times.
What about coconet/signingMessages ? for the moment at least.

ineiti: It should be deleted in app/conode/defs/messg as it is not used anymore. Perhaps it could be removed here, too...

PR #41, opened 2015-10-28
Make conode lib

ineiti:
Moved verification and structures to it's own lib so we can use it from the certificate-transparency repository.


lib/conode/consts.go
```diff
@@ -1,3 +1,3 @@
-package defs
+package conode
 
 var DefaultPort string = "2000"
```
nikkolasg: We should think about removing that i think. It should be embedded in the application using this lib.

lib/conode/verification_test.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func TestVerifyChallenge(t *testing.T) {
+	readTestSig(test_sig_1)
+
+	err := conode.VerifyChallenge(suite, &reply)
+	if err == nil {
+		dbg.Lvl2("Verification passed")
+	} else {
+		dbg.Fatal("Verification failed")
```
nikkolasg: we should use t.Error instead. (and the rest of the tests methods below)

nikkolasg:
I want to code what was discussed in the conode pull request about the Signature struct + splatting out the fields in StampReply (no more sigbroad embedded). Shall we do it there or in the conode pull request ?
I think this pull request should be merge to conode and not development directly since conode is not even merged !

ineiti:
Le 28/10/2015 17:13, nikkolasg a écrit :
>
> I want to code what was discussed in the conode pull request about the 
> Signature struct + splatting out the fields in StampReply (no more 
> sigbroad embedded). Shall we do it there or in the conode pull request ?
>
Can we postpone this to after-IETF? I don't want to handle the fall-out 
with the other conodes. If they can run peacefully for the next 10 days, 
I'd be more than happy.
>
> I think this pull request should be merge to conode and not 
> development directly since conode is not even merged !
>
Of course - I think that's what the pull-request is about:

ineiti  wants to merge 2 commits into conode from make_conode_lib

PR #45, opened 2015-10-30
Exit upon validation

ineiti:
Added a sub-command to ```conode check``` that allows the remote conode that is in validation-mode to exit, update the config.toml and restart.


nikkolasg:
OK

PR #46, opened 2015-11-01
Add conode server to library

ineiti:
Instead of having the server in sign, stamp and conode; the stampserver in stamp and conode, put this into the conode-library. This is also usable for Eleftherios!


nikkolasg:
Should we start thinking ahead and try to set up the base code for having a validation mode here ?
It is explained in the Cothority paper but in essence, we need that each signers contributions (hash of a file, certificate request or transactions) is being validated by each nodes in the tree. When receiving another peer's contributions, a peer validate it or not by sending up a ValidateTransaction message of some kind. 
This is not clear in the paper when that validation thing is supposed to happen.
Actually Eleftherios has already this need of a validation mode as he wants a block to be verified by each signers of the tree.

ineiti:
Yes, good idea. What is needed to finish the move of the server and stamper to the library? Then we could add a return-value to the ‘CommitFunc’ that indicates whether the message should be included or not.

nikkolasg:
Actually I do not know if it can be done this way. From the way I understand it, the validation takes place for EACH signer for EACH contributions. When the root makes the announcement, he does not have all the contributions of all the nodes in the tree. So when a node will respond with a commit, he can only validate the announcement message which can be whatever the root want. For now the announcement message is the current timestamp.
It is not clear in the paper WHEN this validation takes place, maybe you could ask bryan. Either we only validate the MESSAGE of the announcement (so only the root can announce something and that something ONLY will be validated). Or we want to validate EVERYTHING for a specific round, where each contributions (timestamp request) will be broadcasted and validated in the tree by each signers.

nikkolasg:
Apart this problem, it is ok for me to MERGE this branch.

ineiti:
OK, as I opened the pull request, I let you merge it - is that OK as workflow? The opener of the pull-request can't merge it (emergencies aside), it always needs to pass by somebody else.

PR #53, opened 2015-11-04
Fix app stamp error

ineiti:
Cleaning up renaming of DoneFunc to OnDoneFunc and making app/stamp work without error on last round.


nikkolasg:
Old error using the "valid()" method. Simply removed that since it will be changed with the monitor branch.

PR #57, opened 2015-11-06
Fixed shamir bug

nikkolasg:
Temp fix is to change the suite used to directly set the suite to edwards.NewAES128SHA256Ed25519(false) (not using the full group).


ineiti:
And what is the long-term-fix? Is this the bug mentioned in #42 or another one?

nikkolasg:
Yes it is the bug mentioned in #42 .
The longterm fix is to use NewAES128SHA256Ed25519(false) as suite in the whole code. As soon as we do that, we must ask every people in the current cothority tree to  renew their key as they will be invalid.

PR #59, opened 2015-11-09
Implement new callbacks

ineiti:
This is a big rewrite of the proto/sign code to use the new ```Callbacks``` interface. Instead of registering functions, an interface has to implement a function for each step of the process defined by the interface.

In this pull request a first implementation of ```CallbacksStamper``` is done, which holds the same as the original code.

Now one can add different ```Callbacks```: bitcosi, gitcosi, ...

This is a complete first draft, up for comments.


ineiti:
Superseeded by #107

PR #60, opened 2015-11-09
testing for deploy

ineiti:
Wrote test for deploy. This should pass before we can merge replace_wallclock_with_rusage

### DIFF
--- a/app/sign/client.go
+++ b/app/sign/client.go
@@ -13,6 +13,7 @@ import (
 	"github.com/dedis/cothority/lib/proof"
 	"strconv"
 	"github.com/dedis/cothority/proto/sign"
+	"github.com/dedis/crypto/abstract"
 )
 
 var MAX_N_SECONDS int = 1 * 60 * 60 // 1 hours' worth of seconds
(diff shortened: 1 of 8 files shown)


nikkolasg:
It is fine for me to merge but I have one question before ( ;) ) : Should it not be better if we stick to the **regular** way of testing ? By that I mean, each package has its set of tests. And we could make a test that launches them all or just use "go test ./...". Each separate **should** use direct methods calls and not binaries (not currently done for the whole set of tests). That way we test exactly what needs to be tested. Regarding this file, it only test the **return value** of our code, which could be different than 0 for a whole lot of reasons (not enough permissions, space or whatever). It only tests if something is launching but not if something is working correctly as expected and for that, it is **not really** a Go test a priori.
On the other hand I see why do you need this kind of quick test in development to check that your new feature does not break anything on any other apps, but I'd rather put it in a small bash script (or ruby!) near deploy.go rather than in \*_test.go file. That way, you can still use it to quickly check that everything is right, and still have the regulars tests that check the code itself.
What do you think ?

ineiti:
Well, if our code returns a value != 0 for error, then the test works. Anyway, you can test on:

- methods - check whether methods return values as expected
- libraries - check more complex usage involving multiple methods and structs
- application - check whether the application per se works fine

Of course it's best to have all three. One should start building tests for methods whenever one writes a new method.
After that the libraries-test can be written which check on more complex behaviour.
Finally an application-test can test the overall behaviour of the application.

So we should have all three. If we have only the methods test, it doesn't show that the application works OK - it just shows that the methods work. If we have the application-test, it doesn't help to debug, because if we break a method, we only know the whole thing broke, not which part.

Issue #62, opened 2015-11-10
Discard first round in monitoring


ineiti:
Often the first round in an experiment still sets up some structures, so it's often best to discard that first round. I propose per default to discard it, but add an argument to ```deploy``` to keep the first round.

Issue #61, opened 2015-11-10
English usage of spaces


ineiti:
In English, there is NO space before the ':' (contrary to French), so you write

```Launching proxy redirecting to localhost: 4001```

Whatsmore the dbg.*-functions that take multiple arguments already enter a space between all arguments, so

```dbg.Lvl1("Launching proxy redirecting to ", host )```

has one space too much after the ```to```.

PR #64, opened 2015-11-10
Quickfixes

nikkolasg:
issue #61 and #62


ineiti:
One issue - one pull-request.
If you change things (like the arguments in Start), this is a third pull-request.
Can you do some git-magic and make three pull-requests? Supposing that each commit you did introduces only 1 set of features, you can use

```git cherry-pick```

To include these commits into a new branch. If you have commits that work on multiple sets, you'll have to rewrite the code.

nikkolasg:
Totally. Let s train the git skills...

PR #67, opened 2015-11-11
Space in dbg

### LINKED ISSUE #61: English usage of spaces

nikkolasg:
Fixing space


ineiti:
```find . -name "*go" | xargs grep '".* :' ```

Pipe that into a file, check all lines are valid, and do some perl-magic

```for file in $( cat files_with_spaces ); do perl -pi -e```

And I let you figure out the regular expression for the perl-part ;)

ineiti:
The following should be corrected:

- no space before a :
- no space before the end of a "

Usage of grep/perl is OK

ineiti:
```find . -name "*go" | xargs egrep ' ",'```

still spews out a lot of superfluous spaces...

PR #69, opened 2015-11-11
Quickfix discard

nikkolasg:
So we can now call "-discard round,verify..." on the command line to discard the first measure for each of these measures.


ineiti:
I would like to have ```-discard round,verify``` as default - do you think that is good or bad? Or could we have it in the platform/deterlab/deter.toml configuration file?

ineiti:
Created pull-request #75

PR #70, opened 2015-11-11
Monitor stamping

nikkolasg:
added StampPerc option in config_coll to say what is the percentage of node we want to request timestamp from.
Code using that is in stamp/clients.go.

NOTE: Stamp receive the stamp request and respond to it, but strangely the client does not appear to get any responses.. I did not try yet to see if there is the same behavior in development but i am pretty sure there is. I will to investigate further on this later.


nikkolasg:
Test for slack

ineiti:
Rename ```StampPerc``` to ```StampNodes``` or similar. Instead of giving a percent-value, give directly the number of nodes to contact per machine.

PR #76, opened 2015-11-13
Reworked proxy-port and added flag for monitor-port

ineiti:
- Changed the occurences of ```Logger``` to ```Monitor```
- Added monitor_port to flags in ```deploy```
- Removed ProxyAddressPort and replaced with monitor_port + 1


nikkolasg:
We do not necessarily need the Monitor + MonitorPort flags on App since apps will only be using the full address to connect. That way you don't have to change every apps. 
We just need to make sure localhost.go + forkexec.go gives the rights address to apps.

nikkolasg:
Forkexec / Users need to give directly the full address now.

ineiti:
OK to merge.

deploy/platform/deterlab/users/users.go
```diff
@@ -124,11 +123,14 @@ func main() {
 
 	// ADDITIONS : the monitoring part
 	// Proxy will listen on Sink:SinkPort and redirect every packet to
-	// RedirectionAddress:RedirectionPort. With remote tunnel forwarding it will
+	// RedirectionAddress:SinkPort-1. With remote tunnel forwarding it will
```
nikkolasg: RedirectionAddress:SinkPort + 1 , no ?

nikkolasg:
Beside my comment, OK to merge for me.

Liamsi:
I also think it is ready to merge.

PR #86, opened 2015-11-16
Change logging

ineiti:
Threw away some ```log.Warn``` and ```log.Fatal``` and put the different Lvl-values to useful defaults. Level:

- 1 - only most important things to be sure it's running
- 2 - more details for debugging
- 3 - messages that grow linearly with number of nodes
- 4 - messages that grow exponentially with number of nodes
- 5 - HUGE messages


nikkolasg:
We should put that descriptions of levels inside the wiki or the coding rules.

Issue #50, opened 2015-11-02
Tests


ineiti:
A lot of tests are failing - we should clean them up.

ineiti:
Just to be sure that the compilation doesn't fail, it could be enough to put a dummy-test in all directories (e.g. app/*), so when running ```go test```, it tests at least whether everything compiles.

nikkolasg:
And/Or we could set up an associated Travis CI account so it automatically tries to build everything at each pushed commit.

ineiti:
Yes, there is quite a list on what we could/should do - I propose we do our monthly meeting early this week (Monday or Tuesday) and set up some more detailed ‘HOWTO’ about our desired workflow.

ineiti:
Closed by #90

PR #90, opened 2015-11-16
testing for conode

ineiti:
Getting all tests to run again...


ineiti:
Closes #50

ineiti:
OK, this seems to be a good first shot. Only test that's not working is

```lib/monitor/monitor_test.go```

Nicolas says it's a timing issue, so I'll let him have a shot at it ;)

PR #91, opened 2015-11-16
Monitor percentile

nikkolasg:
Added parameter : filter_***** = PercentileValue in runconfig
For example filter_round = 80, will take the 80-percentile of the round measurements and discard everything that is above that value. Using the nearest-rank method.


ineiti:
Gorgeous

Issue #87, opened 2015-11-16
Monitor: take all fields that we receive


nikkolasg:
In monitor remove the extraFields and additionals measures so that it displays every fields that it gets and every measures it receives.

nikkolasg:
In fact, I did not think about it before but there exists an incompatibility between this feature and the fact that we want the header of the CSV already written to file before any tests have been made. Either we wait for the first measure to arrive and display all the fields in the header (Still have to be careful that we do not write send new measures for subsequents tests), or we simply continue specifying which fields exists in our tests results.

ineiti:
No problem if the header comes with the first line of measures.

ineiti:
Closed by #94

PR #94, opened 2015-11-16
reworked monitor + test working

nikkolasg:
monitor now is used iwth a structure for more control on the concurrency. Test added + working (included the TestMonitor !)

### DIFF
--- a/lib/monitor/monitor.go
+++ b/lib/monitor/monitor.go
@@ -18,31 +18,51 @@ import (
 var Sink = "0.0.0.0"
 var SinkPort = "10003"
 
-// mutex is used to update the global stats from many connections
-var mutex *sync.Mutex
-
-var done = make(chan bool)
-
-// Connections currently in use
-var conns = make([]net.Conn, 0)
-
 func init() {
-	mutex = &sync.Mutex{}
+}
+
+type Monitor struct {
+	listener net.Listener
+
+	// Current conections
+	conns map[string]monitorConnection
+	// and the mutex to play with it
+	mutexConn sync.Mutex
+
+	// Current stats
+	stats *Stats
+	// and the mutex to play with it
+	mutexStats sync.Mutex
+
+	// channel to give new measures
+	measures chan Measure
+
+	// channel to notify the end of a connection
+	// send the name of the connection when finishd
+	done chan string
+}
+
+func NewMonitor(stats *Stats) Monitor {
+	return Monitor{
+		conns:      make(map[string]monitorConnection),
+		stats:      stats,
+		mutexStats: sync.Mutex{},
+		measures:   make(chan Measure),
+		done:       make(chan string),
+	}
 }
 
 // Monitor will start listening for incoming connections on this address
 // It needs the stats struct pointer to update when measures come
 // Return an error if something went wrong during the connection setup
-func Monitor(stats *Stats) error {
+func (m *Monitor) Listen() error {
 	ln, err := net.Listen("tcp", Sink+":"+SinkPort)
 	if err != nil {
 		return fmt.Errorf("Error while monitor is binding address : %v", err)
 	}
+	m.listener = ln
 	dbg.Lvl2("Monitor listening for stats on ", Sink, ":", SinkPort)
-
-	ch := make(chan net.Conn)
-	var nconn int
-	var finished bool = false
+	finished := false
 	go func() {
 		for {
 			if finished {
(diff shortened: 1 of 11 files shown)


lib/monitor/monitor.go
```diff
@@ -18,31 +18,51 @@ import (
 var Sink = "0.0.0.0"
```
ineiti: Please add a short overview how the ```Monitor```, ```Stats```, ```Measure``` and ```Measurement``` go together. Btw: what is the difference between ```Measure``` and ```Measurement```?

lib/monitor/monitor.go
```diff
@@ -18,31 +18,51 @@ import (
...
+
+	// channel to give new measures
+	measures chan Measure
+
+	// channel to notify the end of a connection
+	// send the name of the connection when finishd
+	done chan string
+}
+
+func NewMonitor(stats *Stats) Monitor {
```
ineiti: Add a short comment

lib/monitor/monitor.go
```diff
@@ -59,46 +79,73 @@ func Monitor(stats *Stats) error {
...
+// to the monitor
+type monitorConnection struct {
+	conn net.Conn
+	// For telling when to stop AND when the connection is closed
+	done chan string
+	// Giving the stats back to monitor
+	stats chan Measure
+}
+
+func (mc *monitorConnection) Stop() {
```
ineiti: Add a short comment

lib/monitor/stats.go
```diff
@@ -231,6 +231,12 @@ func (m *Measurement) Update(measure Measure) {
 	m.System.Store(measure.CPUTimeSys)
 }
 
+func (m *Measurement) Aggregate() {
```
ineiti: Add a short comment - btw: I'm always lost when I have to search for functions that are named the same way for different structures. Is there a way to rename *Measurement.Aggregate* or *Stats.Aggregate*? And is *Aggregate* really the right word? Look at http://www.thesaurus.com/browse/aggregate for other words.

ineiti:
Is #87 in here?

lib/monitor/monitor.go
nikkolasg: I did not find a adequate name between what is sent from the client (a measure) and what the monitor actually measures/takes/stores/computes (a "measurement"). Please if you have any ideas...

nikkolasg: Each struct has some comments before its definition that describe what they are and what they do related to the others structs. Does this seem reasonable ? What is missing ?

ineiti:
I just updated the https://github.com/dedis/coding with references on how to do the comments. We should make sure that each function gets a comment, and also each package.

Besides that the code looks good.

lib/monitor/monitor.go
```diff
@@ -18,31 +29,55 @@ import (
 var Sink = "0.0.0.0"
 var SinkPort = "10003"
 
-// mutex is used to update the global stats from many connections
-var mutex *sync.Mutex
+func init() {
```
Daeinar: Is this empty init() function required for anything?

lib/monitor/stats.go
```diff
@@ -11,6 +11,196 @@ import (
...
+// Return a NewStats with some fields extracted from the platform run config
+// It enforces the default set of measure to do.
+func NewStats(rc map[string]string) *Stats {
+	s := new(Stats).NewStats()
+	s.readRunConfig(rc)
+	return s
+}
+
+// Read a config file and fills up some fields for Stats struct
+func (s *Stats) readRunConfig(rc map[string]string) {
```
Daeinar: Nitpick: Why is it called "readRunConfig" instead of just simply "readConfig"? Moreover, the two dbg.Fatal comments below (lines 59 and 64) speak of a "RunConfig" function but I guess "readRunConfig" is meant?

lib/monitor/monitor.go
Daeinar: As far as I understand a `Measure` is a single data point whereas a `Measurement` is a collections of `Measure`s? Maybe we can express this relation in the naming of the structs? The current naming scheme might lead to confusions.

(`Record` might be a suitable synonym for a single `Measure`.)

Daeinar: One could also go with `Measure` + `Measures` but the difference might be too subtle? Maybe one of the following?
`Measure` + `MeasureCollection`
`Measurement` + `MeasurementCollection`
`Record` + `RecordCollection`
`DataPoint` + `DataPointCollection`
Too clumsy?

ineiti: This is the state of the nation:

- Measure - is what is used in the clients to start a measure
- Value - one measurement
- Measurement - Wall/User/System measurement tuple
- Stats - slice of Measurements

But it's confusing, to say the least...

lib/monitor/monitor.go
ineiti: No, delete it.

lib/monitor/stats.go
ineiti: Well the `RunConfig` is an array of simulations to perform, so readRunConfig is not too bad. And thus `RunConfig` in the dbg-messages should also be OK.

lib/monitor/monitor.go
Daeinar: Ah you removed it already, thanks.

lib/monitor/stats.go
Daeinar: Okay, thanks for clarifying.

Daeinar:
Wouldn't it be simpler in https://github.com/DeDiS/cothority/blob/monitor_stateful/lib/monitor/measure.go#L62-L70 to just have a Toggle() function which switches between enabled/disabled?

ineiti:
Yes, I also asked myself. What is better:

```
Toggle(bool)
or
SetEnabled(bool)
```

Daeinar:
I'd prefer `SetEnabled(bool)` or just `Enable(bool)` to `Toggle(bool)` since the latter does not capture the meaning of what is "toggled".

PR #99, opened 2015-11-19
Dev readme

Daeinar:
Fixes some markdown formatting issues in the conode README.

### DIFF
--- a/app/conode/README.md
+++ b/app/conode/README.md
@@ -1,9 +1,9 @@
-# Conode
+# CoNode
 
-This is the first implementation of a Cothority Node for public usage. When
-built it functions as a node in a static tree. This first implementations
-serves as a public timestamper-server which takes a hash and returns the
-signed hash together with the inclusion-proof.
+This is the first implementation of a cothority node (CoNode) for public usage.
+When built, it functions as a node in a static tree. This first implementation
+serves as a public timestamper-server which takes a hash and returns the signed
+hash together with the inclusion-proof.
 
 A simple program is given that can get the hash of a file signed and check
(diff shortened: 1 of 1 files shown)


app/conode/README.md
```diff
@@ -117,11 +121,12 @@ These are the steps to be part in the EPFL-conode-project:
 
 ### Compile Conode
 
-For Conode to compile, you need to have the dedis/crypto-library in branch
-```shamir``` and the conode in branch ```development```. We suppose you have
-a running 
+We suppose you have a running Go implementation. For Conode to compile, you need
+to have the dedis/crypto-library in branch ```shamir``` and the conode in branch
+```development```. Execute the following steps:
```
ineiti: In fact we don't need the ```shamir``` branch anymore. Can you correct that?

app/conode/README.md
```diff
@@ -136,13 +141,13 @@ Once conode is compiled, you can call it with
 
 ```conode keygen address [-key keyname]```
 
-Where ```address``` is the public IP of your server. This will create two 
+Where ```address``` is the public IP of your server. This will create two
```
ineiti: Perhaps we should use other markings for the arguments?

Daeinar: What exactly do you mean by **markings**?

ineiti: Like do *address* instead of ```address``` ?

Daeinar: Hmm, okay. I usually prefer to have in-line code formatting when referencing other code parts for consistency reasons. But the above works as well for me, whatever you usually use.

PR #102, opened 2015-11-19
deploy: check if deterlab-experiment is active

Daeinar:
Shutdown cothority/deploy if the target experiment is not available on deterlab.


ineiti:
And the branch could be `deterlab_check_active` or such.

Daeinar:
Agreed. I'll rename the branch.

ineiti:
Leave it for now - but for the next one it'll be nice.

Issue #92, opened 2015-11-16
Verify usage of edwards-suite in Conode


ineiti:
Verify that Conode uses the 8x-Edwards suite and do a new binary and tell everybody to redo it's key.

Daeinar:
Ran some tests from https://github.com/dedis/cothority/deploy/simulation and they seem to work with the new settings.

ineiti:
Closed by #106

PR #106, opened 2015-11-20
Verify Usage of Edwards-suite in CoNode

Daeinar:
Moving all Edwards-suites from 
```
edwards.NewAES128SHA256Ed25519(true)
```
to
```
edwards.NewAES128SHA256Ed25519(false)
```

### DIFF
--- a/app/conode/conode.go
+++ b/app/conode/conode.go
@@ -12,7 +12,7 @@ import (
 )
 
 // Which suite to use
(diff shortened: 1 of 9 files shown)


ineiti:
Did you run some of the tests?

Daeinar:
Yes, see my comment at: https://github.com/DeDiS/cothority/issues/92

Issue #118, opened 2015-12-03
Travis fails with test


ineiti:
In `lib/conode` Travis doesn't build the test `TestStamp` for an unknown reason.

Liamsi:
The reason for that is that on travis we currently use go version 1.4.1 while locally we use 1.5.2 (or 1.5.1). I've als verified that the test fails with version 1.4.1 locally.

Liamsi:
Now the test `TestStamp` runs fine with the branch `travis_tests_fail` (see issue #137). But two other tests fail/quit with panic:
- (fails with only with go 1.5.1) https://travis-ci.org/DeDiS/cothority/jobs/95843756#L4875
- and (fails only with with go 1.5.1 and 1.5.2) https://travis-ci.org/DeDiS/cothority/jobs/95843756#L2037
The 2nd seems related to https://groups.google.com/forum/#!topic/golang-nuts/1aZmhhSvwWc

Liamsi:
Silenced the first by using go version 1.5.2 only (which we should do anyway) and the second by removing/commenting a line in dbug_level.go (which seems unnecessary to me anyways)  
If you want you can have a look on #137, comment/merge and close both #137 and this issue.

ineiti:
Cool - thanks. On an administrative level, we need to decide on what to do with regard to `Issue`s and `Pull-request`s.

ineiti:
Implemented in #137

Liamsi:
Reopened: travis fails again

Liamsi:
Finally, I think this can be closed once and for all

PR #121, opened 2015-12-03
added exceptionlist to broadcast message

ineiti:
Adds the exceptionlist to the broadcastmessage

Still needs a test in `lib/conode/stamp_test.go` which:
- sets up 2 peers
- makes the 2nd one raising exceptions
- connects to the stamplistener
- verifies the transmitted message


Liamsi:
This isn't only about the test anymore, but also:
- [ ] implement missing parts ot the exception handling such that the test can pass (pass around and use the exceptionlist where necessary)
- [ ] the unmarshaling of the struct `StampSignature` fails, verfied from [here](https://github.com/DeDiS/cothority/commit/d56226ea3fef37427e2a1c9c9dea6b6f33730724) on; also understand why and fix it (the error message is 
`cannot unmarshal object into Go value of type abstract.Point`

Liamsi:
TODO:
refactor refusal lists (refused to commit/refused to sign) to increase readability and simplify implementing the exception handling 
 - [ ] add a struct refusal/rejected which contains both lists
 - [ ] use this struct where ever `RejectionPublicList` and `RejectionCommitList` are used

nikkolasg:
After further discutions with Ismail, we came to the conclusion that it may not be necessary to do this change now. The reason behind that is that Bryan wants us to use a bitmap of indexes so we can directly relate the indexes to the keys. This is not clear yet on how we may want to do this but since the code here is still pretty readable and working, I don't think it would be worth the effort since it will be refactored later.

Ok for me to merge.

Liamsi:
Both tests `TestStampWithoutException` and `TestStampWithExceptionRaised` fail on  travis but not locally. Nicolas suggested that it might be a problem with some configuration/key file being different (locally and on travis). I can't really confirm that. I only checked the files in `lib/conode/testdata`though. 
Before this PR the first test was called `TestStamp` and the following line
````
tests := exec.Command("go", "test", "-v", "./...", "-run", "[^T][^e][^s][^t][^S][^t][^a][^m][^p]")
```
in [dedis-travis-script.go](https://github.com/DeDiS/cothority/blob/2f05e8bc3d33c06d7bdbdf29907c213f22a8494d/dedis-travis-script.go#L52) deactivated the test. This does not happen anymore. What would be the correct way to handle this? Deactivate both test in the way before or find out why travis fails on those and activate both? 

Besides the open question on travis test failure, I'm also OK with merging.

Liamsi:
Update: There is another issue #118 which is about travis and failing tests. Hence, I'm OK with merging, too.

ineiti:
OK, merge it if you feel like it's OK.

PR #124, opened 2015-12-04
Tree phase comments

ineiti:
Open for first comments of the new tree/peer structures.

- tree: peers (nodes) and trees
- phase: different parts of an algorithm (announce, commit, challenge, response for CoSi)


lib/tree/peer.go
```diff
@@ -0,0 +1,74 @@
...
+// Peer represents a Cothority-member
+type Peer struct {
+	// The hostname of the peer
+	Name string
+	// The port-number
+	Port int
+	// The public-key
+	PubKey abstract.Point
+	// The private-key - mostly for ourselves
+	PrivKey abstract.Secret
```
bford: I don't think this should be in the Peer struct, given that it seems like it's definitely *only* for ourselves and should probably be "known" by ourselves in some other way.

lib/tree/peer.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/dedis/crypto/abstract"
+	"strconv"
+)
+
+// Peer represents a Cothority-member
+type Peer struct {
+	// The hostname of the peer
+	Name string
+	// The port-number
+	Port int
```
bford: Perhaps better just to use a single string, which can be in "hostname:port" syntax (perhaps with port optional)?  I think the Go network code already supports that syntax, doesn't it?  That would make dependence on the exact structure of endpoint IDs less pervasively hard-coded.

lib/tree/peer.go
```diff
@@ -0,0 +1,74 @@
...
+	Peers []*Peer
+	// The hash-id of this list
+	Hash hashid.HashId
+	// The suite used in this list
+	Suite abstract.Suite
+}
+
+// NewPeerListLocalhost creates a list of n localhosts, starting
+// with port p and random keys.
+func NewPeerListLocalhost(s abstract.Suite, n, port int) *PeerList {
```
bford: Is this just for testing purposes?  It sounds like something that would only be useful for test deployments anyway.  If so, shouldn't it be in a _test.go file?

lib/tree/tree.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+/*
+The tree library holds all hosts and allows for different ways to
+retrieve the hosts - either in a n-ary tree, broadcast or just
+ a selection.
+*/
+
+// Tree holds the structure about the current tree
+type Tree struct {
```
bford: Are these structs intended to be protobuf-encodable?  If so, mention that in comments.  If not, what is the encoding plan for communicating trees?  And is it possible to communicate just part of a tree rather than the whole thing?  (Might not be critical from the start but worth thinking about.)

lib/tree/peer.go
ineiti: That's true. I just thought that it's the most convenient place to put it by putting similar things together. Else I'll have to put the private key somewhere in the `Node` structure (which doesn't exist yet)

lib/tree/peer.go
ineiti: OK, I can do that.

lib/tree/peer.go
ineiti: This is also used when doing tests with `deploy`. But in fact it has to support both localhosts-only and different hosts (for deployment to deterlab). I'll change the signature to take a list of hostnames which will be used in a round-robin fashion to set up the list.

lib/tree/tree.go
ineiti: You have to convince me first that protobuf is really a good idea ;) For the moment I prefer JSON - it also covers the `map[string]`-type, which is not covered by protobuf, but which I'd like to have to transfer the messages between the peers, so that each round-type can transport it's own messages.

I switched between different tree-structures and settled on this one as it is the most convenient to use in the code. It will probably be marsahlled into a list of indexes to the PeerList, as you suggested.

And if I understood the protobuf-implementation correctly, it doesn't work with different versions of structures at the sending and receiving end. But I might have misread the code.

lib/tree/peer.go
```diff
@@ -0,0 +1,74 @@
...
+
+// PeerList regroup a number of peers in a list. One peer can be
+// member of more than one PeerList.
+type PeerList struct {
+	// A list of all peers that are part of this list
+	Peers []*Peer
+	// The hash-id of this list
+	Hash hashid.HashId
+	// The suite used in this list
+	Suite abstract.Suite
```
nikkolasg: Following our discussion, I think suite should not be here. Should it not be higher up, in the definition of what application we want to run(stamp,bitcosi etc) ?

lib/tree/tree.go
```diff
@@ -0,0 +1,98 @@
...
+	// The top-most peer of this tree
+	Root *TreeEntry
+	// The hash-id of this tree - same for the whole tree
+	Hash []byte
+	// The hash of the peer-list
+	HashPL []byte
+	// The branching factor
+	BF int
```
nikkolasg: We may not want to impose a tree topology restriction with BF. The tree should be created the way the initiator wants it to be.

ineiti:
Will start new

PR #133, opened 2015-12-09
Conode add update

ineiti:
Adding script to have a hook to do some things with the conodes during the update process. Everybody needs to re-create their key, as we're not using the full-group anymore.


ineiti:
OK, that should be it.

app/conode/start-conode.sh
```diff
@@ -9,22 +9,28 @@ main(){
   setup)
     if [ -f key.pub ]; then
       echo -e "\n*** Key.pub already exists - if you want to re-create, please delete it first\n"
+      exit
     else
-      ./conode keygen $2
-      cat key.pub | mail linus.gasser@epfl.ch
+      if ./conode keygen $2; then
+        echo Sending public-key to linus.gasser@epfl.ch
+        cat key.pub | mail linus.gasser@epfl.ch
```
nikkolasg: Are you sure about using mail ? It's not even configured on mine. 
The manual way is still the easiest way to do this. (Send email by hand).

Later, I think that when verification is done, we could make these operations automatic. Imagine the following workflow:
 - leader -> node node saying "Ok you're good to go,verification is OK, send me your public key now"
 - node -> leader : "this is my public key <adress> <publickey>"
 - leader computes the config.toml, waits for any other nodes if he wants then
    - leader -> node :" this is the config.toml you should use, run now please !"

But this is something related to bryan wishes of the tree being stateless also so it is only an idea to be incorporated into futur changes..

app/conode/exit_conodes.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
-HOSTS=$( cat config.toml | grep Hosts | sed -e "s/.*\[\"\(.*\)\"\]/\1/" | perl -pe "s/\", \"/\n/g" )
+CONFIG=config.toml
+if [ "$1" ]; then
+  CONFIG=$1/$CONFIG
+fi
+HOSTS=$( cat $CONFIG | grep Hosts | sed -e "s/.*\[\"\(.*\)\"\]/\1/" | perl -pe "s/\", \"/\n/g" )
```
nikkolasg: For things like that, a small comment would be welcomed ;)

nikkolasg:
As a general note, a bit more comments would be very welcomed (yeah I know, Linus, code speaks by itself, but bash speaks badly sometimes :D)  At least one comment at the beginning of "start-conode.sh" saying in short what it does, its differents steps / state etc.

ineiti:
Added more comments and scripts - needs to be tested first.

ineiti:
Oups - pull-request against master.

Issue #112, opened 2015-11-26
Change all log to dbg


ineiti:
We're still using log in some parts of `lib/sign` - this should be changed to `dbg` calls. But first `dbg` needs methods for:

- dbg.Error
- dbg.Errorf
- dbg.Warn
- dbg.Warnf

None of these is `panic`ing or anything else. Just print `Error:` and the message or `Warn:` and the message.

PR #140, opened 2015-12-10
log calls to dbg

nikkolasg:
Resolving issue #112


Liamsi:
Looks very OK. I was just wondering why `dbg.Error` just prints an error message and `dbg.Errorf`does an `os.Exit(1)` additionally. If found that a little confusing but looking at the code it seems intended (and it's not really related to your changes).

PR #142, opened 2015-12-11
Conode add update

ineiti:
Update and putting the port to 7770


nikkolasg:
OK to me.

app/conode/exit_conodes.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
-HOSTS=$( cat config.toml | grep Hosts | sed -e "s/.*\[\"\(.*\)\"\]/\1/" | perl -pe "s/\", \"/\n/g" )
+CONFIG=config.toml
+if [ "$1" ]; then
+  CONFIG=$1/$CONFIG
+fi
+HOSTS=$( grep Hosts $CONFIG | sed -e "s/.*\[\"\(.*\)\"\]/\1/" | perl -pe "s/\", \"/\n/g" | tail -r )
```
Liamsi: Note: after merging we've recognized that `tail -r` won't work on most linux machines...

app/conode/update.sh
```diff
@@ -0,0 +1,80 @@
...
+        ./update.sh update_version
+        exec ./start-conode.sh run
+        ;;
+    esac
+}
+
+# Fetches the latest version and untars it here
+update(){
+  if [ ! -e NO_UPDATE ]; then
+    RELEASE=$( wget -q -O- https://github.com/dedis/cothority/releases/latest | grep DeDiS/cothority/releases/download | sed -e "s/.*href=.\(.*\). rel.*/\1/" )
```
Liamsi: `grep -i DeDiS/cothority/releases/download`

app/conode/start-conode.sh
```diff
@@ -33,45 +48,27 @@ main(){
...
     echo $0 run
+    echo "or if you want to manually update:"
+    echo $0 update
     echo
     ;;
   esac
 }
 
-update(){
-  RELEASE=$( wget -q -O- https://github.com/dedis/cothority/releases/latest | grep DeDiS/cothority/releases/download | sed -e "s/.*href=.\(.*\). rel.*/\1/" )
```
Liamsi: case insenstive grep, too

app/conode/exit_conodes.sh
ineiti: Le 16/12/2015 14:21, Ismail a écrit :
>
> In app/conode/exit_conodes.sh 
> <https://github.com/dedis/cothority/pull/142#discussion_r47774281>:
>
> > @@ -1,6 +1,10 @@
> >  #!/bin/bash
> >
> > -HOSTS=$( cat config.toml | grep Hosts | sed -e "s/.*\[\"\(.*\)\"\]/\1/" | perl -pe "s/\", \"/\n/g" )
> > +CONFIG=config.toml
> > +if [ "$1" ]; then
> > +  CONFIG=$1/$CONFIG
> > +fi
> > +HOSTS=$( grep Hosts $CONFIG | sed -e "s/.*\[\"\(.*\)\"\]/\1/" | perl -pe "s/\", \"/\n/g" | tail -r )
>
> Note: after merging we've recognized that |tail -r| won't work on most 
> linux machines...
>
>
There is `tac` - but I think that doesn't exist everywhere, neither. At 
least it doesn't on a mac. I wanted to exit the leaf-nodes as first 
nodes, then the root-node. But in fact it doesn't do it hierarchly 
anyway. So you might as well just throw away the `tail -r`.

Liamsi: Thanks @ineiti :-)

PR #145, opened 2015-12-14
Network library

nikkolasg:
Rework of the lib/network :
* use of Protobuf for packets (auto-instantiation/Decoding of slices (hashmaps will be another feature to add,it's already in the official docs), compressed binary format that portable enough..)
* No need to have the suite in order to decode using the "Constructors" mechanisms from protobuf ! No need to instantiate packet before
* Can do event-driven packet processing without relying on the big "sign.SigningMessage" struct that contains all the reference to all the packets possible. 

With this in place, we can move forward to the stateless protocol / tree approach.

TODO: 
+ make the net_test.go works. OK
+ Adds a default type when the we receive an unknown type OK
+ make the rest of the code use lib/network PARTIAL OK (only lib/ uses network for now)
+ make all test working PARTIAL OK (see commit d8e5b7dd84ab5a184092570b2d48f0d905dc9fc5)

ADDITIONS:
While I was at it, I was kind of forced to remove the use of the lib/graphs stuff that annoyed us for some time now. Some of the utilities are now in lib/tree (bad name, but a temporary name, something like config should be better suited). The views are also in there.


ineiti:
The `lib/tree` needs to be cleaned up - it was used for *Tree phase*. Please

* rename it to something else
* delete things from *Tree phase*

ineiti:
Still has some conflict with the dev-branch

lib/network/net_test.go
```diff
@@ -1,46 +1,73 @@
...
+	// encounters a non basic type, we first check if we have a constructor for
+	// it, if yes, calls it to get a initialized object, otherwise just call
+	// reflect.New(type). This enfors the use of only ONE suite for example
+	// for all the connections related to one host. This behavior can maybe
+	// change in the future with the use of the Context thing currently
+	// implementing in crypto branch cipher
+	var suite = edwards.NewAES128SHA256Ed25519(false)
+	constructors = make(protobuf.Constructors)
+	var point abstract.Point
+	var secret abstract.Secret
```
ineiti: Use the pre-defined global `suite`

ineiti:
It should pass in travis

nikkolasg:
Travis CI passes !!

PR #148, opened 2015-12-22
Fix skeleton / stamp / sign error

nikkolasg:
fixed channels change from 9101261e4c38bd529ec39d05757e696f848f76f0 so that every apps use the same approach.


Liamsi:
Looks good! Thanks

PR #156, opened 2016-01-12
removed LastSeenVote parameter

nikkolasg:
This was used during voting process that we don't do know, old legacy code.
Don't use Atomic.LoadInt64 with 32-bit architecture on ARM ;)


Liamsi:
Nice, ready to merge.

PR #161, opened 2016-01-14
Reenable commented failing test

### LINKED ISSUE #157: Monitor proxy_test.go
In proxy_test.go, there are two tests: TestProxy and TestReadyProxy. 
ReadyProxy only tests for the Ready functionality, but this is already tested in TestProxy. The issue in Travis VM is that running both tests sequentially fails, but running one after the other does not. It may be a closing connection - related issue, not sure. It works fine on my Mac anyway so that's strange.
For the moment, I just commented out TestReadyProxy.

Liamsi:
will  fix #157


Liamsi:
Closed: Will create a new PR where branch is based on development and not on sda_1

PR #162, opened 2016-01-14
Re-enable commented test and make it pass on travis ...

### LINKED ISSUE #157: Monitor proxy_test.go

Liamsi:
Should fix #157

### DIFF
--- a/lib/monitor/measure.go
+++ b/lib/monitor/measure.go
@@ -12,10 +12,11 @@ package monitor
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/dedis/cothority/lib/dbg"
 	"net"
 	"syscall"
 	"time"
+
+	"github.com/dedis/cothority/lib/dbg"
 )
 
 // Sink is the server address where all measures are transmitted to for
@@ -61,7 +62,10 @@ func ConnectSink(addr string) error {
 }
 
 func StopSink() {
-	connection.Close()
+	if err := connection.Close(); err != nil {
+		// at least tell that we could not close the connection:
+		dbg.Error("Could not close connecttion:", err)
+	}
 	encoder = nil
 }
 
@@ -107,6 +111,7 @@ func send(v interface{}) {
 	if !enabled {
 		return
 	}
+	// XXX comment needed:
 	for wait := 500; wait < 1000; wait += 100 {
 		if err := encoder.Encode(v); err == nil {
 			return
@@ -171,7 +176,9 @@ func (m *Measure) reset() {
 // Prints a message to end the logging.
 func End() {
 	send(Measure{Name: "end"})
-	connection.Close()
+	if err := connection.Close(); err != nil {
+		dbg.Error("Could not close connection:", err)
+	}
 }
 
 // Converts microseconds to seconds.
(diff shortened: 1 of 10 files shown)


Liamsi:
We really should mock `net.Connection` instead of using real connections. Otherwise in every test we open ports, we also test if the port is open etc.  

Done. Please review.

Liamsi:
Golang has a [default mocking framework](https://github.com/golang/mock/tree/master/sample) btw

lib/monitor/measure.go
```diff
@@ -107,6 +111,7 @@ func send(v interface{}) {
 	if !enabled {
 		return
 	}
+	// XXX comment needed:
```
ineiti: For a large number of clients (˜10'000), the connection phase can take some time. This is a linear backoff to enable connection even when there are a lot of requests.

lib/monitor/proxy_test.go
```diff
@@ -80,69 +80,67 @@ func TestProxy(t *testing.T) {
...
+	done := make(chan bool)
+	go func() {
+		if err := monitor.Listen(); err != nil {
+			dbg.Error("Could not start listener:", err)
+			dbg.Error("Retry once in 500ms ...")
+			time.Sleep(500 * time.Millisecond)
+			if err := monitor.Listen(); err != nil {
+				dbg.Fatal("Could not start listener:", err)
+			}
+		}
```
ineiti: Does it work without this? Will say, is that a remnant of a try to make it work, or is it needed? 
If needed {why? Is it not possible to make TestProxy wait for correct closing of the ports?} else { remove }

lib/monitor/proxy_test.go
```diff
@@ -80,69 +80,67 @@ func TestProxy(t *testing.T) {
...
+	proxyAddr := "localhost:" + strconv.Itoa(SinkPort)
+	if err := ConnectSink(proxyAddr); err != nil {
+		dbg.Error("Could not connect to proxy:", err)
+		dbg.Error("Retry once in 500ms ...")
+		time.Sleep(500 * time.Millisecond)
+		if err := ConnectSink(proxyAddr); err != nil {
+			t.Error(fmt.Sprintf("Can not connect to proxy : %s", err))
+			return
+		}
+	}
```
ineiti: Same as with monitor.Listen - needed or not?

lib/monitor/proxy_test.go
Liamsi: This will be part of #163

lib/monitor/proxy_test.go
Liamsi: Also part of #163

Liamsi:
Ready for merging

Liamsi:
BTW: I think the error is related to this: https://github.com/golang/go/issues/11872#issuecomment-126183460
because the first test closes the connection but we still get a `getsockopt: connection refused`

nikkolasg:
Since there will be quite some refactoring in #163 , for me it is OK to merge.

PR #168, opened 2016-01-19
working on tests for nice Travis-report

ineiti:
Adapted tests for new network-library.


lib/network/net_test.go
```diff
@@ -80,37 +91,53 @@ func TestSecureMultiClose(t *testing.T) {
...
 	go func() {
 		err := h1.Listen(fn)
 		if err != nil {
 			t.Fatal("Listening failed for h1:", err)
 		}
+		close(done)
 	}()
-	h2.Open(entity1)
-	err := h1.Close()
+	time.Sleep(time.Second)
```
Liamsi: Nitpicking: I think it would be nice to have a short comment explaining why we need to wait (and why one second).

ineiti: Hmm, in fact it should not be necessary anymore. I'll check it out tomorrow.

And Travis fails again - in a new, unknown place...

nikkolasg: I'll try to look at it this afternoon if I finish JVSS ( not sure ).

lib/sign/messagessign.go
```diff
@@ -16,43 +16,20 @@ Over the network they are sent as byte slices, so each message
...
+var Response = network.RegisterMessageType(ResponseMessage{})
+var SignatureBroadcast = network.RegisterMessageType(SignatureBroadcastMessage{})
+var StatusReturn = network.RegisterMessageType(StatusReturnMessage{})
+var CatchUpReq = network.RegisterMessageType(CatchUpRequest{})
+var CatchUpResp = network.RegisterMessageType(CatchUpResponse{})
+var GroupChanged = network.RegisterMessageType(GroupChangedMessage{})
+var VoteRequest = network.RegisterMessageType(VoteRequestMessage{})
+var CloseAll = network.RegisterMessageType(CloseAllMessage{})
+var Identity = network.RegisterMessageType(IdentityMessage{})
+
```
Liamsi: More nitpicking (not sure): Do we need this values to be global variables? If so can they have more concise names (like AnnouncmentType, etc) and a comment: "Use as 'read only'"?

Liamsi:
Besides the nitpicking: Looks good but we should fix the failing test before merging.

lib/sign/messagessign.go
ineiti: Good question... we need to
- register the structures in the Network-library 
- be able to do a `switch` on the structure-type

In fact the second point is probably not necessary before `sda`, but for sure in `sda`.

Another possibility would be to add a `network.MessageType(IdentityMessage{})` which would return the ID of that message, and use that in a `switch`-statement. But this seems clumsy, too. We could also play around with `string`s instead of `uuid.UUID`s.
Personally I think the global variable approach is the best.

nikkolasg: >Personally I think the global variable approach is the best.

I also agree with that. Golang does not provide us the ability to have constant expressions unfortunatly, so for me it is *by default* the best choice.

Liamsi:
Btw I also get failing tests on my machine (not only in travis): Sometimes `TestProtocolAggregation` and sometimes `TestHostClose` and sometimes even `TestMakeConfig`

lib/sign/messagessign.go
Liamsi: OK. Then we keep it that way (and if we can come up with a non-clumsy, easy solution, we can change it later)

nikkolasg: But we can definitely make clear somewhere what are those global variables and that you should not change them anyhow.

ineiti:
Le 20/01/2016 11:38, Ismail a écrit :
>
> Btw I also get failing tests on my machine (not only in travis): 
> Sometimes |TestProtocolAggregation| and sometimes |TestHostClose| and 
> sometimes even |TestMakeConfig|
>
OK - more testing for Thursday. This and getting `development` to work - 
wonder if it's really worth it for the `development`-branch, as it will 
be overwritten with `sda` and most of the problems I have now are in the 
`lib/sign`-library which will have to be rewritten, too...

Liamsi:
Not sure what you mean by worth the effort for the `development` brnach. 
After pulling master of dedis/crypto (this is also what travis will do) and fixing the issues with `NewKeyPair` on this branch (`sda_clean_test`), the only test that fail are sometimes `TestProtocolAggregation` and sometimes `TestNode2` and a few times `TestMakeConfig`. 
Find attached a patch, for making this branch compile with the latest dedis/crypto.
[keypair.patch.txt](https://github.com/dedis/cothority/files/97316/keypair.patch.txt)

PR #173, opened 2016-01-21
Sda msgchan

ineiti:


lib/sda/overlay.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+}
+
+// TransmitMsg takes a message received from the host and treats it. It might
+// - ask for the identityList
+// - ask for the Tree
+// - create a new protocolInstance
+// - pass it to a given protocolInstance
+func (o *Overlay) TransmitMsg(msg *SDAData) error {
+	node, ok := o.nodes[*(msg.To)]
```
nikkolasg: We can directly take msg.To.Id(), so we avoid redirection and it's quicker and more elegant.

lib/network/encoding.go
```diff
@@ -66,6 +68,24 @@ func TypeFromData(msg ProtocolMessage) uuid.UUID {
 	return msgType
 }
 
+// TypeToUUID Converts MsgType to uuid
+func TypeToUUID(msg ProtocolMessage) uuid.UUID {
+	val := reflect.ValueOf(msg)
+	if val.Kind() == reflect.Ptr {
+		val = val.Elem()
+	}
+	url := "https://dedis.epfl.ch/protocolType/" + val.Type().String()
```
Liamsi: Nitpicking: Maybe a constant with the base url would be better? This would make sure that 1) The bas url can easily be changed and 2) it's the same everywhere (no typos).

nikkolasg: I think we don't even need that actually. Using NewV5 (line just below), we already have to provide in which *namespace* we are, so NamespaceURL + MsgType is quite ... UUIdentifiable!

lib/sda/local.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/dedis/cothority/lib/dbg"
+	"github.com/dedis/cothority/lib/network"
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/config"
+	"github.com/satori/go.uuid"
+	"strconv"
+	"testing"
+)
+
+type Local struct {
```
Liamsi: Nitpicking: Comment needed. What is a `Local`?

lib/sda/node.go
```diff
@@ -0,0 +1,270 @@
...
+	// aggregate messages in order to dispatch them at once in the protocol
+	// instance
+	msgQueue map[uuid.UUID][]*SDAData
+	// Holds flags that influence the behaviour of the node
+	flags uint32
+}
+
+const (
+	NCAggregateMessages = 1 << iota
+	NCTimeout           = 1 << iota
```
Liamsi: Ist this correct? If yes, write a comment. If no, correct it.

lib/sda/node.go
```diff
@@ -0,0 +1,270 @@
...
+		channels: make(map[uuid.UUID]interface{}),
+		handlers: make(map[uuid.UUID]MsgHandler),
+		msgQueue: make(map[uuid.UUID][]*SDAData),
+		flags:    NCAggregateMessages,
+	}
+	return n, n.protocolInstantiate()
+}
+
+// TreeNode gets the treeNode of this node
+func (n *Node) TreeNode() *TreeNode {
```
Liamsi: add to comment: `if there is no TreeNode for the Token of this node, the function will return nil` (alternative: return an error and let caller handle it)

lib/sda/node.go
```diff
@@ -0,0 +1,270 @@
...
+	dbg.Lvl3(cr.Elem().Field(0).Type)
+	if cr.Elem().Field(0).Type != reflect.TypeOf(TreeNode{}) {
+		return errors.New("Input-channel doesn't have TreeNode as element")
+	}
+	typ := network.RegisterMessageUUID(network.RTypeToUUID(cr.Elem().Field(1).Type),
+		cr.Elem().Field(1).Type)
+	n.channels[typ] = c
+	dbg.Lvl3("Registered channel", typ)
+	return nil
+}
```
Liamsi: Cool :-)

lib/sda/node.go
```diff
@@ -0,0 +1,270 @@
...
+		return errors.New("Input is not channel of structure")
+	}
+	if cr.Elem().NumField() != 2 {
+		return errors.New("Input is not channel of structure with 2 elements")
+	}
+	dbg.Lvl3(cr.Elem().Field(0).Type)
+	if cr.Elem().Field(0).Type != reflect.TypeOf(TreeNode{}) {
+		return errors.New("Input-channel doesn't have TreeNode as element")
+	}
+	typ := network.RegisterMessageUUID(network.RTypeToUUID(cr.Elem().Field(1).Type),
```
nikkolasg: Do we need to register to the network ? That way, in host.go, the message will be decoded ?
But are we not already registering our type somewhere else ? Every message that applications uses are already supposed to be registered, no ?  I think you only want to get the type out of the message here and rely on network.RegisterMessageUUID which should not be used for that.  Or did I missed something ?

lib/sda/node_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/satori/go.uuid"
+	"testing"
+	"time"
+)
+
+func TestNodeChannel(t *testing.T) {
+	dbg.TestOutput(testing.Verbose(), 4)
+	names := genLocalhostPeerNames(10, 2000)
+	peerList := genEntityList(tSuite, names)
+	// Generate two example topology
```
Liamsi: two example topologies ? or "an example topology (a binary tree)" or just skip the comment

lib/sda/node_test.go
```diff
@@ -0,0 +1,221 @@
...
+		t.Fatal("Message should contain '3'")
+	}
+}
+
+// Test instantiation of Node
+func TestNewNode(t *testing.T) {
+	sda.ProtocolRegister(testID, NewProtocolTest)
+	h1, h2 := SetupTwoHosts(t, false)
+	// Add tree + entitylist
+	//el := GenEntityListFrom(h1.Suite(), genLocalhostPeerNames(10, 2000))
```
Liamsi: remove obsolete comment

lib/sda/node_test.go
```diff
@@ -0,0 +1,221 @@
...
+	el := sda.NewEntityList([]*network.Entity{h1.Entity, h2.Entity})
+	h1.AddEntityList(el)
+	tree, _ := el.GenerateBinaryTree()
+	h1.AddTree(tree)
+
+	// Try directly StartNewProtocol
```
Liamsi: remove obsolete comment or update

lib/sda/overlay.go
```diff
@@ -0,0 +1,251 @@
...
+	// add children
+	for _, c := range treeNode.Children {
+		mm[c.Id] = c
+	}
+	// add cache
+	tnc[tree.Id] = mm
+}
+
+// GetFromToken returns the TreeNode that the token is pointing at.
+func (tnc TreeNodeCache) GetFromToken(tok *Token) *TreeNode {
```
Liamsi: or nil if there is none for this token

lib/sda/sda.go
```diff
@@ -0,0 +1,20 @@
+package sda
+
+/*
+SDA is the Secure Distributed API which offers a simple framework for generating
+your own distributed systems. It is based on a description of your protocol
```
Liamsi: nitpicking: 'a simple framework for generating your own distributed systems' is a litte to general, isn't it?

Liamsi:
Besides the comments (mostly nitpicking), looks good. Ready to be merged.

lib/sda/local.go
ineiti: `LocalTest` - I'll change the name

lib/sda/node.go
ineiti: Yep, gives 1, 2, 4, 8, 16, ... Using iota only once is OK if you just want to linearly increase. Else you have to repeat the iota.

https://golang.org/ref/spec#Iota

lib/sda/node.go
ineiti: I didn't want to repeat the decoding-routine in `lib/network`, so everything goes through that same library and we need to register everything in there. Should be correct. As the `RegisterMessageUUID` returns if the `struct` is already known, this shouldn't be a problem.

nikkolasg: Yeah ok but there is a function in network library that gives you the type from the message no ?  I'll try do use this function and see if it works.

lib/network/encoding.go
ineiti: done

nikkolasg:
example do not work

lib/sda/local.go
ineiti: done

lib/sda/node.go
ineiti: done

lib/sda/node.go
ineiti: done

lib/sda/node.go
nikkolasg: you don't have to: https://splice.com/blog/iota-elegant-constants-golang/
Look at "expression".

lib/sda/node.go
ineiti: Probably - but here I had the `reflect` of the message-type, so I didn't know how to create the correct message and just went with a second method in `network` that takes directly the UUID.

lib/sda/node_test.go
ineiti: done

lib/sda/node_test.go
ineiti: done

lib/sda/node_test.go
ineiti: done

lib/sda/overlay.go
ineiti: done

lib/sda/sda.go
ineiti: we're not THAT great? You sure?

nikkolasg:
OK for me

lib/sda/node.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	_, ok = n.TreeNode()
+	if !ok {
+		return errors.New("We are not represented in the tree")
+	}
+	n.instance = p(n)
+	return nil
+}
+
+func (n *Node) DispatchFunction(msg []*SDAData) error {
```
nikkolasg: All of these "Dispatch*****" methods should be non exported. Users should be doing anything with that.

lib/sda/node.go
```diff
@@ -0,0 +1,302 @@
...
+		return mt, msgs, true
+	}
+	// no we still have to wait!
+	dbg.Lvl3("Number of msgs:", len(msgs), "number of children:", len(n.Children()))
+	return mt, nil, false
+}
+
+// Start calls the start-method on the protocol which in turn will initiate
+// the first message to its children
+func (n *Node) Start() error {
```
nikkolasg: same as for "Dispatch****", "Start()" should be un-exported.

nikkolasg:
Problem: we can not use node.Done() function.
Why: For example, I am coding ProtocolCothority, which only uses the four steps (Announce,commit,challenge,response). But later, I'd like to implement a ProtocolStamper which will use ProtocolCothority for the four steps, and then will have only one method that handles the SignatureBroadcast. But if ProtocolCothority calls ``` node.Done() ```, ProtocolCothority will "disappear" and no further messages will be dispatched to it. ...
Solution: We could have a ``` Done() ``` function that is being called FROM the node each time a protocol receive a message. That way, ProtocolStamper will override the Done() of ProtocolCothority and will be able to still receive message after.

PR #175, opened 2016-01-21
Jvss porting to SDA

nikkolasg:
This is JVSS ported to SDA.
Basically, one protocol instance setup a longterm distributed key as soon as it is created. Then we can call Sign() on it. That will create another random shared secret, then will make the actual signing using both shared secret.

I did not want to make one-time use only JVSS protocol, because one shared secret takes time to generate, so if I want to sign something using JVSS, I want to use protocols that are already established (i.e. longterm key already generated) and not generate a new longterm key each time I want to sign.

Even I agree that it is clearly not a such beautiful code but I had a hard time coping with the asynchronous way of the Protocol Dispatch() method => concurrency issues mostly !
TODO:
* add comments
* re-refactor ? I can do it without much time now that it works (and I've understood my pitfalls / previous errors). What do you think ? If you think it is OK enough, then we leave it as it is.


lib/sda/host.go
```diff
@@ -656,3 +661,27 @@ func (h *Host) checkPendingTreeMarshal(el *EntityList) {
 	}
 	h.pendingTreeLock.Unlock()
 }
+
+// Testing helpers functions
+// NewHostMock returns a freshly generated host at this address with a random
+// keypair
+func NewHostMock(s abstract.Suite, address string) *Host {
```
Liamsi: If this is intended to be used in tests only related to this package,  do not export it (make it 'private') or move it to the _test file. Same with `SetupHostMock`

protocols/jvss/jvss.go
```diff
@@ -14,24 +21,559 @@ type JVSSProtocol struct {
...
+func (jv *JVSSProtocol) waitForResponses() {
+	for sda := range jv.respChan {
+		dbg.Lvl3("JVSS (", jv.index, ") Received Response")
+		sigResponse := sda.Msg.(SignatureResponse)
+		var requestBuff *RequestBuffer
+		var ok bool
+		if requestBuff, ok = jv.requests[sigResponse.RequestNo]; !ok {
+			// Not good, someone ask for a request we did not produce a shared
+			// secret before .. ??
+			dbg.Error("Receive signature request with request number nto matching any shared secret...")
```
Liamsi: typo

lib/sda/host.go
nikkolasg: Do you know how can you import theses functions from another package if they are in a *_test package ? I can't find ( I did not search too long either ). Here I need theses functions outside SDA, and I can't put it in sda_test because I don't know how to import them. I could put them in jvss_test.go but then everyone will have to repeat the same code over and over...

protocols/jvss/jvss.go
```diff
@@ -10,28 +17,554 @@ import (
 // JVSS Protocol Instance structure holding the information for a longterm JVSS
 // signing mechanism
 type JVSSProtocol struct {
-	// The host we are running on
-	Host *sda.Host
-	// The tree we are using
-	Tree *sda.Tree
+	// The TreeNode denating ourself in the tree
```
Liamsi: typo

protocols/jvss/jvss.go
```diff
@@ -10,28 +17,554 @@ import (
...
+	publicList []abstract.Point
+	// keys of the Host set as config.KeyPair
+	key config.KeyPair
+	// The longterm shared private public key pair used in this JVSS.
+	// The idea is that you can keep this protocol instance as a longterm JVSS
+	// using it to distributively sign anything as long as it runs.
+	longterm *poly.SharedSecret
+	// The schnorr struct used to sign / verify using the longterm key
+	schnorr *poly.Schnorr
+	// polyinfo related to how the shares are generated and reconstructed
```
Liamsi: outdated comment? `polyinfo`

protocols/jvss/jvss.go
```diff
@@ -10,28 +17,554 @@ import (
...
+	// lastrequestnumber seen or executed
+	lastRequestNo int
+	// more tmp variables  that are needed because of the way we use protocol
+	// isntance. For example, you want to generate a distributed secret, you
+	// need a poly.Receiver struct. You first create that struct in Start()
+	// since you are the leader, and then receive the dealers in Dispatch(),
+	// so you need that intermediate variable.
+	// longterm-receiver of the shares
+	ltReceiver *poly.Receiver
+	// channel through we notify that we have sucessfully computed the longterm
```
Liamsi: language: channel through which we .. or: channel that indicates if longterm ...

protocols/jvss/jvss.go
```diff
@@ -10,28 +17,554 @@ import (
...
+	nodes := tree.ListNodes()
+	for i := range tree.EntityList.List {
+		for _, n := range nodes {
+			if n.Entity.Equal(tree.EntityList.Get(i)) {
+				maps[n.Id] = i
+			}
+		}
+	}
+	if idx == -1 {
+		panic("JVSSProtocol could not find itself into the entitylist")
```
Liamsi: language: ~~into~~ in the entity list

protocols/jvss/packets.go
```diff
@@ -0,0 +1,61 @@
...
+//  SignatureRequest is used when we want to sign something
+// This is to be sent by the root to everyone else
+type SignatureRequest struct {
+	Msg       []byte
+	RequestNo int
+}
+
+var SignatureRequestType = network.RegisterMessageType(SignatureRequest{})
+
+type SignatureResponse struct {
```
Liamsi: add comment, like you did for other msg-types

Liamsi:
Besides some language/comment nitpicking: OK to merge.

Liamsi:
Also, the code shouldn't issue errors if ran with `go test -race` (at least we should check that we are actually not dealing with a real race condition, but 'false positives' are very rare). You might ignore warnings if they are from other packages/libs or related to writing to std out. But you shouldn't ignore the warnings in your package. 
Of course, you might decide that it is better to merge the code first and open another issue.

Liamsi:
I have to correct myself: the race condition checker does not issue false positives!

PR #177, opened 2016-01-25
Sda cothority

nikkolasg:
Porting of Cothority to SDA framework.
It will include basic Cothority steps with Rounds (maybe some refactoring on this side also is needed).
It will later cover the Stamper application with the extra SignatureBroadcast step.


ineiti:
Closed to re-open against sda_1

PR #182, opened 2016-01-29
Redo ieee graphs

ineiti:
Re-calculation of experiments from November 2015


deploy/simulation/sign_multi_adapted.toml
```diff
@@ -11,8 +11,8 @@ Machines, ppm, bf
...
-16, 32, 10
-16, 64, 10
-16, 128, 10
-16, 256, 10
-16, 512, 10
+16, 32, 12
+16, 64, 13
+16, 128, 14
+16, 256, 15
+16, 512, 16
```
nikkolasg: why changing the BF factor at the same time? Or why do you do it now and not before?  Just out of curiosity.

nikkolasg:
Well @ineiti it's your python script, can't say much except than it's OK for me ;)

deploy/simulation/sign_multi_adapted.toml
ineiti: It was always changing, just optimising what the optimal change is. It is about `ln2(nbr_hosts)` - the broader the tree, the more calculations, but the faster the transmission (no network delay). So it doesn't have to be broad neither too deep.

ineiti:
If you think it's OK, merge it.

Liamsi:
Hmm, I do not quite understand what you did here or why you did it. Does it have anything to do with the sda? I will close and merge it.

ineiti:
Le 03/02/2016 15:13, Ismail a écrit :
>
> Hmm, I do not quite understand what you did here or why you did it. 
> Does it have anything to do with the sda? I will close and merge it.
>
> —
> <https://github.com/dedis/cothority/pull/182#issuecomment-179258236>.
>
re-writing python-matplotlib-scripts and updating simulation-values. But 
this is from November 2015 and quite old, just hanging around.

PR #185, opened 2016-02-01
SDA-simul

ineiti:
Working simulation on localhost for first protocols.


ineiti:
Some things that need to be decided:
1. do we need both ```channels``` and ```handlers```?
2. is there a better way than to have the ```struct {*sda.TreeNode; MessageAnnounce}``` for channels and handlers?
3. can we further simplify the registration of channels/handlers? Now we have to register first the message-type, then the channels/handlers using that message-type.
4. should we change the ```ProtocolInstance.Dispatch``` from ```sda/protocol``` to a void-fuction that is called as a go-routine to handle the channels?

Daeinar:
@ineiti: The PR looks good. Some comments:
1. I am fine with keeping channels or both channels and handlers. I think the channels-approach is not overly complex though, i.e. I am not sure if there is such a big advantage in keeping also handlers. But maybe that's just me ... :) Furthermore, I would probably rename the skeleton protocol back to `example` (since it's more intuitive, no?) and restructure the folders to either `example/channels/` and `example/handlers/` (if we keep both channels and handlers) or just to `example/` (if we only keep channels). 
2+3. Ideally I think registration of message-types should be the responsibility of the SDA-user (=protocol-implementer), i.e. as it is now, and registration of channels should be handled by SDA automatically. This could be done either in parallel to message-type registration or after `init()`. I think users of SDA should only worry about the message types they need and the rest should be handled by SDA. I guess if one could automatically handle registration of the channels one could also get rid of `StructAnnounce` (with `*sda.Treenode`) etc. I am not sure however if that approach is possible without another bigger restructuring of SDA?
4. Sounds like a good idea, but I am not sure if there are any deeper consequences. Haven't thought about that too much tbh. ;)

ineiti:
Moved skeleton to example-files. I will leave for the moment the registration up to the protocol-implementer, even though I'd like to make it more simple.

lib/sda/overlay.go
```diff
@@ -65,7 +65,16 @@ func (o *Overlay) TransmitMsg(sdaMsg *SDAData) error {
 		}
 		node = o.nodes[sdaMsg.To.Id()]
 	}
-	err := node.DispatchMsg(sdaMsg)
+
+	t, msg, err := network.UnmarshalRegisteredType(sdaMsg.MsgSlice, network.DefaultConstructors(network.Suite))
```
nikkolasg: Cool thing that it is now in overlay !

nikkolasg:
Some comments:
 * cothority.go: from what I understood, it is a central point where you can launch simulations or simply a SDA node (and you decide if you want to start it or not). But ... What if I want to start my protocol using some specific configurations from a file (or stg else) ? It seems this file handle very good the simulation part but not quite the "standalone" part, and furthermore, there are mixed global variables that are about simulation AND standalone parts... I think it could be very better to clearly separate both purposes.
 * lib/sda/simulation.go: ```LoadSimulationConfig(dir,ha string)``` Since it is for testing, it is not so much a big deal, but I really think we should take the habit to stick with *Entity* and stop using clear addresses. (maybe use index in a EntityList to designate someone)
 * lib/sda/overlay.go : Why put overlay as public ? Nobody should use it directly. Or else put Host as private but I'm not sure about keeping both public...  Anyway, we should do an issue about SDA that contains WAY too many public fields / methods.
 * lib/sda.go:129 (and all related stuff): Do we have to keep simulation and sda related ? Can't we solely pass some informations to the simulation interface, and they decide what to do ? Yes it's true that we will mostly simulate using protocols that uses sda, but if we were to separate the two, it could become much *cleaner* and the interaction between much *clearer*.  One example of this strange interaction, in sda/simulation.go:194, in the Node() function, you register the EntityList + Tree to the overlay, for convenience reason, but using a SimulationBFTree as the receiver of the function. For me this is strange.

More to follow tomorrow .. Gonna go home ;)
Question: Where is the code to deploy to deterlab ?

nikkolasg:
>1. do we need both channels and handlers?
I agree with @Daeinar here, I don't see much differences between the two (hence no real need to keep both).
>2. is there a better way than to have the struct {*sda.TreeNode; MessageAnnounce} >for channels and handlers?
Since we decided to go with only TreeNode & Node, and to not expose the Token to the client, it seems a good solution overall. I tried to put the message as a pointer but without success.
>3. can we further simplify the registration of channels/handlers? Now we have to >register first the message-type, then the channels/handlers using that message-type.
Do we have to register the message type ? I thought you registered automatically when using channels.
>4. should we change the ProtocolInstance.Dispatch from sda/protocol to a void-fuction that is called as a go-routine to handle the channels?
Great idea, that way it is explicit what the user must do. +1

ineiti:
OK, so I'll move the `Dispatch`-function to be a go-routine called when the protocol is instantiated.

ineiti:
`LoadSimulationConfig` doesn't have yet the entity-IDs, only the addresses. Difficult to do otherwise.

ineiti:
`simulation` needs to access a lot of structures of the `SDA` to function correctly. So if we want to make even more methods private, it will be very difficult to separate the two.

ineiti:
Private/public:
- users should only access `Node`, nothing else
- `Overlay` and `Host` are mostly used by `Simulation`, so one could simplify things there.

nikkolasg:
OK for me.

nikkolasg:
In fact, before merging we should decide on wether keeping both channels + handlers  or only one... !!!

ineiti:
Le 03/02/2016 15:00, nikkolasg a écrit :
>
> In fact, before merging we should decide on wether keeping both 
> channels + handlers or only one... !!!
>
>
I definitively prefer handlers (as they're NOT callbacks ;) Channels are 
only good if you need a timeout.

Liamsi:
Besides the usual nitpicking (see below) I'm also OK with merging. I agree with Nicolas and Philipp that we should have one mechanism to work with (either handlers or channels), but I would be OK to merge it now, so that we can continue on the sda and later improve or streamline the API.
Nitpicking: 
ports do not get closed properly in the tests (on linux/travis) and the output seems broken (e.g. `Couldn't listen in localhost:2010 : No address worked for listening on this host%!(EXTRA *errors.errorString=Error opening listener: listen tcp 0.0.0.0:2010: bind: address already in use)` )

Liamsi:
> I definitively prefer handlers (as they're NOT callbacks ;) Channels are
only good if you need a timeout.

Funny to see you prefer callbacks ;P

ineiti:
because they're NOT callbacks. If you want to cut it to one, we have to take channels, as handlers don't support timeouts. But please think about all those protocols that will have an 'empty' select-section just to pass the values of the channels to the handlers...

ineiti:
about port-listening problems: we can merge it into `sda_1` and open an issue. Another one is the fact that the `SingleHost`-flag doesn't work yet. But then again, this is only necessary with more than 32'000 hosts. Perhaps not now.

Liamsi:
I'm sure there would be some way to use "NOT-callbacks" together with timeouts. I think it is OK to have some protocols with an empty select. As I said, I am not religious about one or the other approach and if you want to keep both I think this is also OK for now (you have to convince Philipp and Nicols). Regarding the tests: I think it's definitely OK to ignore the port problems for now. I wouldn't even open an issue for that (we won't ignore travis while merging with development anyways)

ineiti:
+1 for keeping both: the added code is not that big, usability for students is improved.

If you need timeouts -> use channels
else -> use handlers

PR #186, opened 2016-02-02
Sda simul deterlab

ineiti:
Working version of running SDA in Deterlab. Using a simple protocol there is a maximum of 31767 nodes on 16 servers.


nikkolasg:
Please include some tests files *.toml so we can test.Are they the same as before ? Apparently now the folder must be simul/runconfig/ ?

simul/cothority/cothority.go
```diff
@@ -0,0 +1,127 @@
...
+
+// The address of this host - if there is only one host in the config
+// file, it will be derived from it automatically
+var HostAddress string
+
+// ip addr of the logger to connect to
+var Monitor string
+
+// Simul is != "" if this node needs to start a simulation of that protocol
+var Simul string
```
Liamsi: Make the global vars private/non-exported (at least from the code there seems no good reason to have them exported but I might have missed sth)

Liamsi:
To whoever is also reviewing this a little tip:
use https://github.com/dedis/cothority/compare/sda_simul_dist...sda_simul_deterlab 
instead of the default files changed view (at least as long as `sda_simul_dist` isn't merged)

simul/platform/deterlab.go
```diff
@@ -47,10 +48,14 @@ type Deterlab struct {
...
+	// Number of available servers
+	Servers int
+
+	//
 	// Name of the simulation
 	Simulation string
-	// Directory of applications
-	AppDir string
+	// Directory holding the cothority-go-file
+	CothorityDir string
```
Liamsi: Isn't this constant?

simul/platform/deterlab.go
```diff
@@ -74,43 +79,32 @@ type Deterlab struct {
...
-	TestConnect bool
 }
 
+var simulConfig *sda.SimulationConfig
+
 func (d *Deterlab) Configure() {
 	// Directory setup - would also be possible in /tmp
 	pwd, _ := os.Getwd()
+	d.CothorityDir = pwd + "/cothority"
 	d.DeterDir = pwd + "/platform/deterlab"
```
Liamsi: Insn't this constant, too?

Liamsi:
Besides some unnecessary global vars, I think this is ready to merge. I just read the code though (did not try to run the simulation on deterlab)

simul/platform/deterlab.go
ineiti: Le 03/02/2016 15:52, Ismail a écrit :
>
> In simul/platform/deterlab.go 
> <https://github.com/dedis/cothority/pull/186#discussion_r51729963>:
>
> >  func (d *Deterlab) Configure() {
> >  	// Directory setup - would also be possible in /tmp
> >  	pwd, _ := os.Getwd()
> > +	d.CothorityDir = pwd + "/cothority"
> >  	d.DeterDir = pwd + "/platform/deterlab"
>
> Insn't this constant, too?
>
> —
> <https://github.com/dedis/cothority/pull/186/files#r51729963>.
>
the `pwd`-part is not constant, this is why I create these here.

Liamsi: Ah, OK. I though relative to the directory they are always the same (but as the comment says, one could change from pwd to tmp, oversaw that one ...) 
And I guess there is also a good reason for making them exported global vars (like sth. deterlab specific)?
Anyways, this is just nitpicking, too.

simul/cothority/cothority.go
ineiti: privatised a lot of variables. As this structure is written/read using toml, the elements that have to be written/read need to be public.

simul/platform/deterlab.go
ineiti: put the directories private, as they're recreated every time the simulation is run.

ineiti:
OK, this should do it. Thanks to @nikkolasg for alpha-testing and finding missing files...

Liamsi:
Looks OK!

Issue #192, opened 2016-02-04
Commit-Tree


Liamsi:
As discussed, we want to create a very basic merkle-tree implementation (for n-ary trees). Currently, we do not care for reconstructing the tree nor about the tree structure. We just want a hash function which can take an arbitrary number of BinaryMarshalers (Points or g^vs) and an arbitrary number of already computed hashes (slices of slices of bytes).

PR #193, opened 2016-02-04
Basic commit tree implementation (issue #192)

Liamsi:
Basic commit-tree implementation used (resolves issue  #192)

### DIFF
--- a/lib/merkle/merkle.go
+++ b/lib/merkle/merkle.go
@@ -1,0 +1,48 @@
+// Package merkle contains a n-ary merkletree implementation
+// (basically a hash function)
+package merkle
+
+import (
+	"encoding"
+
+	"github.com/dedis/crypto/abstract"
+)
+
+// Merkle contains all information needed to hash the data into a commit-tree
+type Merkle struct {
+	suite abstract.Suite
+}
+
+// Hash is necessary to implment the BinaryMarshaler interface on []byte
+type Hash []byte
+
+// NewMerkle initializes
+func NewMerkle(s abstract.Suite) *Merkle {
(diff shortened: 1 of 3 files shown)


ineiti:
Cool - can you add a test whether this passes the `lib/network` Marshal/Unmarshaler?

Liamsi:
> Cool - can you add a test whether this passes the lib/network Marshal/Unmarshaler?

Sure. Do you mean to put the resulting hash in a `NetworkMessage` to see if `lib/network/encoding.go` can (Un)MarshalBinary correctly?

nikkolasg:
in lib/network, you have the public methods ```MarshalRegisteredType(...) error``` & ```UnmarshalRegisteredType(...) error```, you'll see if it reports an error then.

Liamsi:
Thanks for the hint @nikkolasg. Added a test for that. Please review and merge.

ineiti:
Good for me, ready for merge.

lib/merkle/merkle.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"encoding"
+
+	"github.com/dedis/crypto/abstract"
+)
+
+// Merkle contains all information needed to hash the data into a commit-tree
+type Merkle struct {
+	suite abstract.Suite
```
nikkolasg: Do we really need the suite ? To make it more portable, can't we just take a Hash function as the field ? Since we only use the Hash function anyway + we do not instantiate anything with the suite.

Liamsi: done

PR #202, opened 2016-02-05
CloseAll doesn't work for 8191 hosts - PR

ineiti:


lib/sda/host.go
```diff
@@ -175,7 +175,7 @@ func (h *Host) Listen() {
 		dbg.Lvl3(h.workingAddress, "Accepted Connection from", c.Remote())
 		// register the connection once we know it's ok
 		h.registerConnection(c)
-		h.handleConn(c)
+		go h.handleConn(c)
```
nikkolasg: You can remove the go as the function is already supposed to happens in a separate goroutine (at least in my lib/network version)

protocols/manage/count.go
```diff
@@ -49,18 +58,23 @@ func NewCount(n *sda.Node) (sda.ProtocolInstance, error) {
...
 		select {
 		case pc := <-p.PrepareCountChan:
-			dbg.Lvl3("Received from", pc.TreeNode.Entity.Addresses,
-				pc.TreeNode.Id)
+			dbg.Lvl3(p.Myself(), "received from", pc.TreeNode.Entity.Addresses,
+				pc.Timeout)
+			p.Timeout = pc.Timeout
 			p.FuncPC()
 		case c := <-p.CountChan:
 			p.FuncC(c)
```
nikkolasg: I know you like it short, but FuncC / FuncPC are *really* short ... what do they do ?

lib/network/struct.go
```diff
@@ -18,7 +18,7 @@ import (
 
 // How many times should we try to connect
 const maxRetry = 10
-const waitRetry = 1 * time.Second
+const waitRetry = 100 * time.Millisecond
 const timeOut = 5 * time.Second
```
Liamsi: Is this on purpose? Before we waited 1 sec for retrying and now 100ms (makes sense to me because one second is rather long but I am not sure)

lib/sda/host_test.go
```diff
@@ -9,6 +9,25 @@ import (
 	"time"
 )
 
+func TestCloseChannel(t *testing.T) {
```
Liamsi: Remove this test (it doesn't test anything we wrote)

lib/sda/host.go
ineiti: OK - thought it's necessary. Testing now.

simul/cothority/cothority.go
```diff
@@ -77,25 +78,33 @@ func main() {
...
 			case count := <-node.ProtocolInstance().(*manage.ProtocolCount).Count:
 				if count == rootSC.Tree.Size() {
-					dbg.Lvl2("Found all", count, "children")
+					dbg.Lvl1("Found all", count, "children")
 					wait = false
 				} else {
 					dbg.Lvl1("Found only", count, "children, counting again")
 				}
-			case <-time.After(time.Second * 10):
+			case <-time.After(time.Millisecond * time.Duration(timeout) * 2):
```
Liamsi: Can we have short comments why the time(out) is like that (why `* 2` because most of the time it is a binary tree, or does it have other reasons?)

protocols/manage/count.go
ineiti: Commented.

lib/network/struct.go
ineiti: Yep, else the creation of the network-communications of the tree were really slow.

lib/sda/host_test.go
ineiti: Done

Liamsi:
Looks good. I did not try to run it though. For every timeout etc. it would be nice to have a (very short) comment why the time was chosen like that (branching-factor is 2/sum of bla will give a general timeframe of 10 min, which is the max time we want to wait for an experiment to finish). Otherwise they look like randomly chosen constants.

simul/cothority/cothority.go
```diff
@@ -104,24 +113,37 @@ func main() {
...
+		}
+		allClosed <- true
+	}()
 	select {
-	case <-rootSC.Host.Closed:
-		dbg.Lvl2(hostAddress, ": first host closed - quitting")
-		monitor.End()
+	case <-allClosed:
+		dbg.Lvl2(hostAddress, ": all hosts closed")
+	case <-time.After(time.Minute * 2):
```
Liamsi: Make this timeout a global var (const) or as you suggested put them into toml files (with reasonable default if there is nothing set in the toml)

simul/cothority/cothority.go
ineiti: Added comment

simul/cothority/cothority.go
ineiti: Done

ineiti:
Also added a `simul/runfiles/README.md` which explains the runfiles.

nikkolasg:
OK for me.

Liamsi:
Also OK. The readme is helpful, too.

PR #203, opened 2016-02-05
Merge message structs and their SDA-wrappers - PR

ineiti:
On the `sda_unify_msg_struct`-branch, in `lib/sda`, run

```
go test -v -run TestNodeChannelCreate\$
```

and see the errors. Important functions in `lib/sda/node.go`:

- `RegisterChannel` - where the correct structure is checked
- `DispatchChannel` - where the received message is received and prepared for the protocol

The best would be:

- to completely omit the `*TreeNode`-embedding in our structure if we just don't care where the packets come from
- if we add the `*TreeNode`-structure in our message, `DispatchChannel` and `DispatchHandler` will initialise it upon *reception*

### DIFF
--- a/lib/sda/node.go
+++ b/lib/sda/node.go
@@ -108,6 +108,20 @@ func (n *Node) SendTo(to *TreeNode, msg interface{}) error {
 	if to == nil {
 		return errors.New("Sent to a nil TreeNode")
 	}
+	// Check whether we know that type of message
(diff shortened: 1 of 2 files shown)


ineiti:
The branch is `sda_unify_msg_struct`

ineiti:
sda_1 is obsolete, will open this against development

PR #208, opened 2016-02-08
BizCoin to SDA

nikkolasg:
BizCoin porting to SDA + new features (in 7 days!)


Liamsi:
My Todos:
- [ ] port the [client](https://github.com/LefKok/cothority/blob/BitCoSi_round/app/skeleton/client.go#L23)
- [ ] merge sda_1 (after issue #210 is merged)
- [ ] write a simulation (see the round*.go in https://github.com/LefKok/cothority/blob/BitCoSi_round/app/skeleton/)

nikkolasg:
Added some comments. For me OK to merge. The changes in the sda/ library + net/, are some small improvements (better dbg output, error handling, concurrency issues, network buffering issues etc) but no big changes.

ineiti:
Can you add a `README.md` in `protocols` with a short description of each protocol in it?
What about the travis-failures, is this something we need to take care of in the sda-branch?

nikkolasg:
the travis-failure are here since a long time and are caused by the many changes we made in the network library. This should be fixed when merging sda_1 into development later.

lib/crypto/schnorr.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+	cipher := suite.Cipher(rvBuff)
+	cipher.Message(nil, nil, msg)
+	e := suite.Secret().Pick(cipher)
+
+	if !e.Equal(sig.Challenge) {
+		return errors.New("Challenge reconstructed is not equal to one given in signature")
+	}
+	//  'verything OK
```
Liamsi: verything?

lib/crypto/schnorr.go
```diff
@@ -0,0 +1,70 @@
...
+	cbuf, err := ss.Challenge.MarshalBinary()
+	if err != nil {
+		return nil, err
+	}
+	rbuf, err := ss.Response.MarshalBinary()
+	return append(cbuf, rbuf...), err
+}
+
+// SchnorrSignature (named from wikipedia page)
+func SignSchnorr(suite abstract.Suite, private abstract.Secret, msg []byte) (SchnorrSig, error) {
```
Liamsi: comment doesn't fit with method name

lib/network/encoding.go
```diff
@@ -120,8 +125,9 @@ func (am *NetworkMessage) SetError(err error) {
 }
 
 var typeRegistry = make(map[uuid.UUID]reflect.Type)
+var registryLock = new(sync.Mutex)
```
Liamsi: What happens if `UnmarshalRegisteredType`is called as a go-routine several times? And two (or more) go-routines call `registryLock.Lock()`? registry.DeadLock() ...

lib/network/net.go
```diff
@@ -70,9 +71,9 @@ func (t *TcpHost) Listen(addr string, fn func(Conn)) error {
 // Close will close every connection this host has opened
 func (t *TcpHost) Close() error {
 	for _, c := range t.peers {
-		dbg.Lvl3("Closing peer", c)
+		dbg.Lvl4("Closing peer", c)
 		if err := c.Close(); err != nil {
-			dbg.Lvl3("Error while closing peers")
+			dbg.Error("Error while closing peers")
 			return handleError(err)
```
Liamsi: If you are on it, should be: Error while closing peer (not peers)

lib/network/net.go
```diff
@@ -121,25 +122,38 @@ func (c *TcpConn) Receive(ctx context.Context) (NetworkMessage, error) {
...
+		buffer.Write(b[:n])
+		// if we could not read everything yet
+		if Size(buffer.Len()) < s {
+			// make b size = bytes that we still need to read (no more no less)
+			b = b[:s-read]
+		}
 	}
 	defer func() {
 		if e := recover(); e != nil {
 			debug.PrintStack()
```
Liamsi: remove PrintStack or the whole block

lib/network/net.go
```diff
@@ -151,6 +165,9 @@ func (c *TcpConn) Receive(ctx context.Context) (NetworkMessage, error) {
 	return am, nil
 }
 
+// how many bytes do we write at once on the socket
+var maxChunkSize Size = 4096
+
```
Liamsi: could be a global `const`

lib/network/net.go
```diff
@@ -159,24 +176,47 @@ func (c *TcpConn) Send(ctx context.Context, obj ProtocolMessage) error {
...
-	var buffer bytes.Buffer
-	size := Size(len(b))
-	if err := binary.Write(&buffer, globalOrder, size); err != nil {
+	packetSize := Size(len(b))
+	if err := binary.Write(c.Conn, globalOrder, packetSize); err != nil {
 		return err
 	}
 	// Then send everything through the connection
-	buffer.Write(b)
-	c.Conn.SetWriteDeadline(time.Now().Add(timeOut))
```
Liamsi: Shouldn't it still be possible to timeout the write? (otherwise it will block)

lib/network/encoding.go
Liamsi: Also: if it is not used in go-routines, we won't need a lock here.

lib/network/struct.go
```diff
@@ -19,7 +19,7 @@ import (
 // How many times should we try to connect
 const maxRetry = 10
 const waitRetry = 100 * time.Millisecond
-const timeOut = 5 * time.Second
+const timeOut = 3000 * time.Second
```
Liamsi: This doesn't look like a "on-purpose change"

lib/sda/host.go
```diff
@@ -94,7 +94,7 @@ func NewHost(e *network.Entity, pkey abstract.Secret) *Host {
 		host:               network.NewSecureTcpHost(pkey, e),
 		private:            pkey,
 		suite:              network.Suite,
-		networkChan:        make(chan network.NetworkMessage, 1),
+		networkChan:        make(chan network.NetworkMessage, 10),
```
Liamsi: Why is 10 a good choice for the buffer, here?

lib/sda/host.go
```diff
@@ -248,7 +248,9 @@ func (h *Host) SendRaw(e *network.Entity, msg network.ProtocolMessage) error {
 	h.networkLock.Unlock()
 
 	dbg.Lvl4(h.Entity.Addresses, "sends to", e)
-	c.Send(context.TODO(), msg)
+	if err := c.Send(context.TODO(), msg); err != nil && err != network.ErrClosed {
+		dbg.Error("ERROR Sending to connection:", err)
```
Liamsi: Maybe add `c.LocalAdress` or whatever is available

lib/network/encoding.go
```diff
@@ -120,8 +125,9 @@ func (am *NetworkMessage) SetError(err error) {
...
```
Liamsi: Maybe, put this in a struct, which embeds a mutex and only is accessible through functions for reading and writing to the map `typeRegistry` (and which handles the (un)locking)

protocols/bizcoin/bizcoin.go
```diff
@@ -0,0 +1,819 @@
...
+// ChallengeCommit round is started
+func (bz *BizCoin) OnChallengeCommit(fn func()) {
+	bz.onChallengeCommit = fn
+}
+
+// onResponseCommitDone registers a function which will be called when
+// onResponseCommit round is finished
+func (bz *BizCoin) OnResponseCommitDone(fn func()) {
+	bz.onResponseCommitDone = fn
+}
```
Liamsi: I think all these callbacks can be removed?

protocols/bizcoin/ntree.go
```diff
@@ -0,0 +1,332 @@
...
+	dbg.Lvl3(nt.Name(), "Start Signature Request")
+	sigRequest := &RoundSignatureRequest{msg}
+	go nt.verifySignatureRequest(sigRequest)
+	for _, tn := range nt.Children() {
+		nt.SendTo(tn, sigRequest)
+	}
+}
+
+// Go routine that will do the verification of the signature request in
+// parrallele
```
Liamsi: typo

protocols/bizcoin/ntree_simulation.go
```diff
@@ -0,0 +1,91 @@
...
+			dbg.Lvl3("NtreeProtocol DONE")
+			done <- true
+			// TODO verification of signatures
+			/*for {*/
+			//if err := verifyBlockSignature(node.Suite(), node.EntityList().Aggregate, sig); err != nil {
+			//dbg.Lvl1("Round", round, " FAILED:", err)
+			//} else {
+			//dbg.Lvl1("Round", round, " SUCCESS")
+			//}
+			/*}*/
```
Liamsi: Open a new issue if the verification of the signature still fails.

protocols/viewchange/viewchange.go
```diff
@@ -0,0 +1,158 @@
...
+
+type viewMsgChan struct {
+	*sda.TreeNode
+	View
+}
+
+type View struct {
+	EntityList *sda.EntityList
+	Tree       *sda.TreeMarshal
+}
```
Liamsi: We don't use this viewchange code anywhere and we are not sure if it works. We should delete it before merging.

simul/simul.go
```diff
@@ -41,7 +41,7 @@ var nobuild = false
 var clean = true
 var build = ""
 var machines = 3
-var monitorPort = 10000
+var monitorPort = 6789
```
Liamsi: Was this change made on purpose?

Liamsi:
As you can see I also added a bunch of comments (and also did some minor changes)

> What about the travis-failures, is this something we need to take care of in the sda-branch?

I think it is not the best practise that we are ignoring travis failures since such a long time (since we started the sda branch). The whole idea of **continuous** integration gets subverted by ignoring it continuously and seeing it as a kind of 'end-boss' which we face when merging into master.

ineiti:
Move `protocols/bizcoin/*toml` to `simul/runfiles`

lib/network/net.go
nikkolasg: We could but then we need to make tests with that functionality, and I'm afraid when running multiple hosts on one server that the timeout fails before ? or with big messages that it fails before ? 
What do you think ?

ineiti:
Think how to handle the `BlocksDir`-parameter, so that the simulations can be properly run:

* where to download
* does it need a script to initialise these?
* put it into Git or outside?

lib/network/struct.go
nikkolasg: not even used anymore. I remove it.

ineiti:
Do we want to have all `.gitignore` in the root-directory or like `protocols/bizcoin/.gitignore` in the sub-directories?

lib/sda/host.go
nikkolasg: just experimenting because of the marshalling error. Set it back to one.

ineiti:
Is there a way to cut the huge structure of `bizcoin.BizCoin` into smaller parts?

ineiti:
What's the difference between `NewBizCoinProtocol` and `NewBizCoinRootProtocol`?

In general: can we have some documentation at the top of `bizcoin.go`?

Why is it BizCoin and not ByzCoin?

ineiti:
shouldn't `bizcoin.listen` be directly put into `Dispatch`? I remember we 'auto-goify' the `Dispatch`-method in `sda.protocolInstantiate`, no?

ineiti:
Comments about `case` should be inside the case and not before, like in `bizcoin.listen`

ineiti:
Why do you use `Type` in `bizcoin.BizCoinAnnounce` instead of having two types of messages?

nikkolasg:
>Think how to handle the BlocksDir-parameter, so that the simulations can be properly run:
>where to download
It's already copied in deterlab within my account folder ***/remote/blocks. Anyone can access it.
>does it need a script to initialise these?
Nope, it's done by the simulation.
>put it into Git or outside?
Since the size, putting it in Git(hub) won't work. But Git-annex is a nice addon for git for managing big files. we could setup one here maybe.

nikkolasg:
>shouldn't bizcoin.listen be directly put into Dispatch? I remember we 'auto-goify' the Dispatch-method in sda.protocolInstantiate, no?

Yes, when you use the automatic instantiation from node.go instance. Look in bizcoin/server.go, this is like the "platform" that links protocol instances together. I first create the protocol instance, set up some transactions, then link it with the Node. Since I do that for the root, I preferred to launch the listening go routine for everyone.

ineiti:
About `BlocksDir`: I would still like to have a publicly downloadable place where a blockchain (if it's what this is) can be downloaded. Just in case anybody not having access to Deterlab wants to use it.

In the best case, the `bizcoin_simulation` would handle the download of that file.

protocols/bizcoin/bizcoin.go
```diff
@@ -0,0 +1,819 @@
...
+	err := bz.SendTo(bz.Parent(), commitment)
+	dbg.Print(bz.Name(), "HandleCommit() sent to parent!")
+	return err
+}
+
+// startPrepareChallenge create the challenge and send its down the tree
+func (bz *BizCoin) startChallengePrepare() error {
+	// make the challenge out of it
+	trblock := bz.tempBlock
+	marshalled, err := json.Marshal(trblock)
```
ineiti: Why is the trblock marshalled again? Can't this be taken care of by the SDA?

nikkolasg:
>Why do you use Type in bizcoin.BizCoinAnnounce instead of having two types of messages?

No real reasons, it seemed to me at the time that,since we were doing exactly the same thing for announcement and commitment, there will be less code using a simple type than a explicit struct for that (which implies creating the wrapper structure around it, another channel using it). But that's not stg that can't be changed definitely.

protocols/bizcoin/bizcoin.go
```diff
@@ -0,0 +1,819 @@
...
+	bizChal := &BizCoinChallengePrepare{
+		TYPE:      ROUND_PREPARE,
+		Challenge: ch,
+		TrBlock:   trblock,
+	}
+
+	go verifyBlock(bz.tempBlock, bz.lastBlock, bz.lastKeyBlock, bz.verifyBlockChan)
+	dbg.Lvl3(bz.Name(), "BizCoin Start Challenge PREPARE")
+	// send to children
+	for _, tn := range bz.Children() {
```
ineiti: I think we need a `sda.SendToAllChildren`

protocols/bizcoin/bizcoin.go
nikkolasg: Yeah we just used the same functions that lefteris did using the old code. He is using that as a Hash function. That's not right I agree, but for the simulation that was sufficient.

protocols/bizcoin/bizcoin.go
```diff
@@ -0,0 +1,819 @@
...
+	return bzr, true
+}
+
+// verifyBlock is a simulation of a real verification block algorithm
+func verifyBlock(block *blockchain.TrBlock, lastBlock, lastKeyBlock string, done chan bool) {
+	//We measure the average block verification delays is 174ms for an average
+	//block of 500kB.
+	//To simulate the verification cost of bigger blocks we multiply 174ms
+	//times the size/500*1024
+	b, _ := json.Marshal(block)
```
ineiti: How would a correct verification look like? Is it difficult to implement? Was this mentioned in the paper that there is only a 'wait' to 'simulate' the verification?

protocols/bizcoin/bizcoin.go
```diff
@@ -0,0 +1,819 @@
...
+// protocol.
+func (bz *BizCoin) Signature() *BlockSignature {
+	return &BlockSignature{
+		Sig:        bz.commit.Signature(),
+		Block:      bz.tempBlock,
+		Exceptions: bz.tempExceptions,
+	}
+}
+
+func (bz *BizCoin) RegisterOnDone(fn func()) {
```
ineiti: Is this callback better than having a channel? If so, why?

protocols/bizcoin/bizcoin_simulation.go
```diff
@@ -0,0 +1,166 @@
...
+				dbg.Lvl1("Round", round, " FAILED:", err)
+			} else {
+				dbg.Lvl1("Round", round, " SUCCESS")
+			}
+		})
+
+		// Register when the protocol is finished (all the nodes have finished)
+		done := make(chan bool)
+		bz.RegisterOnDone(func() {
+			dbg.Print("SIMULATION ON DONE CALLED")
```
ineiti: `dbg.Print` is only to be used for direct debugging and then to be taken away. No allcaps, please.

And: this callback-function could definitively gain by being replaced by a channel.

protocols/bizcoin/bizcoin_simulation.go
```diff
@@ -0,0 +1,166 @@
...
+			done <- true
+		})
+		//bz.OnChallengeCommit(func() {
+		//rChallComm.NewMeasure("round_challenge_commit")
```
ineiti: Is this still used? Yes -> uncomment. No -> remove

protocols/bizcoin/blockchain/blkparser/blockchain.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+type Blockchain struct {
+	Path        string
+	Magic       [4]byte
+	CurrentFile *os.File
+	CurrentId   uint32
+}
+
+func NewBlockchain(path string, magic [4]byte) (blockchain *Blockchain, err error) {
```
ineiti: All functions in here could take some documentation.

protocols/bizcoin/bizcoin.go
nikkolasg: It's more to have a unified way of doing things. For the rest of the protocols we always used callbacks each time the protocol has finished (the OnDone). For clarity we just use the same mechanism. The fact that it's only used in server.go to pass that to a channel does not matter to bizcoin.go. One could use bizcoin.go in other ways.
Why we use a channel after is only for the simulation, it's easier to wait on a channel that a callback.

protocols/bizcoin/blockchain/blkparser/tx.go
```diff
@@ -0,0 +1,127 @@
...
+	Sequence  uint32
+}
+
+type TxOut struct {
+	Addr     string
+	Value    uint64
+	Pkscript []byte
+}
+
+func ParseTxs(txsraw []byte) (txs []*Tx, err error) {
```
ineiti: At least give short comments for exported functions.

protocols/bizcoin/bizcoin.go
nikkolasg: I think one would have to keep tracks of transactions already validated ? and verify if the hash has a leading numbers of '0' in front ? Seems more complicated than what's needed for the moment, but could be done, sure.
I don't know about the paper.

Liamsi:
> About BlocksDir: I would still like to have a publicly downloadable place where a blockchain (if it's what this is) can be downloaded. Just in case anybody not having access to Deterlab wants to use it.

Lefteris ran a bitcoin node and provided us with the files. But here is a link Lefteris gave me: 
http://bitcoin.stackexchange.com/questions/37927/what-happened-to-the-bootstrap-dat-provided-by-bitcoin-org

but he also said I might be faster (5min)  to start a full node

> In the best case, the bizcoin_simulation would handle the download of that file.

Given the above info this isn't sth. we should put to any effort to get that working.

protocols/bizcoin/bizcoin.go
nikkolasg: +1

protocols/bizcoin/blockchain/messg.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+type TransactionAnnouncment struct {
+	Val blkparser.Tx // Trasaction to be included in a block
+}
+
+// NOT: In order to decoe correctly the Proof, we need to the get the suite
+// somehow. We could just simply add it as a field and not (un)marhsal it
+// We'd just make sure that the suite is setup before unmarshaling.
+type BlockReply struct {
```
ineiti: So how is it done finally? 'We could' means: you didn't (at least not now). "We'd" means any of: we had, we should, or we would.

So where does the `SuiteStr` come from?

protocols/bizcoin/blockchain/messg.go
```diff
@@ -0,0 +1,132 @@
...
+	Response      abstract.Secret // Aggregate response
+	Challenge     abstract.Secret // Aggregate challenge
+	AggCommit     abstract.Point  // Aggregate commitment key
+	AggPublic     abstract.Point  // Aggregate public key (use for easy troubleshooting)
+	SignatureInfo []byte          // All other elements necessary
+}
+
+type BitCoSiMessage struct {
+	ReqNo SeqNo // Request sequence number
+	// ErrorReply *ErrorReply // Generic error reply to any request
```
ineiti: Is this still used?

protocols/bizcoin/blockchain/messg.go
```diff
@@ -0,0 +1,132 @@
...
+	Type MessageType
+	Treq *TransactionAnnouncment
+	Brep *BlockReply
+}
+
+func (sr *BlockReply) MarshalJSON() ([]byte, error) {
```
ineiti: Sorry, again: why is this marshalled to/from JSON?

protocols/bizcoin/blockchain/blkparser/blockchain.go
nikkolasg: done what I could but this is a third party library adapted by @lefkoko  so, I'm not the expert on it.

protocols/bizcoin/blockchain/blkparser/tx.go
nikkolasg: Not  a file we wrote, again third party library used by @lefkoko

protocols/bizcoin/blockchain/trblock.go
```diff
@@ -0,0 +1,106 @@
...
+func HashHeader(h *Header) string {
+	data := fmt.Sprintf("%v", h)
+	sha := sha256.New()
+	sha.Write([]byte(data))
+	hash := sha.Sum(nil)
+	return hex.EncodeToString(hash)
+}
+
+func (trb *TrBlock) Print() {
+	log.Println("Header:")
```
ineiti: Can this be changed to use `dbg` instead of `log`?

protocols/bizcoin/blockchain/messg.go
nikkolasg: copied files from lefteris old code. not even used i think.

protocols/bizcoin/blockchain/messg.go
nikkolasg: copied files from lefteris old code. not even used i think.

protocols/bizcoin/blockchain/messg.go
nikkolasg: copied files from lefteris old code. not even used i think.

protocols/bizcoin/blockchain/trblock.go
nikkolasg: copied files from lefteris old code. not even used i think.

ineiti: Then delete it.

ineiti:
Change all `dbg.Print`: either delete or make `dbg.Lvl2` or `dbg.Lvl3` or such.

ineiti:
Why is `ntree` not in a subdirectory?

protocols/bizcoin/blockchain/messg.go
ineiti: OK, delete it.

protocols/bizcoin/blockchain/messg.go
ineiti: delete it.

protocols/bizcoin/blockchain/messg.go
ineiti: Delete

protocols/bizcoin/blockchain/blkparser/tx.go
ineiti: Delete

nikkolasg:
>Why is ntree not in a subdirectory?
ntree uses bizcoin/server.go for the simulation. 
What about the inverse: why is pbft in a subpackage ? Everything here is related to consensus. Golang packages are themselves pretty huge and do not separate everything into subpackages the way java does for example.
Also if ntree is in a sub-package, then it would have to include bizcoin/, I find it rather strange. 
But if you think it's *cleaner*, then it's also a good reason for me, no problem.

ineiti:
OK, besides the nitpicks, I think the most important actions are:
* handle the blockchain correctly and automatically so that anybody who wants to re-do the simulation can do so without having to rely on Nicolas directory on Deterlab
* delete unused code (Ismail seems to be going at that already)

ineiti:
Oh yes, and rename `bizcoin` to `byzcoin`

protocols/bizcoin/blockchain/blkparser/tx.go
Liamsi: ParseTxs is used to parse the transactions  (from a .dat file to create a Block struct)

protocols/bizcoin/bizcoin.go
Liamsi: Me neither. We can make a more realistic simulation/standalone application. But this should definitely be a (possibly low-priority) issue/task.

lib/network/net.go
Liamsi: @ineiti : new issue? I guess this is important?

Liamsi:
I think it is ready to merge. Any objections?

ineiti:
As I said, please add some instructions on how to get everything needed to run the byzcoin with regard to the blockchain-downloading. If there is a .torrent-file needed, please add this to the git, too.

Is there a way to change the `BlocksDir` to something more relative?

Liamsi:
> As I said, please add some instructions on how to get everything needed to run the byzcoin with regard to the blockchain-downloading. If there is a .torrent-file needed, please add this to the git, too.
So I guess this isn't sufficient (hint in the error message):
https://github.com/dedis/cothority/commit/6891980d1daaf571bd18960bd9c8d2523fc69579
I can put a [torrent]() into the repo but then anyone who ones to re-run the experiment needs to download 21GB of data. The best approach would be to host the files somewhere. But where?

> Is there a way to change the BlocksDir to something more relative?

You can put a relative directory in BlocksDir (or what do you mean).

Issue #199, opened 2016-02-05
Add Simulation for JVSS


Liamsi:
Add a simulation struct for jvss (like in https://github.com/dedis/cothority/blob/sda_1/protocols/example/channels/simulation.go)

Liamsi:
Merged

PR #210, opened 2016-02-08
Simulation for jvss - PR

Liamsi:
Resolves issue #199 (and some minor changes to jvss.go and lib/sda/host.go)


nikkolasg:
We could additionally request the time it takes to setup the longterm key.
But since this is not the priority here, we should move on. The rest is fine for me, you even did some housecleaning, that's perfect,  I say OK to merge ;)

Liamsi:
I just re-ran the simulation and found a very strange behaviour: If I run the with `-debug 3` there are no measurements (not on deterlab and not on localhost). Do you know why this could happen? /cc @nikkolasg 
And independent from that the simulation only works *sometimes*. The more nodes you add the more probably it will fail (no joke). If it fails I see 
```
0: (    jvss.(*JVSSProtocol).waitForRequests: 270) - JVSS ( 3 ) Received signature request ( 7 ) with no secret generated :/
```
I suspect a race condition (but only because it sometimes works and sometimes it doesn't).

Liamsi:
Here is the output of `go test -race` in the directory `protocols/jvss`:
```
...
Found 10 data race(s)
```
What do you think? Merge it and open a new PR for the race conditions in jvss (and maybe other packages) or should we prioritise this? /cc: @Daeinar @nikkolasg

nikkolasg:
No need to prioritise this for the moment, we must focus on BizCoin. However, next time you run it please provide the full output, because some of the races are due to global variables so not necessarily related to JVSS.
But yes, we should definitely look into the race conditions stuff later, and for ALL our code (sda will be the first victim). I think that has its place into a separate PR.

Liamsi:
OK, thanks for your input. I can send you the output by mail (or just run `go test -race` in the directory `protocols/jvss`)

Liamsi:
I removed all relevant race conditions from `jvss.go` and `host.go` (I call those 'irrelevant' which only are about the debug library or in tests. We shouldn't really care about them). The simulation still fails sometimes.

Liamsi:
OK, please merge it. I opened another issue for fixing JVSS #212 (the simulation works fine)

Liamsi:
Done, and also fixed #212

PR #217, opened 2016-02-20
More debugging/trying to find the error:

nikkolasg:
Created the PR so we can discuss about it

### DIFF
--- a/lib/network/net.go
+++ b/lib/network/net.go
@@ -117,15 +117,13 @@ func (c *TcpConn) Remote() string {
 // the ApplicationMessage **decoded** and an error if something
 // wrong occured
 func (c *TcpConn) Receive(ctx context.Context) (NetworkMessage, error) {
+	fmt.Println("Receiving")
 	var am NetworkMessage
 	am.Constructors = c.host.constructors
 	var err error
 	//c.Conn.SetReadDeadline(time.Now().Add(timeOut))
 	// First read the size
 
-	// FIXME Read the size as an UvarInt
-	c.conMut.Lock()
-	defer c.conMut.Unlock()
 	s, err := binary.ReadUvarint(bufio.NewReader(c.Conn))
 	dbg.Print(c.Conn.LocalAddr(), "Receiving s =====================================", s)
 	if err != nil {
@@ -142,7 +140,6 @@ func (c *TcpConn) Receive(ctx context.Context) (NetworkMessage, error) {
 		// read the size of the next packet
 
 		n, err := c.Conn.Read(b)
-
 		if err != nil {
 			fmt.Println("Error while reading parts of the buffer", err)
 		}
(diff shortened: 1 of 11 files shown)


lib/network/net.go
```diff
@@ -185,15 +200,24 @@ func (c *TcpConn) Send(ctx context.Context, obj ProtocolMessage) error {
 	}
 	//c.Conn.SetWriteDeadline(time.Now().Add(timeOut))
 	// First write the size
-	packetSize := Size(len(b))
-	if err := binary.Write(c.Conn, globalOrder, packetSize); err != nil {
+	packetSize := uint64(len(b))
+
+	// FIXME Write the size as an UvarInt
+	var sb = make([]byte, binary.MaxVarintLen64)
+	n := binary.PutUvarint(sb[:], packetSize)
```
nikkolasg: That's not necessary, it does not change anything. if you look at the source of binary.Write ([code](https://golang.org/src/encoding/binary/binary.go?s=6397:6461#L224), it uses PutUVarint as the underlying function when you give it a uint64.

Liamsi: Yepp, saw that. It's even more strange, that it again behaves differently from before.

Liamsi: I did `git checkout sda_1 -- net.go`  and the same with struct.go and encoding.go to see if messed up sth: As expected it does not block but I now get a `networfatal error: bad map state` [sic] and `invalid memory address or nil pointer dereference` and much more. So at least our changes didn't make it worse...

Liamsi: > That's not necessary, it does not change anything. if you look at the source of binary.Write (code, it uses PutUVarint as the underlying function when you give it a uint64.
See the next commit

ineiti:
Is there a failing test-case? Ways to reproduce? Why is this in 'Ready4Merge'?

Liamsi:
> Is there a failing test-case? Ways to reproduce? Why is this in 'Ready4Merge'?

Run the pbft simulation with ~90 hosts and blocksize > 2000 on deterlab OR with even larger blocksize (and fewer machines) on localhost

Liamsi:
> Why is this in 'Ready4Merge'?

Only because it's the default stage. Nicolas and just needed a PR so that we can continue discussing on code changes/experiments during the weekend.

Liamsi:
Closed as this PR lead to some confusion in a sandstorm... The error persists though: using a broadcast protocol (ignores the tree) sending around large pieces of data (some MB), the Receive function cannot read the package size correctly. In worst case this means it reads more than it should (we read some blockchain data/preprepare  message together with other small messages sent later ). Before the package size received was even negative (smelled like an integer overflow), with our changes it now is just too big.this only happens with a larger number of hosts. It took us days and adding a lot of debug output (in all parts of the code, even in protobuf) to figure that out. 
I will open a new issue with a more precise description if we decide to postpone working in this bug.

PR #219, opened 2016-02-25
Network error

ineiti:
Tried to reproduce the error from PBFT - unfortunately this doesn't trigger it. But it is a good stress-test anyway, so I propose we merge it.


nikkolasg:
I'm also for pushing it but we must be careful since it will be run a Travis container, and I don't know the perf / limits of the containers.
* can it support 20*19/2 =190 connections ?
* can it support 190 MB in transit in the loopback interface ?
We'll have to see...

Liamsi:
I also think it's a good idea. Regarding the travis limits: We could give it a try on the travis image we have. Or we could temporarily change the travis config to only run this test (or we could fix the other errors first)

PR #221, opened 2016-02-25
Color nary tree

ineiti:
Correctly colors the nary-trees now


lib/sda/tree.go
```diff
@@ -285,12 +286,22 @@ func (il *EntityList) GenerateBigNaryTree(N, nodes int) *Tree {
...
 			}
 			parent.Children = make([]*TreeNode, children)
+			parentHost, _, _ := net.SplitHostPort(parent.Entity.Addresses[0])
 			for n := 0; n < children; n++ {
-				// TODO: this should check on the host-address
-				// of the Entity and not only the Id.
-				for il.List[elIndex].Id == parent.Entity.Id &&
+				// Check on host-address, so that no child is
+				// on the same host as the parent.
+				childHost, _, _ := net.SplitHostPort(il.List[elIndex].Addresses[0])
```
Liamsi: nice, some standard library function instead of our home-brew one's :-)

lib/sda/tree.go
```diff
@@ -285,12 +286,22 @@ func (il *EntityList) GenerateBigNaryTree(N, nodes int) *Tree {
...
+				elIndexFirst := elIndex
+				for childHost == parentHost &&
 					ilLen > 1 {
```
Liamsi: why the line break?

lib/sda/tree.go
```diff
@@ -285,12 +286,22 @@ func (il *EntityList) GenerateBigNaryTree(N, nodes int) *Tree {
...
 					elIndex = (elIndex + 1) % ilLen
+					// If we tried all hosts, it means we're using
+					// just one hostname, as we didn't find any
+					// other name
+					if elIndex == elIndexFirst {
+						break
+					}
+					childHost, _, _ = net.SplitHostPort(il.List[elIndex].Addresses[0])
```
Liamsi: make this a local var `net.SplitHostPort(il.List[elIndex].Addresses[0])` (used several times), why is `n` never used in this loop?

Liamsi: Ah OK I see. You change `elIndex` before by `elIndex = (elIndex + 1) % ilLen` so it's not the same value

Liamsi:
In my (currently tired) eyes, this is ready to merge.

ineiti:
OK, removed the line-break. Merge it if it seems OK.

lib/sda/tree_test.go
```diff
@@ -241,6 +242,21 @@ func TestBigNaryTree(t *testing.T) {
...
+func TestTreeIsColored(t *testing.T) {
+	dbg.TestOutput(testing.Verbose(), 4)
+	names := []string{"local1:1000", "local1:1001", "local2:1000", "local2:1001"}
+	peerList := genEntityList(tSuite, names)
+	tree := peerList.GenerateBigNaryTree(3, 13)
+	root := tree.Root
+	rootHost, _, _ := net.SplitHostPort(root.Entity.Addresses[0])
+	for _, child := range root.Children {
+		childHost, _, _ := net.SplitHostPort(child.Entity.Addresses[0])
+		if rootHost == childHost {
```
Liamsi: One more thing/question before merging: Shouldn't we test against the address *and* the entities ID? And why does the test only test the first level (only root and its children). Same with `TestBigNaryTree`.

ineiti: If the address is different, so will be the entity-ID.

If the first level is good, we can suppose that the other levels are good, too. Here we have two hosts `local1` and `local2`, so if the root is on `local1`, all it's children will be on `local2`.

ineiti: GenerateBigNaryTree tests on the Entity.IDs, which is not sufficient. This is why I added TestTreeIsColored. But perhaps the two tests should be merged?

Liamsi: Hmm, if the "color" is the the Entity.ID and the addresss the test that checks if the tree is colored correctly should indeed test for both. But I don't think that is very important.

PR #225, opened 2016-02-26
Reimplement count

ineiti:
Working around the timeout-issue by sending a packet directly to the root.


ineiti:
Also fails if 2047 hosts on deterlab with the infamous 'wrong wiretype'

ineiti:
Cross-branch mixup, sorry. Should compile and work now. How did you do the race-test? AFAIS, there is no race anymore with regard to the timeout. But I might be wrong...

PR #230, opened 2016-03-03
nice colored output and working localhost/deterlab

ineiti:
unfortunately fmt.Printf mixes up when run from different processes


Liamsi:
WIP again, tests fail

ineiti:
End-monster successfully defeated.

PR #231, opened 2016-03-03
made Deterlab work again

ineiti:
Added a non-waiting `host.Listen`-function so that starting 32767 nodes doesn't take a *very long* time.


Liamsi:
WIP: travis fails. Maybe related to #230

PR #232, opened 2016-03-03
Fix #218: Verification of Schnorr signatures works (and is tested)

### LINKED ISSUE #218: lib/crypto: need tests
Wrote a small subpackage crypto library (instead of everywhere writing our own SchnorrSig + SchnorrVerify). Did not wrote tests for the moments and stg might fail.

Liamsi:
changes: public key wasn't used for verification (now it is), some refactoring, added test


ineiti:
Looks good to me.

PR #234, opened 2016-03-04
Test localhost deterlab

ineiti:
Re-enabled testing of protocols using

```
cd simul
./do_tests.sh
```

Only `local_byzcoin.toml` fails with `Round 0 failed: Reconstructed challenge not equal to one given` - leaving this for next week.


ineiti:
@Liamsi - do you want to fix the 'local_byzcoin'-error first? I think this error should be done in the sda_1-branch, as it is not specific to this branch.

ineiti:
OK for me

Liamsi:
Loks good

Issue #209, opened 2016-02-08
CoSi protocol : useless field


nikkolasg:
In the message that CoSi is passing, there is the field "From" that is included. This is defenitely useless as we already know who is sending the message, so it should be deleted. (I honestly don't know what has passed through my mind to think we needed that .. ><)

PR #236, opened 2016-03-04
Remove obsolete field (#209)

Liamsi:
(Correct branch for PR)


ineiti:
```
$ ./simul runfiles/cosi_simple.toml
```

Yields an error, but I think this is due to the number of cosi-nodes. Should be fixed in #234

For me this is OK to merge

nikkolasg:
Cool. OK for me.

PR #241, opened 2016-03-08
Rename simulations

ineiti:
Renaming the simulations according to our discussion


ineiti:
Passes test, moving to Ready4Merge

PR #242, opened 2016-03-08
Ported naive and ntree to sda

Liamsi:


protocols/ntree/ntree.go
```diff
@@ -0,0 +1,130 @@
...
+func NewRootProtocol(msg []byte, node *sda.Node) (*Protocol, error) {
+	dbg.Print(msg)
+	dbg.Print(node)
+
+	p, err := NewProtocol(node)
+	if err != nil {
+		return nil, err
+	}
+	dbg.Print(p)
+	//p.message = make([]byte, len(msg))
```
ineiti: Delete?

Liamsi: It is work in progress ...

protocols/ntree/ntree.go
```diff
@@ -0,0 +1,130 @@
...
+		for _, c := range p.Children() {
+			err := p.SendTo(c, &msg.Message)
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		// If we're the leaf, start to reply
+		return p.SendTo(p.Parent(), &SignatureReply{Signatures: []crypto.SchnorrSig{p.signature}})
+	}
```
ineiti: Don't you have to call `Done` here?

protocols/ntree/ntree.go
```diff
@@ -0,0 +1,130 @@
...
+		for _, s := range reply {
+			aggSignatures = append(aggSignatures, s.Signatures...)
+		}
+		aggSignatures = append(aggSignatures, p.signature)
+		return p.SendTo(p.Parent(), &SignatureReply{Signatures: aggSignatures})
+	} else {
+		//
+		dbg.Lvl1("Leader got", len(reply), "signatures. Children:", len(p.Children()))
+	}
+	return nil
```
ineiti: Also call `Done` here?

protocols/ntree/ntree.go
Liamsi: But thanks for the hint

protocols/ntree/ntree.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	err = p.RegisterHandler(p.HandleSignReply)
+	if err != nil {
+		return nil, fmt.Errorf("Couldn't register handler %v", err)
+	}
+	return p, nil
+}
+
+func NewRootProtocol(msg []byte, node *sda.Node) (*Protocol, error) {
+	dbg.Print(msg)
```
ineiti: I still didn't understand why you have `newRootProtocol`. Should probably be obvious...

protocols/ntree/ntree.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	return nil
+}
+
+func (p *Protocol) HandleSignReply(reply []StructSignatureReply) error {
+	if !p.IsRoot() {
+		dbg.Lvl3("Appending our signature to the collected ones and send to parent")
+		aggSignatures := make([]crypto.SchnorrSig, len(p.Children())+1)
+		for _, s := range reply {
+			aggSignatures = append(aggSignatures, s.Signatures...)
```
ineiti: shouldn't this be
```
for i, s := range reply {
		aggSignatures[i] = s.Signatures
}
```

protocols/ntree/ntree.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+		aggSignatures = append(aggSignatures, p.signature)
```
ineiti:
```
aggSignatures[len(p.Children())] = p.signature
```

protocols/ntree/ntree.go
```diff
@@ -0,0 +1,130 @@
...
+		return p.SendTo(p.Parent(), &SignatureReply{Signatures: aggSignatures})
+	} else {
+		//
+		dbg.Lvl1("Leader got", len(reply), "signatures. Children:", len(p.Children()))
+	}
+	return nil
+}
+
+// ----- network messages that will be send around ------- //
+type Message struct {
```
ineiti: sen*t* around

protocols/ntree/simulation.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"github.com/BurntSushi/toml"
+	"github.com/dedis/cothority/lib/dbg"
+	"github.com/dedis/cothority/lib/monitor"
+	"github.com/dedis/cothority/lib/sda"
+)
+
+func init() {
+	sda.SimulationRegister("NaiveTreeSimulation", NewSimulation)
+	sda.ProtocolRegisterName("CosiNtree", func(node *sda.Node) (sda.ProtocolInstance, error) { return NewProtocol(node) })
```
ineiti: I propose we use `CoSi` throughout the code.

protocols/ntree/simulation.go
```diff
@@ -0,0 +1,68 @@
...
+			return err
+		}
+		proto, err := NewRootProtocol(msg, node)
+		if err != nil {
+			dbg.Error("Quitting the simuation....", err)
+			return err
+		}
+		node.SetProtocolInstance(proto)
+		err = proto.Start()
+		if err != nil {
```
ineiti: I prefer

```
node, err := config.Overlay.CreateNewNodeName("CoSiNtree", config.Tree)
// Error handling
err = node.ProtocolInstance.(Protocol).SetMsg( msg )
// Error handling
```

here and incorporate the `NewRootProtocol` as `SetMsg`

protocols/ntree/ntree.go
ineiti: OK, got it now - see my comment in the `simulation.go`

protocols/ntree/ntree.go
Liamsi: No: `aggSignatures` is an array of signatures *and* `s.Signatures` is an array. I will let you know when to review this by putting it to Ready4Merge. It is still WIP as you can see from the board.

ineiti:
sorry for the early report then...

Liamsi:
No problem. There were also some very good hints that I can incorporate earlier now ;-)

protocols/ntree/simulation.go
ineiti: Upper-case the 'S', so it's CoSi instead of Cosi.

protocols/ntree/simulation.go
Liamsi: Question on that: Doesn't `CreateNewNodeName` start the protocol (and some gorountines)? In that way (using a setter) I would need to use a lock etc, no? Or is this only a problem with channels?

protocols/ntree/ntree.go
```diff
@@ -0,0 +1,125 @@
...
+	if !p.IsRoot() {
+		dbg.Lvl3("Appending our signature to the collected ones and send to parent")
+		count := 0
+		for _, s := range reply {
+			count += len(s.Signatures)
+		}
+		aggSignatures := make([]crypto.SchnorrSig, count+1)
+		for i, sigs := range reply {
+			for j, sig := range sigs.Signatures {
+				aggSignatures[i+j] = sig
```
ineiti: won't work - needs a counter or something, else `i + j` will overwrite values as i increases by steps of 1.

protocols/ntree/simulation.go
ineiti: No, it doesn't start. `StartNewNodeName` calls `CreateNewNodeName` and the calls `Start` on the created node. So if you only call `CreateNewNodeName` you have the node, the ProtocolInstance, but no `Start` yet. This is why I don't like the `NewNodeEmptyName` and would like to get rid of it.

ineiti:
Root-node misses verification - can we merge this into sda_1 and continue on my `redo_cosi_graphs` branch?

Liamsi:
OK for me.

protocols/ntree/ntree.go
```diff
@@ -0,0 +1,132 @@
...
+func (p *Protocol) Start() error {
+	if p.IsRoot() {
+		dbg.Lvl3("Starting ntree/naive")
+		return p.HandleSignRequest(structMessage{p.TreeNode(), Message{p.message}})
+	} else {
+		return fmt.Errorf("Called Start() on non-root ProtocolInstance")
+	}
+}
+
+func (p *Protocol) HandleSignRequest(msg structMessage) error {
```
nikkolasg: About HandleSignRequest + HandleSignReply
* No need for them to be exported. handleSign* is enough I think.
* A small comment would be most welcomed. Same for handleSignReply.

protocols/ntree/ntree.go
```diff
@@ -0,0 +1,132 @@
...
+			childPub := sigs.Entity.Public
+			childSig := sigs.Signatures[0]
+			if err := crypto.VerifySchnorr(network.Suite, childPub, p.message, childSig); err != nil {
+				dbg.Error(err)
+			}
+			for _, sig := range sigs.Signatures {
+				aggSignatures[countReplies] = sig
+				countReplies++
+			}
+		}
```
nikkolasg: This is nitpicking surely but can't you do it in one line ?
```
// only setting the *capacity* to the right size, the *length* is 0
aggSignatures := make([]crypto.SchnorrSig,0,count+1)
aggSignatures[0] = p.signature
for _,sigs := range reply {
      /* verification */
      // the one-line part
      aggSignatures = append(aggSignatures,sigs.Signatures...)
}
```
Using `append` won't grow if it's not necessary, and since the *capacity* is already at the right size, then it's fine.

ineiti: off to read `append`

PR #250, opened 2016-03-09
Redo cosi graphs

ineiti:
Working on having new graphs for the paper


protocols/ntree/ntree.go
```diff
@@ -74,6 +74,15 @@ func (p *Protocol) HandleSignRequest(msg structMessage) error {
 }
 
 func (p *Protocol) HandleSignReply(reply []structSignatureReply) error {
```
nikkolasg: About HandleSignRequest + HandleSignReply
* No need for them to be exported. handleSign* is enough I think.
* A small comment would be most welcomed. Same for handleSignReply.

protocols/ntree/ntree.go
```diff
@@ -84,11 +93,6 @@ func (p *Protocol) HandleSignReply(reply []structSignatureReply) error {
 		aggSignatures[0] = p.signature
 		countReplies := 1
 		for _, sigs := range reply {
-			childPub := sigs.Entity.Public
-			childSig := sigs.Signatures[0]
-			if err := crypto.VerifySchnorr(network.Suite, childPub, p.message, childSig); err != nil {
-				dbg.Error(err)
-			}
 			for _, sig := range sigs.Signatures {
```
nikkolasg: This is nitpicking surely but can't you do it in one line ?
```
// only setting the *capacity* to the right size, the *length* is 0
aggSignatures := make([]crypto.SchnorrSig,0,count+1)
aggSignatures[0] = p.signature
for _,sigs := range reply {
      /* verification */
      // the one-line part
      aggSignatures = append(aggSignatures,sigs.Signatures...)
}
```
Using `append` won't grow if it's not necessary, and since the *capacity* is already at the right size, then it's fine.

ineiti: +1

ineiti:
Making new PR against development

Issue #252, opened 2016-03-09
MonitorPort is not propagated to platforms


ineiti:
the `-mport`-flag is rendered unusable by the fact that the chosen port is not passed to the platforms. This happened sometimes during refactoring the `monitor`-package...

PR #253, opened 2016-03-09
MonitorPort is not propagated to platforms

ineiti:
Implements #252 again.


nikkolasg:
Ok for me. Is is tested ? With two experiments running at the same time ?

ineiti:
Thanks for helping me test - it runs now. Waiting for Travis.

nikkolasg:
Ok for me.

PR #255, opened 2016-03-09
Cosi : verifcation at level of tree

nikkolasg:
Nothing more to add that the title does not already say... ;)

### DIFF
--- a/lib/sda/tree.go
+++ b/lib/sda/tree.go
@@ -37,11 +37,15 @@ var TreeType = network.RegisterMessageType(Tree{})
 // also generates the id.
 func NewTree(il *EntityList, r *TreeNode) *Tree {
 	url := network.UuidURL + "tree/" + il.Id.String() + r.Id.String()
-	return &Tree{
+	t := &Tree{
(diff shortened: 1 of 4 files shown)


lib/sda/tree.go
```diff
@@ -403,9 +428,14 @@ type TreeNode struct {
...
-	Entity   *network.Entity
-	Parent   *TreeNode
+	Entity *network.Entity
+	// Parent link
+	Parent *TreeNode
+	// Children links
 	Children []*TreeNode
+	// Aggregate public key for *this* subtree,i.e. this node's public key + the
+	// aggregate of all its children's aggregate public key
+	PublicAggregateSubTree abstract.Point
```
ineiti: Just to be sure: `CoSi` verifies against the aggregated public key in the same way, that is, all children PLUS the current host?

ineiti:
Tried it in my measurements and performs OK. Good for me.

lib/sda/tree.go
nikkolasg: Yes

PR #257, opened 2016-03-11
Redo cosi graphs

ineiti:
PR against development


simul/matplotlib/Paper_cosi2/plotit.py
```diff
@@ -0,0 +1,219 @@
...
+    return data_label
+
+
+# Plots a Cothority and a JVSS run with regard to their averages. Supposes that
+# the last two values from JVSS are off-grid and writes them with arrows
+# directly on the plot
+def plotRoundtime():
+    plotData([['jvss', 'naive_cosi', 'ntree_cosi', 'cosi_depth_3'],
+              ['JVSS', 'Naive', 'NTree', 'CoSi']], 'comparison_roundtime',
+             xticks=[4, 8, 16, 32, 64, 128, 256, 512, 1024, 4096, 16384, 65536],
```
Liamsi: Is there a reason that there is only the jvss.csv pushed to the branch? Can't find the others (calling the python script will fail without redoing all the simulations...)

lib/dbg/debug_lvl.go
```diff
@@ -272,3 +271,40 @@ func LLvlf2(f string, args ...interface{}) { Lvlf(-1, f, args...) }
...
 func LLvlf5(f string, args ...interface{}) { Lvlf(-1, f, args...) }
+
+// SetDebugVisible set the global debug output level in a go-rountine-safe way
+func SetDebugVisible(lvl int) {
+	debugMut.Lock()
+	defer debugMut.Unlock()
+	debugVisible = lvl
+}
+
+func DebugVisible() int {
```
nikkolasg: Sorry but since we started on this go-lint / travis stuff, we need at least a one-line comment on each of the following functions.. ! (DebugVisible, SetShowTime,DebugTime, SetUseColors, UseColors)

protocols/byzcoin/byzcoin.go
```diff
@@ -172,7 +172,7 @@ func NewByzCoinRootProtocol(n *sda.Node, transactions []blkparser.Tx, timeOutMs
 	if err != nil {
 		return nil, err
 	}
-	bz.tempBlock, err = getBlock(transactions, bz.lastBlock, bz.lastKeyBlock)
+	bz.tempBlock, err = GetBlock(transactions, bz.lastBlock, bz.lastKeyBlock)
```
nikkolasg: Nitpicking: Is GetBlock not in the same package ? Can't it be unexported then ?

dist/dist.go
```diff
@@ -0,0 +1,10 @@
+/*
+Distribution the cothority-program in binary form. At least the
+following two distributions should be available:
+
+	* .tar.gz
+	* docker
+
+Others may or may not follow.
+*/
+package dist
```
Liamsi: Looks like you mixed up different branches here, or not? I would delete this file (currently doesn't contain anything useful)

simul/cothority/cothority.go
```diff
@@ -156,11 +153,7 @@ func main() {
 		allClosed <- true
 	}()
 	dbg.Lvl3(hostAddress, scs[0].Host.Entity.First(), "is waiting for all hosts to close")
-	select {
-	case <-allClosed:
-		dbg.Lvl2(hostAddress, ": all hosts closed")
-	case <-time.After(time.Second * time.Duration(scs[0].GetCloseWait())):
-		dbg.Lvl2(hostAddress, ": didn't close after", scs[0].GetCloseWait(), " seconds")
-	}
+	<-allClosed
```
nikkolasg: So are we sure it will never block then ? Why removing the timeout ?

dist/dist.go
ineiti: Ach so was. I hesitate to delete it - it's kind of the future distribution of cothority. I deleted it before but put it back. If you think it's not worth it, I'll delete it again ;)

simul/cothority/cothority.go
ineiti: Because the `simul.go` has a timeout, and I didn't want to multiply timeouts. If the host doesn't close, `simul.go` will do a `Cleanup` of the platform. This way we only have two timeouts, as seen in `simul/runfiles/README.md`. If we add a timeout here, there would be overlapping timeouts.

simul/matplotlib/stats.py
```diff
@@ -6,7 +6,8 @@
 
 # Our CSVs have a space after the comma, so we need a new 'dialect', here
 # called 'deploy'
-csv.register_dialect('deploy', delimiter=',', doublequote=False, quotechar='', lineterminator='\n', escapechar='',
+csv.register_dialect('deploy', delimiter=',', doublequote=False, quotechar='',
+                     lineterminator='\n', escapechar='',
                      quoting=csv.QUOTE_NONE, skipinitialspace=True)
```
nikkolasg: For all the python scripts, I guess we can make an exception out of rule of comments and stuff ;)

simul/matplotlib/Paper_cosi2/plotit.py
ineiti: Pushed everything now.

simul/simul.go
```diff
@@ -260,3 +304,32 @@ func getStartStop(rcs int) (int, int) {
 	dbg.Lvl2("Range is", start, ":", stop)
 	return start, stop
 }
+
+// getRunWait returns either the command-line value or the value from the runconfig
+// file
+func getRunWait(rc platform.RunConfig) int {
```
nikkolasg: I'm OK with that for the moment, but at some point later we should start extending this `RunConfig` struct. Having something like:
```
// Returns the value of the field designated by `name` or the `default` value if the fields does not exists.
func (rc *RunConfig) GetInt(name string, default int) int {
    ...
}

... 
rc.GetInt("runwait",100)
````

The same for `string`, `float64` etc

simul/cothority/cothority.go
nikkolasg: Great !

simul/simul.go
ineiti: OK - here I also 'abuse' it to set a default variable. In the case of using `rc.GetInt`, we'd have to define the default-values at another place, too.

nikkolasg: ? You just pass the default values into the `GetInt` function ? 
You go from `getRunWait(rc)` to `rc.GetInt("runwait", runwait)` 
Writing an issue for it.

ineiti: True, here I use it only once. Want me to include that in this branch?

nikkolasg: No need. I've submitted an issue.

simul/matplotlib/stats.py
ineiti: I put some comments in plot.py and mplot.py as I hope other people can use them, too.

ineiti:
Passes tests locally.

Liamsi:
There are way too many changes not related at all to the plot library. This PR touches almost every corner of the code base. 
For a good review this would quite some time. How do you both think we should handle this? Full-fledged ,time consuming review, or merge and one big review before development gets merged into master?

In general I propose to keep changes made in the branch related to the PR (except of small clean-ups, comments, fixing typos or smaller bugs). Otherwise reviews will take too long and or aren't done properly.

ineiti:
So shall I split it up into smaller PRs?

ineiti:
Split up into separate commits

Issue #260, opened 2016-03-11
Checking-flags


ineiti:
We need a 'checking'-flag for CoSi and its baselines:
- CoSi - if checking (in .toml) is 
  - 0, don't check
  - 1 check at root
  - 2 (default) check everywhere
- Ntree
  - 0, don't check
  - 1, check only direct children
  - 2 (default), check the whole sub-tree
- JVSS
  - 0, don't check
  - 1 (default), root checks all signatures

Liamsi:
Done in PR #264

PR #264, opened 2016-03-14
Check flag on simulation

nikkolasg:
Refers to https://github.com/dedis/cothority/issues/260


protocols/cosi/packets.go
```diff
@@ -15,6 +15,17 @@ import (
...
+	// XXX XXX UGLY HACK XXX XXX
+	// Is to be removed during the next versions of SDA. Need that for the
+	// moment because no way to have external configuration for the intermediate
+	// and leaf nodes.
+	// see https://github.com/dedis/cothority/issues/260
+	// 0 - no check at all
+	// 1 - check only at root
+	// 2 - check at each level of the tree
+	VerifyResponse int
+
```
ineiti: Instead of giving it as a parameter in the message, you can also use a setter and call that setter in https://github.com/dedis/cothority/blob/master/lib/sda/simulation.go#L43 - which will be somewhere in https://github.com/dedis/cothority/blob/check/protocols/cosi/simulation.go#L83 - if you do that, be sure to call

```
cs.SimulationBFTree.Node( config )
```

before, so that the tree and the entityList are set up.

nikkolasg: Please, make an example because I don't see how can I access each node in this function.
Moreover, when we will attack the external api issue, a lot of things will be able to pass through that.
Anyway, it's only for simulation anyway, all theses protocols should *not* be used as something else...!

ineiti: Here you go for an example:

https://github.com/dedis/cothority/pull/273

lib/sda/tree_test.go
```diff
@@ -196,6 +197,33 @@ func TestBinaryTree(t *testing.T) {
...
+func TestTreeNodeEntityIndex(t *testing.T) {
+	defer dbg.AfterTest(t)
+	dbg.TestOutput(testing.Verbose(), 4)
+	names := genLocalhostPeerNames(13, 2000)
+	peerList := genEntityList(tSuite, names)
+	tree := peerList.GenerateNaryTree(3)
+
+	ln := tree.ListNodes()
+	randomNode := ln[rand.Intn(len(ln))]
+	var idx int
```
Liamsi: Nice, some serious testing :-)

protocols/cosi/packets.go
```diff
@@ -11,6 +11,12 @@ import (
 
 // The main messages used by CoSi
 
+// see https://github.com/dedis/cothority/issues/260
+// 0 - no check at all
+// 1 - check only at root
+// 2 - check at each level of the tree
+var VerifyResponse = 1
```
Liamsi: make this `const`

protocols/cosi/simulation.go
```diff
@@ -19,14 +19,19 @@ func init() {
 
 type CoSiSimulation struct {
 	sda.SimulationBFTree
+
+	// See https://github.com/dedis/cothority/issues/260
```
Liamsi: Briefly describe `Checking` (not only by a link)

protocols/cosi/packets.go
nikkolasg: this is not a const. 
Linus made this change here:
https://github.com/dedis/cothority/pull/264/commits/a04ab0a85f14fd67c1884ef03c63cf3ff4e9f67d
so it's not passed as a message. Anyway it's for the simulation so..

Liamsi: Oh, I see this isn't a constant. There is not other way to handle this then by a global (exported) variable?

protocols/cosi/simulation.go
Liamsi: Same in other protocols.

protocols/jvss/simulation.go
```diff
@@ -59,15 +60,17 @@ func (jv *JvssSimulation) Run(config *sda.SimulationConfig) error {
...
-			dbg.Error("Got invalid signature")
-			return err
+		if jv.Checking == 1 {
+			err = proto.Verify(msg, sig)
+			if err != nil {
+				dbg.Error("Got invalid signature")
+				return err
+			}
+			dbg.Lvl3("Signature is OK")
 		}
```
Liamsi: Checking == 2 can't happen in JVSS? If yes, comment.

protocols/ntree/ntree.go
```diff
@@ -109,16 +129,47 @@ func (p *Protocol) SetMessage(msg []byte) {
...
+//   signature
 type SignatureReply struct {
-	Signatures []crypto.SchnorrSig
+	Sig   crypto.SchnorrSig
+	Index int
+}
+
+// SignatureBundle represent the signature that one children will pass up to its
+// parent. It contains:
+//  * The signature reply of this children (sig + index)
```
Liamsi: Not clear what this means. What is the "signature reply of this children"?

protocols/ntree/ntree.go
```diff
@@ -109,16 +129,47 @@ func (p *Protocol) SetMessage(msg []byte) {
...
+//  * The whole set of signature reply of its sub tree
+type SignatureBundle struct {
+	OwnSig  *SignatureReply
+	SubSigs []*SignatureReply
```
Liamsi: Add a comment for both (and possibly rename `OwnSig`)

protocols/cosi/packets.go
nikkolasg: Yes there was the previous way of sending the parameter in the message. Both are strange but the "external api" issue is not yet discussed so .. it will change once we move onto that subject.

protocols/jvss/simulation.go
nikkolasg: No it can't happen.

protocols/cosi/packets.go
Liamsi: OK

protocols/cosi/simulation.go
Liamsi: I think the code should be self-contained (also documentation-wise). But as this is only in the simulation code, I don't really care that much.

ineiti:
+1 for merge

Liamsi:
Not sure if this related to changes made here, but running the simulation with debug level 3 shows quite some race conditions.

ineiti:
Thanks for catching that - has to be fixed. What simulation were you running?

Liamsi:
`test_jvss.toml`

nikkolasg:
Race conditions on CoSi is removed from another PR now merged into development. 
Race conditions for JVSS is to be examined, and there is already an issue for refactoring JVSS. However, I'd say this is not high priority right now ..

Liamsi:
> Race conditions for JVSS is to be examined, and there is already an issue for refactoring JVSS. However, I'd say this is not high priority right now ..

Agreed.

Issue #261, opened 2016-03-13
Localhost simulation build instructions do not work as documented


grempe:
When following the instructions in the README things don't work as expected:

```
$ go get ./...
$ cd simul
$ go build
$ ./simul runfiles/sign_single.toml
```

It needs to be mentioned (for people who may not use Go everyday) that the repository needs to first be moved into the $GOPATH.

e.g. in my case I needed to clone the repo to:

```
~/go/src/github.com/cothority
```

This allowed the `go get ./...` and `go build` instructions to complete successfully.

However the final command references a file `runfiles/sign_single.toml` which no longer exists.  Picking other files from the `runtime` dir seems to work.  However it is not clear what one is accomplishing by running the simulations other than exercising tests on the code.  What is needed is higher level docs that describe the architecture, and how one would go about using any of this code in a real situation.

Liamsi:
Thanks a lot for your feedback. Indeed, the file `sign_single.toml` doesn't exist anymore. I already changed this. Thanks for pointing out the other documentation shortcomings. Will be fixed (very) soon.

ineiti:
Our goal is to move from simulation/research code to application-code during the next few weeks. As such I would very much appreciate if you could give us your use-case on how you would like to use collective signing:

- what application
- what language
- what you would like to have signed

Liamsi:
> It needs to be mentioned (for people who may not use Go everyday) that the repository needs to first be moved into the $GOPATH.

I changed the instructions to be less ambiguous (`go get` clones the directory directly in the GOPATH). 
@ineiti @nikkolasg please review and merge: #265; best see how it looks here: https://github.com/dedis/cothority/tree/readme_started_rewrite) 

We really need to create a standalone application (similar to the one that existed before the SDA merge) as soon as possible and add a description on how-to run, deploy it etc. Similar to what @bford described recently.

@grempe If you have any further questions, feedback please let us know. Thanks again.

ineiti:
Yes, we can discuss about how 'prototype' one once the checking is tested to run.

ineiti:
Closed by #265

PR #265, opened 2016-03-14
Rewrite/update readme

Liamsi:


Deterlab.md
```diff
@@ -0,0 +1,49 @@
...
+Afterwards you only need to unlock your SSH-agent / keychain once (per 
+session) and can access all your stored keys without typing the 
+passphrase each time.
+
+**OSX:**
+
+You can store your SSH-key directly in the OSX-keychain by executing:
+
+```
+$ /usr/bin/ssh-add -K ~/.ssh/<your private ssh key>
```
ineiti: I thought that \`\`\` already implies this is a command, so is it OK to add the '$' here?

Liamsi: Yes, ```  implies a code block. It doesn't really say which code though.  '$' is often used to describe [a shell prompt](https://en.wikibooks.org/wiki/Linux_For_Newbies/Command_Line#Shell_Prompt). I can remove it as it is quite clear from  the context. There isn't really right or wrong here.

ineiti: Just to avoid the "I tried to run `$ /usr/bin` and it didn't work" comment...

ineiti:
Does that include #261?

Liamsi:
Yes, i've updated the (missing file) and it should work out of the box now (with golang properly installed). Bit I didn't test it yet.

Deterlab.md
Liamsi: done

Deterlab.md
```diff
@@ -0,0 +1,49 @@
...
+simpler when deploying to DeterLab:
+
+* `-nobuild`: don't build any of the helpers which is useful if you're 
+working on the main code
+* `-build "helper1,helper2"`: only build the helpers, separated by a 
+",", which speeds up recompiling
+* `-range start:end`: runs only the simulation-lines including `start` 
+and `end`. 
+Counts from 0, start and end can be omitted and represent beginning and 
+end of lines, respectively.
```
ineiti: *beginning and end of lines* -> *first and last line of simulation*

Deterlab.md
ineiti: Heh - you changed it in the main README.md, but not in this README.md...

ineiti:
+1 for merge

PR #266, opened 2016-03-14
added CreateNewNode and changed test in lib/cosi

ineiti:
Finally adding `CreateNewNode` in `local`, so we don't have to register a new protocol just to pass some parameters to the protocol.

Especially @nikkolasg - what do you think?


protocols/cosi/cosi.go
```diff
@@ -86,14 +90,6 @@ func NewProtocolCosi(node *sda.Node) (*ProtocolCosi, error) {
 	return pc, err
 }
 
-// NewRootProtocolCosi is used by the root to collectively sign this message
-// (vanilla version of the protocol where no contributions are done)
-func NewRootProtocolCosi(msg []byte, node *sda.Node) (*ProtocolCosi, error) {
-	pc, err := NewProtocolCosi(node)
-	pc.message = msg
-	return pc, err
-}
```
nikkolasg: I am not so sure about removing that: Creating a RootProtocolCoSi means that this node is the *leader*, so it, and only it, knows about the message to sign yet. In the test you give it to every node on the creation, it feels wrong since the message is actually to be passed in the first message, why would you want to cut that out ?

nikkolasg: Ok got it.

nikkolasg:
Ok for me to Merge.

Liamsi:
Did you run the simulation with the race detector on? Even if it's only the root who gets the message: making it an exported, makes it easier to use it in a wrong way (but maybe this is for later).

Liamsi:
Also OK

Liamsi:
Not sure how we want to handle this in general: `./simul  -race -debug 3  -platform localhost runfiles/test_cosi.toml` also shows some race conditions.

ineiti:
That's a bug - put it back into WIP. But after merging with `development`, it is running without complaining. So keep it in R4M if Travis passes.

Liamsi:
> That's a bug - put it back into WIP. But after merging with development, it is running without complaining. So keep it in R4M if Travis passes.

If I try to run it again it's even worse:
```
E : (                            main.RunTest: 165) - Didn't find simulation CoSiSimulation
F : (                           main.RunTests: 128) - error running test: Didn't find simulation CoSiSimulation
```
Are you sure you ran the same simulation?

ineiti:
Playing ping-pong: Are you sure you pulled the branch? I tried it in two users, and the following line works fine:

```
go build && ./simul -race -debug 3 runfiles/test_cosi.toml 
```

Liamsi:
You are right. Must have been on different branch.

ineiti:
So you put it back into R4M?

ineiti:
That's another solution ;)

PR #273, opened 2016-03-16
Check with simulation node

ineiti:
Just as an example of how to use `Simulation.Node` - here I use a global variable in `lib/cosi` for the verification setting, which is of course not the very best way to do. But `Simulation.Node` is called before the Node is started. Should perhaps be renamed to `Simulation.Host` or `Simulation.Overlay`.


nikkolasg:
Ok I am ok to merge this. Again as discussed, I think things will change when we attack the external api issue.
+1

ineiti:
That's your part of the code, so if you're OK with it, just merge it. I PRd against your branch.

Issue #269, opened 2016-03-15
Lint our code


ineiti:
Make sure all code is run through `golint` in Travis. Discuss what options we want to fail on.

Liamsi:
I looked into the golint code: Unfortunately, you cannot manually set what to lint and what not. Furthermore the API for linting consits of unexported functions: https://github.com/golang/lint/blob/master/lint.go#L172-L190

One easy way to still have linting, would be to fork the linter and only enable the lints we need.

Liamsi:
Another way would be to use `grep` as a kind of white list (probably easier) I will see if this (still) returns too many false positives.

nikkolasg:
"In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process." [1](https://github.com/golang/lint)
The day where golint proposes us something bad, what do we do ?
do you think we are self-disciplined enough to do it ourself when coding AND reviewing a PR ?

Liamsi:
> do you think we are self-disciplined enough to do it ourself when coding AND reviewing a PR ?

I thought so. But we have so many undocumented exported functions [1] even in interfaces that have been merged during code reviews. To me it looks like some automated help can't harm. Another way to achieve at least documented exported functions, would be to turn on golint in your editor (that way you decide what to ignore and what not; instead of letting `grep` decide)

> "In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process." 

I know. But we could still use it for one particular lint: Reminding you of uncommented exported functions. It doesn't mean, we do not need to review and actually read the comments. An automatic tool might beat the best human go-player but it still can't tell if the existing comment makes sense ;-) 

[1] Currently, there seem to be around 261 uncommented, exported functions (`golint ./...| grep " should have comment or be unexported" | wc -l`)

ineiti:
I do like the idea of having a whitelist of things to grep and then do something like

```
if lint | grep -q (one|two|three); then
  echo Lint failed
  exit 1
fi
```

Liamsi:
OK. I will add something like this to travis. Please make sure to review all comments then. If we start writing comments etc., just to silence the linter, this would be counter productive (in my opinion, no comment is better/more honest than a bad, misleading comment for instance)

PR #274, opened 2016-03-16
Added linting to travis

Liamsi:


Liamsi:
OK, this should implement #269 (similar to what was asked for in https://github.com/dedis/cothority/issues/269#issuecomment-197251933).
Currently, the linter fails on two kinds of checks. Here are two examples:
```
protocols/cosi/packets.go:15:6: type name will be used as cosi.CosiAnnouncement by other packages, and that stutters; consider calling this Announcement 
```
and
```
simul/platform/localhost.go:107:1: exported method Localhost.Deploy should have comment or be unexported
```

Liamsi:
I think the two lints are the only reasonable. Do you all agree @nikkolasg @ineiti @Daeinar ?
Anyways, it is easy to add remove or disable the linter.

Liamsi:
I don't know if it's clever to merge it (as travis will fail). Better would be to comment out the linter script before, and to open an issue which we *all* work on after the more urgent tasks are done (everyone fixes the failing checks of the code he created).

ineiti:
There seem to be two errors - or are they related? Lint fails (of course), but the go-test fails, too...

Besides that I propose we comment out the Travis-inclusion but let the script inside, merge and then we work on documenting stuff.

nikkolasg:
Agree with @Liamsi and @ineiti

Liamsi:
> There seem to be two errors - or are they related? Lint fails (of course), but the go-test fails, too...

I saw that, too. I thought the problem is that the lint-script fails and hence the other parts of the `install:` section aren't done correctly. I will experiment with that and comment out the script if I know exactly if the failing tests are related.

Liamsi:
Moving the installation of golint to `before_install:` helps with the tests (otherwise it overwrites the default `install` routine.

Liamsi:
> There seem to be two errors - or are they related? Lint fails (of course), but the go-test fails, too...

Fixed (see [last travis build](https://travis-ci.org/dedis/cothority/builds/116418580)). I commented out the linter in travis now.

ineiti:
OK, cool. Did you open the issue `Everybody get his code documented`?

Merge +1

Liamsi:
> Did you open the issue Everybody get his code documented?

#275
I merged this myself (as it basically doesn't change any relevant parts of the code and the relevant part in travis.yml is commented)

PR #289, opened 2016-03-17
Asynchronous message handling

nikkolasg:
Finally !
The incoming network messages are now buffered in the node and then dispatched as soon as possible to the protocol instance of the node.
Some might say "why don't just use buffered channel", well, you might be right BUT:
* With this approach we don't allocate as more than we need, and we can actually control our allocation if let's say there was a burst at some point, allocated a huge amount of memory, then it went back to normal, we *could* reduce the amount of memory by just taking a smaller chunk of the buffer and let the GC doing its job.
* It's inspired from [here](http://marcio.io/2015/07/handling-1-million-requests-per-minute-with-golang/). I kinda like their approach and took a simplified version. But in the long run, for all the messages that we are going to have (external api + tree/list handling + *many* protocol instances etc), it might make sense to implement something like this at a more lower level, at host.go with many *workers* doing the dispatching. So I see it as a training!
* I had fun doing it even if there was one ***** stupid bug.

If at any point you feel this is **overkill**, please **say so**! I am also OK to use a buffered channel and checking the `len(channel)` each time a new message arrives. Actually I did not know about the fact that we can query the actual *length* of a buffered channel before the jogging of this afternoon... :)

### DIFF
--- a/lib/sda/node.go
+++ b/lib/sda/node.go
@@ -35,6 +35,9 @@ type Node struct {
 	msgQueue map[uuid.UUID][]*SDAData
 	// done callback
 	onDoneCallback func() bool
+	// dispatcher of the messages so overlay can give message to node without
+	// blocking even if the overlaying protocol is still blocking.
+	dispatcher *dispatcher
 }
 
 // AggregateMessages (if set) tells to aggregate messages from all children
@@ -66,6 +69,9 @@ func NewNodeEmpty(o *Overlay, tok *Token) (*Node, error) {
 		messageTypeFlags: make(map[uuid.UUID]uint32),
 		treeNode:         nil,
 	}
+	// creates and runs the dispatcher
+	n.dispatcher = newDispatcher(n.dispatchMsgToProtocol)
+	go n.dispatcher.run()
 	var err error
 	n.treeNode, err = n.overlay.TreeNodeFromToken(n.token)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


ineiti:
+1 for overkill ;) I have my own solution in #290 - a simple queueReader with a queueWriter does the same.

I copied your test, but then it passed even without me changing anything in the `node.go`... So I did change it a bit.

Liamsi:
> +1 for overkill ;) I have my own solution in #290 - a simple queueReader with a queueWriter does the same.

I though you read in the bus ;-) @ineiti's solution looks indeed simpler (easier to understand -> less error prone).

lib/sda/node.go
```diff
@@ -347,9 +359,15 @@ func (n *Node) DispatchMsg(sdaMsg *SDAData) error {
...
 	default:
-		return errors.New("This message-type is not handled by this protocol")
+		err = errors.New("This message-type is not handled by this protocol")
+	}
+	// XXX Should at some point define a central way / point to handle errors
+	// ... ?
+	// Before the error was just going up to the sda.Host and nothing else was
+	// taken into account.
+	if err != nil {
+		dbg.Error(err)
```
Liamsi: At least print it out (silently omitting it, can create boring, hence seemingly endless debugging sessions). Pass it up if you think there are reasons why the caller should handle it.

Liamsi:
If you both can't agree on what approach is better: Let's benchmark both ;-)

ineiti:
Whatever approach we take, there is a lot of work to do afterwards. I spent most of yesterday's evening debugging all concurrency issues now that we can have multiple messages waiting. Locks are now all over the place.

Liamsi:
Let @nikkolasg fix the test. I'll write a benchmark for both and let's discuss and decide afterwards.

nikkolasg:
No need to benchmark @Liamsi it's fine let's go with the simpler approach for the moment ;)
However I'll comment on #290 there's some deadlock issues.

nikkolasg:
Droped in favor of #290

Issue #160, opened 2016-01-13
Insufficient error handling


Liamsi:
In some src files in the deploy package there is very limited error handling: 
Function which return `error`s are often called without handling the potentially occurring error. This lead for instance to #17.

ineiti:
Could be a quickfix

Liamsi:
> Could be a quickfix 
Only if we care about the file deploy.go. There seem to be a lot of places with no, or not sufficient error handling. It might be more reasonable to wait for the sda to be finished and than project-wide search for all function calls which might return errors but where the error handling is ignored.

Liamsi:
Still way to many errors ignored. It's OK to ignore errors in test code but not in production cod, as it makes it obviously very hard to find the error (and we already spent hours searching for bugs, which would have been obvious, if the error was at least logged). 
You can hunt for missing error handling using: https://github.com/kisielk/errcheck 
(run `errcheck ./... | grep -v "_test.go"` to exclude test files; currently yields 125 lines)

Liamsi:
I will take care of this.

Liamsi:
Will work on this tomorrow. Or does it make sense to wait for the services-sda-rewrite?

nikkolasg:
closed with #347

PR #290, opened 2016-03-17
Async 2

ineiti:
More lightweight implementation of the asynchronous message-passing into `Node`.

### DIFF
--- a/lib/sda/host_msg.go
+++ b/lib/sda/host_msg.go
@@ -45,9 +45,12 @@ type Token struct {
 	cacheMutex   sync.Mutex
 }
 
+var tokenMutex sync.Mutex
+
 // Returns the Id of a token so we can put that in a map easily
 func (t *Token) Id() uuid.UUID {
-	t.cacheMutex.Lock()
+	tokenMutex.Lock()
+	defer tokenMutex.Unlock()
 	if t.cacheId == uuid.Nil {
 		url := network.UuidURL + "token/" + t.EntityListID.String() +
 			t.RoundID.String() + t.ProtocolID.String() + t.TreeID.String() +
@@ -55,16 +58,14 @@ func (t *Token) Id() uuid.UUID {
 		t.cacheId = uuid.NewV5(uuid.NamespaceURL, url)
 	}
 	id := t.cacheId
-	t.cacheMutex.Unlock()
 	return id
 }
 
 // Return a new Token contianing a reference to the given TreeNode
 func (t *Token) ChangeTreeNodeID(newid uuid.UUID) *Token {
-	t.cacheMutex.Lock()
-	defer t.cacheMutex.Unlock()
+	tokenMutex.Lock()
+	defer tokenMutex.Unlock()
 	t_other := *t
-	defer t_other.cacheMutex.Unlock()
 	t_other.TreeNodeID = newid
 	t_other.cacheId = uuid.Nil
 	return &t_other
--- a/lib/sda/host_msg.go
+++ b/lib/sda/host_msg.go
@@ -44,6 +44,8 @@ type Token struct {
 	cacheId      uuid.UUID
 }
 
+// Global mutex when we're working on Tokens. Needed because we
+// copy Tokens in ChangeTreeNodeID.
 var tokenMutex sync.Mutex
 
 // Returns the Id of a token so we can put that in a map easily
@@ -56,8 +58,7 @@ func (t *Token) Id() uuid.UUID {
 			t.TreeNodeID.String()
 		t.cacheId = uuid.NewV5(uuid.NamespaceURL, url)
 	}
-	id := t.cacheId
-	return id
+	return t.cacheId
 }
 
 // Return a new Token contianing a reference to the given TreeNode
--- a/lib/sda/host_msg.go
+++ b/lib/sda/host_msg.go
@@ -62,10 +62,11 @@ func (t *Token) Id() uuid.UUID {
 // Return a new Token contianing a reference to the given TreeNode
 func (t *Token) ChangeTreeNodeID(newid uuid.UUID) *Token {
 	t.cacheMutex.Lock()
+	defer t.cacheMutex.Unlock()
 	t_other := *t
+	defer t_other.cacheMutex.Unlock()
 	t_other.TreeNodeID = newid
 	t_other.cacheId = uuid.Nil
-	t.cacheMutex.Unlock()
 	return &t_other
 }
 
--- a/lib/sda/host_msg.go
+++ b/lib/sda/host_msg.go
@@ -3,6 +3,7 @@ package sda
 import (
 	"github.com/dedis/cothority/lib/network"
 	"github.com/satori/go.uuid"
+	"sync"
 )
 
 // Our message-types used in sda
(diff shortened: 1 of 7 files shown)


lib/sda/overlay.go
```diff
@@ -283,7 +280,7 @@ func (o *Overlay) Close() {
 	o.nodeLock.RLock()
 	defer o.nodeLock.RUnlock()
 	for _, n := range o.nodes {
-		if err := n.ProtocolInstance().Shutdown(); err != nil {
+		if err := n.Close(); err != nil {
```
Liamsi: Doesn't this introduce some problems (or do we plan to get rid of `Shutdown` again)?

lib/sda/node.go
```diff
@@ -313,8 +333,42 @@ func (n *Node) DispatchChannel(msgSlice []*SDAData) error {
...
+
+func (n *Node) dispatchMsgReader() {
+	for {
+		n.msgDispatchQueueMutex.Lock()
+		if len(n.msgDispatchQueue) > 0 {
+			dbg.Lvl3("Read message and dispatching it")
+			msg := n.msgDispatchQueue[0]
+			n.msgDispatchQueue = n.msgDispatchQueue[1:]
+			n.msgDispatchQueueMutex.Unlock()
+			n.dispatchMsgToProtocol(msg)
```
Liamsi: potenial returned error (from `dispatchMsgToProtocol`) ignored on purpose?

Liamsi: https://github.com/dedis/cothority/pull/289#discussion-diff-56555772

lib/sda/overlay.go
ineiti: Well now `node` has a goroutine, so we need to shut it down correctly, where before we could just tear it down. So instead of `overlay` calling `Shutdown` on `ProtocolInstance`, we have `overlay` calling `Close` on `node`, which will call `Shutdown` on `ProtocolInstance`.

Sheesh, making a long channel seems easier ;)

lib/sda/node.go
```diff
@@ -313,8 +333,44 @@ func (n *Node) DispatchChannel(msgSlice []*SDAData) error {
...
+func (n *Node) dispatchMsgReader() {
+	for {
+		n.msgDispatchQueueMutex.Lock()
+		if len(n.msgDispatchQueue) > 0 {
+			dbg.Lvl3(n.Myself(), "Read message and dispatching it",
+				len(n.msgDispatchQueue))
+			msg := n.msgDispatchQueue[0]
+			n.msgDispatchQueue = n.msgDispatchQueue[1:]
+			n.msgDispatchQueueMutex.Unlock()
+			n.dispatchMsgToProtocol(msg)
```
Liamsi: Error handling still missing (on purpose? at least doesn't look like ...)

ineiti: What shall I do with the error anyway? Display? I thought about it, but as the messages are asynchronous... perhaps at least display the error.

Liamsi: Yes, at least display it. Otherwise, if something behaves not as expected, we have to search and find this line I guess (and this might take some time). There is even an open issue related to this: #160

Liamsi: But you might know better, what makes sense here.

lib/sda/host_msg.go
```diff
@@ -56,6 +63,8 @@ func (t *Token) Id() uuid.UUID {
 
 // Return a new Token contianing a reference to the given TreeNode
 func (t *Token) ChangeTreeNodeID(newid uuid.UUID) *Token {
+	tokenMutex.Lock()
+	defer tokenMutex.Unlock()
```
nikkolasg: can you explain why we need a mutex here ?
It seems to me that the only reason we need the token is because of this `ChangeTreeNodeID` ?
If so, we should find another way to solve the data race problem because token.Id() is **often** used and since Token *should* be immutable (even if we did not enforce it *yet*), there *should* be no need of mutex.

lib/sda/local.go
```diff
@@ -173,7 +179,11 @@ func (l *LocalTest) NewNode(tn *TreeNode, protName string) (*Node, error) {
 		TreeNodeID:   tn.Id,
 		RoundID:      uuid.NewV4(),
 	}
-	return NewNode(o, tok)
+	node, err := NewNode(o, tok)
+	if err == nil {
+		l.Nodes = append(l.Nodes, node)
+	}
+	return node, err
```
nikkolasg: Should we not register the node also in the overlay ? Will it not bring some issues if we don't ? Like some messages not going to the right P.I. because the Overlay did not know about the right Node ?

lib/sda/node.go
```diff
@@ -255,6 +268,13 @@ func (n *Node) Shutdown() error {
 	return nil
 }
 
+// Close shuts down the go-routine and calls the protocolInstance-shutdown
+func (n *Node) Close() error {
+	dbg.Lvl3("Closing node", n.Myself())
```
nikkolasg: `MySelf()` ? 
I'm not sure this is a good idea, `Name()` is very explicit and the callers knows what to expect: a string. With `Myself()`, we don't know what to expect, is it a struct? a string? For some people coming from C++ / Java, `self` has another meaning.

lib/sda/overlay.go
```diff
@@ -279,13 +292,13 @@ func (o *Overlay) Suite() abstract.Suite {
...
-	o.nodeLock.RLock()
-	defer o.nodeLock.RUnlock()
+	o.nodeLock.Lock()
+	defer o.nodeLock.Unlock()
 	for _, n := range o.nodes {
-		if err := n.ProtocolInstance().Shutdown(); err != nil {
-			dbg.Error("Error shutting down protocol", err)
-		}
+		dbg.Lvl4("Closing node", n.TokenID())
+		o.nodeDelete(n.Token())
```
nikkolasg: You don't shutdown the protocols anymore ? Should we not do it ?

lib/sda/host_msg.go
ineiti: The race detector told me so... I think the problem comes if you create an Id and then call ChangeTreeNodeID on it. But I didn't follow the exact call-path to know how this happens.

lib/sda/local.go
ineiti: This is the `Local`-module which is only to be used in tests. The case here is if we test a Node that is not bound to anything else. so the `Local`-module has to take care about this case itself.

lib/sda/node.go
ineiti: Whatever. `MySelf` does give some additional information that can be useful during debugging (namely the Id of the node, IIRC).

lib/sda/overlay.go
ineiti: `nodeDelete` calls `node.Close` which will take care of the `Shutdown`.

lib/sda/node.go
nikkolasg: `Name()` is still fine even with this additional info. Otherwise, what do you think about `Info()` ?

lib/sda/overlay.go
nikkolasg: Cool ! thx.

lib/sda/node.go
ineiti: OK.

lib/sda/local.go
nikkolasg: but why should we want to do that ... ? I'm sorry for nitpicking but for me it does not make sense to test having a Node non-registered to an Overlay since it will never happen in practice ... ?
But OK, it's for local test anyway.

ineiti: Unit-testing

lib/sda/node.go
```diff
@@ -313,8 +333,47 @@ func (n *Node) DispatchChannel(msgSlice []*SDAData) error {
...
+	n.msgDispatchQueue = append(n.msgDispatchQueue, msg)
+	dbg.Lvl3(n.Myself(), "DispatchQueue-length is", len(n.msgDispatchQueue))
+	if len(n.msgDispatchQueue) == 1 && len(n.msgDispatchQueueWait) == 0 {
+		n.msgDispatchQueueWait <- true
+	}
+	n.msgDispatchQueueMutex.Unlock()
+}
+
+func (n *Node) dispatchMsgReader() {
+	for {
```
ineiti: Needs to check if the `msgDispatchClose` is still available or not. Else messages could be dispatched even though the node is closed. Probably a boolean is better instead of a channel, then set the boolean to true and send a msgDispatchQueueWait.

lib/sda/host_msg.go
Liamsi: If you want something more lightweight (multiple readers at the same time), you can also go with [RWMutex](https://golang.org/pkg/sync/#RWMutex)

lib/sda/node.go
nikkolasg: For the moment, if you use the reguler Overlay => Node configuration (so **not** as using `Local`), then it should be fine. Overlay has this `nodeInfo` map that says if a `Node` has been closed or not. So the overlay won't dispatch a message to a closed `Node`.

ineiti: I changed it anyway so that the for-loop checks if the node already has been closed. If there are some queued up messages, it will quit right away, without trying to dispatch the messages.

ineiti:
Any comments about merging?

nikkolasg:
OK for me

Liamsi:
Looks good to me.

Issue #270, opened 2016-03-15
Signing app


ineiti:
Re-create a signing app using the CoSi protocol. What we need to discuss:

- how to interact with the Cothorities
  - using the same network-port as other communications or another port
  - separation of tasks allowed anonymously and tasks only allowed once people are authenticated
- protocol to use / defining the API

PR #292, opened 2016-03-21
Prototype signing-application

nikkolasg:
Coding the Command Line Interface in a very basic / hacky / fast way to make a sda.Host launch a CoSi round and sign something using the CoSi protocol. The client gets the final signature at the end.


ineiti:
I think we do need to pass the list of cothorities in the request, too.

ineiti:
This is for #270

app/cosid/main.go
```diff
@@ -0,0 +1,143 @@
...
+# Writing your own protocol
+
+If you want to experiment with a protocol of your own, have a look at
+the protocols-package-documentation.
+
+# Deploying
+
+Unfortunately it's not possible to deploy the Cothority as a stand-alon app.
+Well, it is possible, but there is no way to start a protocol.
+*/
```
ineiti: Adjust the comments

app/cosid/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"bytes"
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func runCommand(cmd string, input io.Reader, args ...string) error {
```
ineiti: Helper functions at bottom.

lib/sda/host.go
```diff
@@ -371,6 +368,11 @@ func (h *Host) processMessages() {
 				h.checkPendingTreeMarshal(&il)
 			}
 			dbg.Lvl4("Received new entityList")
+			// Standalone CoSi hack!
```
ineiti: Comment for case after `case`.

lib/sda/host.go
```diff
@@ -380,6 +382,54 @@ func (h *Host) processMessages() {
...
+	tree := cr.EntityList.GenerateBinaryTree()
+	h.overlay.RegisterTree(tree)
+
+	// run the CoSi protocol
+	node, err := h.overlay.CreateNewNodeName("ProtocolCosi", tree)
+	if err != nil {
+		dbg.Error("Error creating tree upon client request:", err)
+		return
+	}
+	node.ProtocolInstance().SigningMessage(cr.Message)
```
ineiti: Love it! Wonder how we will do it in real life...

lib/sda/host_test.go
```diff
@@ -1,15 +1,76 @@
...
+	h1.StartProcessMessages()
+	defer h1.Close()
+	h2 := sda.NewLocalHost(2010)
+	h2.Listen()
+	h2.StartProcessMessages()
+	defer h2.Close()
+	h3 := sda.NewLocalHost(2020)
+	h3.Listen()
+	h3.StartProcessMessages()
+	defer h3.Close()
```
ineiti: Why not use `LocalTest` here?

lib/sda/host_test.go
```diff
@@ -1,15 +1,76 @@
...
 
+// Test hacky way of launching CoSi protocol externally
+func TestHackyCosi(t *testing.T) {
+	// TODO: github.com/btcsuite leaks 2 go routine and is imported by byzcoin
+	// in blkparser.
+	//	defer dbg.AfterTest(t)
+	dbg.TestOutput(testing.Verbose(), 4)
+	// make up the hosts and the entity list
+	h1 := sda.NewLocalHost(2000)
+	h1.ListenNoblock()
```
ineiti: This should be `Listen`, too, no?

lib/sda/host_test.go
nikkolasg: It's not clear for me what all theses boolean flags are when you call "GenLocalTree(...)", I was lazy to dive into the "local api" to get to know exactly what's happenning.

lib/sda/host_test.go
nikkolasg: Right.

ineiti:
Hashing is difficult - try #308

app/cosid/main_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"bytes"
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func TestCompileAndRun(t *testing.T) {
```
Liamsi: Test doesn't clean up correctly. Compiled binary gets deleted but the process (`cosid`) is still running after the test finishes.

ineiti:
Ouch - where is this `seelog` used?

app/cosi/main.go
```diff
@@ -0,0 +1,253 @@
...
+	os.Exit(1)
+}
+
+var f *flag.FlagSet
+var m *flag.FlagSet
+
+func init() {
+	f = flag.NewFlagSet("f", flag.ContinueOnError)
+	m = flag.NewFlagSet("m", flag.ContinueOnError)
+	//dbg.SetDebugVisible(3)
```
ineiti: Remove - if you merge #301 there will be a `dbg.AddFlags()` method to automatically add the `-debug`, `-debug-color` and `-debug-time` flags. It even evaluates environment-variables.

Liamsi:
> Ouch - where is this seelog used?
We don't use it directly. It's buried somewhere in btcsuite's code. (Nicolas and me searched for an `init` function which starts the leaking go-routines but couldn't find them)

app/cosi/main.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	// create a throw-away key pair:
+	kp := config.NewKeyPair(network.Suite)
+
+	// create a throw-away entity with an empty  address:
+	e := network.NewEntity(kp.Public, "")
+	client := network.NewSecureTcpHost(kp.Secret, e)
+	req := &sda.CosiRequest{
+		EntityList: el,
+		Message:    msgB,
```
ineiti: Do we want the message to be signed or the hash of the message? What if the file is huge?

app/cosi/main.go
```diff
@@ -0,0 +1,253 @@
...
+	sig := &sda.CosiResponse{}
+	if err := json.Unmarshal(sb, sig); err != nil {
+		return err
+	}
+	if bytes.Equal(sig.Sum, fHash) {
+
+	} else {
+		return fmt.Errorf("You are trying to verify a signature " +
+			"belongig to another file. (The hash provided by the signature " +
+			"doesn't match with with the hash of the file.")
```
ineiti: Missing closing bracket in Error-string.

app/cosi/main.go
nikkolasg: For the moment I think it's best to keep it like that. When @Liamsi made the remark about going more on the *stream* side, there had been issues raised up:
* In the case the witnesses don't need to verify the *content* of the message, we can pass directly the hash
* In case the witnesses *want* to verify the content of the message (for example a certificate, to verify that the domain is correct etc), the witnesses will have to get the whole file and not only the hash. That is when the streaming ability would be a real asset here. And I guess that will be the most used case.

app/cosid/main.go
```diff
@@ -0,0 +1,106 @@
...
+var debugVisible int
+
+// Initialize before 'init' so we can directly use the fields as parameters
+// to 'Flag'
+func init() {
+	flag.StringVar(&ConfigFile, "config", "config.toml", "which config-file to use")
+	flag.IntVar(&debugVisible, "debug", 1, "verbosity: 0-5")
+}
+
+const releaseVersion = "1.0-alpha"
```
ineiti: Optimistic ;)

Liamsi:
> Do we want the message to be signed or the hash of the message? What if the file is huge?

I thought we decided (together) to sign the message (because we would want to do further verifications etc.). I would also be fine with the hash (because for the moment we just the signature and nothing else).

app/utils.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			return nil, err
+		}
+		entities = append(entities, en)
+	}
+	el := sda.NewEntityList(entities)
+	return el, nil
+}
+
+func NewGroupToml(servers ...*ServerToml) *GroupToml {
```
ineiti: Exported function not documented.

app/utils.go
```diff
@@ -0,0 +1,110 @@
...
+	return &GroupToml{
+		Servers: servers,
+	}
+}
+
+// Returns the TOML representation of this GroupToml
+func (gt *GroupToml) String() string {
```
ineiti: Comment needs to start with name of exported function

lib/sda/host_msg.go
```diff
@@ -90,3 +100,79 @@ type EntityListUnknown struct {
 type SendEntity struct {
 	Name string
 }
+
+// CLI Part of SDA
+
+// The Request to CoSi: it is used by the client to send something to sda that
+// will in turn give that to the CoSi system. It contains the message the client
+// wants to sign.
+type CosiRequest struct {
```
ineiti: Comment on exported structure needs to start with structure-name

lib/sda/node.go
```diff
@@ -534,3 +534,14 @@ func (n *Node) Host() *Host {
 func (n *Node) SetProtocolInstance(pi ProtocolInstance) {
 	n.instance = pi
 }
+
+// HACKY / TEMPORARY way for accessing Cosi functionality
+func (n *Node) SigningMessage(msg []byte) {
```
ineiti: Please apply godoc-practices, even for this hacky/temporary code.

app/cosi/main.go
Liamsi: done

ineiti:
Can we have a standard file-name for the config-output? Like `config.toml`?

Liamsi:
> Can we have a standard file-name for the config-output? Like config.toml?
+1

ineiti:
About the Travis-failure: I see that every test that has  `NewMonitor().Listen()` waits for 100ms after calling `Listen()`. So perhaps just add this wait to `TestStatsString`?

Liamsi:
> About the Travis-failure: I see that every test that has NewMonitor().Listen() waits for 100ms after calling Listen(). So perhaps just add this wait to TestStatsString?

Of course I can add that here. But now all the tests pass (and it's not related at all to this issue).

ineiti:
Usage issues:

- ./cosid doesn't have `config.toml` as default output-configuration-file
- ./cosi if not called with arguments is missing final newline
- ./cosi can't be called with -c, -m order - but the arguments should ignore order
- ./cosi doesn't work with a simple example of only one cosid
- I want scripts - I added a branch with some compilation-scripts in #315

Liamsi:
> ./cosid doesn't have config.toml as default output-configuration-file

Are you sure? I tried it and it does ....

> ./cosi if not called with arguments is missing final newline

In my shell it doesn't show up like that, but I'll defineltly add a new-line to the usage output. 

> ./cosi can't be called with -c, -m order - but the arguments should ignore order

I don not really agree on that, the usage/documentation is clear that you should use them in that order. Maybe it's better to use a 'commands-like' syntax `cosi sign` and `cosi verify` instead of the current approach, but I guess this is another issue?

> ./cosi doesn't work with a simple example of only one cosid

This is true for every CoSi run (even in the simulation). We can make the documentation (and error output) clear, so that this can't happen (tell the user you need at least 2 nodes)?
  
> I want scripts - I added a branch with some compilation-scripts in #315

OK

ineiti:
OK, config.toml is default-config-file - `git pull` fixed that.

```
Example usuage (create a file my-local-group.toml using cosid first):
cosi -f my-file.txt -c my-local-group.tomlicsil1noteb81:cosi ligasser$ 
```

Argument-order is usually not important for most of the command-line-tools I use. So as a user I'm not used to see it different here. *Do not surprise the user*

ineiti:
Single-node cothority handled in #318 - I propose to merge this in, too.

ineiti:
When calling `cosi -m`, the output lacks a final newline.

ineiti:
You could also add `dbg.AddFlags` to the main.go, then the debug-flags would be included.

ineiti:
`dbg.AddFlags` doesn't play well with codegansta's cli...

nikkolasg:
hard to follow the big restructuring at the end but it seems perfectly fine for a POC ;) OK to merge !

Liamsi:
The changes at the end look good. Merge :+1:

ineiti:
I just want to finish the 'checkConfig', then I'll put it back into R4M.

ineiti:
OK, I stop working on it, I like it too much ;)
There is one new command, `check`, which contacts each server individually, then pairs of servers, to check for possible problems. I saw that Ismail's and Daeinar's server don't want to talk to another if Ismail's is `root`. The other way around it works OK ?!?

Should this be merged against `development`? Or against `master`?

ineiti:
I told you I'll stop working on it. Last question: is the description about cosi and cosid well placed in the main README.md, or should it go into `app/README.md`?

nikkolasg:
* Merge to development then master
* And the app description *for me* should be placed under the apps folder WITH a link to it placed under the main readme.

Liamsi:
I basically agree with @nikkolasg . But also see the discussion on slack. I think it could also be a good idea to have a CoSi-only repository. The question then is: how do we keep the CoSi repo in sync with the cothority repo (regularly)?

ineiti:
Will tag it and make a binary release.

Liamsi:
That's even better. Add a link to the README to the documentation in the branch/tag. Or should I do it?

ineiti:
I just added #330

ineiti:
You can help finding typos in https://github.com/dedis/cothority/tree/add_cosi_cli_readme

bford:
Looking good.  I haven't had time to review the code or discussions thoroughly, but just one quick note:  I know I was the one to suggest 'cosid' as the command to start a server, but no need always to take me literally. :)  It occurred to me later that, since the 'cosi' binary already needs to support multiple commands (e.g., 'cosi sign', 'cosi verify'), would it perhaps be simpler and make more sense just to combine 'cosid' into the 'cosi' binary in the form of, say, a 'cosi server' command?  Up to you, and definitely not a high-priority change if it would be a pain for any reason.

PR #296, opened 2016-03-23
Platform mininet

ineiti:
Opening new PR against development.


ineiti:
Using mininet-implementation of old cothority is too complicated. Need to come back later.

ineiti:
Replaced by #443

PR #300, opened 2016-03-23
adjusted protocol-names and moving around byzcoin-files

ineiti:
Moving around files in ByzCoin, renaming the protocol-names


protocols/manage/count.go
```diff
@@ -106,9 +106,11 @@ func (p *ProtocolCount) FuncPC() {
 		p.SendTo(p.Parent(), &NodeIsUp{})
 	}
 	if !p.IsLeaf() {
-		for _, c := range p.Children() {
-			dbg.Lvl3(p.Myself(), "sending to", c.Entity.Addresses, c.Id, p.timeout)
-			p.SendTo(c, &PrepareCount{Timeout: p.timeout})
+		for _, child := range p.Children() {
+			go func(c *sda.TreeNode) {
```
nikkolasg: Out of curiosity: why using go routines ? I don't mind except I prefer *not* using go routines where this is *not* necessary.

ineiti: time of creation for connections - usually the count-protocol is used when no connections are set up yet. If it has to wait for all connections to be set up in a serial way, the timeout is reached too fast. Used in simulations where there are more than 2000 nodes involved.

protocols/manage/count.go
```diff
@@ -106,9 +106,11 @@ func (p *ProtocolCount) FuncPC() {
 		p.SendTo(p.Parent(), &NodeIsUp{})
 	}
 	if !p.IsLeaf() {
-		for _, c := range p.Children() {
-			dbg.Lvl3(p.Info(), "sending to", c.Entity.Addresses, c.Id, p.timeout)
-			p.SendTo(c, &PrepareCount{Timeout: p.timeout})
+		for _, child := range p.Children() {
+			go func(c *sda.TreeNode) {
```
nikkolasg: just out of curiosity (already asked that question I think but can't find it anywhere):
Why use a go routine ? I'm not against go routine at all but it should be avoided when it can be avoided.

ineiti: Just click on "Show 2 comments":

time of creation for connections - usually the count-protocol is used when no connections are set up yet. If it has to wait for all connections to be set up in a serial way, the timeout is reached too fast. Used in simulations where there are more than 2000 nodes involved.

nikkolasg: Ok.

nikkolasg:
Rest is ok for me !

PR #304, opened 2016-03-23
copied from redo_cosi_graph

ineiti:
Added two timeouts for simulations. Also calculation of third parameter when only giving two from "bf, depth, hosts".


nikkolasg:
Cool ! OK for me

PR #305, opened 2016-03-23
Refactoring of tree.go

Daeinar:
This patch does some refactoring of functions and documentation in `tree.go`:
- `(t *Tree) ListNodes()` => `(t *Tree) List()`
- `(t *Tree) GetTreeNode(...)` => `(t *Tree) Search(...)`


lib/sda/tree.go
```diff
@@ -258,7 +259,7 @@ func (tm TreeMarshal) MakeTree(il *EntityList) (*Tree, error) {
 
 // MakeTreeFromList creates a sub-tree given an EntityList
 func (tm *TreeMarshal) MakeTreeFromList(parent *TreeNode, il *EntityList) *TreeNode {
-	idx, ent := il.Search(tm.EntityId)
+	idx, ent := il.SearchEntity(tm.EntityId)
```
ineiti: So you still search the Entity-index. I would call it `EntityIdx` or something, to make it clear it's not a `TreeNodeIdx`. And put a comment in the structure saying that it's possible that one `EntityIdx` doesn't have to be unique in a tree!

ineiti: In fact the `TreeNode`-structure does have `EntityIdx`. But I'd still add a comment there saying that `EntityIdx` is not unique.

Daeinar: Wait, I thought the `EntityIdx` is unique since it maps 1-to-1 to the public key? The `TreeNodeIdx` wouldn't be unique since you can run more than one `Tree` on an `Entity`.

nikkolasg: @ineiti  said :
>  that it's possible that one EntityIdx doesn't have to be unique in a tree!

For my part, the function name I would keep `Search()`, as you call this function from an `EntityList`, so it's twice the same information. It's like method name for packages, if you have `schnorr.SignSchorr`, I find it quite redundant...

ineiti: `SignSchnorr` is a 'static' function from a packet, whereas `SearchEntity` is a function defined on a structure. As such I wouldn't treat them the same way for naming. Of course `entity.SearchEntity` is redundant, but if I have an entityList `il`, saying `il.SearchEntity` is less redundant. Just a bit...

nikkolasg: I guess at this point, I could say that naming its variable `il` is not a very good name and should be named more `entList` or (my favorite) `entityList`. So you can have `entityList.Search(...)`.
Anyway, that was my 2 cent, I won't push for it so as you like.

Daeinar:
Okay worked in your comments accordingly. I decided to stick with the short form `el` and not to go with the long one alias `entityList` as (most of) the functions are short enough so that you always see the definition of `el` in the function header.

Daeinar:
Travis passes. Everything done here from my side.

ineiti:
Seems good - except if someone thinks that `tree.List` is a bit terse and should be changed back into `tree.ListNodes` - but I only have a slight preference for `tree.ListNodes`.

nikkolasg:
OK

PR #308, opened 2016-03-24
added hashfile-function to lib/crypto

ineiti:
Added a HashFile function to the `lib/crypto` that always returns 32 bytes.


nikkolasg:
A remark coming from Ismail's remark: we should start handling everything more as a "stream" rather than a simple bulk of bytes. For example here, we could redesign using a `io.Reader` as argument and reading chunk by chunk so the whole file does not have to get into memory.

Liamsi:
Any idea why travis is failing here? 
I realized that we already hash correctly in the cli branch (must have been outdated output we saw on Thursday). Furthermore, it could be a better approach to simply have an public `func Hash(b []byte) ([]byte, error)` (can be used for the files, too) somewhere instead of `HashFile`?

ineiti:
If you want to have a `Hash`-function, you won't be able to do 'streaming' on a file.

nikkolasg:
Something like:
```
// do the *real* hashing
func Hash(b []byte) []byte {
   ...
}
func HashStream(stream io.Reader) []byte {
     for {
        // Something like...
        b := stream.Read()
        Hash(b)
    }
}
func HashFile(filename string) []byte {
   f := file.Open(filename)
   return HashStream(f)
}
```
Avoiding the brittleness in our code ... !

ineiti:
But how do you keep the state of the hash during multiple calls to `Hash`?

nikkolasg:
When you have a `Hash` object, calling `Sum(nil)` returns the current state <=> hash of the input, but it does not change the *state* of the Hash. You can still input something to the hash after.
```
        // Sum appends the current hash to b and returns the resulting slice.
        // It does not change the underlying hash state.
        Sum(b []byte) []byte
```
Source: https://golang.org/pkg/hash/

ineiti:
And how do you handle two goroutines hashing their file?

nikkolasg:
See updated snippet code.

ineiti:
Is this worth being pursued or do we just close it?

Liamsi:
> Is this worth being pursued or do we just close it?

I think what Nicaols proposed makes sense (and we will need sth. like in different places anyways). But I'm also OK to just close it, if you both agree.

ineiti:
Can you just add an issue in `Features`, please? Then close and delete this branch.

nikkolasg:
Just done it, tell me if that seems ok otherwise we rollback and merge/close.

lib/crypto/hash.go
```diff
@@ -4,34 +4,69 @@ import (
...
+// HashStream returns the hash of the stream reading from it chunk by chunk of
+// size DefaultChunkSize
+func HashStream(suite abstract.Suite, stream io.Reader) ([]byte, error) {
+	return hashStream(suite, stream, DefaultChunkSize)
+}
+
+// Hash a stream using chunks of size
+func HashStreamChunk(suite abstract.Suite, stream io.Reader, chunkSize int) ([]byte, error) {
+	if chunkSize < 1 {
+		return nil, errors.New("Wront chunksize value")
```
ineiti: Typo - "wrong"

lib/crypto/hash.go
```diff
@@ -4,34 +4,69 @@ import (
...
 	return hash.Sum(nil), nil
 }
 
+// HashStream returns the hash of the stream reading from it chunk by chunk of
+// size DefaultChunkSize
+func HashStream(suite abstract.Suite, stream io.Reader) ([]byte, error) {
+	return hashStream(suite, stream, DefaultChunkSize)
+}
+
+// Hash a stream using chunks of size
```
ineiti: Docu-error: first word has to be the name of the function
enumerate arguments

Liamsi:
For me it would be OK to merge it

lib/crypto/hash.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+// Hash simply returns the Hash of the slice of bytes given
+func Hash(hash h.Hash, buff []byte) ([]byte, error) {
+	_, err := hash.Write(buff)
+	return hash.Sum(nil), err
+}
+
+// Size of the chunk  used to read a stream
+const DefaultChunkSize = 1024
```
ineiti:
```hash.go:18:1: comment on exported const DefaultChunkSize should be of the form "DefaultChunkSize ..."
```

nikkolasg: Damn. Also forgot to put the lint verification. Done. Thx.

PR #315, opened 2016-03-30
some scripts for convenient app-building

ineiti:
Couldn't resist to make some scripts for easy deployment. Tested it on my server (the EPFL-server is down) and it doesn't work - see issue in cosi_cli.


nikkolasg:
OK

PR #318, opened 2016-03-30
added correct running for one node only

ineiti:
Corner-case: only one node in the Cothority. Correct signature.

### DIFF
--- a/protocols/cosi/cosi.go
+++ b/protocols/cosi/cosi.go
@@ -1,4 +1,4 @@
-// Implementation of a round of a Collective Signing protocol.
+// Package cosi implements a round of a Collective Signing protocol.
 package cosi
 
 import (
@@ -28,7 +28,7 @@ type ProtocolCosi struct {
 	// The node that represents us
 	*sda.Node
 	// TreeNodeId cached
-	treeNodeId uuid.UUID
+	treeNodeID uuid.UUID
 	// the cosi struct we use (since it is a cosi protocol)
 	// Public because we will need it from other protocols.
 	Cosi *cosi.Cosi
@@ -62,6 +62,18 @@ type ProtocolCosi struct {
 	challengeHook    ChallengeHook
 }
 
+// AnnouncementHook allows for handling what should happen upon an
+// announcement
+type AnnouncementHook func(in *CosiAnnouncement) error
+
+// CommitmentHook allows for handling what should happen when a
+// commitment is received
+type CommitmentHook func(in []*CosiCommitment) error
+
+// ChallengeHook allows for handling what should happen when a
+// challenge is received
+type ChallengeHook func(*CosiChallenge) error
+
 // NewProtocolCosi returns a ProtocolCosi with the node set with the right channels.
 // Use this function like this:
 // ```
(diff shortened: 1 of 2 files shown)


protocols/cosi/cosi.go
```diff
@@ -289,39 +266,22 @@ func (pc *ProtocolCosi) sendChallenge(out *CosiChallenge) error {
...
-	pc.tempResponse = append(pc.tempResponse, in)
-	pc.tempResponseLock.Unlock()
-	// do we have enough ?
-	dbg.Lvl3(pc.Node.Name(), "has", len(pc.tempResponse), "responses")
-	if len(pc.tempResponse) < len(pc.Children()) {
-		return nil
+	if !pc.IsLeaf() {
+		// add to temporary
+		pc.tempResponseLock.Lock()
+		pc.tempResponse = append(pc.tempResponse, in)
```
Liamsi: Do we want to append `nil` here? It doesn't have any side-effects? (Just asking because `handleRepsonse(nil)` is called.

Liamsi: same for `handleCommitment`

Liamsi:
Besides the two questions, looks good.

protocols/cosi/cosi.go
```diff
@@ -234,18 +221,8 @@ func (pc *ProtocolCosi) handleCommitment(in *CosiCommitment) error {
...
 // StartChallenge start the challenge phase. Typically called by the Root ;)
 func (pc *ProtocolCosi) StartChallenge() error {
-	// first check the hook
-	/*if pc.challengeHook != nil {*/
-	//return pc.challengeHook(nil)
-	/*}*/
-
-	if pc.Message == nil {
-		return fmt.Errorf("%s StartChallenge() called without message (=%v)", pc.Node.Name(), pc.Message)
-	}
```
nikkolasg: Why remove that check ?
If we start a CoSi round without signing anything, then the user of the CoSi must surely do something wrong and should be alerted of that ...

protocols/cosi/cosi.go
nikkolasg: Before, lib/cosi  +protocols/cosi had specifics calls when you start a phase => `StartCommitment` or `StartReponse`. Linus just made them obsolete by passing *nil*, but it's the same thing. 
Whether one approach is better than the other one is subject to another discussion but since it looks urgent, it's OK ;)

Liamsi: Yes, I understood that. I was just wondering if it isn't obsolete to add the `nil` to the underlying `pending`... as it comes with costly  `Lock`/`Unlock` (nah, not really that costly, but if obsolete it can be removed)

nikkolasg: the section with the locks is never called because the one that calls with `nil` is always a leaf.

Liamsi: OK, thanks!

protocols/cosi/cosi.go
ineiti: Perhaps we should change `StartAnnouncement` to take a `msg` and dispose of `SigningMessage`. But I think an empty message is not a problem.

nikkolasg: The reason it was put here is because there can be two modes: 
* either put the message in the Announcement
* either put the message in the Challenge. 
The second approach has the benefit of being able to correlate rounds like BizCoin where the Response of the first round is sent as the message in the Challenge of the second round. The first approach lets users make the verification as soon as possible.
So Both are correct and only the latter is used here.
Regarding the empty message... I thought it would be considered invalid but in fact, it seems gpg and others don't complain with an empty message, so why should we ... ! OK

PR #319, opened 2016-03-30
changed name to reflect default nowait

ineiti:
As we saw in the case of `cosi_cli`, it can be confusing to have a `host.Listen` that waits to be able to connect to itself. So this does rename

- `host.Listen` to `host.ListenAndWait`
- `host.ListenNoBlock` to `host.Listen`

### DIFF
--- a/cothority.go
+++ b/cothority.go
@@ -66,7 +66,7 @@ func main() {
 	if err != nil {
 		dbg.Fatal("Couldn't get host:", err)
 	}
-	host.ListenAndWait()
+	host.ListenAndBind()
 	host.StartProcessMessages()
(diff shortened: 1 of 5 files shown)


nikkolasg:
Not sure about `ListenAndWait`, as it is not the *purpose* of the function. For me it seems this function listens on a port and waits for whatever, and the other does not wait. This is not what is happenning
The *purpose* of the function is to connect to itself to be sure the port is opened. sooo.... `ListenAndConnectSelf` or `ListenAndConnect` or `ListenAndWaitBinding` or ... ?

nikkolasg:
OK for me

Issue #190, opened 2016-02-04
Differentiate uuid.UUID


ineiti:
`uuid.UUID`s are used for different IDs. To name just a few:
- TreeNodeID
- EntityID
- TreeID
- others

We should use types for each ID which point to uuid.UUIDs, so that we don't mix up the different IDs. This has to be done for all the code.

Daeinar:
+1

That would make the code much clearer!

Liamsi:
I can work on this one. 
Besides the above mentioned UUIDs, do you all agree on naming the one returned by `RegisterMessageType` as `MessageType` (instead of MessageID)?

Liamsi:
I started with this one. I have a question here: Would you prefer two types `EntityID` and `EntityListID`? Currently they are used interchangeably (possible because everything is just a UUID), see here: https://github.com/dedis/cothority/blob/6484a74cf3e99521821636bcc108831476a40244/lib/sda/tree.go#L212-L214

ineiti:
The example you give is a hack - blame says it's mine :( I remember having done that to avoid having a field that is only used for the root. Bad programming, sorry.

There should be two IDs for Entity and EntityList, but then that part has to be somewhat rewritten.

Liamsi:
OK. I will see how much changes are necessary for that. The easiest way would be to cast from one type to the oder (which would also be a hacky way...)

ineiti:
Or store the EntityListID somewhere else - it's only used in the root-node of the TreeMarshal anyway.

Liamsi:
It seems like it is already stored somewhere (or is this another one): https://github.com/dedis/cothority/blob/6484a74cf3e99521821636bcc108831476a40244/lib/sda/tree.go#L276-L277

Liamsi:
Note for my slef: Sometime we compare uuids using `uuid.Equal` sometimes we use the `==` 
-> write a test to  make sure both work as expected

Liamsi:
> Or store the EntityListID somewhere else - it's only used in the root-node of the TreeMarshal anyway.

OK, just created a new field for this purpose. 

But we have similar problems for `NodeId` which also (sometimes) identifies the `TreeId` ...

Liamsi:
> But we have similar problems for NodeId which also (sometimes) identifies the TreeId ...

And sometimes it even identifies the `TreeNode`? 
[...]

PR #320, opened 2016-03-30
Differentiate uuid.UUID

Liamsi:
PR for #190

### DIFF
--- a/lib/sda/host_test.go
+++ b/lib/sda/host_test.go
@@ -341,8 +341,8 @@ func TestTokenId(t *testing.T) {
 	}
 	id1 := t1.Id()
 	t2 := &sda.Token{
-		EntityListID: uuid.NewV1(),
-		TreeID:       uuid.NewV1(),
+		EntityListID: sda.EntityListID(uuid.NewV1()),
+		TreeID:       sda.TreeID(uuid.NewV1()),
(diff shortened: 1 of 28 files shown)


ineiti:
Skimmed through it and looks very good. +1 for merge.

lib/sda/host_test.go
```diff
@@ -250,7 +250,7 @@ func TestTreePropagation(t *testing.T) {
 		network.DumpTypes()
 		t.Fatal("h1 didn't receive SendTree type:", msg.MsgType)
 	}
-	if msg.Msg.(sda.TreeMarshal).EntityId != uuid.Nil {
+	if msg.Msg.(sda.TreeMarshal).EntityListID != sda.EntityListID(uuid.Nil) {
```
nikkolasg: mhhh.. Is there not a way to remove some noise in these checks ? 
Having some kind of generic checker for specific UUID ?
 Along the lines of ``` EntityListID.Equal(uuid.Nil) ``` 
I don't know I haven't tried it but *if* this is possible that could be a real **plus** .

Liamsi: I added an `Equals()` method, if we were comparing them in production code. But I didn't focus on trying to make the tests "less noisy". I would prefer to merge it and open another issue: #324

PR #322, opened 2016-03-31
Services

nikkolasg:
This is a first try at implementing the Services idea.
The main idea is to have a Service which is a longterm running object that performs mainly two tasks:
* Instantiating protocol instances (instead of using a static registered function). It can now give any external data needed by the ProtocolInstance.
* Process sda-external requests coming from a CLI or others sda servers. It defines what actions can be done with what services: *Launch a CoSi protoocol with this Tree and Entity List and send the output back to me*. The services themselves define their own API.

**Ideas**: 
* Have a basic Service interface as implemented here so we can alredy use it ( TO BE DISCUSSED )
* Move some of the host functionality into Overlay and rename Host to Dispatcher as it is really **only** a Dispatcher anymore. ( TO BE DISCUSSED ) 

**Only food-for-thoughts ideas**:
* separate the ProcessRequest message into a separate interface that any struct can implement => Overlay would register for specific messages the same way as Services does.

This way our diagram for me seems cleaner:

(Disclaimer: **hard** to do diagram in markdown :D )
 
```
                         Node / ProtocolInstance     
                (dispatch) |              | (instantiate)
                       Overlay           Service      <=>      External world
                             \            /
                              Dispatcher
                                   |
                                Network
```

**Questions**:
* With Linus that occured: How to use embedded protocols with this scheme.
=> If the root uses ByzCoin service that uses CoSi service to instantiate the underlying CoSi protocol, *all* the others nodes need to do the same. I guess it's automatically guessed by the `ByzCoin` service. On all others nodes, the host / dispatcher will call `ByzCoinService.InstantiateProtocol()`. Then the `ByzCoin` service will get the CoSi service and call also ` InstantiateProtocol`. Is that OK ?
* Still using ServiceID alongside with String for registration and providing helpers to go back and forth from each type, the same way we did for Node... It seems bloated and uncessary, it would be better if we could think of *one way* of doing things and not two. Using uuid has the advantages of being unique and fixed size, and string has the advantages of being very explicit and easy to debug... What do yo you think ?

**Implementation**:
- [ ] Service interface (dispatching of request, creation of protocol instances)
- [ ] Implementation of CoSi service
- [ ] Move more function of Host into Overlay (Tree creation etc)
- [ ] Rename Host to Dispatcher (?)


protocols/ntree/simulation.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 func init() {
+	sda.RegisterNewProtocol("CoSiNtree", func(node *sda.Node) (sda.ProtocolInstance, error) { return NewProtocol(node) })
```
ineiti: Please avoid these function-registrations and use
```
sda.RegisterNewProtocol("CoSiNtree", NewProtocol(node))
```

The same goes for the other occurrences. The simpler, the better.

nikkolasg: Sure. I simply changed from `ProtocolRegisterName` to `RegisterNewProtocol`, so I guess I must have taken that from a  old version :D

lib/sda/service.go
```diff
@@ -0,0 +1,387 @@
...
+// ProtocolInstance
+type NewProtocol func(*Node) (ProtocolInstance, error)
+
+// ProtocolConstructor is an interface that can instantiate a given protocol
+type ProtocolConstructor interface {
+	NewProtocol(*Node) (ProtocolInstance, error)
+}
+
+// Service is a generic interface to define any type of services.
+type Service interface {
```
ineiti: Should loading and saving of configuration-variables be in here? Something like

```
Load(file string) error
Save(file string) error
```

Where the `sda` provides a filename to the service that can be used to save the configuration? It could also be a directory where the service can store more than one file.

lib/sda/overlay.go
```diff
@@ -125,71 +127,42 @@ func (o *Overlay) EntityList(elid EntityListID) *EntityList {
...
+	n, err := o.createNewNodeID(protocol, tree)
 	if err != nil {
 		return nil, err
 	}
-	o.nodeLock.Lock()
-	defer o.nodeLock.Unlock()
-	o.nodes[node.token.Id()] = node
-	o.nodeInfo[node.token.Id()] = false
-	return node, node.protocolInstantiate()
+	go n.StartProtocol()
```
ineiti: I suppose that the merge won't catch that line. Remove the `go`

ineiti:
Can this be closed then?

ineiti:
Closing for #346

Issue #189, opened 2016-02-03
Merge message structs and their SDA-wrappers


Daeinar:
I think it would be good to merge message structs, as used by an SDA-protocol, and the wrappers around those structs, as required by the SDA-framework. The current solution requires a protocol-designer to specify two structs for a message instead of just one. In other words, move from this

```
// The protocol message
type Msg struct {
    // message fields
}

// SDA-wrapper around Msg
type WMsg struct {
    *sda.TreeNode
    Msg
} 
```

to this

```
// The protocol message
type Msg struct {
    *sda.TreeNode
    // message fields
}
```

ineiti:
Le 03/02/2016 15:24, Philipp Jovanovic a écrit :
>
> I think it would be good to merge message structs, as used by an 
> SDA-protocol, and the wrappers around those structs, as required by 
> the SDA-framework The current solution requires a protocol-designer to 
> specify two structs for a message instead of just one In other words, 
> move from this
>
> |// The protocol message type Msg struct { // message fields } // 
> SDA-wrapper around Msg type WMsg struct { *sdaTreeNode Msg } |
>
> to this
>
> |// The protocol message type Msg struct { *sdaTreeNode // message 
> fields } |
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/dedis/cothority/issues/189>.
>
good call - I thought about the same. If everybody is OK with that, I 
can do that tomorrow (as I wear the 'reflect'-hat for the moment :)

Daeinar:
@ineiti: Feel free to go ahead, you have the 'assign'-hat as well now. ;)

ineiti:
Next try ;)

ineiti:
Look if an error-handling on `Handler`s and `Channel`s is possible.

ineiti:
In the services we started to use

`func Process(e *network.Entity, msg *Processing)`

which, when used also in the protocols, would be much more easy than fiddling around with the structures in reflect.

ineiti:
Followed by https://github.com/dedis/onet/issues/207

PR #325, opened 2016-04-01
Ntree add test

ineiti:
Tests for NTree - will also add tests for `protocols/manage`.


ineiti:
Also added tests for the `protocols/manage`-protocols. Couldn't resist to rewrite the broadcast protocol which sent around too many messages.

Liamsi:
Looks OK

protocols/manage/broadcast.go
```diff
@@ -14,146 +16,82 @@ func init() {
...
+	dbg.Lvl3(b.Info(), "Received message from", msg.TreeNode.String())
+	if msg.TreeNode.Id == b.Root().Id {
+		dbg.Lvl3(b.Info(), "Contacting everybody")
+		// Connect to all nodes that are later in the TreeNodeList, but only if
+		// the message comes from root
+		for _, tn := range b.Tree().List()[b.tnIndex+1:] {
+			dbg.Lvl3("Connecting to", tn.String())
+			err := b.SendTo(tn, &ContactNodes{})
+			if err != nil {
+				return err
```
nikkolasg: In fact, having a handler does not necessarily have to return an error. `RegisterHandler` takes an interface, and we don't check for error, so a `dbg.Error` could do the trick.

nikkolasg:
Except comment, OK for me too.

ineiti:
Removed error-handling (sorry Liamsi...) but added a note to good old #189

ineiti:
Better now - there was a handler that was also used by `Start`, so it should return an error.

PR #327, opened 2016-04-01
Merging `lib/crypto` from `cosi_cli` to `development`

ineiti:
If we go with merging `cosi_cli` only into `master`, then we should have `lib/crypto` in `development`, as it simplifies somewhat the hashing.

Next step is to find all places where hashes are used and get them to use these hash-functions.


Liamsi:
Well done @nikkolasg

Issue #282, opened 2016-03-17
Refactor JVSS


ineiti:
As it was written in a hurry, clean it up.

Daeinar:
Merged #333.

PR #333, opened 2016-04-03
JVSS Refactoring

Daeinar:
Attempt to refactor the JVSS protocol, see #282.


Daeinar:
Done.

protocols/jvss/jvss.go
```diff
@@ -15,645 +21,249 @@ import (
...
-	done bool
-	// done lock
-	doneLock *sync.Mutex
-	// how many deals are ok
-	goodDeal int
-}
+// Identifiers for long- and short-term shared secrets.
+const (
+	LTSS = "LTSS"
+	STSS = "STSS"
```
Liamsi: Nitpicking: Wouldn't it make sense to make this an enum instead of string constants?
Then you can also refer to the identifiers type where it is needed (for instance in `initSecret `)

Liamsi: (See for instance here; the "custom types" example: https://blog.denevell.org/golang-constants-enums.html)

ineiti: For debugging reasons it might be OK to have string constants.

protocols/jvss/jvss.go
```diff
@@ -15,645 +21,249 @@ import (
...
+	// Setup message handlers
+	handlers := []interface{}{
+		jv.handleSecInit,
+		jv.handleSecConf,
+		jv.handleSigReq,
+		jv.handleSigResp,
+	}
+	for _, h := range handlers {
+		if err := jv.RegisterHandler(h); err != nil {
+			return nil, fmt.Errorf("Error, could not register handler: " + err.Error())
```
ineiti: Do we need to put that into the `Node`? Having a `Node.RegisterHandlers` with a varargs?

protocols/jvss/jvss.go
```diff
@@ -1,11 +1,17 @@
...
+// JVSS provides a threshold signing scheme based on Shamir's joint verifiable
+// secret sharing algorithm and Schnorr signatures. The protocl runs in two
+// phases. During the protocol setup a long-term shared secret is establised
+// between all participants. Afterwards, any of the members can request a
+// signature, which triggers the creation of another, short-term shared secret.
+// Each member then sends its partial signature to the requester which finally
+// puts everything together to get the final Schnorr signature. To verify a
+// given Schnorr signature a member still has to be able to access the
+// long-term shared secret from which that particular signature was created.
+
```
Liamsi: Remove this newline. Otherwise the generated `godoc` will show an empty package description (verify yourself. by running `godoc -http=:6060` and visiting  http://localhost:6060/pkg/github.com/dedis/cothority/protocols/jvss/ )

Liamsi:
Besides the two comments: Looks clean and good!

protocols/jvss/jvss.go
```diff
@@ -15,645 +21,249 @@ import (
...
-func (jv *JVSSProtocol) handleRequestSecret(requestBuff *RequestBuffer) (*RequestBuffer, error) {
-	// prepare our deal
-	doneChan := make(chan *poly.SharedSecret)
-	requestBuff.setSecretChan(doneChan)
-	deal := jv.newDeal()
-	requestBuff.addDeal(jv.index, deal)
-	// send to everyone
-	buf, err := deal.MarshalBinary()
-	if err != nil {
+	if err := jv.broadcast(req); err != nil {
```
ineiti: Another function to be added to `Node`

protocols/jvss/jvss.go
Liamsi: OK, makes sense, too.

protocols/jvss/jvss.go
ineiti: Can you put that into a "1.", "2.", ... list? Do you have any URL for further reading?

protocols/jvss/jvss.go
```diff
@@ -15,645 +21,249 @@ import (
...
-func (jv *JVSSProtocol) Shutdown() error {
-	close(jv.reqChan)
-	close(jv.respChan)
-	close(jv.ltChan)
-	close(jv.rdChan)
+// Start initiates the JVSS protocol by setting up a long-term shared secret
+// which can be used later on by the JVSS group to sign and verify messages.
+func (jv *JVSS) Start() error {
+	jv.initSecret(LTSS)
+	<-jv.secretsDone
```
ineiti: The normal `Start` method doesn't wait for the protocol to finish. @nikkolasg , do you think it is OK to 'delete' the `jv.secretsDone` here?

You can either make `secretsDone` public (but then you have to make sure it's empty before you write something into it, else you'll block), make a `WaitForDone` method or do a `RegisterDoneCallback`.

protocols/jvss/jvss.go
```diff
@@ -15,645 +21,249 @@ import (
...
+}
+
+// Secret contains all information for long- and short-term shared secrets.
+type Secret struct {
+	secret   *poly.SharedSecret              // Shared secret
+	receiver *poly.Receiver                  // Receiver to aggregate deals
+	deals    map[int]*poly.Deal              // Buffer for deals
+	sigs     map[int]*poly.SchnorrPartialSig // Buffer for partial signatures
+	numConfs int                             // Number of collected confirmations that shared secrets are ready
+	mtx      *sync.Mutex                     // Mutex to sync access to numConfs
```
ineiti: Mutex are exceptions with regard to the *use pointers* rule. If you define them as values, you don't even have to initialize them (like in line 167)

Liamsi: +1

protocols/jvss/jvss.go
```diff
@@ -15,645 +21,249 @@ import (
...
-				return
-			}
-			dbg.Lvl3("JVSS (", ") Generated Signature Response")
-			rb.signature = sign
-			rb.sigGend = true
-			// notify interested party
-			if rb.sigChan != nil {
-				go func() { rb.sigChan <- sign }()
+		for i := 0; i < len(secret.deals); i++ {
+			if _, err := secret.receiver.AddDeal(jv.nodeIdx(), secret.deals[i]); err != nil {
```
ineiti:
```
for _, deal := secret.deals {
	if _, err := secret.receiver.AddDeal(jv.nodeIdx(), deal); err != nil {
```

ineiti:
Nice work

protocols/jvss/jvss.go
Daeinar: Oh nice, I didn't know that one!

protocols/jvss/jvss.go
Daeinar: That would be a good idea I guess.

protocols/jvss/jvss.go
Daeinar: +1

protocols/jvss/jvss.go
Liamsi: I think there is a missig `range` keyword above.

Daeinar: Yes, indeed.

protocols/jvss/handlers.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	secret.mtx.Lock()
+	secret.numConfs++
+	secret.mtx.Unlock()
+
+	dbg.Lvl2(fmt.Sprintf("Node %d: %s confirmations %d/%d", jv.nodeIdx(), msg.SID, secret.numConfs, len(jv.nodeList)))
+
+	// Check if we have enough confirmations to proceed
+	if (secret.numConfs == len(jv.nodeList)) && (msg.SID == LTSS || msg.SID == fmt.Sprintf("%s%d", STSS, jv.nodeIdx())) {
```
Daeinar: Btw. I find this construction a bit awkward, especially the second part where I check which SID the message has in order to be able to notify the correct node that all confirmations (for establishing the shared secret) have arrived.

If you have a better idea how to solve that, I am open for suggestions. :)

Daeinar:
I think everything is done here from my side.

ineiti:
can you also add a `simul/runconfigs/test_jvss.toml` for inclusion in the simul-tests? It should run for 5-10 rounds for 2-3 values of 'hosts'.

PR #336, opened 2016-04-04
taking 'go'-call away from StartNewNode

ineiti:
that way a

```
StartNewNode
```

is the same as

```
n := CreateNewNode
n.StartProtocol
```


Daeinar:
+1

nikkolasg:
+1

Liamsi:
If the simulations still run, for me it's OK.

PR #338, opened 2016-04-04
better stdoutput for sign msg

ineiti:
There was a stray 'dbg'-message when signing a message - took it away and cleaned up some of the debug-msgs.


Liamsi:
Looks good

nikkolasg:
+1

PR #339, opened 2016-04-04
Osdi ssh

ineiti:
Implementing a simple ssh-key-management using the cosi_cli binary/libraries.


ineiti:
This is now mergeable into `development` without too much hassle! In fact there is a `Host.RegisterExternalMessage` that can register a handler. It is used for the cosi-signature as well as the ssh-ks-functionality. And it could also be used for the SkipChains.

ineiti:
You can check out the branch and do a

```
git diff development lib/sda
```

to see the main differences.

Liamsi:
Or https://github.com/dedis/cothority/compare/development...osdi_ssh if you prefer reviewing it here

nikkolasg:
Question out of the content: Should we merge that into development ? Some features + tests OK, but for cosi_cli, we kept it in its branch no ? Maybe we should do the same here ? Looking at the content righ tnow...

ineiti:
@Liamsi - thanks for the link.
@nikkolasg - as the changes to `lib/sda` are even smaller than with `cosi_cli`, I would +1 for merging it in development. It would also help SkipChains...

app/ssh-kss/ssh-kss.go
```diff
@@ -0,0 +1,90 @@
...
+			if err != nil {
+				dbg.Fatal("Couldn't write config:", err)
+			}
+		}
+		serverApp.Config.List()
+		err = serverApp.Start()
+		if err != nil {
+			dbg.Fatal("Couldn't start server:", err)
+		}
+		<-make(chan bool)
```
nikkolasg: What is this strange dark golang magic ?
Since you are also changing lib/sda/host.go, maybe we could add a Handler `CloseHandler` or a channel, `CloseChannel` which is called / used when the host is closing.
Or make the serverApp.Start() waiting.

lib/ssh-ks/client_app.go
```diff
@@ -0,0 +1,345 @@
...
+
+// ClientApp represents one client and holds all necessary structures
+type ClientApp struct {
+	// This points to the client holding this structure
+	This *Client
+	// Config holds all servers and clients
+	Config *Config
+	// Private is our private key
+	Private abstract.Secret
+}
```
nikkolasg: A lot of "App" appears here while still being in a "lib" folder ... It seems a bit too much on lib and not enough on app/ssh_* (only code for parsing command line). Is it not possible to harmonize it a bit? Keep `Client` in lib and put `ClientApp` in app ? I don't know maybe it's much refactoring I can't tell.

lib/ssh-ks/client_app.go
```diff
@@ -0,0 +1,345 @@
...
+
+// NetworkAddServer adds a new server to the Config
+func (ca *ClientApp) NetworkAddServer(s *Server) error {
+	if ca.Config == nil {
+		return errors.New("No config available yet")
+	}
+	dbg.Lvl3("Servers are", ca.Config.Servers)
+	for _, srv := range ca.Config.Servers {
+		// Add the new server to all servers
+		resp, err := networkSendAnonymous(srv.Entity.Addresses[0],
```
nikkolasg: yeah. we should have that in lib/network => Send anonymously anything you want ! :D

lib/ssh-ks/server_app.go
```diff
@@ -0,0 +1,284 @@
...
+
+// DelServer removes a server from the configuration-list
+func (sa *ServerApp) DelServer(s *Server) error {
+	sa.Config.DelServer(s)
+	return nil
+}
+
+// Start opens the port indicated for listening
+func (sa *ServerApp) Start() error {
+	sa.host = sda.NewHost(sa.This.Entity, sa.Private)
```
nikkolasg: Same comments about the separation between app & lib here.

protocols/cosi/public_api.go
```diff
@@ -0,0 +1,459 @@
...
+		}
+		return &response, nil
+	case <-time.After(time.Second * 10):
+		return nil, errors.New("Timeout on signing")
+	}
+}
+
+// verify takes a file and a group-definition, calls the signature
+// verification and prints the result
+func Verify(fileName, groupToml string) error {
```
nikkolasg: Could you put that directly into lib/cosi ? As there is no network communication nothing complex, it can be put directly in lib/cosi I think, and it's only a wrapper around the existing method that exists to verify (`cosi.VerifySignature`...)
PS: only doing high level comments since it's a very big PR.

protocols/cosi/public_api.go
```diff
@@ -0,0 +1,459 @@
...
+			"doesn't match with the hash of the file.)")
+	}
+	if err := cosi.VerifySignature(network.Suite, fHash, el.Aggregate, sig.Challenge, sig.Response); err != nil {
+		return errors.New("Invalid sig:" + err.Error())
+	}
+	return nil
+}
+
+// GroupToml represents the structure of the group.toml file given to the cli.
+type GroupToml struct {
```
nikkolasg: Same here, I think it should be placed under lib/cosi as it provide nice helpers.

Liamsi:
Besides what @nikkolasg pointed out, I think this can be merged. But I also skimmed through it (too many changes). But I can read it more thoroughly if you prefer that.

app/ssh-kss/ssh-kss.go
ineiti: Yep, will do the same as in `app/cosid/cosid.go`.

lib/ssh-ks/client_app.go
ineiti: Or I put it in `app/ssh-ks-lib`, as it won't be used by others.

lib/ssh-ks/client_app.go
ineiti: It's also useful if you don't know the servers public key as an entry point.

protocols/cosi/public_api.go
ineiti: I wouldn't put too much file-handling and groupToml stuff into `lib/cosi`.

ineiti:
The changes I really would want to have feedback on are in `lib/sda` - this is what I think is critical and might interfere with other stuff. The rest is refactoring `app/cosi` and adding `app/ssh-ks*` and `lib/ssh-ks`, while the latter might be moved to `app/libssh-ks` or something.

lib/sda/host.go
```diff
@@ -20,10 +20,11 @@ import (
 	"golang.org/x/net/context"
 )
 
-/*
-Host is the structure responsible for holding information about the current
- state
-*/
+// ExternalAPI can be used to handle external message-calls
+type ExternalAPI func(*network.Message) network.ProtocolMessage
```
nikkolasg: Would you not need the origin of the message ? `network.Entity`

ineiti: one could add it - here it just replies with a message. Then SDA takes care of sending it back:

https://github.com/dedis/cothority/blob/osdi_ssh/lib/sda/host.go#L408

nikkolasg: Oh yes. OK right.

ineiti:
After fiddling around with `lib/ssh-ks` and `app/libsshks` I finished with `services/ssh-ks` and also added `services/cosi` for completion. The latter needs some cleanup so that it doesn't have to import both `protocols/cosi` and `lib/cosi`.

lib/sda/host.go
```diff
@@ -368,77 +376,46 @@ func (h *Host) processMessages() {
...
+// that will be called if this message arrives
+func (h *Host) RegisterExternalMessage(mt network.ProtocolMessage, f ExternalAPI) network.MessageTypeID {
+	mtID := network.RegisterMessageType(mt)
+	h.externalMsgs[mtID] = f
+	return mtID
+}
+
+// ProcessUnknownMessage takes a message and looks it up in
+// the unknown messages
+func (h *Host) ProcessUnknownMessage(m *network.Message) error {
```
Liamsi: Isn't this name a little misleading? It processes if the message is registered (know) in the externalMsgs map.

Liamsi:
> The changes I really would want to have feedback on are in lib/sda - this is what I think is critical and might interfere with other code. 

OK, I reviewed lib/sda and lib/network: The `ExternalAPI` is a nice and simple idea. We could even use this (or sth similar) for bftcosi (here, we would need such a registered function for every node and not just the root node).

lib/sda/host.go
ineiti: The `Host` has a number of known messages that it processes directly. All other are 'unknown' until they will be recognized by this function.

In `services2`, all messages are 'unknown' or need to be registered. So even the messages now known directly by `Host` will be registered by `Overlay`. Plus the messages from the services.

ineiti:
I hope it won't be too much changes to port it over to `services2`. But depending on the finalisation of `services2` I plan to implement the SkipChain-services with `ExternalAPI`, too.

Issue #340, opened 2016-04-04
Valid signature if file doesn't exist?


Liamsi:
Because of this line https://github.com/dedis/cothority/commit/a8d5a67dc0bd423014d5cba3cc23493bba18b882#diff-f48bbb70f976f52a3bb4bec6863fa7ddL146
we get a valid signature even if the file doesn't exist. Example output:
```
E : (                           main.signFile: 146) - Couldn't read file to be signed: open /var/folders/z7/qx3bqsrx79755g3tzql2s8200051m1/T//data-to-sign.bin: no such file or directory
1 : (                           main.signFile: 150) - &{[93 246 224 226 118 19 89 211 10 130 117 5 142 41 159 204 3 129 83 69 69 245 92 244 62 65 152 63 93 76 148 86] 0e511d32b7c2ef90a583d856cff94be0a23b6e32a3ee704e3cd9e2b90da6e951 0fc8b0de24d58372233917a97912f2b64a14d70ae7573b27957cefe19e08af3d}
1 : (                           main.signFile: 155) - Signature written to: /var/folders/z7/qx3bqsrx79755g3tzql2s8200051m1/T//data-to-sign.bin.sig
```

Liamsi:
Should be `handleErrorAndExit("Couldn't read file to be signed", err)`

PR #343, opened 2016-04-05
Consistent, reasonable error handling

Liamsi:
Exit if we stumbled on an error from which we can't recover from (missing file for instance). Also have spaces between (high level) error user messages and (low-level) errors.Error messages
see #340


ineiti:
OK for me

nikkolasg:
+1

Issue #335, opened 2016-04-04
New functions in `Node`


ineiti:
Some functions that could be added to `Node` - add more if needed.:

- Broadcast - sends to all other nodes
- RegisterHandlers - takes a vararg of functions to register

Daeinar:
I can do that, I've written the relevant code already in JVSS. Should be mostly c&p.

Liamsi:
Done in #345

Issue #295, opened 2016-03-22
Identifying Positions of TreeNodes in Trees


Daeinar:
In certain situations it would be nice to quickly identify (1) the `index` of a given `TreeNode` in a given `Tree` **or** (2) given an `index` and a `Tree` get a pointer to the corresponding `TreeNode`. By `index` I mean here the unique integer when you flatten the `Tree` to a list either via DFS or BFS. I think `ListNodes()` uses DFS.

Is it worth handling such an identification through a general approach embedded into SDA, or do you think that use case is too rare and should instead be handled by the protocol designers themselves (or because some designers iterate through their `Tree` via BFS instead of DFS for whatever reason)?

For the case that we want to add that, what is the proper place to add such a functionality? `Overlay`?

(Iirc we briefly discussed that issue one or two weeks ago when Nicolas ran into that very same problem for JVSS)

Daeinar:
Another (probably obvious) thought: From a usability perspective, it would be probably best to make the above functionality as transparent as possible. Meaning, from within protocol A one should only be able to get information on the location of the `TreeNode` in A's `Tree` and not in the `Tree` of protocol B which might also run in parallel on the same machine.  Maybe it also has security advantages but I am not yet sure about that.

Thoughts?

ineiti:
I saw @nikkolasg adding some index to `TreeNode`s. But I don't remember anymore in what branch.

nikkolasg:
Also related to https://github.com/dedis/cothority/issues/249

nikkolasg:
And added already indexes in the development branch in TreeNode:
https://github.com/dedis/cothority/blob/development/lib/sda/tree.go#L435

Daeinar:
@nikkolasg: Thanks! Didn't see #249.

Daeinar:
The issue with [1] is that it uses Entities instead of TreeNodes. So I guess `EntityIdx` should be renamed to `TreeNodeIdx` and for setting up the latter one should use probably `ListNodes()`, see [2], instead of the `EntityList`.

Thoughts?

[1] https://github.com/dedis/cothority/blob/development/lib/sda/tree.go#L435
[2] https://github.com/dedis/cothority/blob/development/lib/sda/tree.go#L131

nikkolasg:
In fact, these are two different indexes. The one I made is about "give me the index of this `TreeNode`'s public key in the `EntityList`" (I took the simplest approach since we still never uses more than one `TreeNode` per `Entity`).
[...]

Issue #249, opened 2016-03-09
List of nodes on a tree


nikkolasg:
Currently when  a protocols needs to operate without the tree structure, it calls ListNode() which is basically a DFS look up on the tree.
One optimization that sda could benefit is to have the tree as an array of TreeNode.
In this array, each TreeNode would have indexes for its parent & children pointing to elements of this array.  Exactly like some implementations of *heap* data structures.
Benefits:
 *  Tree structure is preserved
 *  the use of all the nodes as a list is *direct*
 * Length of the tree is constant operation

This is *only* an optimization ;)

ineiti:
Is this related to #214?

nikkolasg:
No, #214 is related to how to map a `TreeNode` to its `Entity`. Here I'm only talking about the *construction* of the whole `Tree`. Both optimizations can be made in parallel.

Liamsi:
Done in #345

PR #345, opened 2016-04-06
Node Helper Functions

Daeinar:
This PR introduces various helper functions and constructions in `node.go` and addresses #249, #295,  #335.
- Added a field `treeNodeList` in `node.go`
- Added the following functions:
    - `SendToParent(...)`
    - `SendToChildren(...)`
    - `SendToRoot(...)`
    - `Broadcast(...)`
    - `List()`: To get a list of all TreeNodes
    - `Index()`: To get the index of the calling node in the above list / `EntityList`
    - `RegisterHandlers(...)`: To register a list of SDA handlers

Moreover, it includes the changes in JVSS and RandHound.

### DIFF
--- a/lib/sda/node.go
+++ b/lib/sda/node.go
@@ -18,6 +18,8 @@ type Node struct {
 	token   *Token
 	// cache for the TreeNode this Node is representing
 	treeNode *TreeNode
+	// cached list of all TreeNodes
+	treeNodeList []*TreeNode
 	// channels holds all channels available for the different message-types
 	channels map[network.MessageTypeID]interface{}
 	// registered handler-functions for that protocol
@@ -117,12 +119,56 @@ func (n *Node) IsLeaf() bool {
 	return len(n.treeNode.Children) == 0
 }
 
-// SendTo sends to a given node
+// SendTo sends a given message to a given node
 func (n *Node) SendTo(to *TreeNode, msg interface{}) error {
 	if to == nil {
 		return errors.New("Sent to a nil TreeNode")
 	}
 	return n.overlay.SendToTreeNode(n.token, to, msg)
+}
+
+// SendToParent sends a given message to the parent of the calling node (unless it is the root)
+func (n *Node) SendToParent(msg interface{}) error {
+	if !n.IsRoot() {
+		if err := n.SendTo(n.Parent(), msg); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// SendToChildren sends a given message to all children of the calling node (unless it is a leaf)
+func (n *Node) SendToChildren(msg interface{}) error {
+	if !n.IsLeaf() {
+		for _, node := range n.Children() {
+			if err := n.SendTo(node, msg); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// SendToRoot sends a given message to the root node of the tree (unless the calling node is the root itself)
+func (n *Node) SendToRoot(msg interface{}) error {
+	if !n.IsRoot() {
+		if err := n.SendTo(n.Tree().Root, msg); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Broadcast sends a given message from the calling node directly to all other TreeNodes
+func (n *Node) Broadcast(msg interface{}) error {
+	for _, node := range n.List() {
+		if node != n.TreeNode() {
+			if err := n.SendTo(node, msg); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
 
 // Tree returns the tree of that node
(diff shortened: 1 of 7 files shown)


protocols/randhound/utils.go
```diff
@@ -90,17 +90,8 @@ func (rh *RandHound) hash(bytes ...[]byte) []byte {
...
-
-func (rh *RandHound) sendToChildren(msg interface{}) error {
-	for _, c := range rh.Children() {
-		if err := rh.SendTo(c, msg); err != nil {
-			return err
-		}
-	}
-	return nil
+func (rh *RandHound) index() uint32 {
+	return uint32(rh.Node.Index())
```
Liamsi: Why do you need to cast the index to uint32?

Liamsi:
Looks good

lib/sda/node.go
```diff
@@ -125,6 +127,50 @@ func (n *Node) SendTo(to *TreeNode, msg interface{}) error {
...
 }
 
+// SendToParent sends a given message to the parent of the calling node (unless it is the root)
+func (n *Node) SendToParent(msg interface{}) error {
+	if !n.IsRoot() {
+		if err := n.SendTo(n.Parent(), msg); err != nil {
+			return err
+		}
+	}
+	return nil
```
nikkolasg: Just a stylistic remark: We should prefer doing the *invalid* path first inside the `if` and the rest in the normal track of the function:
```
func (n *Node) SendToParent(msg interface{}) error {
    if n.IsRoot() {
        return nil
    }
    return n.SendTo(n.Parent(),msg)
}
```
I'd say the same for the rest of the functions .

lib/sda/node.go
```diff
@@ -135,6 +181,19 @@ func (n *Node) EntityList() *EntityList {
 	return n.Tree().EntityList
 }
 
+// List returns the list of TreeNodes cached in the node (creating it if necessary)
+func (n *Node) List() []*TreeNode {
+	if n.treeNodeList == nil {
+		n.treeNodeList = n.Tree().List()
+	}
```
nikkolasg: Don't we need a mutex here ? It might be called from different go structures. At least I think...

lib/sda/node.go
ineiti: Shorter and clearer -> +1

protocols/randhound/utils.go
Daeinar: Mostly because it's annoying to Marshal / Hash `int`. So I decided to use `uint32` everywhere in RandHound for indices.

lib/sda/node.go
```diff
@@ -125,6 +127,50 @@ func (n *Node) SendTo(to *TreeNode, msg interface{}) error {
...
+		if err := n.SendTo(n.Parent(), msg); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// SendToChildren sends a given message to all children of the calling node (unless it is a leaf)
+func (n *Node) SendToChildren(msg interface{}) error {
+	if !n.IsLeaf() {
```
ineiti: Same as @nikkolasg remark above

lib/sda/node.go
```diff
@@ -125,6 +127,50 @@ func (n *Node) SendTo(to *TreeNode, msg interface{}) error {
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// SendToRoot sends a given message to the root node of the tree (unless the calling node is the root itself)
+func (n *Node) SendToRoot(msg interface{}) error {
+	if !n.IsRoot() {
```
ineiti: Same as @nikkolasg remark above

lib/sda/node.go
Daeinar: Agreed, I'll change that.

lib/sda/node.go
```diff
@@ -227,6 +286,16 @@ func (n *Node) RegisterHandler(c interface{}) error {
 	return nil
 }
 
+// RegisterHandlers registers a list of given handlers by calling RegisterHandler above
+func (n *Node) RegisterHandlers(handlers []interface{}) error {
```
nikkolasg: Stylistic remark: Instead of slices, why not using `varargs` ? It avoids using a slice of handlers in the protocols.

lib/sda/node.go
ineiti: +1 for a mutex

Daeinar: Or we initialise the variable when `Node` gets created?

lib/sda/node.go
```diff
@@ -227,6 +286,16 @@ func (n *Node) RegisterHandler(c interface{}) error {
...
+	for _, h := range handlers {
+		if err := n.RegisterHandler(h); err != nil {
+			return fmt.Errorf("Error, could not register handler: " + err.Error())
```
ineiti: Use `errors.New("Error, could not register handler: " + err.Error())`

lib/sda/node.go
nikkolasg: As @ineiti said, we might not want to do that for testing reason because when you have 60,000 nodes, it's not something you want to do each time you create a `Node`

Liamsi: I think @nikkolasg wanted to say: from different go routines (intead of structures). But I agree, this should be written in a treadsafe eway. Use RWMutexes (are cheaper) and you write only once, but might read more often (from different go-routines).

ineiti:
Cool - can you just run `simul/do_tests.sh` once you're done?

lib/sda/node.go
Daeinar: Okay. Then I'll add a mutex to Node.

lib/sda/node.go
Daeinar: Agreed.

lib/sda/node.go
nikkolasg: Or 
```
return fmt.Errorf("Error, could not register handler: %s",err.Error())
``` 
:D

Daeinar: OK.

PR #346, opened 2016-04-07
Services - New Try

nikkolasg:
Here comes again the Services branch. After many discussions, design issues, ideology issues and not-enough-whiteboard-space issues, here's a new try at implementing the Services. It includes a lot of changes for the whole lib/sda, as we are now focusing on the "client / daemon" architecture:
* `ProtocolInstance` don't have to register themselves now, since the daemon will never start *itself* a protocol, but only `Services` will start them with some configuration.
* We now define a *interface* that will be the public API that `Service` / `ProtocolInstance` can use to register protocols, send raw messages and in the future, maybe register handlers for certain events and so on..
* `Node` is now `TreeNodeInstance` for more clarity.
* The relation between a `TreeNodeInstance` and a `ProtocolInstance` is now more clear and is enforced by SDA. One `ProtocolInstance` has One `TreeNodeInstance` and it's not possible to put multiple `ProtocolInstance` tied to one `TreeNodeInstance`. The `TreeNodeInstance` is still there to provide some nice utilities such as `SendToParent()`, `IsRoot`, etc.
* `ProtocolInstance` must now implement the `ProcessMessages(msg)` method which is called each time a new message comes for this `ProtocolInstance`. Now the user has the possibility of handling himself the reception / casting of the messages or to use as before the `TreeNodeInstance.RegisterHandler/Channel`. What's important is that the choice is up to the developper of a `ProtocolInstance`.

Some others nice things such a Configurations objects that are passed to the non-root `Services` when they receive the first messages so they know what configuration to pass to the `ProtocolInstance`.

I still link this PR to the previous one #322, but I start a new one from development because it's a lot of refactoring and mostly... the linter is happy so Travis is happy ;)

Ps: of course, doing it in the TDD style.


ineiti:
Did I get that right that `treenode.go` doesn't have any tests yet?

ineiti:
I see this is getting a real big rewrite. Any chance we can do it more linux-kernel-like where the big changes are split up into small ones that each are appliable individually?

Liamsi:
> I see this is getting a real big rewrite. Any chance we can do it more linux-kernel-like where the big changes are split up into small ones that each are appliable individually?

And we could split up the rewrite more easily

nikkolasg:
Each of my commits is a Test failing => Test passing. So each pairs are a small "feature" .
But if I would re-write from scratch, it would definitely be done like that. But here, either I break *everything*  or I add / patch it up little by little.

lib/sda/host.go
```diff
@@ -381,6 +387,20 @@ func (h *Host) processMessages() {
 	}
 }
 
+func (h *Host) processRequest(e *network.Entity, r *Request) {
+	// check if the target service is indeed existing
+	var s Service
+	var ok bool
+	if s, ok = h.serviceStore.serviceByID(r.Service); !ok {
```
ineiti: Beauty-contest: why not write

```
s, ok := h.serviceStore.serviceByID(r.Service); 
if !ok {
```

? It's shorter...

lib/sda/host_msg.go
```diff
@@ -61,10 +67,13 @@ type TokenID uuid.UUID
 type Token struct {
 	EntityListID EntityListID
 	TreeID       TreeID
-	ProtoID      ProtocolID
-	RoundID      RoundID
-	TreeNodeID   TreeNodeID
-	cacheId      TokenID
+	// TO BE REMOVED
```
ineiti: So `ProtoID`, `ServiceID` and `RoundID` will all be stored in `TreeNode`?

BTW: why not `ProtocolID`?

lib/sda/host_msg.go
```diff
@@ -77,13 +86,24 @@ func (t *Token) Id() TokenID {
...
-			t.RoundID.String() + t.ProtoID.String() + t.TreeID.String() +
+			t.RoundID.String() + t.ServiceID.String() + t.ProtoID.String() + t.TreeID.String() +
 			t.TreeNodeID.String()
 		t.cacheId = TokenID(uuid.NewV5(uuid.NamespaceURL, url))
 	}
 	return t.cacheId
 }
 
+func (t *Token) Clone() *Token {
+	t2 := Token{}
```
ineiti: Question: Does `t2 := *t` not work? Will the *IDs not be copied?

lib/sda/overlay.go
```diff
@@ -28,17 +28,26 @@ type Overlay struct {
 	entityListLock sync.Mutex
 	// cache for relating token(~Node) to TreeNode
 	cache TreeNodeCache
+
+	// TreeNodeInstance part
+	instances     map[TokenID]*TreeNodeInstance
+	instancesInfo map[TokenID]bool
```
ineiti: What info is stored in `instancesInfo`?

lib/sda/overlay.go
```diff
@@ -28,17 +28,26 @@ type Overlay struct {
...
+	instancesLock sync.Mutex
+	pis           map[TokenID]ProtocolInstance
```
ineiti: if all other names are written out, I propose to use `protocolInstances` here, too. Even though it's longer...

lib/sda/overlay.go
```diff
@@ -48,41 +57,70 @@ func NewOverlay(h *Host) *Overlay {
...
 		}
-		node = o.nodes[sdaMsg.To.Id()]
-	}
-	// If node is ALREADY DONE => drop packet
-	if isDone {
-		dbg.Lvl2("Dropped message given to node that is done.")
-		o.nodeLock.Unlock()
-		return nil
+		if err := o.RegisterProtocolInstance(pi); err != nil {
+			return errors.New("Error Binding TNI and PI")
```
ineiti: For the error the names should really be written out for clarity.

lib/sda/context.go
```diff
@@ -0,0 +1,76 @@
...
+func (dc *defaultContext) NewTreeNodeInstance(t *Tree, tn *TreeNode) *TreeNodeInstance {
+	return dc.Overlay.newTreeNodeInstance(t, tn, dc.servID)
+}
+
+func (dc *defaultContext) SendRaw(e *network.Entity, msg interface{}) error {
+	return dc.Host.SendRaw(e, msg)
+}
+
+//
+// DESIGN IDEA ... why not !
```
ineiti: Is this what is implemented in `services_scratch`?

lib/sda/overlay.go
```diff
@@ -298,6 +336,55 @@ func (o *Overlay) Close() {
...
+	o.instances[tok.Id()] = tni
+
+	o.RegisterTree(t)
+	o.RegisterEntityList(t.EntityList)
+	dbg.Lvl4(o.host.workingAddress, "Registered NewTreeNodeInstance with Tree + EntityList + ServiceID")
+	return tni
+}
+
+var ErrWrongTreeNodeInstance = errors.New("TreeNodeInstance associated with this ProtocolInstance is already registered")
+var ErrProtocolRegistered = errors.New("A ProtocolInstance already has been registered using this TreeNodeInstance!")
```
ineiti: Question: why are these two errors defined here? Are they re-used elsewhere?

lib/sda/processor.go
```diff
@@ -0,0 +1,50 @@
+package sda
```
ineiti: What does the processor do?

lib/sda/service.go
```diff
@@ -0,0 +1,229 @@
...
+type GenericConfig struct {
+	Type uuid.UUID
+	//Data network.ProtocolMessage
+}
+
+var GenericConfigID = network.RegisterMessageType(GenericConfig{})
+
+// A serviceFactory is used to register a NewServiceFunc
+type serviceFactory struct {
+	cons map[ServiceID]NewServiceFunc
```
ineiti: What is a `cons`?

lib/sda/service.go
```diff
@@ -0,0 +1,229 @@
...
+		names = append(names, n)
+	}
+	return names
+}
+
+// ServiceID returns the ServiceID out of the name of the service
+func (s *serviceFactory) ServiceID(name string) ServiceID {
+	var id ServiceID
+	var ok bool
+	if id, ok = s.translations[name]; !ok {
```
ineiti: Shorter (and other occurences, too):

```
id, ok = s.translations[name]; 
if !ok {
```

lib/sda/service.go
```diff
@@ -0,0 +1,229 @@
...
+// newServiceStore will create a serviceStore out of all the registered Service
+// it creates the path for the config folder of each service. basically
+// ```configFolder / *nameOfService*```
+func newServiceStore(h *Host, o *Overlay) *serviceStore {
+	// check if we have a config folder
+	if err := os.MkdirAll(configFolder, 0777); err != nil {
+		_, ok := err.(*os.PathError)
+		if !ok {
+			// we cannot continue from here
+			panic(err)
```
ineiti: `dbg.Panic`? Also later.

lib/sda/service_test.go
```diff
@@ -0,0 +1,485 @@
...
+	host.Listen()
+	host.StartProcessMessages()
+	dbg.Lvl1("Host created and listening")
+	defer host.Close()
+	// Send a request to the service
+	re := &sda.Request{
+		Service: sda.ServiceFactory.ServiceID("DummyService"),
+		Type:    network.MessageTypeID(uuid.Nil),
+	}
+	// fake a client
```
ineiti: Perhaps we need something like
https://github.com/dedis/cothority/blob/osdi_ssh/services/sshks/api.go#L19
and
https://github.com/dedis/cothority/blob/osdi_ssh/services/sshks/api.go#L28

so a client doesn't need to start up a host to connect to a service.

lib/sda/host.go
nikkolasg: I need the `s Service` variable after. If it's put in the `if` then its scope is binded to the `if` and I can't call `s.ProcessRequest` after.

lib/sda/treenode.go
```diff
@@ -0,0 +1,507 @@
...
+	overlay *Overlay
+	token   *Token
+	// cache for the TreeNode this Node is representing
+	treeNode *TreeNode
+	// channels holds all channels available for the different message-types
+	channels map[network.MessageTypeID]interface{}
+	// registered handler-functions for that protocol
+	handlers map[network.MessageTypeID]interface{}
+	// flags for messages - only one channel/handler possible
+	messageTypeFlags map[network.MessageTypeID]uint32
```
ineiti: When I put `messageTypeFlags` there, I thought we'd have other flags. Now we still don't have any other. Should we change that to `aggregateMessageType map[]bool`? Or is there any hope we'll have other flags?

lib/sda/host_msg.go
nikkolasg: ServiceID and RoundID will be kept in the Token for sure (and the token is given to a TreeNodeInstance, as we are doing now with Node). I am not sure if we still need the ProtocolID since we have already a TreeNodeID , TreeID + EntityListID. I still have to see if that is sufficient to uniquely identify a running PI.

lib/sda/treenode.go
```diff
@@ -0,0 +1,507 @@
...
+}
+
+// EntityList returns the entity-list
+func (n *TreeNodeInstance) EntityList() *EntityList {
+	return n.Tree().EntityList
+}
+
+// Suite can be used to get the current abstract.Suite (currently hardcoded into
+// the network library). Preferably use this function instead of network.Suite
+// when possible.
```
ineiti: Good question - what is preferable and why? Do you already have plans what to do with the `Suite`? If not, I prefer having `network.Suite` all over the code, that makes it more simple to change later.

lib/sda/treenode.go
```diff
@@ -0,0 +1,507 @@
...
+	n.msgDispatchQueueMutex.Lock()
+	n.closing = true
+	if len(n.msgDispatchQueueWait) == 0 {
+		n.msgDispatchQueueWait <- true
+	}
+	n.msgDispatchQueueMutex.Unlock()
+	return n.ProtocolInstance().Shutdown()
+}
+
+func (n *TreeNodeInstance) dispatchHandler(msgSlice []*Data) error {
```
ineiti: Put all private methods to the end of the file.

lib/sda/host_msg.go
nikkolasg: Remember the > 2-days bug in the shamir library, where we even ask for Bryan some help ? 
It was because I also thought that doing the indirection would make a copy but .. Not ! ;)

protocols/cosi/cosi.go
```diff
@@ -323,14 +323,14 @@ func (pc *ProtocolCosi) Cleanup() {
 		pc.DoneCallback(pc.Cosi.GetChallenge(), pc.Cosi.GetAggregateResponse())
 	}
 	close(pc.done)
-	pc.Node.Done()
+	pc.TreeNodeInstance.Done()
```
ineiti: Why can't you call `pc.Done()` here?

lib/sda/overlay.go
nikkolasg: It's not used **yet**. With Node, when a Node is finished, we want to be sure that we drop the messages destined to it that are coming after. Otherwise, the Overlay will just recreate a new Node and dispatch the message to it.
We may be thinking of an other solution, it was a quick hack during the dev of ByzCoin because we had this problem.

lib/sda/host.go
ineiti: This is why I put it in the line above. Then you have it available for later. I can live with the fact that you want to write concise `if`-statements, but if the variables are used outside of the scope of the if, it's bizarre to define the variables beforehand, when you can do it with a `:=`.

lib/sda/overlay.go
nikkolasg: It's for the Service so they know what is wrong instead of just seeing an error text. They can do:
```
if err := c.RegisterProtocolInstance(pi); err == sda.ErrProtocolRegistered {
  // act accordingly
} else err == sda.ErrWrongTreeNodeInstance {
  // act accordingly 
} else {
  // drop everything and run away
}
```
Just for convenience, but not really important.

lib/sda/processor.go
nikkolasg: The same thing I explained but still not implemented. I just layed down the foundations so we can do the things we talked about with 
`ProcessorHandler` and `ProcessorChannels`

lib/sda/service.go
nikkolasg: cons...tructors !

lib/sda/service_test.go
nikkolasg: Yes. I want maybe to push the idea further and having a `Client` struct.  Again. this is only the foundation, I barely  patched up the Services.

lib/sda/treenode.go
nikkolasg: I just simply copied Node => TreeNodeInstance. Did not think about that, but yes I'm guessing we can simplify things a little.

lib/sda/host_msg.go
ineiti: But that was when using pointers, no? Here we just use slices.

```
	tok := &sda.Token{EntityListID: sda.EntityListID(uuid.NewV4())}
	tok2 := *tok
	if tok.EntityListID != tok2.EntityListID {
		t.Fatal("t and t2 should be the same")
	}
	tok2.EntityListID = sda.EntityListID(uuid.NewV4())
	if tok.EntityListID == tok2.EntityListID {
		t.Fatal("t and t2 should differ now")
	}
```
passes. Or do I miss something?

lib/sda/overlay.go
ineiti: Can we name it `instancesSentInfo`? It will be `false` by default, then you can set it to `true` once the info has been sent.

lib/sda/overlay.go
ineiti: I just wonder why you do this here but not in other places. It might be useful, but unity of code is also useful...

lib/sda/host_msg.go
nikkolasg: Try changing the value of an ID in the token to see if it affects the other. If it passes, then it's OK ;)

nikkolasg: Ok . Strange but OK. !

lib/sda/processor.go
ineiti: Can you put experimental things in `services_scratch`? I would be very happy to use your `services2` as soon as possible. We need a service for `SkipChains` and `RandHound`. We can do them using my simple implementation, but if we don't have to rewrite the services once `services2` is done, that would be even more cool.

lib/sda/service.go
ineiti: +1 for writing `constructors` for people like me...

lib/sda/service_test.go
ineiti: OK, just wanted to point out a simple solution to how to access the services.

lib/sda/treenode.go
ineiti: Just would like to have your input on whether you think we'll have other flags one day or not.

lib/sda/host_msg.go
ineiti: Here the IDs are defined as slices which are copied when doing the `:= *`. In the other example you had pointers, which also got copied, but not the structures pointed to... 
If the `Token` would have things like `Tree *Tree`, then the `:= *` would not work.

lib/sda/treenode.go
nikkolasg: Not have precise plan in mind no. But I would have said the contrary ^^ I would prefer to see the code using functions that returns the suite used by the struct, and at the lowest level of the function stack, this `Suite` is defined to be the `network.Suite`. When we want to change, we have to change it only in one place. 
But since we don't really know what we want to do later (support multiple `Suite` ?), I don't know what would be the best solution ..

ineiti: Then let's stick to `network.Suite` and add something else once we know how to do it. Else we have different ways of using the suite in the code.

lib/sda/service.go
Liamsi: I think @nikkolasg wrote `cons` because it is also used in protobuf (ans d I think I saw it in other places too). My personal rule of thumb is: "long" (speaking) names for everything that is exported. For local vars used only in a short scope I use very short var names. For unexported vars/fields (like here), I also tend to use a tradeoff between the two. But clear, "speaking" names never hurt (only while typing ;-))

lib/sda/service.go
```diff
@@ -0,0 +1,229 @@
...
+// start launches a new service
+func (s *serviceFactory) start(name string, c Context, path string) (Service, error) {
+	var id ServiceID
+	var ok bool
+	if id, ok = s.translations[name]; !ok {
+		return nil, errors.New("No Service for this name: " + name)
+	}
+	var fn NewServiceFunc
+	if fn, ok = s.cons[id]; !ok {
+		return nil, errors.New("No Service for this id:" + fmt.Sprintf("%v", id))
```
Liamsi: Later (not important): Use fmt.Sprintf for the whole string (instead of concatenating with +)

Liamsi:
OK, I commented on a very few lines (which are *not* relevant for the big-picture). More comments on the broader design will follow if cosi uses services. In overall: looks good! 
(Btw. Sth. seems to block (see last travis build))

Liamsi:
I will quickly re-enable the -v flag on development, and rerun the last build to see which test is blocking (and hence making travis fail)

Liamsi:
(on my machine everything works ...)

Liamsi:
I tried to make on test work again. I think it will be quite some painful work to get every test running... 
On the other hand `services/cosi_test.go` looks neat and simple. So it's maybe worth the trouble.

lib/sda/host_msg.go
nikkolasg: Ok since we still can instantiate a protocol instance directly (our whole test-code base does that), ProtocolID is kept.

lib/sda/overlay.go
nikkolasg: I don't think we are talking about the same thing. This map is just to store which instances are `Done` so when the messages are received, we don't create a new protocol if its destination is an already `Done` PI. It's not about the config or whatever, so `instancesSentInfo` is misleading for me. If you want to change name, what do you say about `instancesDone` ?

lib/sda/context.go
nikkolasg: design idea ? 
It was just an idea that after evolved into the services_scratch, but it's not implemented.

lib/sda/service_test.go
nikkolasg: I leave it for future discussion, I mainly focus on having this PR mergeable and usable, and future improvements like this can be done in later PR (and we'll surely do what you propose since we can basically copy and paste, while the from_scratch is advancing..)

lib/sda/treenode.go
nikkolasg: From my point of view, I don't think they're gonna be used in the from_scratch branch, since we have this notion of `Processor` interface which can be implemented by `ProcessorHandler` or `ProcessorChannels` or `ProcessorXXX`, the interface themselves will define how to dispatch the messages.
But in this PR I'd keep them as the tests are there to test this functionality and it's working so. Let's finish building this and then we can break other stuff :D

protocols/cosi/cosi.go
nikkolasg: I just changed from `Node` to `TreeNodeInstance`. But I did not check if we registered a callback on TreeNodeInstance; and we don't so I changed to it `pc.Done()`

ineiti:
As discussed WIP till merge with cosi/cosid SkipChains

ineiti:
Merge it...

PR #347, opened 2016-04-08
Errorhandling

Liamsi:
If errors occur do not simply ignore them (only in rare cases or tests). I tried to change the underlying logic of a function as little as possible. Most of the time I just print out the error message using `dbg.Error`. Only where it was obvious I changed the execution logic (for instance if the the function creates a directory using `od.MkDir` and the rest of the code depends on the successful creation of this directory, I just return the error and do not continue the execution of this function)

### DIFF
--- a/lib/crypto/key.go
+++ b/lib/crypto/key.go
@@ -18,17 +18,21 @@ func ReadPub64(suite abstract.Suite, r io.Reader) (abstract.Point, error) {
 // WritePub64 writes a public point to a base64 representation
 func WritePub64(suite abstract.Suite, w io.Writer, point abstract.Point) error {
 	enc := base64.NewEncoder(base64.StdEncoding, w)
-	err := suite.Write(enc, point)
-	enc.Close()
-	return err
+	return write64(suite, enc, point)
+}
+
+func write64(suite abstract.Suite, enc io.WriteCloser, data ...interface{}) {
+	errW := suite.Write(enc, data)
+	if err := enc.Close(); err != nil && errW == nil {
+		return err
+	}
+	return errW
 }
 
 // WriteSecret64 converts a secret key to a Base64-string
 func WriteSecret64(suite abstract.Suite, w io.Writer, secret abstract.Secret) error {
 	enc := base64.NewEncoder(base64.StdEncoding, w)
-	err := suite.Write(enc, secret)
-	enc.Close()
-	return err
+	return write64(suite, enc, secret)
 }
 
 // ReadSecret64 takes a Base64-encoded secret and returns that secret,
--- a/lib/crypto/key.go
+++ b/lib/crypto/key.go
@@ -21,9 +21,9 @@ func WritePub64(suite abstract.Suite, w io.Writer, point abstract.Point) error {
 	return write64(suite, enc, point)
 }
 
-func write64(suite abstract.Suite, enc io.WriteCloser, data ...interface{}) {
-	errW := suite.Write(enc, data)
-	if err := enc.Close(); err != nil && errW == nil {
+func write64(suite abstract.Suite, wc io.WriteCloser, data ...interface{}) error {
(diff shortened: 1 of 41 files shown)


lib/crypto/key.go
```diff
@@ -18,17 +18,21 @@ func ReadPub64(suite abstract.Suite, r io.Reader) (abstract.Point, error) {
...
-	return err
+	return write64(suite, enc, point)
+}
+
+func write64(suite abstract.Suite, wc io.WriteCloser, data ...interface{}) error {
+	errW := suite.Write(wc, data)
+	if err := wc.Close(); err != nil && errW == nil {
+		return err
+	}
+	return errW
```
nikkolasg: Very Stylistic remark (that **definitely** can be > /dev/null), but the simpler the better; you can avoid declaring two variables and have a "one-condition" `if`.
```
func write64(suite abstract.Suite, wc io.WriteCloser, data ...interface{}) error {
	if err := suite.Write(wc, data); err != nil {
               return err
        }
	return wc.Close()
}
```

Liamsi: Good catch. I'll fix the failing tests first, and change that.

Liamsi: I remember why I did this this way. If you do
```
if err := suite.Write(wc, data); err != nil {
               return err
}
``` 
and return in the first `if`: I am not sure if the writer is opened and stays open (`Close` won't be called).

nikkolasg: Godoc on base64:
> the caller must Close the returned encoder to flush any partially written blocks

If the previous blocks are wrong, I guess we don't care much about the end of the buff. The previous bytes are *already* written, it's only for for *end* of the buffer, if the size is not a multiple of 3, that the encoder has to add some padding bytes (see Base64 encoding on Wiki, very well explained ;)), that's why we must call `Close`. But if the rest is faulty, we don't *have* to call `Close`

Liamsi: Ah OK. I thought more about it the file/resource way... then it would need to be closed (not only flushed). But you're right.

nikkolasg: Yeah but your `WriteCloser` comes from the base64 itself, and not from a file or whatever (what we give to `WritePub64` is only a `Writer`, and you create the `WriterCloser` on top of it with base64). So there's no resource non-freed except the internal buffer of the base64 encoder, so that will be GC'd anyway.

Liamsi: That's why I ended with "But you're right" above and immediately changed it ;-)

ineiti:
We could make a `dbg.Err` that takes an err and a vararg of strings, and prints the strings + the error only if the err != nil:

```
if err := h.Close(); err != nil{
    dbg.Error("Couldn't close: ", err)
}
```

would be

```
dbg.Err("Couldn't close: ", h.Close())
```

Liamsi:
> We could make a dbg.Err that takes an err and a vararg of strings, and prints the strings + the error only if the err != nil:

That's a goof idea.

nikkolasg:
It sounds a good idea to reduce code amount but at the same time, I find it very confusing to read
` dbg.Error("...",err)` without printing error. Maybe something like `dbg.ErrorAssert("...",err)`, saying *it will check if the error is non nil and print it* could be cleaner.

Liamsi:
`dbg.IfErr(msg string, err error)`?

nikkolasg:
Whatever name that says clearly what the function is doing is OK for.
For me that PR is ready to merge.

Liamsi:
> Whatever name that says clearly what the function is doing is OK for.
For me that PR is ready to merge.

Cool. Thanks for reviewing /cc @ineiti? 
I'll open a issue for the debug library idea https://github.com/dedis/cothority/issues/349

ineiti:
+1 for merge

PR #352, opened 2016-04-12
changed dbg.Error to dbg.Lvl3 for expected ssh-failures

ineiti:
Some of the ssh-calls in the simulation fail normally, because the cleanup is not necessary. Only when the previous simulation failed the cleanup is needed. So the error should normally be ignored.

### DIFF
--- a/simul/platform/deterlab/users/users.go
+++ b/simul/platform/deterlab/users/users.go
@@ -51,22 +51,13 @@ func main() {
 			defer wg.Done()
 			if kill {
 				dbg.Lvl3("Cleaning up host", h, ".")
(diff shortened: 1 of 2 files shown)


simul/platform/deterlab/users/users.go
```diff
@@ -171,3 +162,10 @@ func deterFromConfig(name ...string) *platform.Deterlab {
 	dbg.SetDebugVisible(d.Debug)
 	return d
 }
+
+// Runs a command on the remote host and outputs eventual error
```
Liamsi: Runs a command on the remote host and outputs eventual error (if debug level >= 3)

Liamsi:
Looks fine

Liamsi:
It's OK to skip the 2-of-3 OKs for this review (barely contains changes)

Issue #348, opened 2016-04-08
ByzCoin BFT


Daeinar:
Implement a standalone version of the ByzCoin consensus algorithm.

Liamsi:
I wanted to start working on that on Monday (probably together with @LefKok). Is it OK if I assign this to myself?

Daeinar:
Sure of course, that's why I added it here. I am not working on that right now. :)

PR #354, opened 2016-04-13
Bftcosi

Liamsi:
related to #348


Liamsi:
Please review (if travis passes ;-) otherwise back to WIP it ...

lib/sda/node.go
```diff
@@ -372,6 +372,7 @@ func (n *Node) dispatchHandler(msgSlice []*Data) error {
 
 func (n *Node) reflectCreate(t reflect.Type, msg *Data) reflect.Value {
 	m := reflect.Indirect(reflect.New(t))
+	// FIXME data race ...
```
ineiti: Is this done now?

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,604 @@
...
+	doneProcessing chan bool
+
+	// finale signature that this BFTCoSi round has produced
+	finalSignature *BFTSignature
+}
+
+// NewBFTCoSiProtocol returns a new bftcosi struct
+func NewBFTCoSiProtocol(n *sda.Node, verify VerificationFunction) (*ProtocolBFTCoSi, error) {
+	// initialize the bftcosi node/protocol-instance
+	bft := new(ProtocolBFTCoSi)
```
ineiti: Please use the standard constructor we use everywhere else:

```
bft := &ProtocolBFTCoSi{
    Node: n,
    suite: n.Suite(),
```

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,604 @@
...
+	bft.commit = cosi.NewCosi(n.Suite(), n.Private())
+	bft.verifyChan = make(chan bool)
+	bft.doneProcessing = make(chan bool, 2)
+	bft.doneSigning = make(chan bool, 1)
+	bft.verificationFun = verify
+	bft.aggregatedPublic = n.EntityList().Aggregate
+	bft.threshold = int(2.0 * math.Ceil(float64(len(bft.Tree().List()))/3.0))
+
+	// register channels
+	n.RegisterChannel(&bft.announceChan)
```
ineiti: There was a pull-request about error-handling ;)

protocols/bftcosi/packets.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// challengeChan is the type of the channel that will be used to catch the
+// challenge messages.
+type challengePrepareChan struct {
+	*sda.TreeNode
+	ChallengePrepare
+}
+
+type challengeCommitChan struct {
```
ineiti: Can we also have some small documentation for this one? yes, it's private. But all other structure have something. Pretty please?

protocols/bftcosi/bftcosi_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/dedis/cothority/lib/dbg"
+	"github.com/dedis/cothority/lib/sda"
+	"github.com/stretchr/testify/assert"
+)
+
+// Dummy verification function: always returns OK/true/no-error on data
+var veriCount int
+var countMut sync.Mutex
+
+func veri(m []byte, ok chan bool) {
```
ineiti: veri -> verification or verify

protocols/bftcosi/bftcosi_test.go
```diff
@@ -0,0 +1,88 @@
...
+	countMut.Lock()
+	veriCount++
+	dbg.Print("Verification called", veriCount, "times")
+	countMut.Unlock()
+	dbg.Print("Ignoring message:", string(m))
+	// everything is OK, always:
+	ok <- true
```
ineiti: +1 for Nicolas idea to return a bool instead of working with channels here.

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,615 @@
...
+		// append response
+		resp, err := bft.prepare.Response(bft.tempPrepareResponse)
+		if err != nil {
+			return err
+		}
+		bzrReturn.Response = resp
+	}
+
+	dbg.Lvl3("BFTCoSi Handle Response PREPARE")
+	// if I'm root, we are finished, let's notify the "commit" round
```
ineiti: Comments about the true if-branch should be inside the if-branch. Here it's 50/50 ;) Either outside

`// Check whether we need to notify the commit round`

or inside

`// Notify 'commit'-round as we're root`

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,615 @@
...
+	err := bft.SendTo(bft.Parent(), r)
+	bft.Done()
+	return err
+}
+
+// handleResponseCommit handles the responses for the commit round during the
+// response phase.
+func (bft *ProtocolBFTCoSi) handleResponseCommit(r *Response) error {
+	// check if we have enough
+	// FIXME possible data race
```
ineiti: Is that very difficult to fix? Either remove the possible data race or the *FIXME*.

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,615 @@
...
+
+	// store the exceptions for later usage
+	bft.tempExceptions = ch.Exceptions
+	dbg.Lvl3("BFTCoSi handle Challenge COMMIT")
+	if bft.IsLeaf() {
+		return bft.startResponseCommit()
+	}
+
+	// send it down
+	for _, tn := range bft.Children() {
```
ineiti: we now have a `Node.SendToChildren`.

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,615 @@
...
+	chal := bft.prepare.Challenge(ch.Challenge)
+	ch.Challenge = chal
+
+	dbg.Lvl3(bft.Name(), "BFTCoSi handle Challenge PREPARE")
+	// go to response if leaf
+	if bft.IsLeaf() {
+		return bft.startResponsePrepare()
+	}
+	var err error
+	for _, tn := range bft.Children() {
```
ineiti: `Node.SendToChildren`

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+
+	// send challenge + signature
+	cc := &ChallengeCommit{
+		TYPE:      RoundCommit,
+		Challenge: c,
+		Signature: bft.prepare.Signature(),
+	}
+	dbg.Lvl3("BFTCoSi Start Challenge COMMIT")
+	for _, tn := range bft.Children() {
```
ineiti: `Node.SendToChildren`

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,615 @@
...
+		doneProcessing:   make(chan bool, 2),
+		doneSigning:      make(chan bool, 1),
+		verificationFun:  verify,
+		aggregatedPublic: n.EntityList().Aggregate,
+		threshold:        int(2.0 * math.Ceil(float64(len(n.Tree().List()))/3.0)),
+	}
+
+	// register channels
+	if err := n.RegisterChannel(&bft.announceChan); err != nil {
+		return bft, err
```
ineiti: Shouldn't it be `return nil, err`? because the `bft` is not valid at this point, no?

protocols/bftcosi/bftcosi.go
Liamsi: Sorry, removed the comment in node.go but forgot this one. It is already fixed...

ineiti: +1

ineiti:
From a syntactical point of view it looks OK. I didn't check the functionality.

protocols/bftcosi/bftcosi.go
Liamsi: I let the caller decide what to do with the half-baked bft instance. But, yes, could also be nil.

protocols/bftcosi/bftcosi.go
Liamsi: Cool, thanks. Will change that.

protocols/bftcosi/bftcosi.go
ineiti: put it nil, so nobody's tempted to use it anyway.

protocols/bftcosi/bftcosi.go
Liamsi: Hmm, on my local test this doesn't change anything. But in reality (and maybe in other places besides the one pointed out), sometimes you would want to continue sending even if one send failed (on child went offline). I still changed it like you proposed (almost) everywhere. But I'm sure we will need another function, which continues sending soon.

ineiti: Valid point. Perhaps we should change `SendToChildren` so

- it sends to all children anyway
- it only returns an error if it was a leaf

protocols/bftcosi/bftcosi.go
Liamsi: OK

Liamsi:
> From a syntactical point of view it looks OK. I didn't check the functionality.

Thanks for the review. @LefKok checked the functionality.

protocols/bftcosi/bftcosi.go
Liamsi: > it sends to all children anyway

OK, just print the error (so we can see, that one or more nodes aren't reachable)

> it only returns an error if it was a leaf

Why would you want to return an error if it was a leaf? This implicates one always needs a n.IsLeaf  before calling SendToChildren (to not cause an error). I'm OK with that, just asking what the reasoning is behind that.

Regarding this issue: I will revert every call which continued sending to manually send again (instead of SendToChildren)

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,615 @@
...
+	ch.Challenge = chal
+
+	dbg.Lvl3(bft.Name(), "BFTCoSi handle Challenge PREPARE")
+	// go to response if leaf
+	if bft.IsLeaf() {
+		return bft.startResponsePrepare()
+	}
+	var err error
+	for _, tn := range bft.Children() {
+		err = bft.SendTo(tn, ch)
```
ineiti: Here you'll only catch the error of the last child.

Liamsi: I know. I will add a function to node which collects the errors and won't stop sending...

ineiti: For me this could even be the default behaviour of the function.

Another question is whether the `SendTo` should go into a go-routine. Useful if you have lots of children. Specially in the set-up phase, where a connection can take quite some time until it is set up (n * 100ms)...

Liamsi: I will do exactly this (I'll resolve it directly on this branch though without opening another issue)

nikkolasg: >Another question is whether the SendTo should go into a go-routine. Useful if you have lots of children
Do you mean One go routine for sending or One go routine / destinations = child.
I guess the first solution is not necessary because the caller can do it itself very easily if it wants to : 
`go bft.SendToChildren(msg)`
The second solution might be a good candidate, but I would also let the user choose by providing him both methods. One that sends sequentially in ONE go routine and the other by using one goroutine for each destination. What do you think ?

ineiti: 2nd solution, add a `go` in the for-loop of `SendToChildren`

protocols/bftcosi/bftcosi.go
nikkolasg: > Why would you want to return an error if it was a leaf? This implicates one always needs a n.IsLeaf before calling SendToChildren (to not cause an error)

It should be the expected behavior, a user of SDA should not send messages to children if it is a leaf. otherwise sda can return to him a `sda.ErrSendToNilChildren` error ( stg like that).

protocols/bftcosi/bftcosi.go
```diff
@@ -0,0 +1,615 @@
...
+func (bft *ProtocolBFTCoSi) nodeDone() bool {
+	dbg.Lvl3(bft.Name(), "nodeDone()")
+	bft.doneProcessing <- true
+	if bft.onDoneCallback != nil { // only true for the root
+		bft.onDoneCallback()
+	}
+	return true
+}
+
+const preparePrefix = 0x1
```
nikkolasg: can you explain what this variable is about ?

nikkolasg:
Except my few comments and of Linus, I'm OK to merge.

protocols/bftcosi/bftcosi.go
Liamsi: Added a function `SendToChildrenInParallel`

protocols/bftcosi/bftcosi.go
Liamsi: It was there to differentiate between the phases. It's not necessary anymore. @LefKok: Are you OK, with removing it completely?

LefKok: Yes we can remove it, however in the prepare round we should sign the hash
of the bft.Msg not the actual Msg.

2016-04-15 17:20 GMT+02:00 Ismail Khoffi <notifications@github.com>:

> In protocols/bftcosi/bftcosi.go
> <https://github.com/dedis/cothority/pull/354#discussion_r59891127>:
>
> > +func (bft *ProtocolBFTCoSi) RegisterOnSignatureDone(fn func(*BFTSignature)) {
> > +	bft.onSignatureDone = fn
> > +}
> > +
> > +// nodeDone is either called by the end of EndProtocol or by the end of the
> > +// response phase of the commit round.
> > +func (bft *ProtocolBFTCoSi) nodeDone() bool {
> > +	dbg.Lvl3(bft.Name(), "nodeDone()")
> > +	bft.doneProcessing <- true
> > +	if bft.onDoneCallback != nil { // only true for the root
> > +		bft.onDoneCallback()
> > +	}
> > +	return true
> > +}
> > +
> > +const preparePrefix = 0x1
>
> It was there to differentiate between the phases. It's not necessary
> anymore. @LefKok <https://github.com/LefKok>: Are you OK, with removing
> it completely?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/dedis/cothority/pull/354/files/f7e372d4f0f0ecddf30556b3495f5b540096bffc#r59891127>
>



-- 

Kokoris-Kogias Eleftherios
PhD Candidate at School of  Computer  and Communication Sciences
EPFL

Liamsi: That's not the case already? Look at `startChallengePrepare` and `startChallengeCommit`?

protocols/bftcosi/packets.go
Liamsi: Of course, why not.

Liamsi:
> Except my few comments and of Linus, I'm OK to merge.

All resolved (and more). Then do it :-)

protocols/bftcosi/bftcosi.go
LefKok: Sorry I missed that. Yes remove the structs no need for them anymore

PR #357, opened 2016-04-18
Services skipchain

ineiti:
Addin skipchains to the services.


ineiti:
OK - I threw away the `gob`-encoder and used `network.Marshal` only. I don't think it's possible to replace our usual tree/encoding/decoding, as we rely on the fact that the `Roster/EntityList` should be transferred separately from the `Tree/TreeNodeList`.

lib/sda/processor.go
```diff
@@ -0,0 +1,126 @@
...
+
+	errI := ret[1].Interface()
+
+	if errI != nil {
+		return &StatusRet{errI.(error).Error()}
+	}
+
+	reply := ret[0].Interface()
+	if reply == nil {
+		reply = &StatusRet{""}
```
nikkolasg: We can return something more explicit such as "Empty return message" or whatelse you think of.

ineiti: `StatusOK` which is simply `&StatusRet{""}` ?

lib/sda/processor.go
```diff
@@ -0,0 +1,126 @@
...
+	reply := p.GetReply(e, cr)
+	if err := p.SendRaw(e, reply); err != nil {
+		dbg.Error(err)
+	}
+}
+
+// ProcessServiceMessage is to implement the Service interface.
+func (p *ServiceProcessor) ProcessServiceMessage(e *network.Entity,
+	s *ServiceMessage) {
+	cr := &ClientRequest{
```
nikkolasg: I'm pretty sure I wrote a comment about that but I can't find it anymore (..?):
In `GetReply` you can directly pass the data as `[]byte` instead of creating a `ClientRequest` for a `ServiceMessage`.
It's small but as said in my forgotten-in-the-abyss comment, if we can keep the number of small hacks like this leaning over 0, that'd be great.

lib/sda/service.go
```diff
@@ -227,11 +231,9 @@ func (s *serviceStore) serviceByID(id ServiceID) (Service, bool) {
 // * Data: contains all the information of the request
 type ClientRequest struct {
 	// Name of the service to direct this request to
-	Service ServiceID `json:"service_id"`
-	// Type is the type of the underlying message
-	Type network.MessageTypeID `json:"type"`
```
nikkolasg: good riddance

lib/sda/processor.go
ineiti: So you want me to rewrite `GetReply` to take `data[]`?

lib/sda/service.go
ineiti: Now the `ClientRequest` only has `data[]` - so we could take it away, no?

services/skipchain/api.go
```diff
@@ -0,0 +1,130 @@
...
+// NewClient instantiates a new client with name 'n'
+func NewClient() *Client {
+	return &Client{Client: sda.NewClient("Skipchain")}
+}
+
+// CreateRootInter creates two Skipchains: a root SkipChain with
+// maximumHeight of maxHRoot and an intermediate SkipChain with
+// maximumHeight of maxHInter. It connects both chains for later
+// reference.
+func (c *Client) CreateRootInter(elRoot, elInter *sda.EntityList, maxHRoot, maxHInter int, ver VerifierID) (root, inter *SkipBlock, err error) {
```
nikkolasg: I guess I'm not familiar with the code enough to deeply understand the design of it. However regarding the "api", I don't understand why we fix with this 3-level structure and don't have more generic structure that would allow greater flexibility. Again, I'm not so familiar so I might be wrong but I'd love to see the api more simpler, stg along the lines of:
```
// Create a skipchain with maximum height, and which
// is of type `chainType` which represents which type of 
// skipchain it is -> Roster Skipchain, or Timevault Skipchain,
// DNS Skipchain, or pure bytes Skipchain, etc.
func (c *Client) CreateSkipchain( height, chainType, ...  ) { ... }

// Appending a block to the skipchain,
// The block can contain a Roster or pure Bytes or whatever, 
// It all depends on the type of the Skipchain.
func (c *Client) AddBlock( chainID, data ) { ... }
 
// DERIVATION of AddData
func (c *Client) AddRoster( chainID, roster) { 
     ...
     c.AddBlock(chainID,rosterBlock)
 }
func (c *Client) AddData( chainID, data) { 
   ...
   c.AddBlock(chainID, dataBlock)
 }

// Add a parent-child link between two skipchains. 
// That creates the relationship between two skipchain 
// so one is the parent of the other.
func (c *Client) AddParentChildLink( parentID, childID ) { ... }
```
With a structure like this, it's easy to produce our 3-level layout but it's also easy to maintain one skiplist only easily (as we want to do for stuff like CONIKS and so on), or a 2-level layout as we might not always need an intermediate.

Again, it's very abstract design and I'm not so much familiar with the code yet, **and there may be subtleties I did not get**, but in general a simpler design leads to a better outcome ! 
Of course, it's not something that have to be done now, but depending on what you all think (feedbacks please!), it might be a good goal for a next iteration on this complex subject...

ineiti: In fact the only thing that is missing from the API is the `AddParentChildLink`. The rest is there, with a twist:
We cannot have a `AddData` and `AddRoster`, as we have to verify the SkipBlocks first. This is why there is a `ProposeData` and `ProposeRoster`.
The `CreateRootInter` and `CreateData` are convenience API-calls because they will probably be used in most of the cases to initialize the SkipChains. What I should do is to have those two functions use `Propose.*`.

services/skipchain/skipchain.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+var skipchainSID sda.ServiceID
+
+// Service handles adding new SkipBlocks
+type Service struct {
+	*sda.ServiceProcessor
+	// SkipBlocks points from SkipBlockID to SkipBlock but SkipBlockID is not a valid
+	// key-type for maps, so we need to cast it to string
+	SkipBlocks map[string]*SkipBlock
```
nikkolasg: I think Golang does not allow this is because a slice is not comparable [0]. One solution to this is to use a fixed size array, and we can do that since SkipBlockID is only a hash, and a hash has a fixed size output (32 with sha256).
Having 
```
type HashID [32]byte
type SkipBlockID HashID
```
can solve it I guess.

[0] : https://blog.golang.org/go-maps-in-action

ineiti: I didn't want to set the HashID as an `array of bytes` because if we ever change the hashing-method, we'd have to adjust the HashID-definition. Which might be OK. But still thinking (hoping) that one day we'll have more than one encryption-suite going at the same time...

services/skipchain/api.go
nikkolasg: Yes I got the terminology wrong, it's `Propose{Data,Roster}`, right.
> The CreateRootInter and CreateData are convenience API-calls

I get it, because this is what we needed right now, but you directly jumped from nothing to these convenience calls instead of building smaller blocks (creating simply one skipchain, adding a simple relationship between two skipchains etc) to get to a `CreateRootInter` etc. I'm not saying it does not have to be there, I'm more saying that it should be decomposed because I'm not sure about your assumption 
> because thy will probably be used in most of the cases

Will it always be true ? We don't know (and the way I understood CONIKS with skipchain it did not use 3-levels but I may be wrong) so I think it's best to build small building blocks and create "convenience" calls on top of it. Exactly how we do our helpers like "TreeNodeInstance" in `lib/sda`, a ProtocolInstance developer can do all the hard work it self if it wants to, or just register easy handlers functions.
**Again**, it's not something to do here, more a thought on the whole design in general that could be good to think about for a next iteration.

services/skipchain/skipchain.go
nikkolasg: Right. But if we have more than one encryption suite, we'll still have the same problem. So it turns out that it may be not the best way to **use** the suite hashes for this, **because** it's not a fixed size. We may be better off saying that SkipBlockID are sha256 IDs and that's it, unrelated to whatever suite we use. I think it's quite common to fix a hashing algorithm for stuff like that. We won't get bitten in the foot when we will want to change to another suite. Like in `lib/sda`, we sticked to `UUID` which are `[16]byte`. What do you think ?

services/skipchain/api.go
nikkolasg: Ok after discussion with @ineiti , I get it that a `Propose{Roster,Data}` can be used as `CreateSkipchain` if the block you give does not have any back links.

services/skipchain/skipchain.go
ineiti: I don't have any idea what is reasonable to do here ;) We can ask Bryan tomorrow morning. Or by mail.

services/skipchain/api.go
ineiti: Will update comments to reflect that better.

services/skipchain/msgs.go
```diff
@@ -0,0 +1,118 @@
...
+	LatestID SkipBlockID
+}
+
+// GetUpdateChainReply - returns the shortest chain to the current SkipBlock,
+// starting from the SkipBlock the client sent
+type GetUpdateChainReply struct {
+	Update []*SkipBlock
+}
+
+// SetChildrenSkipBlock adds a child-SkipBlock to a parent SkipBlock
```
nikkolasg: > adds a child-SkipBlock to a parent SkipBlock
It is not a "parent-child" link ? Or is it itself a Skipblock that makes the link between the two ?

ineiti: `add link to a child-SkipBlock`

services/skipchain/skipchain.go
```diff
@@ -0,0 +1,321 @@
...
+		if s.verifyNewSkipBlock(latest, prop) {
+			s.updateSkipBlockLinks(latest, prop)
+		} else {
+			return nil, errors.New("Verification error")
+		}
+		if err := s.startPropagation(latest); err != nil {
+			return nil, err
+		}
+	} else {
+		// TODO: allow for other verificators
```
nikkolasg: How do you envision such verificators system ? Using some kind of registration like ProtocolInstance and Service, stg like: ` RegisterSkipchainVerificators( ... )` ?
Don't you think it makes sense to implement that in this PR ? We'll likely need it very soon if we want to start using it, no ?

services/skipchain/skipchain.go
```diff
@@ -0,0 +1,321 @@
...
+		proposed.Index++
+		// genesis block has a random back-link:
+		bl := make([]byte, 32)
+		rand.Read(bl)
+		proposed.BackLinkIds[0] = bl
+		// empty forward link:
+		proposed.updateHash()
+	} else {
+		proposed.Index = prev.Index + 1
+		// TODO: add higher backlinks
```
nikkolasg: All the followings `TODO` are not crucial for this PR but I guess we should already create an issue for those before we forget to implement those features no ?

ineiti:
yes, some kind of registration is needed. It depends how fast other people want to use it. If you want to wait for the `identity`-service and `ssh`-application, I'll put the PR back into `Work in progress`. But I think that's at least a week, if not two, in the future.

This PR was a *preview* of the skipchain service ;)

nikkolasg:
To put back in written form what was just discussed:
It all depends on what we want to do next. But I guess we have time until NSDI, so I think it would be better to *finish* this PR by implementing the *few* TODOS that are left, merge it and then move on with the applications. We already know that we need those TODOs (verifiers + height at least)  and we already have an idea of what the application will need from this servce. Having one thing merged instead of 2 in WIP (skipchain + application) is also much more clean. Well, let's also see what Bryan plans to do with us.

services/skipchain/skipchain_test.go
```diff
@@ -0,0 +1,259 @@
...
+	blockCount++
+	assert.Equal(t, blockCount, latest.Index)
+	// the genesis block has a random back-link:
+	assert.Equal(t, 1, len(latest.BackLinkIds))
+	assert.NotEqual(t, 0, latest.BackLinkIds)
+
+	next := NewSkipBlock()
+	next.Data = []byte("And the earth was without form, and void; " +
+		"and darkness was upon the face of the deep. " +
+		"And the Spirit of God moved upon the face of the waters.")
```
nikkolasg: We even have skipchains out of the Bible ! 
We call them ... **Divine Skipchains**.

lib/sda/processor.go
nikkolasg: That means we allow for Service to *not* reply to client all the times. That's ok for me. So yeah let's go with `StatusOK`

ineiti: The clients will get a reply - which is `StatusOK`.

lib/sda/processor.go
nikkolasg: if you also think that makes sense yes. I do for the simple reason that our code should be consistent most of the times and we should avoid using tricks most of the time where it can be avoided. Ex. of where it can *not* be avoided is when dealing with OS, or language limitations or any external API.

lib/sda/service.go
nikkolasg: Well, that's still good to have a wrapper around it (`ClientRequest`). What are you talking about exactly ?

services/skipchain/skipchain_test.go
ineiti: Now we're sure SkipChains didn't start more than 3000 years before now...

nikkolasg: And maybe more importantly that they had not suffered any kind of modifications over the years !!

lib/sda/service.go
ineiti: Taking away the wrapper, so we don't have the `ClientRequest` structure anymore. As it is a structure containing only `data[]`...

services/skipchain/api.go
ineiti: Finally the `Propose` == `Create/Add` did only hold up in the service but was badly reflected in the API. So now we DO have
`CreateRoster`
`ProposeRoster`
`CreateData`
`ProposeData`

mostly because in the case of `Create` we need extra arguments. So instead of cramming it all in 2 usage patterns of one function, I created two functions.

PR #360, opened 2016-04-19
Medco protocol is killed after 2min00s

ChristianMct:
This is where I'm stuck implementing the Medco protocol.

## Situation
Simulating the protocol with the current parameters (10'000 messages) in localhost or on deterlab triggers the bug. 

## Symptoms
The non-root nodes are silently killed after **exactly 2 minutes**.
The number of message sent doesn't seem to be causing the crash, but is indeed influenced by the running time somehow bounded to 2 minutes.


ineiti:
As per slack: add

```
CloseWait = 600
```
to your `.toml`-file

PR #361, opened 2016-04-19
Cosi app using services2

Liamsi:


Liamsi:
@nikkolasg @ineiti : please have a look at https://github.com/dedis/cothority/blob/cosi_app_services2/app/cothorityd/cothorityd_test.go#L46 (starting a daemon like this will create a never ending test) currently I [skipped the test](https://github.com/dedis/cothority/commit/a173ff5adfcd91abdf6c61047eec6aafb0a12a72).

PR #362, opened 2016-04-20
Cosi app into services2

nikkolasg:
refer to #361


app/cothorityd/cothorityd_test.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"bytes"
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+	"testing"
+)
+
+func TestCompileAndRun(t *testing.T) {
```
ineiti: I propose we replace `TestCompileAndRun` with the `test_cosi.sh`-script.

ineiti:
Cool - can we have a running `test_cosi.sh`-script? That would be very nice.

lib/sda/host.go
```diff
@@ -387,14 +336,42 @@ func (h *Host) processMessages() {
 			m := data.Msg.(ServiceMessage)
 			h.processServiceMessage(data.Entity, &m)
 		default:
-			dbg.Error("Didn't recognize message", data.MsgType)
+			err := h.ProcessExternalMessage(&data)
```
ineiti: Oups - where does this comes from? I think this should be removed, now that services is running - no?

services/cosi/encode.go
```diff
@@ -0,0 +1,61 @@
...
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"github.com/dedis/cothority/lib/crypto"
+	"github.com/dedis/cothority/lib/network"
+	"strings"
+)
+
+// MarshalJSON implements golang's JSON marshal interface
+func (s *ServiceResponse) MarshalJSON() ([]byte, error) {
```
ineiti: Do we use JSON or just rely on protobuf for the moment? For me Protobuf would be OK...

app/cosi/cosi.go
```diff
@@ -0,0 +1,374 @@
...
+	res, err := signStatement(r, el)
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+// signStatement can be used to sign the contents passed in the io.Reader
+// (pass an io.File or use an strings.NewReader for strings)
+func signStatement(read io.Reader, el *sda.EntityList) (*s.ServiceResponse,
```
ineiti: I propose we have the services defining an API that can be called from any client wanting to use it - like I did in

https://github.com/dedis/cothority/blob/services_skipchain/services/skipchain/client.go
and
https://github.com/dedis/cothority/blob/services_skipchain/services/skipchain/api.go

what do you think?

services/cosi/encode.go
Liamsi: This is for writing out the received signature. JSON is more human readable (and for instance easier to parse from a web-frontend)

nikkolasg: Erf. I'm struggling here. On one side, I also think we should stick to one encoding technique. But on the other side, a JSON signature is really really nicer than a protobuf signature. I'm not sure if we should impose stg strict here as long as we did not resolve our issue about defining clear boundaries between External API Services vs Internal messages.

ineiti: If it's for writing into a `.config`-file, let's keep JSON.

PR #364, opened 2016-04-20
Adding service client

ineiti:
Adding a client-structure for easy access to services.


ineiti:
I just thought that we should put `sda.ErrMsg` inside of `sda.Client.Send` and return the 'merged' `err` from `sda.Client.Send`.

lib/sda/service.go
```diff
@@ -278,3 +282,120 @@ func CreateServiceMessage(service string, r interface{}) (*ServiceMessage, error
...
+		Entity:    network.NewEntity(kp.Public, ""),
+		Private:   kp.Secret,
+		ServiceID: ServiceFactory.ServiceID(s),
+	}
+}
+
+// NetworkSend opens the connection to 'dst' and sends the message 'req'. The
+// reply is returned, or an error if the timeout of 10 seconds is reached.
+func (c *Client) Send(dst *network.Entity, msg network.ProtocolMessage) (*network.Message, error) {
+	client := network.NewSecureTCPHost(c.Private, c.Entity)
```
nikkolasg: Why waiting on the `send` ? Why not directly into the `NewClient` ? We may want to use a client many times and not only one. Creating a longterm  `TCPHost` seems more efficient.
Do we want to use one `Client` for different `Services` addresses ? Because, directly connecting to the `Services` at the beginning also seems more efficient, but I don't know what our use cases are. For me, it seems logical that one `Client` connects to one `Service` but then again, that's my opinion. What do you think ?

lib/sda/service.go
```diff
@@ -278,3 +282,120 @@ func CreateServiceMessage(service string, r interface{}) (*ServiceMessage, error
...
+	select {
+	case response := <-pchan:
+		dbg.Lvlf5("Response: %+v", response)
+		// Catch an eventual error
+		err := ErrMsg(&response, nil)
+		if err != nil {
+			return nil, err
+		}
+		return &response, nil
+	case <-time.After(time.Second * 10):
```
nikkolasg: make it a field of `Client` so we don't ever have comments like "my simulation timeouted !" :D

lib/sda/service_test.go
```diff
@@ -444,6 +449,34 @@ func TestServiceBackForthProtocol(t *testing.T) {
 	assert.Equal(t, resp.Val, 10)
 }
 
+func TestClient_Send(t *testing.T) {
```
nikkolasg: Where did we start writing test using underscores ? Golang is all about CamelCase...

lib/sda/service.go
```diff
@@ -278,3 +282,120 @@ func CreateServiceMessage(service string, r interface{}) (*ServiceMessage, error
...
+holds the private and public key and allows to connect to a service through
+the network.
+The error-handling is done using the ErrorRet structure which can be returned
+in place of the standard reply. The Client.Send method will catch that and return
+ the appropriate error.
+*/
+
+// Client for a service
+type Client struct {
+	Private abstract.Secret
```
nikkolasg: I would put `Private` to `private`, or even not keeping at as a field. But here, it says it may be useful to a user to play with that `Private` while it's not.

lib/sda/service.go
ineiti: `Client` will be implemented in each service, so there is a `cosi.Client`. It is not necessary it always connects to the same `ServerIdentity`(tm) but may change. Else it should be called `ClientConnection`.

lib/sda/service.go
ineiti: Field of `Client` and then a setter like `SetTimeout`?

lib/sda/service_test.go
ineiti: It's IntelliJs doing. The case here is that `Client` is a structure and `Send` is a method on that structure. It follows https://golang.org/pkg/testing/#pkg-overview

lib/sda/service.go
ineiti: Put it private, as we need it for setting up the tcphost.

lib/sda/service.go
nikkolasg: Stg like that yes.

lib/sda/service_test.go
nikkolasg: Where did you see any non-CamelCase code there ? 
The golang codebase use everywhere CamelCase, these are the conventions ...

ineiti: Just follow the link I provided...

Liamsi: Hi, I won't vote if there should be an _ or not here (just because I don't find it very important) ... But: @ineiti the link you provided does not say it is convention to put a `_` in between (only for examples: `Example_ForBlaBlaEtc`). In the google code itself you won't find an underscore (only in the Example tests). But I'm OK to make this a convention in *our* code (as it's not wrong/not a problem to put an `_` in between Struct and member function).

ineiti: You have to scroll down and do an interpolation of `Example` to `Test`. But I think it does hold:

The naming convention to declare examples for the package, a function F, a type T and method M on type T are:

```
func Example() { ... }
func ExampleF() { ... }
func ExampleT() { ... }
func ExampleT_M() { ... }
```

ineiti: Yes, running `egrep -r "func.*Test.*_.*\(" .` on the go-source doesn't give back as many examples for Tests including '_' as I thought, so this argument doesn't hold.

As it's an issue of style, I propose we just vote on it (nice we are three ;) - no blank bulletins. I like the idea of having `TestT_M`, so +1 for me.

nikkolasg: I like `TestM` +1

PR #365, opened 2016-04-21
Time Vaults

Daeinar:
Implementation of the time vaults protocol which allows to encrypt a message with a shared secret that is  decrypted automatically after a given amount of time.


ineiti:
According to discussion with @Daeinar

PR #367, opened 2016-04-21
Fixed Readme

Liamsi:
install on packages works (instead of single files `go install cosi/cosi.go` which fails)


ineiti:
You introduced a data-race by changing a README.md? Well done ;) If you add me to the Travis-admins, I'll re-launch the Travis.

ineiti:
Will probably pass using #368

PR #368, opened 2016-04-25
Add service processor

ineiti:
Adds a service processor, so that services can register functions as do protocols. This is a multi-package. It also does:
- change the TreeNodeCache to a struct and add a mutex
- make Network a little less chatty in debug-mode

### DIFF
--- a/lib/sda/processor.go
+++ b/lib/sda/processor.go
@@ -10,7 +10,7 @@ import (
 
 // ServiceProcessor allows for an easy integration of external messages
 // into the Services. You have to embed it into your Service-structer,
-// then it will offer an 'AddMessage'-method that takes a message of type
+// then it will offer an 'RegisterMessage'-method that takes a message of type
 // 	func ReceiveMsg(e *network.Entity, msg *anyMessageType)(error, *replyMsg)
 // where 'ReceiveMsg' is any name and 'anyMessageType' will be registered
 // with the network. Once 'anyMessageType' is received by the service,
@@ -29,8 +29,8 @@ func NewServiceProcessor(c Context) *ServiceProcessor {
 	}
 }
 
-// AddMessage puts a new message in the message-handler
-func (p *ServiceProcessor) AddMessage(f interface{}) error {
+// RegisterMessage puts a new message in the message-handler
+func (p *ServiceProcessor) RegisterMessage(f interface{}) error {
 	ft := reflect.TypeOf(f)
 	// Check we have the correct channel-type
 	if ft.Kind() != reflect.Func {
--- a/lib/sda/processor.go
+++ b/lib/sda/processor.go
@@ -1,0 +1,116 @@
+package sda
+
+import (
+	"errors"
+	"reflect"
+
+	"github.com/dedis/cothority/lib/dbg"
+	"github.com/dedis/cothority/lib/network"
+)
+
+type ServiceProcessor struct {
+	functions map[network.MessageTypeID]interface{}
+	Context
+}
+
+func NewProcessor(c Context) *ServiceProcessor {
+	return &ServiceProcessor{
+		functions: make(map[network.MessageTypeID]interface{}),
+		Context:   c,
+	}
+}
+
+// AddMessage puts a new message in the message-handler
+func (p *ServiceProcessor) AddMessage(f interface{}) error {
+	ft := reflect.TypeOf(f)
+	// Check we have the correct channel-type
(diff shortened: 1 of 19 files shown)


lib/sda/processor.go
```diff
@@ -0,0 +1,125 @@
...
+
+	if err := p.SendRaw(e, reply); err != nil {
+		dbg.Error(err)
+	}
+}
+
+// ProcessServiceMessage is to implement the Service interface.
+func (p *ServiceProcessor) ProcessServiceMessage(e *network.Entity,
+	s *ServiceMessage) {
+	cr := &ClientRequest{
```
nikkolasg: `ClientRequest` for `ServiceMessage` ?

lib/sda/processor.go
```diff
@@ -0,0 +1,125 @@
...
+// NewServiceProcessor initializes your ServiceProcessor.
+func NewServiceProcessor(c Context) *ServiceProcessor {
+	return &ServiceProcessor{
+		functions: make(map[network.MessageTypeID]interface{}),
+		Context:   c,
+	}
+}
+
+// AddMessage puts a new message in the message-handler
+func (p *ServiceProcessor) AddMessage(f interface{}) error {
```
nikkolasg: The name `AddMessage` is confusing for me in this context. What do you think of `processor.RegisterHandler(fn)` (so it's clear we expect a function that handles a certain type of message), or `processor.Handler(fn)` (short minimalistic golang-alike version ... )

lib/sda/processor.go
```diff
@@ -0,0 +1,125 @@
...
+		return errors.New("Need 2 return values: network.ProtocolMessage and *error*")
+	}
+	// Automatic registration of the message to the network library.
+	dbg.Lvl3("Registering handler", cr1.String())
+	typ := network.RegisterMessageUUID(network.RTypeToMessageTypeID(
+		cr1.Elem()),
+		cr1.Elem())
+	p.functions[typ] = f
+	return nil
+}
```
Liamsi: I'm not sure, but doesn't this repeat exactly the same (reflection-)code buried somewhere in the network library or somewhere else in the sda framework? If yes, I would prefer to refactor it such that we have the common code appear in one place, if I'm wrong, then we can keep it like this.

lib/sda/processor.go
ineiti: `GetReply` wants to have a `ClientRequest`, which in the case of a `ServiceMessage` has only the `Data`-field set.

lib/sda/processor.go
ineiti: `AddMessage` is bad, `RegisterHandler` reminds me too much of the `Protocol`s. Can we take `RegisterMessage`?

lib/sda/processor.go
ineiti: No, this one is more elaborate ;) In fact the `Protocol`-reflection code searches for a specific message-structure (that I REALLY would like to change but never managed to), which is different here. Also do we have return values that need to be checked and used, so it's not the same code.

lib/sda/processor.go
nikkolasg: I understand but instead of using another type, the `GetReply` method only uses the type and the buffer (`Data`), so why not passing that directly ?  Moreover, I made a mistake when defining the `ClientRequest` struct, because since we use `lib/network`, the `type` is already in the buffer, and there's no need to have a field explicitely telling the type of the message; it's error prone (one thing at two places, your saying!). (I originally made that because I thought we could use JSON encoding but it turns out that's it's very painful).
Something like  `CallHandlers(entity,buffer)`  should suffice for both cases (client messages and internal service messages)

It's not **essential** but I think it's important to keep the number of hacks like this (transform input parameter into another one so we can call a function with it) should be kept to a minimum and should tend towards 0...

ineiti: OK, sounds reasonable. Shall I also take out the `Type` from the `ClientRequest`?

lib/sda/processor.go
nikkolasg: We can think of it as a third way of processing messages, although this one is ... straightforward as it is only a one-to-one mapping (one message input => one message output).

lib/sda/processor.go
nikkolasg: Cool! Yes definitely. but that might require some changes elsewhere, you'll have to see if it's big or not, otherwise you can just ignore this field and we'll remove it gradually.

PR #371, opened 2016-04-27
Add service processor to cosi

ineiti:
This gives us a running example of how to use the service processors.


Liamsi:
OK for me. Did you test if the CLI still works?

nikkolasg:
Fine for me also as long as the CLI still works.

ineiti:
Tested using the `app/test_cosi.sh` and it says 'success'.

Issue #372, opened 2016-04-28
network.Message field values are null value after decoding


ChristianMct:
It looks like the `network.Message` instances are not decoded correctly, the actual fields values after decoding is the null value.

This bug I found tracking down two other bugs:

1. The `*network.Entity` passed to service's handlers points to a "null-valued" `Entity`. It's not so clear to me whether it should point to the Client's or the Service's `Entity`, but it seems to be the only way the handler can know on which `Entity` it is running so I would opt for the latter. (?). The reason for this bug is that the `Entity` passed to the handler is supposed to be decoded from the `service.ClientRequest` message, but there is a bug there.
2. Returned value from the `sda.Client.Send` method are decoded to the correct struct but with the null-valued fields values.

I'm not a reflect expert so I could not find the origin of the bug. Pull request #373

ChristianMct:
1. Not a bug, the `Entity` of a `Client` has empty address field.
2. Only exported field are taken into account by protobuf.

PR #373, opened 2016-04-28
Bug situation of issue #372 

ChristianMct:
This is sets up the situation explained in issue #372.

Lines of interest:
[in host.go](https://github.com/ChristianMct/cothority/blob/message_null_value/lib/sda/host.go#L326)
[in MedcoClient.go](https://github.com/ChristianMct/cothority/blob/message_null_value/services/medco/medco_client.go#L31)


ChristianMct:
#372 has been solved more quicker than I expected.

PR #374, opened 2016-04-28
Add the network.Entity to the Context

ChristianMct:
It would be very useful  to have the Service's `network.Entity` as part of the `Context` so we can access it from the Services.


nikkolasg:
OK for me.

Liamsi:
Also OK for me

PR #377, opened 2016-05-13
dbg-library outputs to stderr for dbg.Print,Warning,Error,Fatal,Panic

ineiti:
As title says...

### DIFF
--- a/lib/dbg/debug_lvl.go
+++ b/lib/dbg/debug_lvl.go
@@ -84,11 +84,11 @@ var StaticMsg = ""
 var regexpPaths, _ = regexp.Compile(".*/")
 
 const (
-	lvlPrint = iota - 10
-	lvlWarning
+	lvlWarning = iota - 10
 	lvlError
 	lvlFatal
 	lvlPanic
+	lvlPrint
 )
 
 func lvl(lvl int, args ...interface{}) {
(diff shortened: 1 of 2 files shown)


nikkolasg:
Why stderr for `dbg.Print` ?

ineiti:
Because `dbg.Print` should only be used short-term to get some debugging messages and then disappear. But if you think it should go into stdout, that's also possible.

nikkolasg:
Yes but that does not mean it should be on stderr, at least the name does not imply anything like that. +1 for stdout. Otherwise, OK for merging !

nikkolasg:
BTW, Bryan did not mention it but do you think it makes sense to also include the possibility of writing to a file ?  Like most daemon simply write their logs to a file (or syslog, whatever). Not sure yet.

ineiti:
He did mention something like that. So if you want to add a flag, that's good for me.

nikkolasg:
Do you think it makes sense to it on this PR or another one ?

ineiti:
I think the writing to a file is the job of the cosi-binary, not of the debugging-library. If you want me to add a 'logging'-possibility by redirecting stdout and stderr to log-files, that's possible. But I didn't understand your request to be that.

nikkolasg:
Closed . Refers to https://github.com/dedis/cothority/pull/380

Issue #375, opened 2016-05-12
Usability issues in standalone cosi/cothorityd


Liamsi:
There are a bunch of cosi-cli usability issues.
@nikkolasg started to work on the server part in the [usability branch](https://github.com/dedis/cothority/tree/usability).

I will work on the client part. After that we'll merge the `cosi` and the `cothorityd` into one binary. After testing we will move the code to a seperate cosi repo.

Liamsi:
TODOs (mainly for myself) on the client:

- [ ] `cosi check` outputs some obscure error message (coming from codegangsta cli)
- [ ] `cosi sign -help` usage instructions are confusing (`-sig` option)/"The ‘sign’ and ‘verify’ commands should be consistent" & use stdin/stdout “by default”
- [ ] don't use the `dbg` library; or at least ditch the colors
- [ ] keep the std output clean of other output (if everything is fine, otherwise use std err) 
- [ ] rename -servers to ‘-group’ or ‘-def’ (group definition) 
- [ ] update README according to all points above

ineiti:
You can keep the `dbg`-library for debug output, but then do
```
if ! flag("-v"){
    dbg.SetDebugVisible(0)
} else {
    dbg.SetDebugVisible(3)
}
dbg.SetUseColors(false)
```

Liamsi:
If we add a `-v` flag as above the user still won't be able to use stdin/stdout with that flag turned on.

ineiti:
which might be acceptable - using `-v` is for debugging purposes mainly. I'm trying to write a test for the `dbg`-library to have `dbg.Error` and others on stderr - but it feels quite awkward...

Liamsi:
Found a new (minor) usability glitch. If you are running a server and type in the IP with IP non-reachable from outside. You'll see sth like:
```
[+] Enter the IP address you would like others cothority servers and client to contact you: 
[-] Invalid IP address given (%!!(MISSING)(EXTRA string=, string=))
```

Liamsi:
I merged both applications (`cosi` and `cothirtyd`) into one binary. The `cothorityd` functionality is now available through `cosi server`. It needs more (human) testing though.

I also created a [cosi repo](https://github.com/dedis/cosi). I didn't put anything into it yet, as we first need to make sure that the packages the cosi app imports from the `cothority` (or later maybe only from `sda` or whatever it will be called will exist even if the code changes (by using https://gopkg.in for instance). 
[...]

PR #379, opened 2016-05-13
Usability

Liamsi:
Related to #375


app/cosi/cosi.go
```diff
@@ -98,139 +99,144 @@ func main() {
...
 	if len(el.List) > 1 {
 		// Then check pairs of servers
 		for i, first := range el.List {
 			for _, second := range el.List[i+1:] {
 				es := []*network.Entity{first, second}
-				checkList(sda.NewEntityList(es))
+				go checkList(sda.NewEntityList(es), &wg)
 				es[0], es[1] = es[1], es[0]
-				checkList(sda.NewEntityList(es))
+				go checkList(sda.NewEntityList(es), &wg)
```
ineiti: What's the advantage of parallelizing that? I would rather have the tests go sequential to avoid problems.

nikkolasg: That's what @bford asked for and I think it makes sense because it's faster. These are all independent checks so why not ?

app/cothorityd/cothorityd.go
```diff
@@ -1,127 +1,400 @@
-/*
```
ineiti: You should add some documentation to the cothorityd, so it shows up nicely in `godoc`.

app/cothorityd/cothorityd.go
```diff
@@ -1,127 +1,400 @@
...
+	}
+	host.ListenAndBind()
+	host.StartProcessMessages()
+	host.WaitForClose()
+
+}
+
+// CreateCothoritydConfig will ask through the command line to create a Private / Public
+// key, what is the listening address
+func interactiveConfig() {
```
ineiti: That looks like a huge function - can it be split up somehow?

ineiti:
Are the bash-script-tests still running?

app/cothorityd/cothorityd.go
nikkolasg: Usually I would agree but here everything is closely related and it's a *one shot* function. If I split it, there would be a part about writing the config file, generating the keys and asking for (maybe 2) ip addresses, but none of these functions would be called elsewhere, because they are only needed here.
But yes, everything is splittable anyway, so I can try to split it up :D

nikkolasg:
@ineiti yes test_cosi.sh returns success on branch usability

app/cosi/cosi.go
ineiti: If the master says so - merge at least the two for-loops and put the `wg.Add(2)` before line 133.

ineiti:
In fact it was `interactiveConfig` I thought is too long - but well...

ineiti:
shall I merge it?

nikkolasg:
`interactiveConfig` may be long and it's something we can discuss, is that OK if we *postpone* this disucssion until we have something working ?
And I'm just re-writing the for-loops thing before you can merge.

ineiti:
Go for it. Will be put on the 'eternal TODO'-list like so many others...

PR #380, opened 2016-05-17
Usability dbg stdout

nikkolasg:
Directly to development because we merged usability before this one.
Refer to #377 for details.


ineiti:
Hah - that's my PR - so it's your decision ;)

nikkolasg:
Hehe. Right. Merging then.

PR #382, opened 2016-05-18
Identity service for ssh-keystorage

ineiti:
This replaces #376 for better naming and no dependencies on the old `skipchain`-branch.


ineiti:
Will re-open against #383

PR #383, opened 2016-05-19
SDA-changes and new propagation-protocol

ineiti:
Some aesthetical changes to SDA and a new protocol for propagating data across a tree - very useful in services.


lib/sda/local.go
```diff
@@ -58,10 +58,10 @@ func (l *LocalTest) StartProtocol(name string, t *Tree) (ProtocolInstance, error
 
 // CreateNewNodeName takes a name and a tree and will create a
 // new Node with the protocol 'name' without running it
-func (l *LocalTest) CreateProtocol(name string, t *Tree) (ProtocolInstance, error) {
+func (l *LocalTest) CreateProtocol(t *Tree, name string) (ProtocolInstance, error) {
```
ineiti: Make this compatible with `overlay.CreateProtocol`

lib/network/struct.go
```diff
@@ -201,8 +201,8 @@ type Entity struct {
 // EntityID uniquely identifies an Entity struct
 type EntityID uuid.UUID
 
-// Equals returns true if both EntityID are equal or false otherwise.
-func (eid EntityID) Equals(other EntityID) bool {
+// Equal returns true if both EntityID are equal or false otherwise.
+func (eid EntityID) Equal(other EntityID) bool {
```
ineiti: All other structures have `Equal`, so follow common naming.

lib/sda/context.go
```diff
@@ -1,16 +1,17 @@
...
-import (
-	"github.com/dedis/cothority/lib/network"
-)
+import "github.com/dedis/cothority/lib/network"
 
-// Context is the interface that is given to an Service
+// Context is the interface that is given to a Service
 type Context interface {
-	NewTreeNodeInstance(*Tree, *TreeNode) *TreeNodeInstance
+	NewTreeNodeInstance(*Tree, *TreeNode, string) *TreeNodeInstance
```
ineiti: The service-name is needed so the `Token` can be used correctly.

PR #388, opened 2016-05-24
searching se bug

ineiti:
There is (at least one) bug with regard to sending multiple parallel requests from the same host.

### DIFF
--- a/lib/sda/service_test.go
+++ b/lib/sda/service_test.go
@@ -9,6 +9,7 @@ import (
 	"gopkg.in/dedis/cothority.v0/lib/dbg"
 	"gopkg.in/dedis/cothority.v0/lib/network"
 	"gopkg.in/dedis/cothority.v0/lib/sda"
+	"sync"
 )
 
 type DummyProtocol struct {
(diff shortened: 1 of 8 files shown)


lib/sda/service_test.go
```diff
@@ -113,8 +114,6 @@ func (ds *DummyService) ProcessServiceMessage(e *network.Entity, s *sda.ServiceM
 }
 
 func TestServiceNew(t *testing.T) {
-	defer dbg.AfterTest(t)
-	dbg.TestOutput(testing.Verbose(), 4)
 	ds := &DummyService{
```
Liamsi: Just a question: Why is it reasonable to remove:
```
defer dbg.AfterTest(t)		
dbg.TestOutput(testing.Verbose(), 4)
``` 
here?

Liamsi:
Looks good!

lib/sda/service_test.go
ineiti: Because it is done in `TestMain` in `sda_test.go`

PR #391, opened 2016-05-25
Host verify connection

ineiti:
Handle connections better:
- timeout to prune stale connections
- if a write fails, try to re-open it

### DIFF
--- a/lib/sda/export_test.go
+++ b/lib/sda/export_test.go
@@ -55,3 +55,9 @@ func (h *Host) AbortConnections() error {
 	close(h.ProcessMessagesQuit)
 	return h.host.Close()
 }
+
+func (h *Host) CloseConnections() error {
+	h.networkLock.Lock()
+	defer h.networkLock.Unlock()
+	return h.closeConnections()
+}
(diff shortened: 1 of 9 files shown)


ineiti:
This is just in the case a `SendRaw` fails to re-open the connection and to try again.

lib/sda/export_test.go
```diff
@@ -47,3 +47,21 @@ func (o *Overlay) TokenToNode(tok *Token) (*TreeNodeInstance, bool) {
...
+
+func (h *Host) AbortConnections() error {
+	for _, c := range h.connections {
+		h.unregisterConnection(c)
+	}
+	close(h.ProcessMessagesQuit)
+	return h.host.Close()
+}
+
+func (h *Host) CloseConnections() error {
```
nikkolasg: CloseConnection should it not unregister the connection too ? Since it's closed we won't use it anymore ..

ineiti: Line 53 is not what you're looking for?

ineiti: Sorry - `CloseConnection` initialises the `connections`-map anyway.

ineiti:
Superseeded by #395

PR #393, opened 2016-05-25
Host verify connection 

nikkolasg:
But without the forceSendError private field.


nikkolasg:
Travis failing because of v0 stuff

PR #400, opened 2016-05-27
Clean 1/2 - `app/`

ineiti:
Split up the big `service_sda` PR - 1/3: cleanup of the `app/`-directory


app/lib/server/server.go
```diff
@@ -0,0 +1,499 @@
...
+	"strings"
+	"time"
+
+	"github.com/dedis/cothority/app/lib/config"
+	"github.com/dedis/cothority/lib/crypto"
+	"github.com/dedis/cothority/lib/dbg"
+	"github.com/dedis/cothority/lib/network"
+	"github.com/dedis/cothority/lib/sda"
+	// Empty imports to have the init-functions called which should
+	// register the protocol
```
Liamsi: Move this comment in front of empty imports

app/lib/server/server.go
```diff
@@ -0,0 +1,499 @@
...
+	oi.Info("Setting up a cothority-server.")
+	oi.Inputf("Please enter the [address:]PORT for incoming requests [%d]: ", DefaultPort)
+	reader := bufio.NewReader(os.Stdin)
+	var str = readString(reader)
+	// let's dissect the port / IP
+	var hostStr string
+	var ipProvided = true
+	var portStr string
+	var serverBinding string
+	splitted := strings.Split(str, ":")
```
Liamsi: use golang's well-tested internal functions instead: https://golang.org/pkg/net/#SplitHostPort

app/lib/oi/oi.go
```diff
@@ -0,0 +1,102 @@
+package oi
```
Liamsi: We might need a better name for this package: output (?) or as we just discussed merge it with package `dbg`

doc.go
```diff
@@ -0,0 +1,14 @@
+package cothority
+
+/*
+This repository implements a of the collective authority (cothority) framework.
```
Liamsi: typo/grammar

app/lib/server/server.go
```diff
@@ -0,0 +1,499 @@
+package server
+
```
Liamsi: is this just renaming the file `app/cothorityd/config.go` or are there any other changes?

ineiti: Mostly renaming plus some things learnt from the `cosi`-binary. Sooner or later `cosi` should use this library, too.

app/lib/server/server.go
ineiti: Long live `goimports`

doc.go
ineiti: I just copy/pasted it from the README.md - also corrected that one...

app/lib/server/server.go
```diff
@@ -70,31 +70,28 @@ func InteractiveConfig(binaryName string, ed25519 bool) {
 	var ipProvided = true
 	var portStr string
 	var serverBinding string
-	splitted := strings.Split(str, ":")
+	if !strings.Contains(str, ":") {
+		str = ":" + str
+	}
+	host, port, err := net.SplitHostPort(str)
```
nikkolasg: did you tested it ?

Liamsi: If you tried if this still works I'm OK with merging this PR.

app/lib/ui/ui_test.go
```diff
@@ -25,6 +25,10 @@ func TestInput(t *testing.T) {
 	assert.Equal(t, "[?] Question [def]: ", getOutput())
 	setInput("")
 	assert.Equal(t, "def", Input("def", "Question"))
+	dbg.Print("two-lines-input")
+	setInput("1\n2")
+	assert.Equal(t, "1", Input("", "Question1"))
+	assert.Equal(t, "2", Input("1", "Question2"))
```
Liamsi: Do we need/use that?

simul/runfiles/nottest_byzcoin_ntree.toml
```diff
@@ -0,0 +1,7 @@
+Servers = 5
+Simulation = "ByzCoinNtree"
+Rounds = 2
+BF = 2
```
Liamsi: A general remark as this seem to happen quite often: It's strange that files you rename appear as deleted (in one commit) and create in another. It's difficult to see the changes/diff if we do it like this (which is bad for reviewing the code).

Liamsi:
OK for me

app/lib/config/config.go
```diff
@@ -128,25 +128,34 @@ type ServerToml struct {
 	Description string
 }
 
-// ReadGroupToml reads a group.toml file and returns the list of Entity
-// described in the file.
-func ReadGroupToml(f io.Reader) (*sda.EntityList, error) {
+// ReadGroupDescToml reads a group.toml file and returns the list of Entities
+// and descriptions in the file.
+func ReadGroupDescToml(f io.Reader) (*sda.EntityList, []string, error) {
```
nikkolasg: Should we create a `Server` & `Group` struct that contains the decoded data instead of passing each data differently ?
Not sure, but if we ever add a new field to a `Server` (like certificate when we'll do TLS) then we have to return again an additional return argument...

app/lib/server/server.go
```diff
@@ -0,0 +1,474 @@
...
+
+// Service used to get the port connection service
+const whatsMyIP = "http://www.whatsmyip.org/"
+
+// How long we're willing to wait for a signature
+var RequestTimeOut = time.Second * 1
+
+// interactiveConfig will ask through the command line to create a Private / Public
+// key, what is the listening address
+func InteractiveConfig(binaryName string, ed25519 bool) {
```
nikkolasg: Instead of a `bool`, I would have prefered a separate function that clearly indicates the difference, since it is *hack*. But we still have to discuss with Bryan about these ed25519 issues so it may not be necessary after (if for example we implement a *real* Ed25519 suite)

app/lib/server/server.go
```diff
@@ -0,0 +1,474 @@
...
+		ui.Fatal("Error formating private key to hexadecimal. Abort.")
+	}
+	var point abstract.Point
+	if ed25519 {
+		// use the transformation for ed25519 signatures
+		//point = cosi.Ed25519Public(network.Suite, kp.Secret)
+	} else {
+		point = kp.Public
+	}
+	pubStr, err := crypto.PubHex(network.Suite, point)
```
Liamsi: `point` is not initialized if `ed25519==true`. Is this on purpose?

app/lib/server/server.go
Liamsi: Also it's not clear what the purpose of this bool flag is (at least to me). It's only used below (here https://github.com/dedis/cothority/pull/400/files#r65341628)

app/lib/server/server.go
nikkolasg: Since it is now in a library, it should not I guess

app/lib/server/server.go
ineiti: As discussed, will remove it and wait for a decision on how to go forward with regard to ed25519 formats.

app/lib/server/server.go
```diff
@@ -0,0 +1,474 @@
...
+		return err
+	}
+
+	buffer, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	if !bytes.Contains(buffer, []byte("1")) {
+		return errors.New("Address unrechable")
```
nikkolasg: My bad but you copied it so ... typo error here *unreAchable*

app/lib/config/config.go
ineiti: Can you give a short code example? I don't understand what you want to do.

app/lib/server/server.go
ineiti: Probably got lost somewhere between copying in the 5 branches, sorry.

simul/runfiles/nottest_byzcoin_ntree.toml
ineiti: yes, checkout/merge between 5 branches is difficult.

app/lib/server/server.go
ineiti: same as above - will be removed.

app/lib/server/server.go
ineiti: I'll have to teach IntelliJ to look into strings for typos.

app/lib/config/config.go
ineiti: something like

```
type Group struct{
  EL *sda.EntityList
  Description map[sda.EntityID]string
}
```
?

nikkolasg: More like :
```
type Group struct {
   Description string
   // either an array or a map, not sure yet. 
   // maps performs poorly when put under huge loads
   // do we need to address a server directly from the group from its EntityID ?
   // I think we only need that from the entityList PoV
   Servers []Server
}
type Server struct {
   Description string
   Entity *network.Entity
   Certificate ...
   Owner ...
   ...
}
```
What do think ?

ineiti:
Merge?

app/lib/server/server.go
Liamsi: Is this resolved?

Liamsi:
> Merge?

Did you review the now outdated comments?

app/lib/server/server.go
Liamsi: Did you?

nikkolasg:
Ok for me. merging.

PR #401, opened 2016-05-27
Cleanup 2/2 - `lib/`

ineiti:
Split up the big `service_sda` PR - 2/3: cleanup of the `lib/`-directory and also the protocols and services who depend on them


lib/cosi/cosi.go
```diff
@@ -104,6 +104,14 @@ type Signature struct {
 	Response  abstract.Secret
 }
 
+// NewSignature returns a pre-initialized signature with Secret.One
+func NewSignature(s abstract.Suite) *Signature {
```
nikkolasg: Why should you need such a thing ?

ineiti: IIRC to be able to correctly marshal a 'nil'-signature.

nikkolasg: To marshal or unmarshal ? 
* To marshal a nil-signature -> you should not do that and it should throw an error if you try to do that. If you marshal  a `NewSignature()`, it just hides a bug.
* To unmarshal, the network library does it for you with `UnmarshalRegisteredType`.

More generally, a signature should be available only when the `Cosi` struct is ready to deliver one or delivered by the network. I don't really see the point of having a *default signature* (signature of what ? with what cosigners ? etc), but maybe you used it somewhere where it was needed ?  A simple `Ctrl-F` search in the PR did not show anything.

lib/sda/context.go
```diff
@@ -29,15 +30,36 @@ func newDefaultContext(h *Host, o *Overlay, servID ServiceID) *defaultContext {
...
+// CreateProtocol makes a TreeNodeInstance from the root-node of the tree and
+// prepares for a 'name'-protocol. The ProtocolInstance has to be added later.
+func (dc *defaultContext) CreateProtocol(t *Tree, name string) (ProtocolInstance, error) {
+	pi, err := dc.Overlay.CreateProtocolService(dc.servID, t, name)
+	return pi, err
+}
+
+// CreateProtocolAuto is like CreateProtocol but doesn't bind a service to it,
+// so it will be handled automatically by the SDA.
+func (dc *defaultContext) CreateProtocolAuto(t *Tree, name string) (ProtocolInstance, error) {
```
nikkolasg: Mhhh. that's becoming confusing. 
* Why a `Context` should need to create a protocol not tied to a `Service` (using `CreateProtocolAuto`) when a `Context` is *only* used by  a `Service` ? We only create protocols without a service in our tests & simulations.

* I find it confusing to have both but we did not find any better way for the moment so it's fine; but maybe we could be more helpful with a more precise naming ? `CreateProtocolService` or `CreateServiceProtocol` // and `CreateProtocolIndependant` or `CreateIndependantProtocol` ? I don't know, I'm not sure, but both are very similar and it's not easy to remember which one is the right one to use just by looking at the name `CreateProtocolAuto`

lib/sda/local.go
```diff
@@ -229,6 +229,23 @@ func (l *LocalTest) GetPrivate(h *Host) abstract.Secret {
...
+func (l *LocalTest) GetServices(hosts []*Host, sid ServiceID) []Service {
+	services := make([]Service, len(hosts))
+	for i, h := range hosts {
+		services[i] = l.Services[h.Entity.ID][sid]
+	}
+	return services
+}
+
+// MakeHELS is an abbreviation to make a Host, an EntityList, and a service
+func (l *LocalTest) MakeHELS(nbr int, sid ServiceID) ([]*Host, *EntityList, Service) {
```
nikkolasg: Do you use it somewhere ? It just appears on this PR but I can't see any use of it. 
And also, don't you think that'd more useful to have a function that returns the `Service` from the `Host` (or from `LocalTest`) BY Host ? And not just returning the `Service` of the **first** host ? Like 
```
func (l *LocalTest) Service(h *Host,sid *ServiceID) Service
```
Maybe not, I'm not sure, it's depending on the test you've made I guess.  know that in the `service_test.go`, I often needed to have control over the two services (in a 2-hosts setup). 
Orthogonal to that, we could add a comment saying that it returns the `Service` of the first `Host` == root.

lib/sda/protocol.go
```diff
@@ -24,17 +25,13 @@ type ProtocolInstance interface {
 	// Start on it.
 	Start() error
 	// Dispatch is called at the beginning by SDA for listening on the channels
-	// XXX Should remove that => not using ncessarily channels
```
nikkolasg: Why remove the comment ?? It's still accurate, we don't necessarily uses chanels, so why should we have a dispatch method ?

lib/sda/service.go
```diff
@@ -295,7 +297,7 @@ in place of the standard reply. The Client.Send method will catch that and retur
 
 // Client for a service
 type Client struct {
-	private abstract.Secret
+	Private abstract.Secret
```
nikkolasg: Do we have a use-case where a user of a `Client` uses this `Private`,i.e. why putting `Private` as a exported/public field ? This morning I don't see any.

lib/sda/overlay.go
```diff
@@ -234,12 +241,20 @@ func (o *Overlay) Close() {
 // CreateProtocol returns a fresh Protocol Instance with an attached
 // TreeNodeInstance
 func (o *Overlay) CreateProtocol(t *Tree, name string) (ProtocolInstance, error) {
```
nikkolasg: Same here regarding the method name, we should clearly distinguish between creating a protocol tied to a Service and creating a protocol which is independent. `CreateProtocol` is the default here but we don't know its behavior except if we look that there is also `CreateProtocolService`.

lib/sda/processor.go
```diff
@@ -81,20 +82,47 @@ func (p *ServiceProcessor) ProcessClientRequest(e *network.Entity,
...
+		return err
 	}
-	p.GetReply(e, cr)
+	dbg.Lvl4("Raw-sending to", e)
+	return p.SendRaw(e, sm)
+}
+
+// SendISMOthers sends an InterServiceMessage to all other services
+func (p *ServiceProcessor) SendISMOthers(el *EntityList, msg network.ProtocolMessage) error {
+	errStrs := []string{}
```
nikkolasg: Nitpick: we should prefer `var errStrs []string` because it does not allocate anything if we never append on it. 
https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

lib/sda/protocol.go
ineiti: Convenience and backward-compatibility?

nikkolasg: That's exactly why this comment is there: to discuss about that. From my point of view, it's not clear at all we should keep that function. 
We can fill an issue about it and discuss about it there ? But we should definitely stick that comment here, since that's not the goal of this PR to decide about that.

lib/cosi/cosi.go
ineiti: It's used in the skipchain:

https://github.com/dedis/cothority/pull/403/files#diff-8cab4b17d6c04722d8a04ae4a1068cc6R112

I send the skipblocks around before signing them. But I can also move `NewSignature` to `skipchain/struct.go`

lib/sda/context.go
ineiti: 1. for example in https://github.com/dedis/cothority/pull/403/files#diff-795454cdd490b54e84d48b96481f415eR271 the skipchain wants to start a BFT-protocol but doesn't have to control it further, so instead of writing the protocol-creation in its `NewProtocol`-method, it can give this work to the SDA.

2. OK for `CreateProtocolService` and `CreateProtocolSDA`?

lib/sda/protocol.go
ineiti: +1 for opening an issue. -1 for discussing in code.

lib/sda/service.go
ineiti: https://github.com/dedis/cothority/pull/404/files#diff-6510846677b7d460f30e2726622e9869R183

and when saving the client-structure it's useful to have that field public. Else we'll have to write `MarshalBinary`-methods.

lib/sda/overlay.go
ineiti: `CreateProtocolService` and `CreateProtocolSDA`?

lib/cosi/cosi.go
nikkolasg: Mhhh. Well for me it seems strange to send a "nil" `SkipBlock` that gets filled by other peers, why not send a request to get the block directly ? But that's a separate issue - 
In the meantime, yes, that'd be cleaner to have that in `skipchain/struct.go` as a private method. Thanks.

lib/sda/context.go
nikkolasg: Ohhh. Ok that's an use we did not consider. OK right. 
Yep +1 for the renaming that's already a lot better :)

lib/sda/service.go
nikkolasg: Ok from what I see, you use `client.Send` and `client.Private`. You need the `Private` because you sign something but the private key you use for *signing* and the private key you use for the *client part* (`Send`) *should / can* be separate; they have nothing in common. 
Normally the `Private` of a client is never actually used, since it's only given to `TCPHost` but never used - maybe it's something that has to be changed also ? - that was also the reason why to put it private, because a user of the `Client` api should **not care** about keys at all. If he wants to, then he embeds `Client` and use its own fabric.
Wouldn't it be possible to get the same behaviour with something like:
```
// Identity can both follow and update an IdentityList
type Identity struct {
	// client as unexported so it is not marshalled
	client *sda.Client
	// private key as exported so it is marshalled 
	// different role == different key
	Private abstract.Secret 
	ID         ID
	...
}
```
* Clear separation of role between signing and client request/response == different keys
* No need to export `Private` *just for that*

I'm sorry to bother about this but since we are doing an effort to have a clean API, I think it's important to think about those details. Moreover, it's only a suggestion, if any of you have a better idea, please !

lib/network/encoding.go
```diff
@@ -192,6 +192,7 @@ func MarshalRegisteredType(data ProtocolMessage) ([]byte, error) {
 	var err error
 	if buf, err = protobuf.Encode(data); err != nil {
 		dbg.Error("Error for protobuf encoding:", err)
+		dbg.Error(dbg.Stack())
```
Liamsi: Nice, this will help with some nasty protobuf errors. On the other hand, we should be careful not to output the stack if for instance someone sent a corrupt message on purpose or by accident.

lib/network/struct.go
```diff
@@ -194,15 +194,13 @@ type Entity struct {
...
-	iter int
 }
 
 // EntityID uniquely identifies an Entity struct
 type EntityID uuid.UUID
 
-// Equals returns true if both EntityID are equal or false otherwise.
-func (eid EntityID) Equals(other EntityID) bool {
+// Equal returns true if both EntityID are equal or false otherwise.
+func (eid EntityID) Equal(other EntityID) bool {
```
Liamsi: For other uuids this function is still called `Equals` (like for TreeID). I would prefer to keep named like before. If you want to rename it, open an issue and rename all `Equals` functions.

lib/sda/context.go
```diff
@@ -1,16 +1,17 @@
...
 
-// Context is the interface that is given to an Service
+// Context is the interface that is given to a Service
 type Context interface {
-	NewTreeNodeInstance(*Tree, *TreeNode) *TreeNodeInstance
+	NewTreeNodeInstance(*Tree, *TreeNode, string) *TreeNodeInstance
 	RegisterProtocolInstance(ProtocolInstance) error
 	SendRaw(*network.Entity, interface{}) error
+	CreateProtocolService(*Tree, string) (ProtocolInstance, error)
+	CreateProtocolSDA(*Tree, string) (ProtocolInstance, error)
```
Liamsi: (Optional) Maybe having a short comment explaining the purpose and differences of `CreateProtocolService` and `CreateProtocolSDA` could be helpful

lib/network/struct.go
Liamsi: I just saw, you renamed all occurrences. So ignore this one.

lib/sda/host.go
```diff
@@ -342,8 +352,8 @@ func (h *Host) processServiceMessage(e *network.Entity, m *ServiceMessage) {
 		// 404 Service Unknown
 		return
 	}
-	dbg.Lvl3("host", h.Address(), " => Dispatch request to ServiceMessage")
-	s.ProcessServiceMessage(e, m)
+	dbg.Lvl5("host", h.Address(), m)
+	go s.ProcessServiceMessage(e, m)
```
Liamsi: Why is this called as a go-routine now?

lib/network/encoding.go
ineiti: In fact that's probably something we need to comment. I was not sure whether to leave it in or not.

lib/sda/host.go
ineiti: Because a serviceMessageHandler might produce another serviceMessage and wait for its return before continuing.

lib/network/encoding.go
Liamsi: I would probably remove it, or only print it if debug level is high (> 3 for instance)

lib/sda/host.go
Liamsi: OK, makes sense.

lib/sda/context.go
ineiti: the comments are in the functions below.

Liamsi: I saw those, but I thought the interface is the first thing one will look at. Anyways you are right, the (commented) implementations are just a few lines away...

nikkolasg: +1 for comments in the interface !

lib/sda/protocol.go
Liamsi: Related issue #405

ineiti:
Ready4Merge?

Liamsi:
> Ready4Merge?

Of course the changelog says Files changed 0. That's definetly OK with me ;-)

screen shot 2016-06-02 at 09 59 10

nikkolasg:
Failing

PR #404, opened 2016-05-27
Service identity

ineiti:
The identity-service, based on skipchains.


Makefile
```diff
@@ -28,7 +28,7 @@ test_lint:
 	}
 
 test_go:
-	go test -race -p=1 -short ./...
+	go test -race -p=1 -short -v ./...
```
Liamsi: `-v` on purpose change?

app/cisc/commands.go
```diff
@@ -0,0 +1,193 @@
...
+			//			Name:  "tob,toblockchain",
+			//			Usage: "force copy of config-file to blockchain",
+			//		},
+			//		cli.StringFlag{
+			//			Name:  "toc,toconfig",
+			//			Usage: "force copy of blockchain to config-file",
+			//		},
+			//	},
+			//	Action: sshSync,
+			//},
```
Liamsi: remove the comments?

ineiti: got me - this will come in a later PR ;) I'll uncomment them as the methods are still there.

app/cisc/cisc.go
```diff
@@ -0,0 +1,312 @@
...
+
+func idConnect(c *cli.Context) error {
+	log.Info("Connecting")
+	name, err := os.Hostname()
+	log.ErrFatal(err)
+	switch c.NArg() {
+	case 2:
+		// We'll get all arguments after
+	case 3:
+		name = c.Args().Get(2)
```
nikkolasg: If you are only using one argument then why having these different `case` here ?

app/cisc/cisc.go
```diff
@@ -0,0 +1,312 @@
...
+			Usage: "The configuration-directory of the ssh-directory",
+		},
+	}
+	app.Before = func(c *cli.Context) error {
+		configDir := config.TildeToHome(c.String("config"))
+		os.Mkdir(configDir, 0660)
+		log.SetDebugVisible(c.Int("debug"))
+		configFile = configDir + "/config.bin"
+		if err := loadConfig(); err != nil {
+			log.Error("Problems reading config-file. Most probably you\n",
```
nikkolasg: is it possible to continue without being able to read the config ? => should we not `Fatal` here also ?

app/cisc/cisc.go
ineiti: Because I want to:
  - don't do anything if there are 2 arguments
  - set the name if there are 3 arguments
  - fail if there is another number of arguments

app/cisc/cisc.go
```diff
@@ -0,0 +1,312 @@
...
+	"strings"
+
+	"github.com/dedis/cothority/app/lib/config"
+	"github.com/dedis/cothority/log"
+	"github.com/dedis/cothority/services/identity"
+	"gopkg.in/codegangsta/cli.v1"
+)
+
+/*
+Cisc is the Cisc Identity SkipChain to store information in a skipchain and
```
nikkolasg: Put these as the firsts lines of the files so it shows up nicely in godoc (IIRC ... !)

app/cisc/cisc.go
```diff
@@ -0,0 +1,312 @@
...
+		return nil
+	}
+	app.Run(os.Args)
+
+}
+
+/*
+ * Identity-related commands
+ */
+func idCreate(c *cli.Context) error {
```
nikkolasg: you never return an error here, but this signature (having an error return) is necessary because of codegangsta. It seems to me that it'd best if you try to follow their **guidance** and return an error instead of panicking. Their library should handle it accordingly (may need some manual test to make sure)

app/cisc/cisc.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// Get the current configuration
+	sc, err := NewSSHConfigFromFile(sshConfig)
+	log.ErrFatal(err)
+
+	// Add a new host-entry
+	hostname := c.Args().First()
+	alias := c.String("a")
+	if alias == "" {
+		alias = hostname
```
nikkolasg: Is it possible that first argument is empty and the alias also ?

app/cisc/cisc.go
```diff
@@ -0,0 +1,312 @@
...
+func sshDel(c *cli.Context) error {
+	assertCA()
+	if c.NArg() == 0 {
+		log.Fatal("Please give alias or host to delete from ssh")
+	}
+	ah := c.Args().First()
+	if len(kvGetValue("ssh", clientApp.ManagerStr, ah)) == 0 {
+		log.Print("Didn't find alias or host", ah)
+		sshLs(c)
+		log.Fatal("Aboring")
```
nikkolasg: Abor**T**ing

app/cisc/commands.go
```diff
@@ -0,0 +1,193 @@
...
+						Name:  "p,propose",
+						Usage: "will also show proposed config",
+					},
+				},
+			},
+			{
+				Name:    "vote",
+				Aliases: []string{"v"},
+				Usage:   "vote on existing config",
+				Flags: []cli.Flag{
```
nikkolasg: It's not so much important but I would not put the decision of voting `no` as a flag. The command itself suppose by default you vote `yes`, which from my point of view is not really ideal (don't suppose the vote of someone in advance) neither clear (it's not explicit). 
I'd rather see it as a *required* argument `ssh vote {y,n}` **or** even better interactively: show the user the config again (best to be sure what to vote about), and then ask for the vote on the command line.

app/cisc/commands.go
nikkolasg: Usually I'd say yes, but since he already wrote them, and the commands are already there (but with a `log.Fatal("Not implemented")`, I'd keep them here since it's likely to be implemented in the near future...

app/cisc/cisc.go
```diff
@@ -0,0 +1,312 @@
...
+Cisc is the Cisc Identity SkipChain to store information in a skipchain and
+being able to retrieve it.
+
+This is only one part of the system - the other part being the cothority that
+holds the skipchain and answers to requests from the cisc-binary.
+*/
+
+var configFile string
+var sshDir string
+var sshConfig string
```
nikkolasg: Does all the 3 above global variables have to be globals ... ? After you then have methods like `loadConfigfile()` which takes no arguments. In my head, I would have expect the method to take a string to the config file as an argument ... 
We've had so much trouble with globals so if we can avoid them that'd be best. I'm not proficient enough with the code to say if it's possible or not, I let your judgement decide :D

app/cisc/lib.go
```diff
@@ -0,0 +1,172 @@
...
+			return nil
+		}
+		return err
+	}
+	_, msg, err := network.UnmarshalRegistered(buf)
+	if err != nil {
+		return err
+	}
+	var ok bool
+	clientApp, ok = msg.(*identity.Identity)
```
nikkolasg: Same comments as above, here you read the config file into a global variable; it should probably be avoided if possible.
It also seems less *magical*:
```
func loadConfig() error  
```
 => Where does the config file comes from  and what does it read, where do I get the config ?
vs 
```
func loadConfig(config string) (Identity,error) 
```
The latter seems more explicit and less prone to *magic hazards* .
Do you think it's possible ? Again I'm not knowledgeable enough with the code but if possible that'd be really great.

app/cisc/lib.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	buf, err := network.MarshalRegisteredType(clientApp)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(configFile, buf, 0660)
+}
+
+// Asserts that the clientApp exists, else fatals
+func assertCA() {
```
nikkolasg: Then we could also get rid of this.

app/cisc/lib.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+}
+
+// Reads the group-file and returns it
+func getGroup(c *cli.Context) *config.Group {
+	gfile := c.Args().Get(0)
+	gr, err := os.Open(gfile)
+	log.ErrFatal(err)
+	groups, err := config.ReadGroupDescToml(gr)
+	gr.Close()
```
nikkolasg: put the `gr.Close()` before the `log.ErrFatal` as a `defer ...`

nikkolasg: `gr.Close` could be put as a defer , and you should check the second `err`.

app/cisc/lib.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// MakeSSHKeyPair make a pair of public and private keys for SSH access.
+// Public key is encoded in the format for inclusion in an OpenSSH authorized_keys file.
+// Private Key generated is PEM encoded
+// StackOverflow: Greg http://stackoverflow.com/users/328645/greg in
+// http://stackoverflow.com/questions/21151714/go-generate-an-ssh-public-key
+// No licence added
+func makeSSHKeyPair(pubKeyPath, privateKeyPath string) error {
+	privateKey, err := rsa.GenerateKey(rand.Reader, 1024)
```
nikkolasg: C'mon, please. More than 1024 bits !! SSH supports 2048 and 4096 IIRC.

app/cisc/sshconfig.go
```diff
@@ -0,0 +1,169 @@
...
+- sections always start with comment followed by "Host"
+- the configuration-part of a host has to start with a tab
+*/
+
+// SSHConfig holds all host-definitions
+type SSHConfig struct {
+	Host []*SSHHost
+}
+
+// NewSSHConfig takes a string which represents the ssh-config files, parses
```
nikkolasg: I am surely missing something here so don't hesitate to shut me down:
What are you calling ssh config file ? Is it the known_hosts ? In this case, this [library](https://godoc.org/golang.org/x/crypto/ssh#ParseKnownHosts) will do the trick just fine. I'm sure it's not that so can you give a link or something on which files you are reading ?

app/cisc/cisc.go
ineiti: Suppose the config-file is corrupted and you're about to create a new one with a new cothority. So instead of having to delete the old config-file before creating a new cothority, you can do it in one go.

app/cisc/cisc.go
ineiti: Yes, I also thought of it. I decided `log.ErrFatal` is too useful not to use it. So the errors are empty. If you and Ismail think it is better to let the error-handling for the codegangsta, I'll add all those `if err != nil`... lines.

app/cisc/cisc.go
ineiti: First argument has to be there, as it gives the hostname for which the ssh-key is valid. `if c.Narg() != 1` above.

app/cisc/commands.go
ineiti: Hmm - taking awayt the `reject`, have interactive if no `[yn]`-argument is given, and follow `[yn]` else? Does that make sense?

app/cisc/commands.go
```diff
@@ -0,0 +1,193 @@
...
+				ArgsUsage: "key value",
+				Action:    kvAdd,
+			},
+			{
+				Name:      "rm",
+				Aliases:   []string{"ls"},
+				Usage:     "list all values",
+				ArgsUsage: "key",
+				Action:    kvRm,
+			},
```
ineiti: Oups - I see the alias is wrong.

app/cisc/cisc.go
ineiti: OK, I got rid of those. The `clientApp` would be more difficult, so I leave it for now.

Also I'm doing all testing through an external bash-script, so I'm not too dependant on having no global variables (beside the fact that they _are_ evil, of course).

app/cisc/lib.go
ineiti: Yes, or use that to load the configuration if we need it.

app/cisc/lib.go
ineiti: I'll try if 2048 work, too. I copy/pasted the code from Stackoverflow, so I don't know if it will work...

app/cisc/sshconfig.go
ineiti: No, its `.ssh/config`, which can do a lot more. I also want to be able to keep the comments before an entry, as I would like to put some information for `cisc` in there (as they would be incompatible with SSH). So I wrote this library.

Makefile
ineiti: Just for tests.

app/cisc/lib.go
ineiti: So I removed all globals - but there is now code-duplication all over the place with regard to `saveConfig` that is called all over the place where one call before was enough.

app/cisc/lib.go
ineiti: Did it anyway and don't really like it. But well, globals are bad, even though I would've let them in the current case.

app/cisc/commands.go
nikkolasg: If I understood correctly: `vote y` or `vote n` or `vote => interaction`.
If so, then +1

ineiti: Yes, that was the intention.

app/cisc/cisc.go
nikkolasg: Ok ok right, it's not not very important.

services/skipchain/skipchain.go
```diff
@@ -320,7 +320,7 @@ func (s *Service) startBFTSignature(block *SkipBlock) error {
 		if err := block.BlockSig.Verify(network.Suite, el.Publics()); err != nil {
 			return errors.New("Couldn't verify signature")
 		}
-	case <-time.After(time.Second * 3):
+	case <-time.After(time.Second * 10):
```
nikkolasg: This kind of stuff is really nasty because we never really know how long. 10 seconds still looks too short for me, as a signing operation could really take 10 secs with latency, extreme end points geographically, etc. Otherwise, if something has not returned after let's say 1mn or 2mn, then it's definitely something wrong.

services/identity/struct.go
```diff
@@ -0,0 +1,234 @@
...
+			// Create subkey
+			subkey := strings.TrimPrefix(k, start)
+			subkey = strings.SplitN(subkey, ":", 2)[0]
+			ret = append(ret, subkey)
+		}
+	}
+	return sortUniq(ret)
+}
+
+// GetValue returns the value of the key
```
nikkolasg: `and return an empty slice if key is absent`

services/skipchain/skipchain.go
ineiti: Depending on the size of the tree, of course. What is your proposal here? 1 minute?

services/identity/struct.go
```diff
@@ -0,0 +1,234 @@
...
+// NewConfig returns a new List with the first owner initialised
+func NewConfig(threshold int, pub abstract.Point, owner string) *Config {
+	return &Config{
+		Threshold: threshold,
+		Device:    map[string]*Device{owner: {pub}},
+		Data:      make(map[string]string),
+	}
+}
+
+// Copy makes a deep copy of the AccountList
```
nikkolasg: `Copy returns a deep copy of the AccountList` (yeah I know it's small, but it's everyone's right to nitpick !! :D)

services/identity/struct.go
```diff
@@ -0,0 +1,234 @@
...
+
+// Config holds the information about all devices and the data stored in this
+// identity-blockchain
+type Config struct {
+	Threshold int
+	Device    map[string]*Device
+	Data      map[string]string
+}
+
+// Device has write-access to the IdentityList if the threshold is given
```
nikkolasg: `Device is represented by a public key and has write-access ....` ?

services/identity/struct.go
ineiti: it's definitively something that _should_ be added here!

services/identity/api.go
```diff
@@ -0,0 +1,226 @@
...
+	return nil
+}
+
+// ConfigUpdate asks if there is any new config available that has already
+// been approved by others and updates the local configuration
+func (i *Identity) ConfigUpdate() error {
+	if i.Cothority == nil || len(i.Cothority.List) == 0 {
+		return errors.New("Didn't find any list in the cothority")
+	}
+	msg, err := i.Send(i.Cothority.GetRandom(), &ConfigUpdate{ID: i.ID})
```
nikkolasg: `GetRandom()` can we change it to `Random()` ?

services/identity/api.go
```diff
@@ -0,0 +1,226 @@
...
+		&ProposeFetch{},
+		&ConfigUpdate{},
+		&UpdateSkipBlock{},
+		&ProposeVote{},
+	} {
+		network.RegisterMessageType(s)
+	}
+}
+
+// Identity can both follow and update an IdRoster
```
nikkolasg: I know you already wrote quite a lot on the subject but maybe something a bit more detailled would be nice. Either take some paragraph you already wrote, or point a link to a nicer explanation, or both.

services/skipchain/skipchain.go
nikkolasg: Why not. If you think that makes sense. I'm just afrait 10sec is too short, but I would not know the best timeout to put; so 1mn seems reasonably long enough for me.

services/identity/api.go
ineiti: Yes, it returns a `ServerIdentity`. But I like the `Get` here to reflect we're _getting_ something and not doing _RandHound_ or other.

app/cisc/README.md
```diff
@@ -0,0 +1,58 @@
+# CISC - Cisc Identity SkipChain
+
+## Description
+Cisc is based on a personal blockchain handled by the cothority. It
+ can store any key/value pairs, and has a special module for managing
+ ssh-public-keys.
+
+Based upon the well-known skipchains, cisc serves a data-block with different entries that can be handled by a number of devices who propose changes and cryptographically vote to approve or deny those changes. Different data-types exist that will interpret the data-block and offer a service.
```
Liamsi: well-know blockchains (or skipchains ;-) ?

ineiti: yes, internal joke. Should probably get rid of it.

app/cisc/cisc.go
Liamsi: `log.ErrFatal` prints an error and exists, right (no panic-ing)? Also don't think thats very relevant, but if use 3rd party libs we should use them consistently (maybe later).

app/cisc/cisc.go
Liamsi: > don't do anything if there are 2 arguments

Does this make sense (or will it make sense)?

ineiti: Yes, because the arguments are:
`group id [id-name]`
so the `id-name` can be omitted. But as two star-programmers get it wrong, I should perhaps replace that with
`if c.NArg() != 2 && c.NArg() != 3{ log.Fatal("wrong number of arguments) }`
or something like that...

app/cisc/cisc.go
ineiti: Every `log.ErrFatal` would have to be replaced by
```
if err != nil{
    return err
}
```
And personally I prefer `log.ErrFatal`. It makes the code easier to read.

services/identity/api.go
nikkolasg: I know, I even agree. This is the Java way, and even if I can't stand it, it's a nice thing to have a `GetXXX` and `SetXXX` because it's clear and explicit. But in Golang, it's more like `XXX()` and `SetXXX()`. I don't really really mind (really!!) but again, **being consistent** is the more important thing here, as it's almost like a *muscle reflex* just by looking at the method if it's a setter/getter or stg else. 
What do we do usually ?

Looking at other piece of code, for example, `Roster.Publics()`, `TreeNode.AggregatePublic()`,`TreeNodeInstance.Private()`,`TreeNodeInstance.Token()` ... the list goes on.

For golang developpers, it does not seem so much troubling (at least for me) to have a `Cothority.Random()` returning a random member of a cothority.
If you still prefer having the `Get` then we can let the third member of our cyber-democracy having the final decision ... waiting on @Liamsi :)

ineiti: Well, we need more than 2/3 to avoid a split in our BFT-group, so we would have to add Philipp, and then only being able to vote if 3/4 come together...

Instead of `GetRandom`, what about `RandomSI`, `RandomServerIdentity`, `ChoseRandomly`?

nikkolasg: * `Cothority.RandomSI` is confusing for me, 
* `Cothority.RandomServerIdentity` is a bit long but why not, +1
* `Cothority.ChoseRandomly` : seems like we are doing Ruby ;)

Another proposition: `Cothority.RandomMember` ?

Liamsi: It is quite uncommon in golang to work with getters.

services/identity/service.go
```diff
@@ -0,0 +1,316 @@
...
+type Service struct {
+	*sda.ServiceProcessor
+	identities      map[string]*Storage
+	identitiesMutex sync.Mutex
+	skipchain       *skipchain.Client
+	path            string
+}
+
+// Storage stores one identity together with the skipblocks
+type Storage struct {
```
Liamsi: Looks like this struct is only used internally. Can it be made private?

Liamsi:
Besides some nitpicks, looks good to me.

services/identity/service.go
nikkolasg: +1 for private ! Could you also look for any other struct that is not needed outside the module ?

services/identity/api.go
nikkolasg: > It is quite uncommon in golang to work with getters.

? I don't think so, I mean, it's quite usual to see getters, but they're just no preprended with `Get`...?
But you make me think of something: `Random()` is **not** a Getter in the *strict sense*, because it does not return a field; it computes a random number and return a random element *from* the slice field. It computes something.

Liamsi: By getter I meant java-style like getters (GetXXX).

ineiti:
Resolved last comments. If Travis passes, I'll be OK to merge.

app/cisc/cisc.go
ineiti: Not on purpose I returned an error there, and codegansta just ignores that, the error is never printed :( So I definitively prefer `log.ErrFatal`!

nikkolasg:
OK

PR #407, opened 2016-05-31
Protocol propagate

ineiti:
Re-did pull-request against correct branch


protocols/manage/propagate.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// PropagateReply is sent from the children back to the root
+type PropagateReply struct {
+	Level int
+}
+
+// PropagateStartAndWaitSDA starts the propagation protocol and blocks till everything
+// is OK or the timeout has been reached.
+// This is used when you want to start a protocol without a service.
```
Liamsi: What does it mean: "till everything is OK"?

Liamsi: Remove this or be more specific. Also its not clear from the signature of the function nor from the comment what this function returns.

protocols/manage/propagate.go
```diff
@@ -0,0 +1,199 @@
...
+		return -1, err
+	}
+	return propagateStartAndWait(pi, msg, msec, f)
+}
+
+// PropagateStartAndWaitService starts the propagation protocol and blocks till everything
+// is OK or the timeout has been reached.
+// This is used when you want to start a protocol included in a service.
+func PropagateStartAndWaitService(c sda.Context, el *sda.EntityList, msg network.ProtocolMessage, msec int, f func(network.ProtocolMessage)) (int, error) {
+	tree := el.GenerateNaryTreeWithRoot(8, c.Entity())
```
Liamsi: Same comments as above.

protocols/manage/propagate.go
```diff
@@ -0,0 +1,199 @@
...
+
+// PropagateReply is sent from the children back to the root
+type PropagateReply struct {
+	Level int
+}
+
+// PropagateStartAndWaitSDA starts the propagation protocol and blocks till everything
+// is OK or the timeout has been reached.
+// This is used when you want to start a protocol without a service.
+func PropagateStartAndWaitSDA(o *sda.Overlay, el *sda.EntityList, msg network.ProtocolMessage, msec int, f func(network.ProtocolMessage)) (int, error) {
```
nikkolasg: Further question: why do we allow such propage protocol using *only* SDA ? The whole point of the propagate protocol is to broadcast a data through the tree right ? What's the use of using only sda if  the data will be lost after ? lost means no `Service` will use it somehow ? I'm sure you have a use case, what is it :D ?

protocols/manage/propagate.go
```diff
@@ -0,0 +1,199 @@
...
+	CreateProtocol(t *sda.Tree, name string) (sda.ProtocolInstance, error)
+	Entity() *network.Entity
+}
+
+// PropagateSendData is the message to pass the data to the children
+type PropagateSendData struct {
+	// Data is the data to transmit
+	Data []byte
+	// Msec is the timeout in milliseconds
+	Msec int
```
Liamsi: timeout for what event?

nikkolasg: If the whole tree has not received the data before a certain timeout, the propagate returns with an error I guess.

Liamsi: Yes, if I read the code I can see that. But if the comment only says "the timeout" it's not really useful. I would either remove the comment (than it's clear that one must read the code to find out what will happen with `Msec` or explain which timeout so that one doesn't need to read through the code).

protocols/manage/propagate.go
ineiti: It's use-case is testing... I'll make it private.

PR #418, opened 2016-06-08
Closed but not merged PRs

ineiti:
Closed but not merged pull-requests:
- cleanup_1_apps
- cleanup_2_lib
- protocol_propagate

### DIFF
--- a/lib/sda/context.go
+++ b/lib/sda/context.go
@@ -1,16 +1,28 @@
 package sda
 
-import (
-	"github.com/dedis/cothority/lib/network"
-)
+import "github.com/dedis/cothority/lib/network"
 
-// Context is the interface that is given to an Service
+// Context is the interface that is given to a Service
 type Context interface {
-	NewTreeNodeInstance(*Tree, *TreeNode) *TreeNodeInstance
+	// NewTreeNodeInstance implements the Context interface method
+	NewTreeNodeInstance(*Tree, *TreeNode, string) *TreeNodeInstance
+	// RegisterProtocolInstance takes a PI and stores it for dispatching the message
+	// to it.
 	RegisterProtocolInstance(ProtocolInstance) error
+	// SendRaw sends a message to the entity
 	SendRaw(*network.Entity, interface{}) error
+	// CreateProtocolService makes a TreeNodeInstance from the root-node of the tree and
+	// prepares for a 'name'-protocol. The ProtocolInstance has to be added later.
+	CreateProtocolService(*Tree, string) (ProtocolInstance, error)
+	// CreateProtocolSDA is like CreateProtocolService but doesn't bind a service to it,
+	// so it will be handled automatically by the SDA.
+	CreateProtocolSDA(*Tree, string) (ProtocolInstance, error)
+	// Address is the address where this host is listening
 	Address() string
+	// Entity returns the entity the service uses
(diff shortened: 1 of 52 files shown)


nikkolasg:
51 files ... what are the *things* ?
ProtocolName from TreeNodeInstance ?

lib/sda/context.go
```diff
@@ -1,16 +1,28 @@
...
+	CreateProtocolService(*Tree, string) (ProtocolInstance, error)
+	// CreateProtocolSDA is like CreateProtocolService but doesn't bind a service to it,
+	// so it will be handled automatically by the SDA.
+	CreateProtocolSDA(*Tree, string) (ProtocolInstance, error)
+	// Address is the address where this host is listening
 	Address() string
+	// Entity returns the entity the service uses
 	Entity() *network.Entity
+	// GetID returns the service-id
+	GetID() ServiceID
```
nikkolasg: Better named `ServiceID` ?

lib/sda/host_test.go
```diff
@@ -14,6 +14,7 @@ import (
 
 // Test setting up of Host
 func TestHostNew(t *testing.T) {
+	defer dbg.AfterTest(t)
```
nikkolasg: was it not handled directly into the Test.Main thing ?

lib/sda/overlay.go
```diff
@@ -231,26 +238,40 @@ func (o *Overlay) Close() {
...
-	pi, err := o.CreateProtocol(t, name)
+	pi, err := o.CreateProtocolSDA(t, name)
 	if err != nil {
 		return nil, err
 	}
-	go pi.Start()
+	go func() {
+		err := pi.Start()
+		if err != nil {
+			dbg.Error("Error while starting:", err)
```
nikkolasg: Having that only printed and not returned in a function named `StartProtocol` seems to me that we should not expect an error value return in `Start`, because we don't / can't use it.

lib/sda/host_test.go
ineiti: Yes, problem with `AfterTest` is, that you only know after ALL tests whether a go-routine is hanging around. So if the test fails with a go-routine too much, you don't know in *which* test the go-routine was started. Thus I put it back in.

Best would be to have a `before` and `after` for testing, but that doesn't exist in vanilla go-tests...

lib/sda/overlay.go
ineiti: Well it's in the go-routine, so it's difficult to return that value...

nikkolasg: Exactly, so why should have a return value at all ;)

Liamsi: There are several ways how-to handle errors with go-routines. I think the most appropriate way would be through channels (which might be an overkill in this case (?))

nikkolasg: It depends in the situation, here we are forced to call Start() in a go-routine because the go routine can take time, but we don't want the main *thread* to hang up on it waiting on the error response (otherwise it's useless to use channels). we have this chiken-and-egg problem here, and it does not make sense for me to have the Start return an error because we can't use it, and we can't make a meaning out of it (it's not the sda code (network/tree etc), it's a protocol, so it should be a protocol that handle its own error, not sda)

nikkolasg:
Ok for me

PR #421, opened 2016-06-09
Bftcosi

ineiti:
Working bftcosi implementation, even if a node or more fails!


lib/cosi/cosi.go
```diff
@@ -191,15 +191,25 @@ func (c *Cosi) CreateResponse() (*Response, error) {
 
 // Response generates the response from the commitment, challenge and the
 // responses of its children.
-func (c *Cosi) Response(responses []*Response) (*Response, error) {
-	// create your own response
-	if err := c.genResponse(); err != nil {
-		return nil, err
+// if respond == true, will also add our response
+func (c *Cosi) Response(respond bool, responses []*Response) (*Response, error) {
```
nikkolasg: I think you also have to change `VerifyResponses` or did you check if that still works if you make `respond == false` ?

ineiti: I added a call to `VerifyResponses` in `TestVerifyWithException2` and it seems to work.

protocols/bftcosi/bftcosi.go
```diff
@@ -15,13 +14,16 @@ import (
 // VerificationFunction can be passes to each protocol node. It will be called
 // (in a go routine) during the (start/handle) challenge prepare phase of the
 // protocol. The passed message is the same as sent in the challenge phase.
-type VerificationFunction func(Msg []byte) bool
+type VerificationFunction func(Msg []byte, Data []byte) bool
```
nikkolasg: I don't get what is the difference between `Msg` and `Data`. For me the old function signature was fine, you verify a bunch of bytes... why is that new addition required ?

protocols/bftcosi/bftcosi.go
```diff
@@ -187,189 +185,167 @@ func (bft *ProtocolBFTCoSi) Dispatch() error {
...
+
+// RegisterOnDone registers a callback to call when the bftcosi protocols has
+// really finished
+func (bft *ProtocolBFTCoSi) RegisterOnDone(fn func()) {
+	bft.onDoneCallback = fn
+}
+
+// RegisterOnSignatureDone register a callback to call when the bftcosi
+// protocol reached a signature on the block
+func (bft *ProtocolBFTCoSi) RegisterOnSignatureDone(fn func(*BFTSignature)) {
```
nikkolasg: Difference between `RegisterOnDone` and `RegisterOnSignatureDone` ? Once is the signature is done, the protocol still has something else to perform ?

lib/cosi/cosi.go
nikkolasg: I would prefer putting the bool flag at the end of the list arguments: `Response(responses []*Response, respond bool)` as to keep consistent with the `Commit(commits []*Commitment)` function.

nikkolasg: Also, just to be sure, this behavior is not to be implemented now in the crypto library right ? Because this behaviors means that a node can refuse to sign after a commit, that means it has to pass in the final signature its public key and its commit. 
In the current design of dedis/crypto/cosi, we only have the *mask* which represents the missing public keys to substract, but there's nothing to handle a failure after a commit.

ineiti: As just discussed now: I would really like to have this integrated to be able to port BFTCoSi over to the new library.

protocols/bftcosi/bftcosi.go
ineiti: Probably not the nicest thing to do. The `Msg` is signed and verified, and the `Data` is just sent along. So I don't have to send another message before it. Now that I think of it: do we have a working `configuration`-message for services the way we discussed some time ago? I could send the `Data` along there.

protocols/bftcosi/bftcosi.go
ineiti: Good question. `OnSignatureDone` is called as soon as the signature is finished, while `OnDone` is called once the protocol is cleaning up. So either can go away. For the moment only `OnDone` is used, so I propose we keep that as other protocols probably also will have the same. What do you think?

nikkolasg: Ok for `OnDone`, it keep consistency regarding other protocols which have the same method

protocols/bftcosi/bftcosi.go
nikkolasg: Damn I don't think so. that would be a really good thing to do, even better if we can remove the data part here.

protocols/bftcosi/bftcosi.go
```diff
@@ -97,23 +99,39 @@ type ProtocolBFTCoSi struct {
...
+		doneProcessing:       make(chan bool, 2),
+		doneSigning:          make(chan bool, 1),
+		VerificationFunction: verify,
+		AggregatedPublic:     n.Roster().Aggregate,
+		threshold:            len(n.Tree().List()) * 2 / 3,
+		Msg:                  make([]byte, 0),
+		Data:                 make([]byte, 0),
+	}
+
+	idx, _ := n.Roster().Search(bft.ServerIdentity().ID)
```
ineiti: Is this now the default method to get the index in the Roster? Will this return the same value under all circumstances?

protocols/bftcosi/bftcosi.go
```diff
@@ -187,291 +200,300 @@ func (bft *ProtocolBFTCoSi) Dispatch() error {
...
+// RegisterOnDone registers a callback to call when the bftcosi protocols has
+// really finished
+func (bft *ProtocolBFTCoSi) RegisterOnDone(fn func()) {
+	bft.onDoneCallback = fn
 }
 
-// startAnnouncementCommit create the announcement for the commit phase and
+// RegisterOnSignatureDone register a callback to call when the bftcosi
+// protocol reached a signature on the block
+func (bft *ProtocolBFTCoSi) RegisterOnSignatureDone(fn func(*BFTSignature)) {
```
ineiti: I think this must be removed - I don't know how it re-appeared here, pretty sure I took it out at _some_ time.

protocols/bftcosi/bftcosi.go
```diff
@@ -187,291 +200,300 @@ func (bft *ProtocolBFTCoSi) Dispatch() error {
...
-
-	bftChal := &ChallengePrepare{
-		TYPE:      RoundPrepare,
-		Challenge: ch,
-		Msg:       bft.Msg,
+// startChallenge creates the challenge and sends it to its children
+func (bft *ProtocolBFTCoSi) startChallenge(t RoundType) error {
+	if t == RoundPrepare {
+		// create challenge from message's hash:
+		// XXX Why ? Since Challenge() already hash the message
```
ineiti: No discussion in the code please - it's difficult to follow. Please use github-issues or PRs. In that special case we should ask Lefteris.

protocols/bftcosi/bftcosi.go
```diff
@@ -187,291 +200,300 @@ func (bft *ProtocolBFTCoSi) Dispatch() error {
...
+	// append response
+	if !ok {
+		dbg.Lvl3(bft.Roster(), "Refused to sign")
+	}
+
+	// Return if we're not root
+	if !bft.IsRoot() {
+		return bft.SendTo(bft.Parent(), bzrReturn)
+	}
+	// let's modify the cosi signature so it don't account the exception
```
ineiti: Can you elaborate that comment? I don't understand it.

nikkolasg: using `cosi.Signature()` returns the full signature with *all* responses aggregated inside. Because cosi does not support exceptions for the moment, we have to remove the *responses* from the signature which should not be there (i.e. exceptions).

ineiti: Please rewrite the comment, then. I cannot understand it the way it is written.

protocols/bftcosi/bftcosi.go
nikkolasg: Well, I just came up with the first solution that came to my mind. 
It's a good question and unfortunatly, I can't answer yes to it because ... everything is exported :) So one can change the `roster.List` anyway we want. 
If this don't happen, we can still have many `ServerIdentity` in the array, so it will always return the first which may not be the right one, but at least it's consistent.

protocols/bftcosi/bftcosi.go
nikkolasg: Mhh I was also inclined to remove it, but when doing it in code, I wonder if this still could be useful. For example, for non-root nodes, a Service may want to know when the bft is finished ?

As a general rule, do we want to enforce to have a callback / channel to signal when a protocol is finished ? (enforcing means putting it in the interface definition)
If the answer is no, then we can remove it. If yes.. then keep it.
I'm not sure myself.

nikkolasg:
the tests are working on my machine, anyone can test ?

protocols/bftcosi/bftcosi.go
```diff
@@ -97,23 +102,46 @@ type ProtocolBFTCoSi struct {
...
+		TreeNodeInstance:     n,
+		suite:                n.Suite(),
+		prepare:              cosi.NewCosi(n.Suite(), n.Private(), n.Roster().Publics()),
+		commit:               cosi.NewCosi(n.Suite(), n.Private(), n.Roster().Publics()),
+		verifyChan:           make(chan bool),
+		doneProcessing:       make(chan bool, 2),
+		doneSigning:          make(chan bool, 1),
+		VerificationFunction: verify,
+		AggregatedPublic:     n.Roster().Aggregate,
+		threshold:            len(n.Tree().List()) * 2 / 3,
```
Liamsi: Shouldn't you ceil here? Otherwise the threshold could be too low.

Liamsi: Add least this is (sometimes) a little lower then before: https://play.golang.org/p/KNOAJOLtfV

nikkolasg: A safe bet is too take the upper bound (with `Ceil`), the more relaxed bet (meaning it's more easy to obtain) is to use this formula. Maybe lefteris should know better.

ineiti: I wrote a test for it `TestThreshold` - please have a look if it's OK. I'm a bit ashamed it took me more than 5 minutes to do the integer-math. +45 minutes for getting the test to run ;)

Liamsi: Looks good on first sight, but the test doesn't pass?

ineiti: Travis-power - it passes locally :( I'll have a look at it.

protocols/bftcosi/bftcosi.go
ineiti: OK, keep it for the moment. Might go away.

protocols/bftcosi/bftcosi.go
ineiti: Once #455 has been implemented, we can correct that here.

protocols/bftcosi/bftcosi.go
ineiti: Changed it - OK.

ineiti:
Closing and re-opening against master

PR #423, opened 2016-06-14
Two ways of taking control over ProtocolInstance creation

ChristianMct:
When testing protocols that operate on the presence of data on `TreeNode` that are not the root, we need to have access to their `ProtocolInstance` in order to set a pointer to these data.

These are two ways of doing it, with a modified version of example_test.go as use case:

1. Call sda.ProtocolRegisterName for the test. You can actually override SDA's `ProtocolInstace` creation function with a locally provided function. This is nice because it doesn't change anything in SDA. This isn't nice because this survives the test (you need to set the old `NewProtocol` function back).
2. Adding some logic at the host level so that you can register your own `ProtocolInstace` constructor, like the services are doing. This is nice because you don't manipulate global state in SDA. This isn't nice because you change SDA lib.

So the question is, is it worth giving local tests their own way of creating `ProtocolInstance`, or is the little "hack" in a) already good enough ?

### DIFF
--- a/lib/sda/host.go
+++ b/lib/sda/host.go
@@ -65,7 +65,7 @@ type Host struct {
 	serviceStore *serviceStore
 
 	// factory overriding protocol instance creation
-	newProtocol func(*TreeNodeInstance)(ProtocolInstance, error)
+	newProtocol func(*TreeNodeInstance) (ProtocolInstance, error)
 }
 
 // NewHost starts a new Host that will listen on the network for incoming
@@ -250,7 +250,7 @@ func (h *Host) StartProcessMessages() {
 
 // RegisterNewProtocolFactory registers a new constructor function for the overlay to use for creating new
 // ProtocolInstance's
-func (h *Host) RegisterNewProtocol(fn func(*TreeNodeInstance)(ProtocolInstance, error)) {
+func (h *Host) RegisterNewProtocol(fn func(*TreeNodeInstance) (ProtocolInstance, error)) {
 	h.newProtocol = fn
 }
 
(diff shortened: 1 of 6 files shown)


nikkolasg:
> When testing protocols that operate on the presence of data on TreeNode that are not the root, we need to have access to their ProtocolInstance in order to set a pointer to these data.

What data are you talking about ? There's no *raw data* on `TreeNode`. I have trouble understanding your problem ..  could you give an concrete example ? From what I understand, you could also simply register a protocol to a new name with a new creator function and use this one. Stg like:
```
// in normal run
sda.ProtocolInstantiate("BFTCosi")

// in testing function
sda.RegisterNewProtocol("BFTCosiTest",func(*TreeNodeInstance) (sda.ProtocolInstance,error) {
  proto := NewBFTCosi()
  proto.MyData = ...
})
```
Using this, there's no need to put back the original function.

ChristianMct:
A concrete example would be a protocol that performs the sum of variable that is different for every TreeNode. What I'm trying to use SDA for is to have a service that collect data at each host running the service, and then to aggregate all the data using a protocol. In this use case, there is raw data not only at the root, but also at each TreeNode. Isn't it possible with SDA ?

I haven't thought about your third option which I think is the best one. Thanks.

ineiti:
so shall we close it? I didn't dive into it, but it seems like either registering a new protocol or using services you can have your data at each node.

ChristianMct:
We can close it.

> or using services you can have your data at each node
My point was about local testing where it's better to isolate the protocol. In this case we aren't provided with a `NewProtocol` "delegate" function to set the ProtocolInstance up. What we can do instead is to register this function as a new protocol in SDA (as long as the implementation of this protocol doesn't depend on the protocol's name, which should normally not be the case).

nikkolasg:
Sure ? If you feel that's not a good option enough for you, please feel free to pass by the office and try to explain to us what you want.

Issue #408, opened 2016-06-01
Use optimized edwards curve


Liamsi:
Currently this suite is used: https://github.com/dedis/cothority/blob/master/lib/network/encoding.go#L12-L24
we should be using `github.com/dedis/crypto/edwards/ed25519` instead.

ineiti:
As soon as crypto-#48 is merged, we'll use that curve.

PR #427, opened 2016-06-15
Use ed25519 optimzed curve

nikkolasg:
Using the optimized (and formatted) ed25519 curve.
See #408 .

### DIFF
--- a/lib/cosi/cosi_test.go
+++ b/lib/cosi/cosi_test.go
@@ -5,10 +5,10 @@ import (
 	"testing"
 
 	"github.com/dedis/crypto/config"
-	"github.com/dedis/crypto/edwards"
+	"github.com/dedis/crypto/ed25519"
 )
 
-var testSuite = edwards.NewAES128SHA256Ed25519(false)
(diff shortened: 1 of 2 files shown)


ineiti:
Just for completion's sake: there is also a suite in `lib/cosi/cosi_test.go` - we should use network-suite there, even though it will be deleted shortly.

nikkolasg:
Good catch

lib/cosi/cosi_test.go
```diff
@@ -5,10 +5,10 @@ import (
 	"testing"
 
 	"github.com/dedis/crypto/config"
-	"github.com/dedis/crypto/edwards"
+	"github.com/dedis/crypto/ed25519"
 )
 
-var testSuite = edwards.NewAES128SHA256Ed25519(false)
+var testSuite = ed25519.NewAES128SHA256Ed25519(false)
```
ineiti: Should be `network.Suite`, but the code will be deleted anyway, so 'let it be'

PR #429, opened 2016-06-15
added benchmark

evanv03:
Added Benchmark for conversion between Marshaled and Unmarshaled trees

### DIFF
--- a/lib/sda/tree_test.go
+++ b/lib/sda/tree_test.go
@@ -472,6 +472,15 @@ func TestEntityList_Publics(t *testing.T) {
 // - public keys
 // - corner-case: accessing parent/children with multiple instances of the same peer
 // in the graph
+//BenchmarkTreeMarshal will be the benchmark for the conversion between TreeMarshall and Tree
+func BenchmarkTreeMarshal(b *testing.B) {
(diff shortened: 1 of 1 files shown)


lib/sda/tree_test.go
```diff
@@ -472,6 +472,46 @@ func TestEntityList_Publics(t *testing.T) {
 // - public keys
 // - corner-case: accessing parent/children with multiple instances of the same peer
 // in the graph
+//BenchmarkTreeMarshal will be the benchmark for the conversion between TreeMarshall and Tree
+//It takes 37461141 ns before optimization on Macbook Pro
```
ineiti: Now that we don't do the optimisation, we might get rid of that line.

We make comments starting with a space, and the above comments (line 472-474) can be deleted (don't know where they come from).

lib/sda/tree_test.go
```diff
@@ -472,6 +472,46 @@ func TestEntityList_Publics(t *testing.T) {
...
+func BenchmarkTreeMarshal(b *testing.B) {
+	tree, _ := genLocalTree(10000, 2000)
```
ineiti: For having not too long benchmarks, we can put that value to 1000

Issue #438, opened 2016-06-17
Return Stats from servers


evanv03:
We want to be able to retrieve some stats from the servers with regard to connections.
It can possibly be added later that services can add other stats

ineiti:
Here is a TODO-list for the stats-service:
- [ ] Merge with new master and make a PR against `master`
- [ ] Return the IP:Ports of all cothorities this one is connected to
- [ ] Return the name of all available services
- [ ] Return the version of the cothority
- [ ] Propose how the other services can return status-reports

PR #439, opened 2016-06-17
First version of Stats

evanv03:
This the first version of #438


nikkolasg:
I quite don't get the design idea here. Are we creating one application for each service ... ? Why do you create an app `status` ?
I kinda thought we would create a `cothorityd` daemon and `cothority` client app which regroups all protocols/services we have in cothority. One application that can be used for many different services. Having a status application seems rather overkill. Moreover, embedding the group.toml in the app is not ok.
@evanv03 don't worry I'm asking the whole team here :) Whether we keep that design or not, I'm sure you learnt a lot and this is not lost **at all** :)

ineiti:
The goal of this is to being able to retrieve status-information of a running `cothorityd`. This can be included in other `app`s, and a reference implementation is in the `app/status`.

The status returned by the status-service has network-status and can include also status-reports from other services. This latter part is not yet defined.

app/status/group.toml
```diff
@@ -0,0 +1,20 @@
+
+[[servers]]
```
nikkolasg: remove this file

app/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+	dbg.SetUseColors(false)
+}
+
+func main() {
+	app := cli.NewApp()
+	app.Name = "Status"
+	app.Usage = "Get and print status of all servers in a file."
+	app.Commands = []cli.Command{
+		{
+			Name:    "getStatus",
```
nikkolasg: We need to change names here because it makes `./status getStatus`, what about `./status network` ?
Furthermore, this could be defined as the main command on the app so calling `./status` directly returns this information. It can still be kept also a sub command if needed.
what do you think ?

app/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+			Aliases: []string{"g"},
+			Usage:   "Gets status from all entities in group file.",
```
nikkolasg: Better comments saying what is a status

app/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+// signFile will search for the file and sign it
+// it always returns nil as an error
+func getStatus(c *cli.Context) error {
+	groupToml := c.GlobalString(optionGroup)
+	el, err := readGroup(groupToml)
+	ui.ErrFatal(err, "Couldn't read file.")
+	dbg.Lvl3(el)
+	cl := status.NewClient()
+	for i := 0; i < len(el.List); i++ {
+		dbg.Print("Hello: ")
```
nikkolasg: Well it's nice of you to say hello but I think it can be removed.

app/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+	app.Run(os.Args)
+}
+
+// signFile will search for the file and sign it
+// it always returns nil as an error
+func getStatus(c *cli.Context) error {
+	groupToml := c.GlobalString(optionGroup)
+	el, err := readGroup(groupToml)
+	ui.ErrFatal(err, "Couldn't read file.")
+	dbg.Lvl3(el)
```
nikkolasg: prefer to use `Info, Warn` etc for client apps.

app/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+	if len(el.List) <= 0 {
+		return nil, errors.New("Empty or invalid group file:" +
+			tomlFileName)
+	}
+	dbg.Print(el)
+	return el, err
+}
+
+func printConn(e *status.StatusResponse) {
+	fmt.Print("Host: ", e.Serv, "\n")
```
nikkolasg: either use `ui.XXX` or `dbg.XXX` but drop `fmt.Print(ln)`

app/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+		fmt.Print("Connection: ", e.Remote[i], "\n")
+		fmt.Print("Total Packets Recieved: ", e.Received[i], "\n")
+		fmt.Print("Total Packets Sent: ", e.Sent[i], "\n")
+		fmt.Print("\n")
+	}
+	fmt.Print("\n \n")
+}
+
+/*
+func write(r io.Writer, tomlFileName string) {
```
nikkolasg: remove comments if not used

app/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+			tomlFileName)
+	}
+	dbg.Print(el)
+	return el, err
+}
+
+func printConn(e *status.StatusResponse) {
+	fmt.Print("Host: ", e.Serv, "\n")
+	for i := 0; i < len(e.Received); i++ {
+		fmt.Print("\n")
```
nikkolasg: btw you can start using `fmt.Println` and you don't have to do these `"\n"` anymore :)

app/status/status.go
```diff
@@ -0,0 +1,112 @@
...
+	dbg.SetDebugVisible(1)
+	dbg.SetUseColors(false)
+}
+
+func main() {
+	app := cli.NewApp()
+	app.Name = "Status"
+	app.Usage = "Get and print status of all servers in a file."
+	//a status is a list of connections and packets sent and received for each server in the file
+	app.Commands = []cli.Command{
```
ineiti: Use the default command of `codegangsta.cli` instead of a command.

app/status/status.go
```diff
@@ -0,0 +1,112 @@
...
+		sr, _ := cl.GetStatus(el.List[i])
+		printConn(sr)
+		dbg.Print(cl)
+	}
+	return nil
+}
+
+// sign takes a stream and a toml file defining the servers
+func readGroup(tomlFileName string) (*sda.EntityList, error) {
+	dbg.Lvl2("Reading From File")
```
ineiti: `dbg.Print`

app/status/status.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	el, err := config.ReadGroupToml(f)
+	if err != nil {
+		return nil, err
+	}
+	if len(el.List) <= 0 {
+		return nil, errors.New("Empty or invalid group file:" +
+			tomlFileName)
+	}
+	dbg.Print(el)
```
ineiti: `dbg.Lvl3`

app/status/status.go
ineiti: `dbg.Print` - once merged with `development`-branch, all `dbg` will turn into `log` and don't output the method and line-numbers if the debug-level is < 1.

app/test_status.sh
```diff
@@ -0,0 +1,63 @@
...
+
+. lib/test/libtest.sh
+. lib/test/cothorityd.sh
+DBG_SHOW=2
+STATICDIR=test
+
+main(){
+    startTest
+    build
+    #test Build
```
ineiti: uncomment

app/test_status.sh
```diff
@@ -0,0 +1,63 @@
...
+    test Network
+    stopTest
+}
+
+testNetwork(){
+    cothoritySetup
+    cp group.toml cl1
+    testOut "Running network"
+    testOK runCl 1 network
```
ineiti: replace with `testGrep`

ineiti:
I added a TODO-list in #438

ineiti:
My bad - I should've told you how we handle these cases, but with the conference it slipped out of my mind. Changes like these are better put in a separate branch. Best you can do now is to return to the master-branch, updated, make a new branch, then import the changes, and commit this new branch:

```
git checkout master
git pull
git checkout -b context_to_struct
git checkout stat sda/context.go
# Eventually also import other files
git commit -am "imported context-changes"
git push -u origin context_to_struct
```
Then in Github you can create a new pull-request of `context_to_struct` against master and let us look through it.

You'll also have to close this pull-request and re-open it against `master`.

app/status/status.go
```diff
@@ -0,0 +1,107 @@
...
+			Name:  "debug, d",
+			Value: 0,
+			Usage: "debug-level: `integer`: 1 for terse, 5 for maximal",
+		},
+	}
+	app.Before = func(c *cli.Context) error {
+		log.SetDebugVisible(c.GlobalInt("debug"))
+		return nil
+	}
+	app.Action = network
```
ineiti: More something like
```
app.Action = func(c *cli.Context) error {
      network(c)
  }
```

ineiti:
Close and re-open against master.

PR #440, opened 2016-06-17
Liblog3

ineiti:
Finally - merging `app/lib/ui` and `lib/dbg` into `log`. Did all changes, but kept the `dbg`-directory, so that cosi can be done separately.


nikkolasg:
You probably want to merge that into development instead ?

PR #422, opened 2016-06-19
Some syscall fields are not cross platform

mahdiz:
When compiling cothority in Windows, I get the following compile-time error:

```
# github.com/dedis/cothority/lib/monitor
..\..\dedis\cothority\lib\monitor\measure.go:237: undefined: syscall.Getrusage
..\..\dedis\cothority\lib\monitor\measure.go:237: undefined: syscall.RUSAGE_SELF
..\..\dedis\cothority\lib\monitor\measure.go:240: rusage.Stime undefined (type *syscall.Rusage has no field or method Stime)
..\..\dedis\cothority\lib\monitor\measure.go:240: rusage.Utime undefined (type *syscall.Rusage has no field or method Utime)
```

It seems that some fields of `syscall` are not cross platform. Not sure if cothority is supposed to support Windows now. If not, I'm just opening this issue for future support :)


mahdiz:
Another option is to move the function `getRTime()` to a separate go file `rtime_linux.go` with the linux build flag. Then, add a Windows version of the function `getRTime()` and put it in a `rtime_windows.go` file with `// +build windows`. This PR does this.

ineiti:
Looks very nice. Can you also put the `iiToF` into `rtime.go`? This is not used elsewhere anymore.

lib/monitor/rtime_windows.go
```diff
@@ -0,0 +1,21 @@
...
+package monitor
+
+import (
+	"github.com/dedis/cothority/lib/dbg"
+	"syscall"
+)
+
+// Returns the system and the user CPU time used by the current process so far.
+func getRTime() (tSys, tUsr float64) {
+	var creationTime, exitTime, kernelTime, userTime syscall.Filetime
```
nikkolasg: Nicely found: it does not show in the [syscall package](https://golang.org/pkg/syscall/).

lib/monitor/rtime_windows.go
```diff
@@ -0,0 +1,21 @@
...
+	hProcess, _ := syscall.GetCurrentProcess()
+	if err := syscall.GetProcessTimes(hProcess, &creationTime, &exitTime, &kernelTime, &userTime); err != nil {
+		dbg.Error("Couldn't get rusage time:", err)
```
nikkolasg: return default value here instead of going further

lib/monitor/rtime.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"github.com/dedis/cothority/lib/dbg"
+	"syscall"
+)
+
+// Returns the system and the user CPU time used by the current process so far.
+func getRTime() (tSys, tUsr float64) {
+	rusage := &syscall.Rusage{}
+	if err := syscall.Getrusage(syscall.RUSAGE_SELF, rusage); err != nil {
+		dbg.Error("Couldn't get rusage time:", err)
```
nikkolasg: Return default value instead of going further

mahdiz:
@nikkolasg: Fixed the return values.
@ineiti: Merged `development` into my branch and added PR #442 against `development`
Thanks :)

PR #441, opened 2016-06-20
Log-library

ineiti:
Merged in development. This merges `app/lib/ui` and `lib/dbg` into `log`


log/common.go
```diff
@@ -0,0 +1,36 @@
+package log
+
+import (
+	"io"
+	"os"
+)
+
+// For debugging purposes we can change the output-writer
```
nikkolasg: Where can we change those ?

log/lvl.go
```diff
@@ -0,0 +1,385 @@
+// Package log is an output-library that can print nicely formatted
```
nikkolasg: How come all of this appear as new files ? and not just renaming + some adjustements ? Is it because of IntelliJ ?
`git mv` should not do that.

log/lvl.go
```diff
@@ -0,0 +1,385 @@
...
+
+// LinePadding of line-numbers for a nice debug-output - used in the same way as
+// NamePadding
+var LinePadding = 3
+
+// Testing variable can have multiple values
+// 0 - no testing
+// 1 - put all line-numbers to 0
+// 2 - like 1, but write to TestString instead of stdout
+var Testing = 0
```
nikkolasg: Can we have that as unexported since it seems for testing ?

log/common.go
ineiti: https://github.com/dedis/cothority/pull/441/files#diff-cd6b3fa641dad75332d64df082d098ccR14

log/lvl.go
ineiti: Because I re-wrote most of these to put all methods together:

- lvl.go - all Lvl.. methods
- ui.go - other log-commands
- common.go - structures and variables

log/lvl.go
ineiti: yes - before `lvl_test.go` was defined as `package lvl_test`, so I needed it to be exported.

log/lvl.go
```diff
@@ -0,0 +1,377 @@
...
+// UseColors returns the actual setting of the color-usage in log
+func UseColors() bool {
+	debugMut.Lock()
+	defer debugMut.Unlock()
+	return useColors
+}
+
+// MainTest can be called from TestMain to set up some default
+// variables
+func MainTest(m *testing.M) {
```
nikkolasg: Let's do a pragmatic programmer approach: it'd be nice to have a comment saying what the *default variables* => what's the output state of `m`

log/lvl.go
```diff
@@ -0,0 +1,377 @@
...
+	code := m.Run()
+	AfterTest(nil)
+	os.Exit(code)
+}
+
+// ParseEnv looks at the following environment-variables:
+// - DEBUG_LVL - for the actual debug-lvl - default is 1
+// - DEBUG_TIME - whether to show the timestamp - default is false
+// - DEBUG_COLOR - whether to color the output - default is true
+func ParseEnv() {
```
nikkolasg: Should it be exported ? Would it make sense to make it unexported and run it on default in `init` ? I mean that's what env. variables are for, if they are defined, use them. I find it a bit too much to have to call this function also in order to use env. variables. 
What do you think ?

log/lvl.go
```diff
@@ -0,0 +1,377 @@
...
+		useColors, err = strconv.ParseBool(dc)
+		Lvl3("Setting useColor to", dc, showTime, err)
+		if err != nil {
+			Error("Couldn't convert", dc, "to boolean")
+		}
+	}
+}
+
+// AddFlags adds the flags and the variables for the debug-control
+func AddFlags() {
```
nikkolasg: Maybe `RegisterFlags` ? 
Also, it'd be nice to specify that it is using the official  `flags` package

log/common.go
```diff
@@ -0,0 +1,36 @@
...
+	lvlWarning = iota - 20
+	lvlError
+	lvlFatal
+	lvlPanic
+	lvlInfo
+	lvlPrint
+)
+
+// Format indicates how the prints will be shown
+var Format = FormatLvl
```
nikkolasg: it seems that whatever the `Format` is, `Lvl{1,2,3,4,5}` will always use the `FormatLvl` format,meaning with the name of the function, line etc.
So if people wants to use `Info,Warn...` they have to explicitly change the format. Am I missing something ?
If not, it may be best to directly put the format to `FormatPython` or `FormatNone`, so people don't have to change it.

log/testutil.go
```diff
@@ -0,0 +1,78 @@
...
+		if t != nil {
+			t.Fatalf("Test leaks %d gorountines.", len(stackCount))
+		} else {
+			Fatal(fmt.Sprintf("Test leaks %d gorountines.", len(stackCount)))
+		}
+	}
+}
+
+// Stack converts []byte to string
+func Stack() string {
```
nikkolasg: Is there a need for that to be exported ?

log/lvl.go
ineiti: OK - probably not the best method to start with, as `testing.M` is golang-intern. But good for the other methods.

log/lvl.go
ineiti: I prefer having this exported, so that the user has control whether he wants this to be automated or not. But if you and @Liamsi think it's OK to put that in `init`, I'll put it there.

log/common.go
ineiti: You miss the point why I wanted to have `dbg` separated from `ui` - because it's two different use-cases: 
- `dbg` for helping in debugging
- `ui` for always interaction with the user

Now we have functions like `Print`, `Info`, ... which are used in both, what to do?

Another approach would be to have only `FormatPython` and `FormatNone`, and set `FormatLvl` by default if the `debugVisible` is > 0. What do you think?

log/testutil.go
ineiti: Yes, this is a convenience function for debugging, used in some places like `network`.

log/lvl.go
ineiti: I would like to discuss a little bit the format of these `contractz` (sic) we want to write. Because it won't be only here, but also in `cosi`, `bftcosi` and others ;)

log/common.go
nikkolasg: I see. Why I wanted to have in one central place is because they do the same thing and it *can* be used in both situations:
* In cothorityd, you want to have debug output AND general purpose output.
* In client apps, you also want the same as you would want to debug also (connection opening, closing, error while verifying signature etc).

In both cases, only the *format* changes.
The way I saw the merging between `ui` and `dbg` was more stg like there is **only** `LvlXXX` or `Info,Warn,Debug` but not both, because like you said, it's inconsistent.
So a client would set the right format `FormatPython` and cothorityd would set `FormatDbg` and both will uses the same *unified* methods (`Info`, `Warn`, `Debug`... I prefer these names because it's clear when to use them unlike `LvlXX` where we don't even use half of it)

Lol we have an issue here, so it may be a good time to discuss with @Liamsi on the whiteboard in some point soon.

log/lvl.go
Liamsi: Huge parts of lvl.go look indeed very familiar. Using git mv here and modifying the file would have helped a lot seeing the changes. Now it looks like a completely new file.

log/common.go
Liamsi: The more time we spend with this whole `dbg` library, the more I'm convinced (again) we should use something which already exists instead of writing our own library. Besides the log framework in the standard golang libs there are plenty of other libraries giving us this functionality.

nikkolasg: mhhh if the golang standard library can give us the function name with line with timestamps and all of that activable, then +1 for that.
-1 for using other external libraries.

Liamsi: But using external libraries like this color output lib https://github.com/dedis/cothority/pull/441/files#diff-642c66c6c818f27e92e6829c75a89d86R56 is +1? Doesn't really make any sense to me.

nikkolasg: Ok i've done some research about logging, here's a good article [0].
It seems the most known third party library for logging (on github) seems very horrible [1], I don't much like it because it looks too complicated. I also want to keep the fact that we don't need to *instantiate* anything using the `dbg` library, using `logrus`, it seems we have to have the same control we need.
Actually it seems the default log library in Golang can support what we need [2], but unfortunatly, their levels are not quite clear, in fact they don't have any levels.

To me, since we already quite a good codebase in the `dbg` library, it does not seem a good idea to throw it away. But it definitely some refactoring so it's usable in both case (*debug view* and *client view* case) in a unified manner (and switchable on/off etc etc). Personally, I prefer the levels using clear words also: `Info Notice Warn Error Debug Fatal Panic` as `logrus` is doing (taken from their github page):
```
log.Debug("Useful debugging information.")
log.Info("Something noteworthy happened!")
log.Warn("You should probably take a look at this.")
log.Error("Something failed but I'm not quitting.")
// Calls os.Exit(1) after logging
log.Fatal("Bye.")
// Calls panic() after logging
log.Panic("I'm bailing.")
```

Theses are *classics* that everyone can understand and use the same way it's been used in other languages/libraries.

[0] http://dave.cheney.net/2015/11/05/lets-talk-about-logging
[1] https://github.com/Sirupsen/logrus
[2] https://golang.org/pkg/log/

ineiti: That works for the `common.go`, but not for the lvl. What about the proposition I made above of throwing `FormatLvl` away?

nikkolasg: Yep let's do that. I filled an issue for that at #444

nikkolasg:
For me it's fine since we want to start over a *sane* base for all repos. I filled the issue #444 which we should discuss soon enough.

Liamsi:
Also OK

ineiti:
Run a randhound to know who clicks on `merge`

PR #448, opened 2016-06-22
added Changelog

ineiti:


nikkolasg:
OK for me but how do you envision we update the changelog ?
* Manually ? Which commits are we supposed to put ? I think one solution would be for every PR into master we should squash the commits into one
* Autmatically ? If we squash the commits then there also exists some automatic tools to generate changelog ...

ineiti:
I added a comment to the CoSi-Changelog saying

# Everything that is changing and might break other people's
# code should be put in here

Does that make sense? We can also add new things available. I would NOT put bug-fixes in there.

nikkolasg:
>I would NOT put bug-fixes in there.
You would not put bug-fixes in the changelog ?
Also I don't understand why you put "Everything that is changing....", as for me, a changelog is just a transcript of the modifications from a version to another. 
What am I missing ?

ineiti:
only write the important things in the Changelog that other developers *need* to know.

Liamsi:
> only write the important things in the Changelog that other developers need to know.

I totally agree with that. We need to make sure that the log makes sense to someone who isn't very, very familiar with the code. Or do you think that's overkill?

Liamsi:
But bugfixes are also relevant to other developers and users.

Liamsi:
As a changelog by it's nature is rather a moving target then something fixed, I'm OK to merge this one. We could start with manually updating the changelog and see hwo well it works out.

Liamsi:
Merged. Please refer to the issue either in a commit or in the PR itself. So that the reviewer doesn't need to search for it.

PR #456, opened 2016-06-24
Enable cosi simul

ineiti:
Adds a protocol and a service to being able to simulate CoSi again. This needs some changes to `dedis/cosi` as in PR https://github.com/dedis/cosi/pull/44


protocols/cosimul/protocol.go
```diff
@@ -0,0 +1,83 @@
...
+		// This is the leaf-node and we can't verify it
+		return
+	}
+
+	verify := false
+	switch VerifyResponse {
+	case 1:
+		verify = c.IsRoot()
+	case 2:
+		verify = !c.IsLeaf()
```
nikkolasg: ? My boolean logic may be broken this morning but 2 <=> everyone verifies:
* IsLeaf() = true   =>  verify  =  false  => !? (all nodes are supposed to verify)
* IsLeaf() = false  =>  verify  =  true   => OK

ineiti: I should update the comment then, because it doesn't make sense for the leafs to verify the response.

nikkolasg: Yes it does, we want to be able to verify the responses at each level so we can know where the error occurred before the response go to the root.

ineiti: What is the purpose of the leaf verifying it's own response?

nikkolasg: Well you can make a special case for the leaves if you want but all the other intermediate + root node can/must verify the responses.

nikkolasg:
When it's working with Travis, I'm all fine.

protocols/cosimul/protocol.go
Liamsi: Hmm, the problem is that the code doesn't really speak for itself. I think it would be more clear if we used an enum instead of integers. @ineiti: do you agree or do you think it's beyond the scope of this issue?

Liamsi:
> When it's working with Travis, I'm all fine.

Just saw that. I put it back to WIP

ineiti:
it needs https://github.com/dedis/cosi/pull/44 first

Liamsi:
Doesn't build currently -> WIP

protocols/cosimul/cosi_test.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/dedis/cothority/log"
+	"github.com/dedis/cothority/sda"
+	"github.com/dedis/crypto/cosi"
+)
+
+func TestCosi(t *testing.T) {
```
Liamsi: nitpick: s/Cosi/CoSi but it's just a test, so it's OK

protocols/cosimul/cosi_test.go
```diff
@@ -0,0 +1,59 @@
+package cosimul
```
Liamsi: +1 for the funny package name :-)

protocols/cosimul/cosi_test.go
ineiti: should be `TestCoSimul` anyway...

protocols/cosimul/protocol.go
```diff
@@ -0,0 +1,90 @@
...
+type VRType int
+
+const (
+	// NoCheck will do no check at all
+	NoCheck = VRType(0)
+	// RootCheck will check only at root
+	RootCheck = VRType(1)
+	// AllCheck check at each level of the tree, except the leafs
+	AllCheck = VRType(2)
+)
```
Liamsi: Could bet simplified to:
```
type VRType int
const (
 	// NoCheck will do no check at all
 	NoCheck VRType = iota
 	// RootCheck will check only at root
 	RootCheck 
 	// AllCheck check at each level of the tree, except the leafs
 	AllCheck 
 )

File: services/cosimul/simulation.go [hunk 5]
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/BurntSushi/toml"
+	"github.com/dedis/cosi/service"
+	"github.com/dedis/cothority/log"
+	"github.com/dedis/cothority/monitor"
+	"github.com/dedis/cothority/sda"
+	"github.com/dedis/crypto/cosi"
+)
+
+func init() {
+	sda.SimulationRegister("CosiService", NewSimulation)
```
Liamsi: Shouldn't this also be "CoSiService"?

ineiti: Should be `CoSimulService` - but in fact it's of no use. Deleted.

Liamsi:
Seems like the linter doesn't pass. Besides that (and the other nitpicks which you might ignore), I think this is ready to merge.

Liamsi:
A general comment: The new file https://github.com/dedis/cothority/blob/45794173820f96e6b39076178a7dd44e87d12d38/protocols/cosimul/simulation.go looks very similar to the deleted one: https://github.com/dedis/cothority/pull/456/commits/10a9cd624bd7a0a96c3df978784cfb9d11d81342 Although here it's just one small file and there were quite some changes [1] use renaming/moving as often as possible. Creating new files just creates noise and it takes more time to review.

[1] https://www.diffchecker.com/i7fa4b4h

PR #460, opened 2016-06-29
remove unnecessary constants

ineiti:
also removed the init-function and moved `app.Before` into `app.Action`.


Liamsi:
related to issue ?

ineiti:
Some cleaning up of the app where I felt bad about bugging Evan all the time ;)

Issue #455, opened 2016-06-23
Make the configuration in service actually work


nikkolasg:
For the moment, the implementation is not done for the configuration parameter in `NewProtocol` for a `Service`. We should do that soon. 
The requirements are that for the first message a "service-managed" protocol sends down the tree, TreeNodeInstance also sends a `GenericConfig` which will be passed into the `NewProtocol` method of the services down the tree.
In order to do that, it seems logical to add a `SetGenericConfig` to `TreeNodeInstance` even if it looks ugly, I did not come up with a better solution yet. Any ideas ?

ineiti:
Done.

PR #464, opened 2016-07-04
Bftcosi

ineiti:
New PR for BftCoSi - This is to make bftcosi work with the new cosi-code from `crypto/cosi`.


protocols/bftcosi/bftcosi.go
```diff
@@ -3,33 +3,38 @@
...
+type VerificationFunction func(Msg []byte, Data []byte) bool
 
 // ProtocolBFTCoSi is the main struct for running the protocol
 type ProtocolBFTCoSi struct {
 	// the node we are represented-in
 	*sda.TreeNodeInstance
+	// The message that will be signed by the BFTCosi
 	Msg []byte
+	// Data going along the msg to the verification
+	Data []byte
```
Liamsi: Question: I do not really understand why this field is necessary. (Why would it be useful for example for ByzCoin?)

protocols/bftcosi/bftcosi.go
```diff
@@ -3,33 +3,38 @@
 package bftcosi
 
 import (
-	"math"
+	"crypto/sha512"
 	"sync"
 
-	"github.com/dedis/cothority/cosi"
+	"gopkg.in/dedis/cothority.v0/lib/dbg"
```
Liamsi: Is there a reason for using the `v0` version here?

protocols/bftcosi/bftcosi.go
```diff
@@ -51,40 +56,40 @@ type ProtocolBFTCoSi struct {
...
 	// protocol when all nodes have finished. Either at the end of the response
 	// phase of the commit round or at the end of a view change.
-	onDoneCallback func()
+	onDone func()
 
 	// onSignatureDone is the callback that will be called when a signature has
 	// been generated ( at the end of the response phase of the commit round)
 	onSignatureDone func(*BFTSignature)
 
 	// verificationFun will be called
```
Liamsi: renamed `verficationFun`

protocols/bftcosi/bftcosi.go
```diff
@@ -51,40 +56,40 @@ type ProtocolBFTCoSi struct {
...
 	// during the (start/handle) challenge prepare phase of the protocol
-	verificationFun VerificationFunction
+	VerificationFunction VerificationFunction
```
Liamsi: Is it necessary for this to be public?

protocols/bftcosi/bftcosi.go
ineiti: Usually the services are supposed to being able to pass data along an instantiation of the protocol. But that part is not done yet, so the most easy one was to add this data-field here. This should go away as soon as the services can send messages for the instantiation of a protocol. See also #455

protocols/bftcosi/bftcosi.go
ineiti: No reason - will be corrected.

ineiti:
updated top comment.

protocols/bftcosi/bftcosi.go
```diff
@@ -97,23 +102,46 @@ type ProtocolBFTCoSi struct {
...
+	// commitSignature is the signature generated during the commit phase
+	// This signature is adapted according to the exceptions that occured during
+	// the commit phase.
+	commitSignature []byte
+
+	// commitCommitDone is to ensure the Challenge phase of the Commitment phase
+	// (the second commit) is done, before doing the challenge.
+	// It's needed because both phases are started *almost* simultaneously and
+	// the commit phase can be de-sync.
+	commitCommitDone chan bool
```
Liamsi: If I understood it correctly the above new lines are only there to have some sort of exception handling. Maybe that's overkill but couldn't this fields be bundled into some struct?

protocols/bftcosi/bftcosi.go
```diff
@@ -97,23 +102,46 @@ type ProtocolBFTCoSi struct {
...
+	// our index in the Roster list
+	index int
+	// prepareSignature is the signature generated during the prepare phase
+	// This signature is adapted according to the exceptions that occured during
+	// the prepare phase.
+	prepareSignature []byte
+	// commitSignature is the signature generated during the commit phase
+	// This signature is adapted according to the exceptions that occured during
+	// the commit phase.
+	commitSignature []byte
```
Liamsi: If I understood it correctly the above new lines are only there to have some sort of exception handling. Maybe that's overkill but couldn't this fields be bundled into some struct?

Liamsi: If I understood it correctly the above new lines are only there to have some sort of exception handling. Maybe that's overkill but couldn't this fields be bundled into some struct?

ineiti: This is for the first round of the protocol where BFTCoSi counts how many nodes are OK to sign off. I also think that the structure could be a little bit more lightweight for what it is. Any suggestions?

Makefile
```diff
@@ -28,7 +28,7 @@ test_lint:
 	}
 
 test_go:
-	go test -race -p=1 ./...
+	go test -race -p=1 -short ./...
```
Liamsi: +1 for faster tests! But I see below that we now `Skip` a lot of short tests, too. Ist that on purpose?

ineiti: I skip mainly the extensive verification-failure testing of the bftcosi protocol. Also because this fails from time to time on Travis, and I never got it to fail on my computer. So I only run the simple bftcosi tests.

protocols/bftcosi/bftcosi.go
Liamsi: Hmm, no really good suggestion, but one idea would be to move the above fields into a (private) struct and all the temp* fields, too. I'm sure there would be nicer ways to make the code more readable but I would need to think more about that.

protocols/bftcosi/bftcosi.go
```diff
@@ -295,105 +312,106 @@ func (bft *ProtocolBFTCoSi) handleCommit(comm Commitment) error {
...
 
-	ch, err := bft.prepare.CreateChallenge(h)
-	if err != nil {
-		return err
-	}
+	log.Lvl4(bft.Name(), "RoundType:", t)
+	if t == RoundPrepare {
+		// need to hash the message before so challenge in both phases are not
+		// the same
+		data := sha512.Sum512(bft.Msg)
```
Liamsi: Is there any reason for not using the suites hash function?

Makefile
Liamsi: Some of the skipped tests look super simple. But maybe I missed their complexity on first sight.

ineiti: For future reference: this is because of #467

protocols/bftcosi/bftcosi.go
ineiti: The `crypto/cosi`-signature uses sha512, as far as I can tell. At least replacing that with `suite.Hash` doesn't work...

protocols/bftcosi/bftcosi.go
```diff
@@ -1,35 +1,51 @@
...
 	"github.com/dedis/cothority/sda"
 	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/cosi"
 )
 
 // VerificationFunction can be passes to each protocol node. It will be called
 // (in a go routine) during the (start/handle) challenge prepare phase of the
 // protocol. The passed message is the same as sent in the challenge phase.
-type VerificationFunction func(Msg []byte) bool
+type VerificationFunction func(Msg []byte, Data []byte) bool
```
nikkolasg: Maybe write a comment saying the `Data` part will be removed later when the `GenericConfig` works ? Because having a signature like this is definitely not something we want to keep. 
Either that or we make a PR for having `GenericConfig` working before merging this one.

ineiti:
Added comment about data - once Travis passed, is this OK to merge?

nikkolasg:
Bbftcosi in itself is not finished, because there's still the "view change" thing that needs to be done: bftcosi will change again. In that perspective, I'm *ok to merge now*. Otherwise I'd had prefer to remove the `data` thing.

ineiti:
Anyway, cothority is development. It should pass the tests, but besides that it's OK to have a somewhat moving target.

Liamsi:
Also OK.

PR #465, opened 2016-07-04
Service skipchain

ineiti:
Re-opened against master
medco is failing - not related to skipchain


services/skipchain/msgs.go
```diff
@@ -0,0 +1,119 @@
...
+		// Propagation
+		&PropagateSkipBlock{},
+		// Requests for data
+		&GetUpdateChain{},
+		&GetUpdateChainReply{},
+		// Data-structures
+		&ForwardSignature{},
+		&SkipBlockFix{},
+		&SkipBlock{},
+	}
```
Liamsi: Didn't we already had and reviewd that? I remember we coded that together before OSDI?

services/skipchain/skipchain.go
```diff
@@ -0,0 +1,474 @@
...
+// If the the latest block given is nil it verify if we are actually creating
+// the first (genesis) block and creates it. If it is called with nil although
+// there already exist previous blocks, it will return an error.
+func (s *Service) ProposeSkipBlock(e *network.ServerIdentity, psbd *ProposeSkipBlock) (network.Body, error) {
+	prop := psbd.Proposed
+	var prev *SkipBlock
+
+	if !psbd.LatestID.IsNull() {
+		// We're appending a block to an existing chain
+		var ok bool
```
Liamsi: why not use `ok :=`

Liamsi: Ah I see, because of prev

services/skipchain/msgs.go
ineiti: Never mad it into the R4M queue I believe :(

services/skipchain/struct.go
```diff
@@ -0,0 +1,229 @@
...
+		return err
+	}
+	//for _, fl := range sb.ForwardLink {
+	//	if err := fl.VerifySignature(sb.Aggregate); err != nil {
+	//		return err
+	//	}
+	//}
+	//if sb.ChildSL != nil && sb.ChildSL.Hash == nil {
+	//	return sb.ChildSL.VerifySignature(sb.Aggregate)
+	//}
```
ineiti: Yes, #478.

ineiti:
AFAIR the OSDI-part was hacked together and did not work at all as it was never used. This is why it hasn't been merged ;) Now it's cleaned-up (remember the structure/interface disaster we had?), uses BFTCoSi and 'only' needs to pass Travis and check all new blocks.

nikkolasg:
OK

PR #466, opened 2016-07-04
Log print correct

ineiti:
Print correct output for `Print` and others when in debug-output.


nikkolasg:
OK for me.

Liamsi:
The changes done here: white output if we are in the info level and  ... ? Is there an issue related to this PR describing what this is about?

ineiti:
Sorry - the problem was that `log.Print` didn't give correct line-informations when in debug-mode. It was off by one line in the callstack... Furthermore `log.Info` didn't get correctly tagged when using debug-mode. This patch fixes both.

Liamsi:
OK for me, too!

Issue #426, opened 2016-06-15
SDA Refactor - Unify the Processor interface


nikkolasg:
I propose we unify the work we started to do at different places in the code, namely our way to process messages. In Service, we have this `ServiceProcessor` and in protocols, we have `TreeNodeInstance` which can be used to register message handlers etc.
We should unify this into a `Processor` interface which could be basically defined as 

```
type Processor interface {
   ProcessMessage(e *network.Entity, msg network.ProtocolMessage)
}
```

This interface will be implemented by a `HandlerProcessor`, a `ChannelProcessor` or whatever comes up to mind, and these implementations will be used both by `Service` and `ProtocolInstance`.

nikkolasg:
Ok so regarding our talk this afternoon, here's some precision about the issue I'm doing:
* Having a `Processor` interface which will be implemented by a `ChannelProcessor` and `HandlerProcessor` so those nice functionalities can be re-used somewhere else and are not tied to `TreeNodeInstance`. 
* Having a  `Dispatcher` which store `Processor` and dispatch message to them. For the moment a simple one is implemented (`BlockingDispatcher`) but later, one with queue, and/or worker/consumer will have to be implemented. This is needed because we are starting to put go-routines everwhere for each new message that `Host` is handling so... 

Basically that's it.

nikkolasg:
UPDATE: The part with `ProtocolInstance`  seems relatively complex to port using generic `Processors` and `Dispatcher` and also there is no clear direct benefits, thus we decided to postpone it until further need.
However, I will continue for the rest of the week to port `Service` as a `Processor` so the `Service` interface is lighter and the `InterServiceMessage` generic packet is hidden from the `Service` developer now.

nikkolasg:
After long discussions and review, this is neither needed at the moment nor really easily implementable in a nice way (with the least amount of casting / reflect)

PR #468, opened 2016-07-11
Refactor SDA - Processor & Dispatcher

nikkolasg:
Here's a PR for #426 . There's little bit more than what's described in the text issue as I did not take into account re-usability in the issue. Namely, there is now a `Dispatcher` whose role is to correctly give message to `Processors`. `Host` embeds now a `Dispatcher`.

I'll edit the .travis.yml file so `dedis/cosi` will be checked-out on the same branch in order to make Travis happy.


Liamsi:
travis fails -> back to WIP (I think it's the race condition which is fixed in #473; the deadlock is still there though)

Makefile
```diff
@@ -1,31 +1,42 @@
...
 		exit 1; \
-	fi; \
-	if ! go vet ./...; then \
+		fi; \
+		if ! go vet ./...; then \
 		exit 1; \
-	fi \
-	}
+		fi \
+		}
```
Liamsi: Were this new tabs intentionally created? If yes, why is the closing bracket on a different level than the opening one?

Makefile
```diff
@@ -1,31 +1,42 @@
...
+		exclude="protocols/byzcoin|_test.go"; \
+		lintfiles=$$( golint ./... | egrep -v "($$exclude)" ); \
+		if [ -n "$$lintfiles" ]; then \
 		echo "Lint errors:"; \
 		echo "$$lintfiles"; \
 		exit 1; \
-	fi \
-	}
+		fi \
+		}
```
Liamsi: same as above

Liamsi:
As travis only fails because of something (seemingly) unrelated to this issue, I'm OK with skipping the jvss test as long as it still fails. So we can continue with this (currently more urgent) work an merge it. In #473 the race condition is resolved but there is still some deadlock and it seems that the order of short-term and long-term secrets isn't always as intended. I think I still need to investigate one or two days on how-to fix the that.

protocols/medco/deterministic_switching_test.go
```diff
@@ -1,17 +1,20 @@
...
 	"github.com/dedis/cothority/sda"
 	. "github.com/dedis/cothority/services/medco/libmedco"
 	"github.com/dedis/crypto/random"
-	"testing"
-	"time"
 )
 
 func TestDeterministicSwitching(t *testing.T) {
+	// see https://github.com/dedis/cothority/issues/479
+	t.Skip()
```
Liamsi: General remark: a short message saying why this is skipped (and ideally that it will be resolved by XXX) would be nice

sda/processor_test.go
```diff
@@ -10,18 +10,102 @@ import (
...
+//agg.aggregate(basicMessageType, &basicMessage{12})
+//agg.aggregate(basicMessageType, &basicMessage{13})
+//agg.aggregate(basicMessageType, &basicMessage{14})
+//agg.aggregate(basicMessageType, &basicMessage{15})
+
+//agg.aggregateType(basicMessageType, 3)
+//assert.Len(t, agg.release(basicMessageType), 3)
+//assert.Nil(t, agg.release(basicMessageType), nil)
+
+//}
```
Liamsi: Why is this here?

sda/service.go
```diff
@@ -155,19 +160,23 @@ func (s *serviceFactory) start(name string, c *Context, path string) (Service, e
 // serviceStore is the place where all instantiated services are stored
 // It gives access to :  all the currently running services and is handling the
 // configuration path for them
-type serviceStore struct {
+type serviceManager struct {
 	// the actual services
 	services map[ServiceID]Service
 	// the config paths
 	paths map[ServiceID]string
+	// the main sda host
```
Liamsi: what differentiates a "main" sda host from a regular one?

sda/service.go
```diff
@@ -189,37 +199,74 @@ func newServiceStore(h *Host, o *Overlay) *serviceStore {
...
+		if !ok {
+			log.Error("Received a request for an unknown service", r.Service)
+			// XXX TODO should reply with some generic response =>
+			// 404 Service Unknown
+			return
+		}
+		go s.ProcessClientRequest(id, &r)
+	default:
+		// will launch a go routine for that message
+		s.Dispatch(data) //s.ProcessServiceMessage(id, &m)
```
Liamsi: Maybe remove: "//s.ProcessServiceMessage(id, &m)"

Makefile
nikkolasg: vim auto formatting :/

sda/processor_test.go
nikkolasg: I spend quite some time on this before letting it down. I still hoped it could be useful but then no.. :D

sda/service.go
nikkolasg: removed *main*

Liamsi:
Like in #473 skipchains fail now: https://travis-ci.org/dedis/cothority/builds/144960803#L203

Liamsi:
Skip the skipchain tests until #428 is resloved? /cc @ineiti

nikkolasg:
I'd say yes.

Liamsi:
> I'd say yes.

As it is already merged, I think it's the only way which makes sense (and of course quickly fixing #482 )

Liamsi:
OK for me.

CHANGELOG
```diff
@@ -8,3 +8,7 @@
 	- Renamed network.Message -> network.Packet
 	- Moved lib/* to parent-directory
 	- Depend on dedis/cosi for the cosi-library
+
+Refactoring version (XXX Set to right number):
+    - unify Processor interface 
+    - ...
```
ineiti: Can you write a short list of the things that changed?

Makefile
```diff
@@ -1,30 +1,41 @@
 install:
+	@export PR=https://api.github.com/repos/$$TRAVIS_REPO_SLUG/pulls/$$TRAVIS_PULL_REQUEST; \
+	export BRANCH=$$(if [ "$$TRAVIS_PULL_REQUEST" == "false" ]; then echo $$TRAVIS_BRANCH; else echo `curl -s $$PR | jq -r .head.ref`; fi); \
+	echo "TRAVIS_BRANCH=$$TRAVIS_BRANCH, PR=$$PR, BRANCH=$$BRANCH"; \
+	pattern="refactor_"; \
+	if [[ $$BRANCH =~ $$pattern ]]; then \
```
ineiti: Something in here is not correct yet:

https://travis-ci.org/dedis/cothority/builds/145013562#L161

Most probably you should write `"$$BRANCH"` and `"$$pattern"`.

Makefile
```diff
@@ -1,30 +1,41 @@
...
+	fi;\
+	cd $$GOPATH/src/github.com/dedis/cothority; \
 	go get -t ./...
 
 test_fmt:
 	@echo Checking correct formatting of files
 	@{ \
-	files=$$( go fmt ./... ); \
-	if [ -n "$$files" ]; then \
+		files=$$( go fmt ./... ); \
```
ineiti: That was probably based on my version, but anyway: can you correct (my) bad and remove the `@{`? I don't remember why it is here and it doesn't make sense to me.

Makefile
```diff
@@ -1,30 +1,41 @@
...
+		fi; \
+		if ! go vet ./...; then \
 		exit 1; \
-	fi \
+		fi \
 	}
 
 test_lint:
 	@echo Checking linting of files
 	@{ \
```
ineiti: same as above - just because we're re-writing the Makefile, it's a good moment to clean this up.

protocols/medco/deterministic_switching_test.go
```diff
@@ -13,6 +13,9 @@ import (
 )
 
 func TestDeterministicSwitching(t *testing.T) {
+	// Issue with deadlock occurences in the medco protocol
+	// see https://github.com/dedis/cothority/issues/479
+	t.Skip()
```
ineiti: Put the issue in the skip itself:
`t.Skip("Issue #479 with deadlock occurences in the medco protocol")`

sda/overlay.go
```diff
@@ -30,20 +31,135 @@ type Overlay struct {
...
+	// messages going to protocol instances
+	h.RegisterProcessor(o, SDADataMessageID)
+	// requesting a tree
+	h.RegisterProcessor(o, RequestTreeMessageID)
+	// sending back a tree
+	h.RegisterProcessor(o, SendTreeMessageID)
+	// requesting a roster
+	h.RegisterProcessor(o, RequestRosterMessageID)
+	// sending back a roster
+	h.RegisterProcessor(o, SendRosterMessageID)
```
ineiti:
```
for _, m := range []interface{ SDADataMessageID, RequestTreeMessageID, ...}{
    h.RegisterProcessor(o, m)
}
```
?

sda/overlay.go
```diff
@@ -30,20 +31,135 @@ type Overlay struct {
...
+	switch data.MsgType {
+	case SDADataMessageID:
+		sdaMsg := data.Msg.(ProtocolMsg)
+		sdaMsg.ServerIdentity = data.ServerIdentity
+		err := o.TransmitMsg(&sdaMsg)
+		if err != nil {
+			log.Error("ProcessSDAMessage returned:", err)
+		}
+
+	// A host has sent us a request to get a tree definition
```
ineiti: I prefer the comments _inside_ the case. Like the comment at the end of `RequestTreeMessageID` doesn't really make sense to me.

sda/overlay.go
```diff
@@ -30,20 +31,135 @@ type Overlay struct {
...
+
+		tree, err := tm.MakeTree(il)
+		if err != nil {
+			log.Error("Couldn't create tree:", err)
+			return
+		}
+		log.Lvl4("Received new tree")
+		o.RegisterTree(tree)
+		o.checkPendingSDA(tree)
+		// Some host requested an Roster
```
ineiti: Put comment inside of corresponding case.

sda/overlay.go
```diff
@@ -30,20 +31,135 @@ type Overlay struct {
...
+			log.Lvl2("Requested entityList that we don't have")
+			err = o.host.SendRaw(data.ServerIdentity, &Roster{})
+		}
+		if err != nil {
+			log.Error("Couldn't send empty entity list from host:",
+				o.host.ServerIdentity.String(),
+				err)
+			return
+		}
+		// Host replied to our request of entitylist
```
ineiti: Put comment inside of corresponding case.

sda/overlay.go
```diff
@@ -123,12 +239,115 @@ func (o *Overlay) TransmitMsg(sdaMsg *ProtocolMsg) error {
...
+// and give it the message!.
+// NOTE: put that as a go routine so the rest of the processing messages are not
+// slowed down, if there are many pending sda message at once (i.e. start many new
+// protocols at same time)
+func (o *Overlay) checkPendingSDA(t *Tree) {
+	go func() {
+		o.pendingSDAsLock.Lock()
+		var newPending []*ProtocolMsg
+		for _, msg := range o.pendingSDAs {
+			// if this message references t
```
ineiti: If the comment describes the whole decision, put it outside. If the comment describes what happens after the decision, put it inside the if:
```
// check whether we received a message for tree t
if ...
```
or
```
if ...
    // received message for the tree t
else
    // not our message, append to pending messages
```

sda/overlay.go
nikkolasg: Yeah I saw that it is quickly becoming annoying to call multiple times the `RegisterProcessor`. I'm more thinking of doing something like :
```
type Dispatcher interface {
   RegisterProcessor( p Processor, types ...network.MessageTypeID,)
}
// then
h.RegisterProcessor(o, SDADataMessageID, RequestTreeMessageID, ...)
...
What do you think ?

File: sda/overlay.go [hunk 17]
```diff
@@ -123,12 +239,115 @@ func (o *Overlay) TransmitMsg(sdaMsg *ProtocolMsg) error {
...
+}
+
+// checkPendingTreeMarshal is called each time we add a new Roster to the
+// system. It checks if some treeMarshal use this entityList so they can be
+// converted to Tree.
+func (o *Overlay) checkPendingTreeMarshal(el *Roster) {
+	o.pendingTreeLock.Lock()
+	sl, ok := o.pendingTreeMarshal[el.ID]
+	if !ok {
+		// no tree for this entitty list
```
ineiti: Good description inside the branch.

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
 	"github.com/dedis/cothority/network"
 )
 
+// Dispatcher is an interface whose sole role is to distribute messages to the
+// right Processor. No processing is done,i.e. no looking at packet content.
+// There are many ways to distribute messages: for the moment, only
+// BlockingDispatcher is implemented, which is a blocking dispatcher.
+// Later, one can easily imagine to have a dispatcher with one worker in a
+// goroutine or a fully fledged producer/consumers pattern in go routines.
+// Each Processor that want to receive all messages of a specific
```
ineiti: s/want/wants/

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+// type must register them self to the dispatcher using `RegisterProcessor()`.
```
ineiti: s/self/itself/

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+// The network layer must call `Dispatch()` each time it received a message, so
```
ineiti: s/received/receives/

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+// the dispatcher is able to dispatch correctly to the right Processor for
+// further analysis.
+type Dispatcher interface {
+	// RegisterProcessor is called by a Processor so it can receive any packet
```
ineiti: s/any packet/packets/

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+	// of type msgType.
+	// **NOTE** In the current version, if a subequent call RegisterProcessor
```
ineiti: s/call/call to/

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+	// happens, for the same msgType, the latest Processor will be used; there
+	// is no *copy* or *duplication* of messages.
+	RegisterProcessor(p Processor, msgType network.MessageTypeID)
+	/* // Dispatch will find the right processor to dispatch the packet to. The id*/
```
ineiti: That's a strange comment. Can you make it like all others, that is `//`?

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+	//// is the identity of the author / sender of the packet.
+	//// It can be called for example by the network layer.
+	//// If no processor is found for this message type, then an error is returned
+	/*Dispatch(packet *network.Packet) error*/
```
ineiti: Is this still used? Will be used? Necessary?

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+}
+
+// Processor is an abstraction to represent any object that want to process
+// packets. It is used in conjunction with Dispatcher:
+// A processor must register itself to a Dispatcher so the Dispatcher will
+// dispatch every messages to the Processor asked for.
+type Processor interface {
+	// Process takes a ServerIdentity as the sender identity and the message
```
ineiti: I only see one argument in the method.

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+	// sent.
+	Process(packet *network.Packet)
+}
+
+// BlockingDispatcher is a Dispatcher that simply calls `p.Process()` on a
+// processor p each time it receives a message with `Dispatch`. It does *not*
+// launch a go routine, or put the message in a queue, etc.
+// It can be re-uswored for more complex dispatcher.
```
ineiti: s/uswored/used/

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+type BlockingDispatcher struct {
+	procs map[network.MessageTypeID]Processor
+}
+
+// NewBlockingDispatcher will return a freshly initialized BlockingDispatcher
```
ineiti: I don't know if we want to be a..l about making complete sentences in comments and end them with a '.'...

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+func NewBlockingDispatcher() *BlockingDispatcher {
+	return &BlockingDispatcher{
+		procs: make(map[network.MessageTypeID]Processor),
+	}
+}
+
+// RegisterProcessor save the given processor in the dispatcher.
```
ineiti: s/save/saves/

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+	var p Processor
+	if p = d.procs[packet.MsgType]; p == nil {
+		return errors.New("No Processor attached to this message type.")
+	}
+	p.Process(packet)
+	return nil
+}
+
+// RoutineDispatcher is a Dispatcher that will dispatch messages to Processor
+// in a go routine. RoutineDispatcher creates one go routine per messages it
```
ineiti: s/messages/message/

sda/processor.go
```diff
@@ -10,6 +10,97 @@ import (
...
+	return &RoutineDispatcher{
+		BlockingDispatcher: NewBlockingDispatcher(),
+	}
+}
+
+// Dispatch implements the Dispatcher interface. It will give the packet to the
+// right Processor in a go routine.
+func (d *RoutineDispatcher) Dispatch(packet *network.Packet) error {
+	var p Processor
+	if p = d.procs[packet.MsgType]; p == nil {
```
ineiti: If you use the `p` afterwards, I prefer the
```
p := d.procs[packet.MsgType]
if p == nil {
```

sda/processor.go
```diff
@@ -25,13 +116,16 @@ type ServiceProcessor struct {
 
 // NewServiceProcessor initializes your ServiceProcessor.
 func NewServiceProcessor(c *Context) *ServiceProcessor {
-	return &ServiceProcessor{
+	s := &ServiceProcessor{
```
ineiti: Is this for debugging purposes? Then add a `log.Lvl3(s)`, else I'd put the return.

sda/processor.go
```diff
@@ -25,13 +116,16 @@ type ServiceProcessor struct {
...
 		functions: make(map[network.MessageTypeID]interface{}),
 		Context:   c,
 	}
+	return s
 }
 
 // RegisterMessage puts a new message in the message-handler
+// XXX More comments are needed as it's not clear whether RegisterMessage waits
```
ineiti: Where will we find these comments?

sda/processor.go
```diff
@@ -69,22 +163,29 @@ func (p *ServiceProcessor) RegisterMessage(f interface{}) error {
 	return nil
 }
 
+// Process implements the Processor interface and dispatch ClientRequest message
```
ineiti: s/dispatch/dispatches/

sda/processor_test.go
```diff
@@ -10,18 +10,78 @@ import (
...
+	dispatcher.Dispatch(&network.Packet{
+		Msg:     basicMessage{10},
+		MsgType: basicMessageType})
+
+	select {
+	case m := <-processor.msgChan:
+		msg, ok := m.Msg.(basicMessage)
+		assert.True(t, ok)
+		assert.Equal(t, msg.Value, 10)
+	default:
```
ineiti: will this ever be called? Don't we need a `case <-time.After(3*time.Second)`?

sda/processor_test.go
```diff
@@ -10,18 +10,78 @@ import (
...
+	h1.Dispatch(&network.Packet{
+		Msg:     basicMessage{10},
+		MsgType: basicMessageType})
+
+	select {
+	case m := <-proc.msgChan:
+		basic, ok := m.Msg.(basicMessage)
+		assert.True(t, ok)
+		assert.Equal(t, basic.Value, 10)
+	default:
```
ineiti: same as above - will this ever be called?

sda/processor_test.go
```diff
@@ -10,18 +10,78 @@ import (
...
 
 func init() {
+	basicMessageType = network.RegisterMessageType(&basicMessage{})
 	RegisterNewService("testService", newTestService)
 	testServiceID = ServiceFactory.ServiceID("testService")
+	testMsgID = network.RegisterMessageType(&testMsg{})
 }
 
 func TestProcessor_AddMessage(t *testing.T) {
 	defer log.AfterTest(t)
```
ineiti: You can add a
```
func TestMain(m *testing.M){
    log.MainTest(m)
}
```

which does all the `AfterTest` and verbose-interpretation.

sda/service.go
```diff
@@ -32,8 +32,13 @@ type Service interface {
 	// receives a request, it looks whether it knows the Service it is for and
 	// then dispatch it through ProcessRequest.
 	ProcessClientRequest(*network.ServerIdentity, *ClientRequest)
-	// ProcessServiceRequest takes a message from another Service
-	ProcessServiceMessage(*network.ServerIdentity, *InterServiceMessage)
+	// Processor makes a Service being able to handle any kind of packets
```
ineiti: That's a comment that is probably worth putting in the general description of `service.go`.

sda/service.go
```diff
@@ -155,19 +160,23 @@ func (s *serviceFactory) start(name string, c *Context, path string) (Service, e
...
 // configuration path for them
-type serviceStore struct {
+type serviceManager struct {
 	// the actual services
 	services map[ServiceID]Service
 	// the config paths
 	paths map[ServiceID]string
+	// the sda host
+	host *Host
+	// serviceManager can take registration of Processors
```
ineiti: Linting: Put correct name.

sda/service.go
```diff
@@ -189,37 +199,74 @@ func newServiceStore(h *Host, o *Overlay) *serviceStore {
...
+			return
+		}
+		go s.ProcessClientRequest(id, &r)
+	default:
+		// will launch a go routine for that message
+		s.Dispatch(data)
+	}
+}
+
+// Register the processor to the service manager and tells the host to dispatch
```
ineiti: s/Register/RegisterProcessor/ - why does the linter not complain?

sda/service.go
```diff
@@ -189,37 +199,74 @@ func newServiceStore(h *Host, o *Overlay) *serviceStore {
...
+func (s *serviceManager) RegisterProcessor(p Processor, msgType network.MessageTypeID) {
+	// delegate message to host so the host will pass the message to ourself
+	s.host.RegisterProcessor(s, msgType)
+	// handle the message ourself (will be launched in a go routine)
+	s.RoutineDispatcher.RegisterProcessor(p, msgType)
 }
 
 // TODO
-func (s *serviceStore) AvailableServices() []string {
+func (s *serviceManager) AvailableServices() []string {
```
ineiti: Hmm, is this difficult to implement?

sda/service.go
```diff
@@ -189,37 +199,74 @@ func newServiceStore(h *Host, o *Overlay) *serviceStore {
...
 	panic("not implemented")
 }
 
 // TODO
-func (s *serviceStore) Service(name string) Service {
+func (s *serviceManager) Service(name string) Service {
```
ineiti: This is done now, no?

sda/service.go
```diff
@@ -189,37 +199,74 @@ func newServiceStore(h *Host, o *Overlay) *serviceStore {
...
 	return s.serviceByString(name)
 }
 
 // TODO
-func (s *serviceStore) serviceByString(name string) Service {
+func (s *serviceManager) serviceByString(name string) Service {
```
ineiti: when will this be done?

sda/service.go
```diff
@@ -241,6 +288,7 @@ type ClientRequest struct {
 }
 
 // RequestID is the type that registered by the network library
+// TODO rename that ClientRequestID
```
ineiti: is now not a good time?

services/medco/service_test.go
```diff
@@ -18,7 +18,9 @@ const numberAttr = 10
 
 // TestService tests medco complete service execution.
 func TestService(t *testing.T) {
-	t.Skip("Still needs to be debugged.")
+	// Issue with deadlock occuring in the medco protocol
+	// See https://github.com/dedis/cothority/issues/479
+	t.Skip()
```
ineiti: `t.Skip("Issue 479 with deadlock occuring in the medco protocol")`

services/skipchain/skipchain_test.go
```diff
@@ -16,7 +16,9 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	log.MainTest(m)
+	// Issue with skipchain service blocking. currently actively being resolved.
```
ineiti: remove comment.

sda/overlay.go
ineiti: `types` would be the function-type? Yes, looks good. I wanted to do the same for the service-messages and other handlers, too.
Perhaps put it in plural like `RegisterProcessors`?

ineiti: No plural, sorry, doesn't make sense.

Makefile
ineiti: I think it even makes the test not working - `exit 1` will most probably just exit the sub-process generated by `{}`. This is why the linter didn't complain...

Makefile
ineiti: Probably more than a cleaning-up...

sda/processor.go
nikkolasg: Not sure about this one, it's multiple messages, not only one... ?

sda/processor.go
nikkolasg: The question is more like "When" ^^ If you have any ideas how to clarify this function, please !

sda/processor_test.go
nikkolasg: Channel is a buffered one and it's a blocking dispatcher so before the `select`, the message is already into the channel. https://play.golang.org/p/aDtiNV4wms

sda/processor_test.go
nikkolasg: see my previous response

sda/service.go
nikkolasg: Maybe because it recognizes it implements an `Interface` and then he does not require to put any comments on top (because godoc could re-use the comment of the `Interface` ?)

sda/service.go
nikkolasg: It should not be really difficult. I'll open an issue for it.

nikkolasg: see #497

sda/service.go
nikkolasg: Done.

sda/service.go
nikkolasg: Done.

sda/service.go
nikkolasg: See #497

sda/processor.go
```diff
@@ -10,6 +10,99 @@ import (
...
+type Dispatcher interface {
+	// RegisterProcessor is called by a Processor so it can receive all packets
+	// of type msgType. If given multiple msgType, the same processor will be
+	// called for each of all the msgType given.
+	// **NOTE** In the current version, if a subequent call to RegisterProcessor
+	// happens, for the same msgType, the latest Processor will be used; there
+	// is no *copy* or *duplication* of messages.
+	RegisterProcessor(p Processor, msgType ...network.MessageTypeID)
+	// Dispatch will find the right processor to dispatch the packet to. The id
+	// is the identity of the author / sender of the packet.
```
Liamsi: Change to: The id identifies the sender of the packet.

sda/processor.go
Liamsi: https://github.com/dedis/cothority/pull/468/files#diff-1fcf0c916aa94004ca28cd7eee12a6d7R102
here it looks like one go-rountine indeed takes one packet (which holds one message only), right?

nikkolasg: Right sorry, I forgot it was `RoutineDispatcher` sorry ... :D

sda/service.go
```diff
@@ -204,37 +220,76 @@ func newServiceStore(h *Host, o *Overlay) *serviceStore {
...
 		services[id] = s
 		configs[id] = configName
-		// !! register to the ProtocolFactory !!
-		//ProtocolFactory.registerService(id, s.NewProtocol)
 	}
 	log.Lvl3(h.workingAddress, "instantiated all services")
-	return &serviceStore{services, configs}
+
+	// registering messages that services are expecting
+	// TODO
```
Liamsi: Why is here an empty `// TODO` ?

sda/overlay.go
```diff
@@ -30,20 +31,132 @@ type Overlay struct {
...
+		host:               h,
+		trees:              make(map[TreeID]*Tree),
+		entityLists:        make(map[RosterID]*Roster),
+		cache:              NewTreeNodeCache(),
+		instances:          make(map[TokenID]*TreeNodeInstance),
+		instancesInfo:      make(map[TokenID]bool),
+		protocolInstances:  make(map[TokenID]ProtocolInstance),
+		transmitMux:        sync.Mutex{},
+		pendingTreeMarshal: make(map[RosterID][]*TreeMarshal),
+		pendingSDAs:        make([]*ProtocolMsg, 0),
```
Liamsi: Could we think of a slightly more speaking name here and below with `checkPendingSDA`?

sda/processor.go
```diff
@@ -32,6 +125,8 @@ func NewServiceProcessor(c *Context) *ServiceProcessor {
 }
 
 // RegisterMessage puts a new message in the message-handler
+// XXX More comments are needed as it's not clear whether RegisterMessage waits
+// for message for/from Clients or for/from Services.
```
Liamsi: Is this for later?

Liamsi:
Besides some nitpicks: Looks good to me!

sda/processor.go
nikkolasg: Well, since you both points it I will do it, I kinda hoped that @ineiti would do it but he's not there so ... :)

sda/overlay.go
nikkolasg: If you have a better idea, propose. For me it's OK but I don't mind changing for a better name.

Liamsi: Hmm, what does it actually do? Some brainstorming/what do you think about:  `pendingTreeData`, `pendingTree`, `internalMsg`

nikkolasg: It actually checks if there are some messages going to protocols where the overlay did not have the tree before and now it has so it can create a `TreeNodeInstance` + `ProtocolInstance` and pass down the message. I changed it to `checkPendingProtocolMessages` , is that all right for you ?

sda/processor.go
```diff
@@ -124,10 +124,17 @@ func NewServiceProcessor(c *Context) *ServiceProcessor {
 	}
 }
 
-// RegisterMessage puts a new message in the message-handler
-// XXX More comments are needed as it's not clear whether RegisterMessage waits
-// for message for/from Clients or for/from Services.
-func (p *ServiceProcessor) RegisterMessage(f interface{}) error {
+// RegisterHandler will store the given handler that will be used by the service.
+// f must be a function that takes two input:
```
Liamsi: two input*s* or arguments

sda/processor.go
```diff
@@ -124,10 +124,17 @@ func NewServiceProcessor(c *Context) *ServiceProcessor {
...
+//  * network.ServerIdentity: from whom the message is coming from
+//  * Pointer to a struct: message that the service is ready to handle.
+// f must have two return values:
+//  * Pointer to a struct: message that the service has generated as a reply and
+//  that will be sent to the requester (the sender).
+//  * Error in any case there is an error.
+// f can be used to treat internal service messages as well as external requests
```
Liamsi: Reading this comment and rereading the code below: why don't we make f a function instead of an `interface{}` then? The comment wouldn't be necessary and it would be more clear how to use it. I'll try later if it works.

sda/processor.go
```diff
@@ -124,10 +124,17 @@ func NewServiceProcessor(c *Context) *ServiceProcessor {
...
+// from clients.
+func (p *ServiceProcessor) RegisterHandler(f interface{}) error {
 	ft := reflect.TypeOf(f)
 	// Check we have the correct channel-type
 	if ft.Kind() != reflect.Func {
```
Liamsi: Also, below, the error messages aren't up to date anymore. For instance: "Need 2 return values: *network.ProtocolMessage* and error" should be "Need 2 return values: *network.Body* and error", right?

sda/processor.go
Liamsi: Still I don't think that is very clear. The only "thing" which want to keep general, is the 2nd argument of the passed (callback)-function. This could instead look sth like this:
`f func(sId *network.ServerIdentity, structPtr interface{})(network.Body, error)` instead of `f interface{}` and checking all properties of the passed argument during runtime. But this is orthogonal to this issue.

sda/overlay.go
Liamsi: Much better. Thanks.

PR #470, opened 2016-07-12
different small changes

ineiti:
- added `TildeToHome` to `app/lib/config`
- using `pcregrep` in `app/lib/test` to allow multi-line grepping
- correct line-reporting on fatal in `log`
- only show log-output if connected in `network`
- fixed `protocols/manage/propagate` to use `sda.Context` as structure and not interface
- print warning when the same service is registered twice


app/lib/config/config.go
```diff
@@ -237,3 +240,13 @@ func (s *ServerToml) String() string {
 	}
 	return buff.String()
 }
+
+// TildeToHome takes a path and replaces an eventual "~" with the home-directory
+func TildeToHome(path string) string {
```
nikkolasg: Mhhh I don't like doing these things manually but apparently it seems difficult to do otherwise. Look [here](https://travis-ci.org/dedis/cothority/builds/143923343).
Any way, in the link, either the accepted solution or [the one below](http://stackoverflow.com/a/17617721) are simpler.
And why would you want to `ErrFatal` ? It seems to me an usual situation...

app/lib/config/config.go
```diff
@@ -237,3 +240,13 @@ func (s *ServerToml) String() string {
...
+	if strings.HasPrefix(path, "~") {
+		usr, err := user.Current()
+		ui.ErrFatal(err)
```
nikkolasg: Moreover, did we not remove that ui library ? Just looked in my local repo, it seems both are duplicated ?

ineiti: Oups - yes, should be `log`.

app/lib/config/config.go
ineiti: The first link is a travis-run...

The second link is about using `~` in shell-commands, here I want to use it in `os.Open` or others, where I cannot do what is written in the accepted solution.

For the `ErrFatal`, I can return an error, would that be better? Or just ignore the error and replace the `~` with a `.` if the user-directory is not available?

sda/protocol.go
```diff
@@ -59,6 +59,10 @@ func ProtocolRegisterName(name string, protocol NewProtocol) ProtocolID {
 		protocols = make(map[ProtocolID]NewProtocol)
 		protocolNames = make(map[ProtocolID]string)
 	}
+	if _, exists := protocolNames[u]; exists {
+		log.Warn("Protocol", u, "already exists - not overwriting")
```
nikkolasg: If we want that behavior, then please justify it (by a comment or something). It's not obvious why it should be like this. 
Moreover, there are similar questions regarding the `Service`, the `StatusReporter` etc... Everything that is registerable; For those, we'll either do keep a consistent behavior or adjust and justify every independant behavior.

ineiti:
also removed `app/lib/ui` - don't know why it was still there.

sda/protocol.go
ineiti: In test/simulation-mode multiple hosts are handled by a cothority. If a protocol is registered more than once, it is mostly due to an error and will not have the desired effect.

My use-case was registering a bftcosi-protocol in the service, but giving it an individual verification-method for each node. This is not possible when having multiple nodes/hosts in the same cothority-binary, but only confusing the user (in that case, me). So give out a warning so the user knows he's doing something that the system won't support.

app/lib/config/config.go
nikkolasg: > For the ErrFatal, I can return an error, would that be better? Or just ignore the error and replace the ~ with a . if the user-directory is not available?

Sorry right, I'm used to that notation yet (no `if` testing!). Ok then indeed we have to fatal otherwise we can't do much, but maybe we can fatal with a more explicit error: here it will be `user.UnknownUserError` [link](https://golang.org/pkg/os/user/#Current) which is weird I think.

sda/protocol.go
nikkolasg: I also agree with you but this should be justified and specified by a comment, so we know what to expect from this function (invariants and so on) ... :) 
(Thanks)

app/lib/config/config.go
nikkolasg: > The second link is about using ~ in shell-commands, here I want to use it in os.Open or others, where I cannot do what is written in the accepted solution.

The second [solution](http://stackoverflow.com/a/17617721) is applicable here to me it seems .. ?

app/lib/config/config.go
```diff
@@ -237,3 +239,16 @@ func (s *ServerToml) String() string {
 	}
 	return buff.String()
 }
+
+// TildeToHome takes a path and replaces an eventual "~" with the home-directory
+func TildeToHome(path string) string {
+	if strings.HasPrefix(path, "~") {
+		usr, err := user.Current()
+		if err != nil {
+			log.Error("Could't get home-directory")
```
nikkolasg: Ah you changed it back to `Error` ? Hehe. I was more OK with `Fatal` because if you can't get the home directory, you can't do much, and the rest of the program should not continue working on some invalid data ... What do you think ?

app/lib/config/config.go
ineiti: Please try out what you're proposing: `something/~/somethingelse` doesn't use the home-directory.
What is the difference between the solution written there and mine?

app/lib/config/config.go
ineiti: I also prefer `Fatal` - changing it back as it was before.

nikkolasg:
OK for me.

PR #472, opened 2016-07-12
Cyber pass

evanv03:
This is the Cyberpass pull request


app/guard/guard.go
```diff
@@ -0,0 +1,251 @@
+package main
+
```
ineiti: A 3-5 line comment presenting the work is always welcome.

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+// Database is a structure that stores Cothority(the list of guard servers), and a list of all users within the database
+type Database struct {
+	Cothority *sda.Roster
+	Users     []User
+}
+
+var db *Database
+
+var mastersalt []byte
```
ineiti: `mastersalt` is only used in `set`, so initialize it there. It doesn't need to be global.

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+		db, ok = msg.(*Database)
+		if !ok {
+			log.Fatal("We are not OK")
+		}
+		return nil
+	}
+	app.Run(os.Args)
+}
+
+func xor(x []byte, y []byte) []byte {
```
ineiti: helper functions to the bottom, most important functions to the top. So `xor` should be at the bottom.

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+			tomlFileName)
+	}
+	log.Lvl3(el)
+	return el, err
+}
+
+func set(c *cli.Context, UID []byte, epoch []byte, password string, userdata []byte) {
+	k := make([]byte, 32)
+	rand.Read(k)
+	//pwhash is the password hash that will be sent to the guard servers
```
ineiti: Visual error: please use a " " after the `//`.

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+	pwhash := abstract.Sum(network.Suite, []byte(password), mastersalt)
+	//secretkeys is the Shamir Secret share of the keys
+	secretkeys := s.Create(2, len(db.Cothority.List), string(k))
+	//salty is a list of salts that will be used to encrypt the pwhash before sending it to the guard, derived from a single
+	//master salt
+	salty := saltgen(mastersalt, len(db.Cothority.List))
+	responses := make([][]byte, len(db.Cothority.List))
+	keys := make([][]byte, len(db.Cothority.List))
+	for i := 0; i < len(db.Cothority.List); i++ {
```
ineiti: you can use
```
for i, l := range db.Cothority.List{
```
here and then replace `db.Cothority.List[i]` with `l`.

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+	block, _ := aes.NewCipher(k)
+	aesgcm, _ := cipher.NewGCM(block)
+	ciphertext := aesgcm.Seal(nil, mastersalt, userdata, nil)
+	db.Users = append(db.Users, User{UID, mastersalt, keys, ciphertext})
+}
+
+// saltgen is a function that generates all the keys and salts given a length and an initial salt
+func saltgen(salt []byte, count int) [][]byte {
+	salts := make([][]byte, count)
+	tlas := salt
```
ineiti: Why don't you use the salt here? `tlas := salt` doesn't make a copy, if that is your intent:

https://play.golang.org/p/XJ8HsoXSnQ

Perhaps

```
salt = abstract.Sum(network.Suite, salt)
```

is what you need here?

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// setup when you setup the password
+func setup(c *cli.Context) error {
+	db = &Database{}
+	groupToml := c.Args().First()
+	log.Print("Setup is working")
+	var err error
+	t, err := readGroup(groupToml)
+	db.Cothority = t
```
ineiti: you can write directly

```
db = &Database{
    Cothority: t
}
```

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+	log.Print("Setup is working")
+	err = ioutil.WriteFile("config.bin", b, 0660)
+	log.ErrFatal(err)
+	return nil
+}
+
+// getuser returns the user that the UID matches
+func getuser(UID []byte) *User {
+
+	for i := 0; i < len(db.Users); i++ {
```
ineiti: `for _, u := range db.Users`

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+	}
+	//this is necessary because there needs to be a return at the end but
+	return nil
+}
+
+// getpass contacts the guard servers, then gets the passwords and reconstructs the secret keys
+func getpass(c *cli.Context, UID []byte, epoch []byte, pass string) {
+	if getuser(UID) == nil {
+		log.ErrFatal(nil, "Wrong Username")
```
ineiti: `log.Fatal("Wrong username")`

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	pwhash := abstract.Sum(network.Suite, []byte(pass), getuser(UID).Salt)
+	salty := saltgen(getuser(UID).Salt, len(db.Cothority.List))
+	responses := make([]string, len(db.Cothority.List))
+	keys := make([]string, len(db.Cothority.List))
+	for i := 0; i < len(db.Cothority.List); i++ {
```
ineiti: `for i, si := range db.Cothority.List`

app/guard/guard.go
```diff
@@ -0,0 +1,256 @@
...
+		cl := guard.NewClient()
+		sendy := abstract.Sum(network.Suite, pwhash, salty[i])
+		rep, err := cl.GetGuard(db.Cothority.List[i], UID, epoch, sendy)
+		log.ErrFatal(err)
+		responses[i] = string(rep.Msg)
+		keys[i] = string(xor(getuser(UID).Keys[i], rep.Msg))
+	}
+	k := s.Combine(keys)
+	if len([]byte(k)) == 0 {
```
ineiti: `if len(k) == 0{`

services/guard/guard.go
```diff
@@ -0,0 +1,93 @@
...
+	Epoch []byte
+	Msg   []byte
+}
+
+// Response is what the Guard service will reply to clients.
+type Response struct {
+	Msg []byte
+}
+
+//[]byte("gmPTvzEuu88GAL_67UJFFZXzaAhWn_N3ZonPU-D272Y=SuNSfNfzkP3CSSNIR3uuERNxDWzN7MwebioZGkZBwdc=")
```
ineiti: Is this still used?

services/guard/guard.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Request treats external request to this service.
+func (st *Guard) Request(e *network.ServerIdentity, req *Request) (network.Body, error) {
+	//hashy computes the hash that should be sent back to the main server H(pwhash, x, UID, Epoch)
+	hashy := abstract.Sum(network.Suite, req.Msg, st.z, req.UID, req.Epoch)
+	key := []byte("example key 1234")
```
ineiti: You'll have to explain once again why you need this key here. I don't see it used anywhere.

services/guard/guard_test.go
```diff
@@ -0,0 +1,46 @@
...
+	local := sda.NewLocalTest()
+	// generate 5 hosts, they don't connect, they process messages, and they
+	// don't register the tree or entitylist
+	_, el, _ := local.GenTree(5, false, true, false)
+	defer local.CloseAll()
+
+	// Send a request to the service
+	client := NewClient()
+	log.Lvl1("Sending request to service...")
+	UID := []byte{'U', 'S', 'E', 'R'}
```
ineiti: `uid := []byte("USER")`

services/guard/guard_test.go
```diff
@@ -0,0 +1,46 @@
...
+	Epoch := []byte{'E', 'P', 'O', 'C', 'H'}
+	msg := []byte{'H', 'e', 'l', 'l', 'o'}
+	Hzi, _ := client.GetGuard(el.List[0], UID, Epoch, msg)
+	//lol1 := make([]string, len(Hzi))
```
ineiti: delete unused code.

app/guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+		rep, err := cl.GetGuard(si, UID, epoch, sendy)
+		log.ErrFatal(err)
+		responses[i] = rep.Msg
+		block, err := aes.NewCipher(responses[i])
+		if err != nil {
+			panic(err)
+		}
+		log.Print("Block Size: ", block.BlockSize())
+		// This part is necessary because you need a key to convert the Hash to a stream. The key is not even important because
+		//all that is required is that the request gives the same output for the same input
```
ineiti: Please put a space between the `//` and the text. See http://lkml.iu.edu/hypermail/linux/kernel/1607.1/00627.html

app/guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+		stream := cipher.NewCTR(block, iv)
+		msg := make([]byte, 88)
+		//this does not work, as it merely appends zeros at the end, this is still necessary to do an xor during the clients
+		//password setting and authentication services
+		log.Print("SecretKey: ", []byte(secretkeys[i]))
```
ineiti: `log.Print` all over the place should be `log.Lvl2` or even `log.Lvl3`.

app/guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+		stream.XORKeyStream(msg, []byte(secretkeys[i]))
+		log.Print("set xorkey after: ", msg)
+		keys[i] = msg
+
+	}
+	// This is the code that seals the user data using the master key and saves it to the db
+	block, _ := aes.NewCipher(k)
+	aesgcm, _ := cipher.NewGCM(block)
+	ciphertext := aesgcm.Seal(nil, mastersalt, userdata, nil)
+	db.Users = append(db.Users, User{UID, mastersalt, keys, ciphertext, iv})
```
ineiti: I would create the user-struct at the beginning of the method and then fill it up as values come in. This would better reflect which variables will be stored, as instead of using `iv` in the code, it would be `user.IV` which would directly show that the IV comes from the user. What do @nikkolasg and @Liamsi think?

app/guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+		responses[i] = rep.Msg
+		block, err := aes.NewCipher(responses[i])
+		if err != nil {
+			panic(err)
+		}
+		log.Print("Block Size: ", block.BlockSize())
+		// This part is necessary because you need a key to convert the Hash to a stream. The key is not even important because
+		//all that is required is that the request gives the same output for the same input
+		stream := cipher.NewCTR(block, iv)
+		msg := make([]byte, 88)
```
ineiti: Are we sure the message is always 88 bytes? Or should this be `len(secretkeys[i]`?

Did you try it out with more than 2 guard-servers?

app/guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+	block, _ := aes.NewCipher(k)
+	aesgcm, _ := cipher.NewGCM(block)
+	ciphertext := aesgcm.Seal(nil, mastersalt, userdata, nil)
+	db.Users = append(db.Users, User{UID, mastersalt, keys, ciphertext, iv})
+}
+
+// saltgen is a function that generates all the keys and salts given a length and an initial salt
+func saltgen(salt []byte, count int) [][]byte {
+	salts := make([][]byte, count)
+	for i := 0; i < count; i++ {
```
ineiti: `for i := range salts`

app/guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+	keys := make([]string, len(db.Cothority.List))
+	for i, si := range db.Cothority.List {
+		cl := guard.NewClient()
+		sendy := abstract.Sum(network.Suite, pwhash, salty[i])
+		rep, err := cl.GetGuard(si, UID, epoch, sendy)
+		log.ErrFatal(err)
+		block, err := aes.NewCipher(rep.Msg)
+		stream := cipher.NewCTR(block, getuser(UID).Iv)
+		msg := make([]byte, 88)
+		//this does not work, as it merely appends zeros at the end, this is still necessary to do an xor during the clients
```
ineiti: Is it now working or not?

app/guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+	log.ErrFatal(err)
+	aesgcm, err := cipher.NewGCM(block)
+	log.ErrFatal(err)
+	plaintext, err := aesgcm.Open(nil, getuser(UID).Salt, getuser(UID).Data, nil)
+	log.ErrFatal(err)
+	log.Print(string(plaintext))
+}
+func setpass(c *cli.Context) error {
+	UID := []byte(c.Args().Get(0))
+	Epoch := []byte{'E', 'P', 'O', 'C', 'H'}
```
ineiti: `Epoch := []byte("EPOCH")`

app/guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+	block, err := aes.NewCipher([]byte(k))
+	log.ErrFatal(err)
+	aesgcm, err := cipher.NewGCM(block)
+	log.ErrFatal(err)
+	plaintext, err := aesgcm.Open(nil, getuser(UID).Salt, getuser(UID).Data, nil)
+	log.ErrFatal(err)
+	log.Print(string(plaintext))
+}
+func setpass(c *cli.Context) error {
+	UID := []byte(c.Args().Get(0))
```
ineiti: all variables inside a method should be lower-case.

services/guard/api.go
```diff
@@ -0,0 +1,35 @@
...
+// NewClient makes a new Client
+func NewClient() *Client {
+	return &Client{Client: sda.NewClient(ServiceName)}
+}
+
+// GetGuard is the function that sends a request to the guard server and creates a client to receive the responses
+func (c *Client) GetGuard(dst *network.ServerIdentity, UID []byte, epoch []byte, t []byte) (*Response, error) {
+	//send request an entity in the network
+	log.Lvl4("Sending Request to ", dst)
+	ServiceReq := &Request{UID, epoch, []byte(t)}
```
ineiti: lowercase variable

services/guard/guard.go
```diff
@@ -0,0 +1,76 @@
...
+	network.RegisterMessageType(&Response{})
+}
+
+//This is the area where Z is generated for a server, it creates z, which is a bytestring of length n for each guard.
+
+// Guard is a structure that stores the guards secret key, z, to be used later in the process of hashing the clients requests
+type Guard struct {
+	*sda.ServiceProcessor
+	path string
+	z    []byte
```
ineiti: Should probably be capital so the structure can be stored later (not yet implemented).

services/guard/guard_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/dedis/cothority/log"
+	"github.com/dedis/cothority/sda"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestServiceGuard(t *testing.T) {
+	defer log.AfterTest(t)
+	log.TestOutput(testing.Verbose(), 4)
+	local := sda.NewLocalTest()
+	// generate 5 hosts, they don't connect, they process messages, and they
```
ineiti: Whole sentence: start with a capital letter and end with a ".".

services/guard/guard_test.go
```diff
@@ -0,0 +1,36 @@
...
+	// Send a request to the service
+	client := NewClient()
+	log.Lvl1("Sending request to service...")
+	UID := []byte("USER")
+	Epoch := []byte("EPOCH")
+	msg := []byte("Hello")
+	Hzi, _ := client.GetGuard(el.List[0], UID, Epoch, msg)
+
+	Hz2, _ := client.GetGuard(el.List[0], UID, Epoch, msg)
+	fmt.Println(Hzi)
```
ineiti: don't use `fmt.Print` - here a `log.Lvl3` is appropriate.

app/guard/guard.go
evanv03: the message is always a multiple of 88 bytes, though i think scales with the secret and not the number of shares

app/guard/guard.go
```diff
@@ -0,0 +1,343 @@
...
+		sendy := blankpoint.Mul(Gu, blankscalar.Mul(pwbytes, blindbytes))
+		rep, err := cl.GetGuard(si, uid, epoch, sendy)
+		log.ErrFatal(err)
+		log.Lvl2("SetRep: ", rep.Msg)
+		reply := blankpoint.Mul(rep.Msg, blankscalar.Inv(blindbytes))
+		responses[i], err = reply.MarshalBinary()
+		log.ErrFatal(err)
+
+		block, err := aes.NewCipher(responses[i])
+		if err != nil {
```
ineiti: `log.ErrFatal`

app/guard/guard.go
```diff
@@ -0,0 +1,343 @@
...
+	block, _ := aes.NewCipher(k)
+	aesgcm, _ := cipher.NewGCM(block)
+	ciphertext := aesgcm.Seal(nil, mastersalt, userdata, nil)
+	db.Users = append(db.Users, User{uid, mastersalt, keys, ciphertext, iv})
+}
+
+// saltgen is a function that generates all the keys and salts given a length and an initial salt.
+func saltgen(salt []byte, count int) [][]byte {
+	salts := make([][]byte, count)
+	for i := 0; i < count; i++ {
```
ineiti: `for i := range salts{`

app/guard/guard.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// setup is called when you setup the password database.
+func setup(c *cli.Context) error {
+
+	groupToml := c.Args().First()
+	log.Lvl2("Setup is working")
+	var err error
+	t, err := readGroup(groupToml)
+	db = &Database{
```
ineiti: `db = &Database{t}`

app/guard/guard.go
```diff
@@ -0,0 +1,343 @@
...
+	usrdata := []byte(c.Args().Get(2))
+	set(c, UID, Epoch, string(Pass), usrdata)
+	b, err := network.MarshalRegisteredType(db)
+	log.ErrFatal(err)
+	err = ioutil.WriteFile("config.bin", b, 0660)
+	return nil
+}
+func get(c *cli.Context) error {
+	uid := []byte(c.Args().Get(0))
+	epoch := []byte{'E', 'P', 'O', 'C', 'H'}
```
ineiti: `epoch := []byte{'EPOCH'}`

Liamsi:
There seem to be compile errors: https://travis-ci.org/dedis/cothority/builds/150617004#L167-L169
Do you also have them locally, too? /cc @evanv03

app/guard/guard.go
Liamsi: In case you struggle with the above proposal: I thinl @ineiti meant `[]byte("EPOCH")` instead.

app/guard/guard.go
```diff
@@ -0,0 +1,315 @@
...
+		if os.IsNotExist(err) {
+			return nil
+		}
+		log.ErrFatal(err, "The config.bin file threw an error")
+		_, msg, err := network.UnmarshalRegistered(b)
+		log.ErrFatal(err, "UnmarshalRegistered messeed up")
+		var ok bool
+		db, ok = msg.(*Database)
+		if !ok {
+			log.Fatal("We are not OK")
```
Liamsi: Can we give the user a hint *what* is not OK?

app/guard/guard.go
```diff
@@ -0,0 +1,315 @@
...
+		return nil, errors.New("Empty or invalid group file:" +
+			tomlFileName)
+	}
+	log.Lvl3(el)
+	return el, err
+}
+
+func set(c *cli.Context, uid []byte, epoch []byte, password string, userdata []byte) {
+	mastersalt := make([]byte, 12)
+	rand.Read(mastersalt)
```
Liamsi: In very rare and ugly cases (almost impossible) rand.Read can return an error (not enough entropy). You would definitely want to catch this cases (and `panic` to be safe) .

Liamsi: also for all other uses of rand.Read

services/guard/guard_test.go
```diff
@@ -0,0 +1,33 @@
...
+	client := NewClient()
+	log.Lvl1("Sending request to service...")
+	UID := []byte("USER")
+	Epoch := []byte("EPOCH")
+	msg := network.Suite.Point()
+
+	Hzi, _ := client.GetGuard(el.List[0], UID, Epoch, msg)
+
+	Hz2, _ := client.GetGuard(el.List[0], UID, Epoch, msg)
+	assert.Equal(t, Hzi, Hz2)
```
Liamsi: What does this test actually test? It looks like we want to test that calling `client.GetGuard(el.List[0], UID, Epoch, msg)` twice returns twice the same result?

ineiti:
OK for me...

services/guard/api.go
```diff
@@ -0,0 +1,36 @@
...
+	*sda.Client
+}
+
+// NewClient makes a new Client
+func NewClient() *Client {
+	return &Client{Client: sda.NewClient(ServiceName)}
+}
+
+// GetGuard is the function that sends a request to the guard server and creates a client to receive the responses
+func (c *Client) GetGuard(dst *network.ServerIdentity, UID []byte, epoch []byte, t abstract.Point) (*Response, error) {
```
Liamsi: rename to GetGuard -> SendToGuard

services/guard/api.go
```diff
@@ -0,0 +1,36 @@
...
+type Client struct {
+	*sda.Client
+}
+
+// NewClient makes a new Client
+func NewClient() *Client {
+	return &Client{Client: sda.NewClient(ServiceName)}
+}
+
+// GetGuard is the function that sends a request to the guard server and creates a client to receive the responses
```
Liamsi: Update comment to reflect what is really happening here. (client sends message ...)

services/guard/guard_test.go
```diff
@@ -0,0 +1,33 @@
...
+	_, el, _ := local.GenTree(5, false, true, false)
+	defer local.CloseAll()
+
+	// Send a request to the service
+	client := NewClient()
+	log.Lvl1("Sending request to service...")
+	UID := []byte("USER")
+	Epoch := []byte("EPOCH")
+	msg := network.Suite.Point()
+
```
Liamsi: Add a comment (sth along the lines of) "Send message twice to see key didn't change in the same epoch" (or similar)

Liamsi:
LGTM besides a few minor comments

PR #480, opened 2016-07-14
Trying out stuff for the jvss race / deadlock condition...

nikkolasg:


Liamsi:
As far as I can tell from travis' output: Looks like you are not running into a race condition bur outputting too much (> 4 Mb). If you make the test less verbose and it succeeds, we are fine :-)

Liamsi:
To me it already looks like you nailed it! (at least before your changes the race-condition occurred during the first or second iteration)

nikkolasg:
I don't get it,  from the output, it still seems to block no ?

Liamsi:
> I don't get it, from the output, it still seems to block no ?

From what I saw yesterday, travis just stops after 4Mb. Did I miss some output stating that it was blocking?

PR #481, opened 2016-07-14
Identity follower

ineiti:
Implemented the follower-part of `cisc`


services/identity/service.go
```diff
@@ -39,14 +43,14 @@ func init() {
 // Service handles identities
 type Service struct {
 	*sda.ServiceProcessor
-	identities      map[string]*storage
+	Identities      map[string]*Storage
 	identitiesMutex sync.Mutex
```
nikkolasg: `Identities` is public but you have an `identitiesMutex` private. I assume here the mutex is for controlling the access to `Identities` ? If `Identities` is public, then any user can access it without the mutex. It would make more sense to either enforce the use of the mutex by getter/setter or just drop the mutex if it's not necessary .

services/identity/service.go
```diff
@@ -281,33 +292,61 @@ func (s *Service) Propagate(msg network.Body) {
...
+func (s *Service) setIdentityStorage(id ID, is *Storage) {
 	s.identitiesMutex.Lock()
 	defer s.identitiesMutex.Unlock()
 	log.Lvlf3("%s %x %v", s.Context.ServerIdentity(), id[0:8], is.Latest.Device)
-	s.identities[string(id)] = is
+	s.Identities[string(id)] = is
+}
+
+// saves the actual identity
+func (s *Service) save() {
```
nikkolasg: It should maybe return an error given that there are two cases when things could go wrong, and the caller can't know that.

services/identity/service.go
```diff
@@ -281,33 +292,61 @@ func (s *Service) Propagate(msg network.Body) {
...
+	log.Lvl3("Saving service")
+	b, err := network.MarshalRegisteredType(s)
+	if err != nil {
+		log.Error("Couldn't marshal service:", err)
+	} else {
+		err = ioutil.WriteFile(s.path+"/identity.bin", b, 0660)
```
nikkolasg: and then we can indent that block to the top block level in this method.

services/identity/service.go
```diff
@@ -281,33 +292,61 @@ func (s *Service) Propagate(msg network.Body) {
...
 	s := &Service{
 		ServiceProcessor: sda.NewServiceProcessor(c),
-		identities:       make(map[string]*storage),
+		Identities:       make(map[string]*Storage),
 		skipchain:        skipchain.NewClient(),
 		path:             path,
 	}
+	b, err := ioutil.ReadFile(path + "/identity.bin")
+	if err != nil && !os.IsNotExist(err) {
+		log.Error("Error while reading", path+"/identity.bin:", err)
```
nikkolasg: if this error occurs, is it safe to assume that the caller can use the returned `sda.Service` ? If not then this method probably should return an error (or panic with the appropriate comment).

services/identity/service.go
```diff
@@ -281,33 +292,61 @@ func (s *Service) Propagate(msg network.Body) {
...
+	} else {
+		if len(b) > 0 {
+			_, msg, err := network.UnmarshalRegistered(b)
+			if err != nil {
+				log.Error("Couldn't unmarshal:", err)
```
nikkolasg: same as above

services/skipchain/skipchain.go
```diff
@@ -456,6 +461,20 @@ func (s *Service) lenSkipBlocks() int {
 	return len(s.SkipBlocks)
 }
 
+// saves the actual identity
+func (s *Service) save() {
```
nikkolasg: Seems this function is exactly the same as in `services/identity/service.go`. Only the file name is changing. Any idea on how it could be reduced to one codebase used twice ? I'm not sure yet since I'm not really too familiar with the identity & skipchains code base

services/identity/service.go
ineiti: the field is public so I can do `Marshal` `Unmarshal` on it. The mutex is to access it from the service.

services/identity/service.go
ineiti: Good question - I don't really know what we should do if the `save` fails. As it is any error is printed but ignored. What would be a good reaction on a failing error?

services/identity/service.go
ineiti: you mean if we decide to return an error? I'm not sure if this is necessary.

services/identity/service.go
ineiti: The `Service` is initialized anyway. Here it is either overwritten with the loaded data, or returned as-is. So if any part of the loading fails (no file, wrong file, ...), it just returns a perfectly usable `Service`.

services/identity/service.go
ineiti: ditto ;)

services/skipchain/skipchain.go
ineiti: It comes back to the discussion we had about what should be done for the services - shall we offer a `save`-method in the `Context`? We'd have two use-cases already!

app/cisc/cisc.go
```diff
@@ -76,9 +80,10 @@ func idCreate(c *cli.Context) error {
 	}
 	log.Info("Creating new blockchain-identity for", name)
 
-	cfg := identity.NewIdentity(group.Roster, 2, name)
+	cfg := &ciscConfig{Identity: identity.NewIdentity(group.Roster, 2, name)}
```
Liamsi: make the `2` a constant with a speaking name

app/cisc/cisc.go
```diff
@@ -97,16 +102,27 @@ func idConnect(c *cli.Context) error {
 	idBytes, err := hex.DecodeString(c.Args().Get(1))
 	log.ErrFatal(err)
 	id := identity.ID(idBytes)
-	cfg := identity.NewIdentity(group.Roster, 2, name)
+	cfg := &ciscConfig{Identity: identity.NewIdentity(group.Roster, 2, name)}
```
Liamsi: same as above

services/identity/service.go
nikkolasg: So you are sending a `Service` through the network ? It seems too late to change any design here but it seems strange... :s
And the mutex is needed because multiple client can ask the service the identities it holds at the same time ?

ineiti: No, that's for loading/saving, and as it's a `map`, it was more easy this way.

ineiti: If I try to marshal the map, this is what I get:

```
Error for protobuf encoding: reflect: NumField of non-struct type
Couldn't marshal service: reflect: NumField of non-struct type
```

app/cisc/lib.go
```diff
@@ -11,50 +11,140 @@ import (
...
+	for k := range cfg.Config.Data {
+		log.Info("Key set", k)
+	}
+}
+
 // Returns the config-file from the configuration
 func getConfig(c *cli.Context) string {
 	configDir := config.TildeToHome(c.GlobalString("config"))
-	os.Mkdir(configDir, 0660)
+	os.Mkdir(configDir, 0770)
```
Liamsi: Can't that fail?

app/cisc/lib.go
```diff
@@ -75,6 +165,7 @@ func getGroup(c *cli.Context) *config.Group {
 // retrieves ssh-config-name and ssh-directory
 func sshDirConfig(c *cli.Context) (string, string) {
 	sshDir := config.TildeToHome(c.GlobalString("cs"))
+	os.Mkdir(sshDir, 0700)
```
Liamsi: same as above?

services/identity/service.go
Liamsi: Isn't this a fatal error?

nikkolasg: It seems to me that decision should be left to the user. Then in the place you use `save()`, because you are the user for the moment you can panic or whatever...

services/identity/service.go
nikkolasg: Yeah but when creating a `Service` with a given data and that fails, that means the returned `Service` is not initialized as expected. Again, that's the same question: should we print & continue or return an error ...
I'm not sure either but it's a question we need to solve .!

ineiti: `s` is initialized at the beginning of the method, and only updated if there are no errors. So either we have the `s` initialized from the beginning of the method and return that, or we update it with the data read from the file.

Does that make sense?

services/skipchain/skipchain.go
nikkolasg: Something like (pseudo code) ...?
```
func (c *Context) Save(file string, i interface{}) error {
   data := network.MarshalRegisteredType(i)
   return fileIO.Write(data)
}
```
You're right, there are already two uses case so it may be time :)

ineiti: Yes, and while we're at it, we could also 
- add a `Load`-method
- add a `Path`-method
- remove the path from the `sda.NewServiceFunc`

should I open an issue or include that in here?

nikkolasg: The path method is for setting or retrieve the path or both ?
Definitely an issue to do later, so we can let these functions as is for the moment

ineiti: Hah - I wrote `GetPath` first, then I thought that you don't like the `Get`, but it might be clearer ;) 
It's a getter, the service should not be able to set the path!

services/identity/service.go
ineiti: Hmm, this is in the cothority-node, and the error is more pointing to a problem of the cothority than the message received through the API. The cothority-node can continue even without saving.

services/skipchain/skipchain.go
nikkolasg: It's clear if we see the  *signature* : `Path() string` :)

ineiti: See #491

ineiti:
One more commit to remove the data-race in `serviceFactory`:

https://travis-ci.org/dedis/cothority/builds/146393746#L203

services/identity/service.go
nikkolasg: There's ways to make it work right (having a struct that contains both the identities and the lock and saving only that struct), but you think it's not the most urgent think, then fine, I won't push for it ;)

services/identity/service.go
nikkolasg: Ok right, that's internal so the client don't have to know something failed.

services/identity/service.go
Liamsi: I agree with everything @nikkolasg said here.

Liamsi:
There is a deadlock -> back to WIP

services/identity/service.go
nikkolasg: Yep but include that as a comment at least for these kind of subtleties.

services/identity/service.go
ineiti: Well I use `Service` as the struct you're describing. So you would prefer having yet another struct and put that in here?

Liamsi:
OK, I see the deadlock doesn't directly seem related to this PR. Did you open an issue for that or check if there is already one related to this?

services/identity/service.go
ineiti: I put it in its own method. I hope it's more clear like that.

ineiti:
I saw that deadlock this morning and corrected it in branch `hpets16`. I ported it back to here.

ineiti:
superseeded by #511

PR #483, opened 2016-07-15
Refactor processor travis

ineiti:
to check why and where this fails.


nikkolasg:
It would make more sense to do the PR against `refactor_processor`, you're pulling all my changes also, which we don't want to comment here.

ineiti:
yes of course - this is WIP for the moment. Just hitting on Travis.

ineiti:
Other tests are in progress to find this error. #496, #507, #499

PR #486, opened 2016-07-20
Hpets16

ineiti:
Added docker-generation and the scripts needed to reproduce the hotpets16-demo


ineiti:
re-opening against identity_follower

Issue #493, opened 2016-07-23
Unify TestMain in _test.go-files


ineiti:
The correct way to use the logging-library is to have a

```
func TestMain(m *testing.M){
    log.MainTest(m)
}
```

which does all the correct settings according to the verbose-flag and checking whether any go-routines are still active.

PR #494, opened 2016-07-23
Unify testmain

ineiti:
Implemented #493. Also adds

- mutex fix in `sda/services.go` (because of https://travis-ci.org/dedis/cothority/builds/146393746#L203)
- fix of `TestStatsString` in `monitor/stats.go`


sda/service.go
```diff
@@ -72,6 +72,7 @@ type serviceFactory struct {
 	translations map[string]ServiceID
 	// Inverse mapping of ServiceId => string
 	inverseTr map[ServiceID]string
+	mutex     sync.Mutex
```
nikkolasg: Can you just remind me again why the deadlock / data race is happening ? Because normally, you register `Service` in every `init()` methods so I don't see how there could be a data race.  Is it because of the tests ?

nikkolasg:
Except for the mutex thing where I still don't understand why we need it, I'm OK

sda/service.go
ineiti: Hah - found it:
https://travis-ci.org/dedis/cothority/builds/146393746#L203
There `go` complains about a race-condition on the `serviceFactory`-structure, so I put in a mutex (and created a deadlock...)

nikkolasg: I guess it's because of the tests then. Ok well, it does not make a huge difference and it's still a security bonus so let's go for it :)

Issue #495, opened 2016-07-23
Make `sda/local` search for open port-numbers


ineiti:
When running things in parallel, applying fixed IP-numbers doesn't work. Let `sda/local` search for open port-numbers.
Also change tests that use a fixed port-number, so that the tests can be run in parallel.

Liamsi:
That's what #170 (which is part of #378). Even if the approach is a different one (mock the connections vs. search for open ports and use real connections) it tries to solve the same problem.

PR #496, opened 2016-07-23
Local search ports

ineiti:
This implements #495 - which also allowed to have a better stress-test for the networking-library:

in `network/net_test.go` there is a `TestStress` which calls repeatedly `Listen` and `Open` on a `SecureTCPHost` until it fails.

Some changes were necessary:
- `SecureTCPHost.Listen` is not blocking anymore! which makes other tests fail :(
- `TCPHost.Listen` sends the actual listening port in a channel if the port-number was `0`

One possible bug already found is in `SecureTCPConn.exchangeServerIdentity where the 

`nm, err = sc.TCPConn.Receive(context.TODO())`

can fail because the connection has been made but the other end didn't send anything yet.

By changing the `nbrHosts` in `network/net_test.go::stressTest` one can make it fail sooner or later.

With all that written, I have no idea for the moment if that is a but in `cothority/network` or if we just swamp the network too badly and it _has_ to fail.


ineiti:
Before merging this needs to pass once Travis is into high usage:

https://www.traviscistatus.com/

ineiti:
I think this is now _really_ ready for merge. BFTCoSi fixes come in #511

Makefile
```diff
@@ -39,16 +39,16 @@ test_lint:
 	}
 
 test_multi:
+	cd network; \
 	for a in $$( seq 10 ); do \
-	  cd services/identity; \
-	  go test -v -race -p=1 -short ./...; \
+	  go test -v -race -run Stress; \
```
Liamsi: Is test `test_multi` only for running stress-tests?

ineiti: Yes, it's a placeholder for any tests you might want to run for a couple of times in Travis. I'll add a comment.

network/net.go
```diff
@@ -35,10 +40,11 @@ import (
 // If constructors == nil, it will take an empty one.
 func NewTCPHost() *TCPHost {
 	return &TCPHost{
-		peers:        make(map[string]Conn),
-		quit:         make(chan bool),
-		constructors: DefaultConstructors(Suite),
-		quitListener: make(chan bool),
+		ListeningPort: make(chan int, 1),
```
Liamsi: Does this new field need to be public? (And is it only used in tests?)

ineiti: made private as it's only used in the tests.

network/net.go
```diff
@@ -238,25 +259,37 @@ func (st *SecureTCPHost) Listen(fn func(SecureConn)) error {
...
-		if err = st.TCPHost.listen(addr, receiver); err != nil {
+		go func() {
+			err = st.TCPHost.listen(addr, receiver)
 			// The listening is over
-			if err == ErrClosed || err == ErrEOF {
-				return nil
+			if err == nil || err == ErrClosed || err == ErrEOF {
+				return
+			}
+			st.TCPHost.ListeningPort <- -1
```
Liamsi: Can't this and the code below lead to `workingAddress` set to `-1`?

ineiti: No, because it checks if `port > 0`.

protocols/bftcosi/bftcosi_test.go
```diff
@@ -15,8 +15,8 @@ var veriCount int
 var countMut sync.Mutex
 
 func TestMain(m *testing.M) {
-	log.Info("skipping test because of https://github.com/dedis/cothority/issues/467")
-	//log.MainTest(m)
+	log.Info("Skipping for https://github.com/dedis/cothority/issues/467")
```
Liamsi: Either skip it, or delete comment.

sda/host.go
```diff
@@ -92,14 +93,18 @@ func (h *Host) listen(wait bool) {
 		h.registerConnection(c)
 		h.handleConn(c)
 	}
-	go func() {
```
Liamsi: I didn't test it, but do you have any idea, why this was in a go-routine prior your changes? I guess `NewHost` isn't blocking now, right?

sda/service_test.go
```diff
@@ -287,7 +287,11 @@ func TestServiceNewProtocol(t *testing.T) {
 		link: make(chan bool),
 	}
 	var count int
+	countMutex := sync.Mutex{}
 	RegisterNewService("DummyService", func(c *Context, path string) Service {
+		countMutex.Lock()
+		log.Lvl2("Creating service", count)
+		defer countMutex.Unlock()
```
Liamsi: put this line directly under the `countMutex.Lock()` one

Liamsi:
Some general remarks/comments:

> in network/net_test.go there is a TestStress which calls repeatedly Listen and Open on a SecureTCPHost until it fails.

As we discussed it would be nicer to have tests which are more like unit tests and really test each aspect of each single function. For instance the other `Receive` bug you found should have been covered by one test which writes/reads more to the underlying (potentially mocked) connection  (thank god it's an interface ;-) etc.
That said, these kind of stress test is still nice to have (if we fail to test our code appropriately, we can still simulate some form of realistic behaviour) 👍 

> TCPHost.Listen sends the actual listening port in a channel if the port-number was 0

That's really cool! What stops us from simplifying the code and always write the actual listening port to this channel instead of only if called with port 0 (maybe I'll find the answer while having a more detailed look myself)?

sda/host.go
ineiti: `Listen` is non-blocking now, as we need to know when the port-number is available.

network/net.go
Liamsi: Thanks!

protocols/bftcosi/bftcosi_test.go
Liamsi: I see, but `log.MainTest(m)` wasn't commented anymore.

ineiti:
The channel is only used internally, so you can't access it anyway. The port-number is always available in the `workingAddress`, whether you indicated a port-number or a ":0".

I'll add a test for the receive in #511 - btw: the interface of `network` doesn't help here, as I'll have to interfere with the golang-connection...

protocols/bftcosi/bftcosi_test.go
ineiti: I know - my bad. Corrected in latest commit.

Liamsi:
> I'll add a test for the receive in #511 - btw: the interface of `network` doesn't help here, as I'll have to interfere with the golang-connection...

Which is the interface I meant (not ours).

network/net.go
```diff
@@ -430,7 +471,9 @@ func (c *TCPConn) Send(ctx context.Context, obj Body) error {
 		log.Lvl4("Sending from", c.conn.LocalAddr(), "to", c.conn.RemoteAddr())
 		n, err := c.conn.Write(b[:length])
 		if err != nil {
-			log.Error("Couldn't write chunk starting at", sent, "size", length, err)
+			log.Error("Couldn't write chunk starting at", sent, n, "size", length, err)
+			log.Error(log.Stack())
+			//time.Sleep(time.Millisecond * 100)
```
Liamsi: Can you please remove this line (commented)?

network/net.go
```diff
@@ -505,29 +549,47 @@ func (sc *SecureTCPConn) ServerIdentity() *ServerIdentity {
...
+	// Try for 1 second to receive the other entity
+	var nm Packet
+	var err error
+	for i := 0; i < 10; i++ {
+		// Receive the other ServerIdentity
+		nm, err = sc.TCPConn.Receive(context.TODO())
+		switch {
+		case err == nil:
+			if i > 0 {
+				log.Warn(sc, "Got a packet after a failure")
```
Liamsi: What does that mean? We received something after retrying?

network/net.go
```diff
@@ -505,29 +549,47 @@ func (sc *SecureTCPConn) ServerIdentity() *ServerIdentity {
...
 		sc.TCPConn.conn.RemoteAddr().String())
 	if err := sc.TCPConn.Send(context.TODO(), ourEnt); err != nil {
-		return fmt.Errorf("Error while sending indentity during negotiation:%s", err)
+		return fmt.Errorf("Error while sending indentity during negotiation: %s", err)
+	}
+
+	// Try for 1 second to receive the other entity
+	var nm Packet
+	var err error
+	for i := 0; i < 10; i++ {
```
Liamsi: Why `10`?

network/net.go
```diff
@@ -505,29 +549,47 @@ func (sc *SecureTCPConn) ServerIdentity() *ServerIdentity {
...
+		// Receive the other ServerIdentity
+		nm, err = sc.TCPConn.Receive(context.TODO())
+		switch {
+		case err == nil:
+			if i > 0 {
+				log.Warn(sc, "Got a packet after a failure")
+			}
+			i = 10
```
Liamsi: Is this some kind of `break` for the loop?

network/net.go
```diff
@@ -505,29 +549,47 @@ func (sc *SecureTCPConn) ServerIdentity() *ServerIdentity {
...
 	// Send our ServerIdentity to the remote endpoint
-	log.Lvl4("Sending our identity", ourEnt.ID, "to",
+	log.Lvlf4("Sending our identity %x to %s", ourEnt.ID,
 		sc.TCPConn.conn.RemoteAddr().String())
 	if err := sc.TCPConn.Send(context.TODO(), ourEnt); err != nil {
-		return fmt.Errorf("Error while sending indentity during negotiation:%s", err)
+		return fmt.Errorf("Error while sending indentity during negotiation: %s", err)
+	}
+
+	// Try for 1 second to receive the other entity
```
Liamsi: Can we make this similar to https://github.com/dedis/cothority/pull/496/files#diff-0083b6929e4a66649c44253981e90af9L166 where we use some consts `WaitRetry`/`MaxRetry`

network/net.go
Liamsi: If yes, use break instead modifying the for-loop's index inside the loop.

protocols/manage/close_all_test.go
```diff
@@ -2,35 +2,19 @@ package manage_test
...
-	select {
-	case children := <-protocol.ChildCount:
-		log.Lvl2("Instance 1 is done")
-		if children != nbrNodes {
-			t.Fatal("Didn't get a child-cound of", nbrNodes)
-		}
-	case <-time.After(timeout):
-		t.Fatal("Didn't finish in time")
-	}
+	pi.Start()
```
Liamsi: What does this test, test now? Is there a way it can fail? It looks like it just calls `pi.Start()` and if this returns, everything is fine, else we will only see that after 10 minutes (or whatever is the limit in golang).

network/net.go
ineiti: Yes, I had this case from time to time, but not lately. So I wanted to check if it still happens. I can also just take it away. 

The handling here happens if the connection is opened and the listener tries to receive bytes before the identity has been sent. Which seems to be plausible to me, but I didn't see it anymore (I did so earlier). Should perhaps again be a test which does this.

network/net.go
ineiti: Make this a constant at the beginning of the file?

Liamsi: Why not. If this info is relevant for the outside world make it even public. I guess the time (how long) and how often it retries here is (or could be) relevant for the caller. Not sure how this is reflected by `10`, though.

network/net.go
ineiti: break in switch doesn't break the for-loop:

https://play.golang.org/p/Po3eME1Ndy

but I'd be happy to see how to do it more clean!

protocols/manage/close_all_test.go
ineiti: It tests the `CloseAll`-protocol. I can wrap it in a go-routine and wait for a channel and timeout if it's not coming back in 10 seconds or so.

Liamsi:
Another general remark. As pointed out in https://github.com/dedis/cothority/issues/495#event-732574553 parts of this PR try to solve exactly the same problem as this pending PR https://github.com/dedis/cothority/pull/499. Hence there will be a large amount of merge conflicts (as both change the same lines).  Any ideas how we want to handle that? 

In other words: Do we want to use the port `0` trick (which by itself is cool and useful) for all the tests or do we want to mock them with channels? I'm undecided and don't have strong preference here. Should we think about pros/cons?

ineiti:
The 0-trick should also work with channels. So I don't see any conflict here. In the `sda`-directory I propose we keep the `NetworkRouter` because we want to test it on a lower level, while in the other directories (`services` and `protocols`) we can rely on `LocalRouter`.

network/net.go
Liamsi: Change to https://play.golang.org/p/LHumjEjsyh

network/net_test.go
```diff
@@ -307,6 +310,100 @@ func TestTcpNetwork(t *testing.T) {
...
+		}(i)
+	}
+	wg.Wait()
+	for i := range hosts {
+		wg.Add(1)
+		go func(nbr int) {
+			log.Lvl2("Opening", nbr, hosts[nbr])
+			c, err := hosts[nbr].Open(hosts[(nbr+1)%nbrHosts].serverIdentity)
+			log.ErrFatal(err)
+			if loadLen > 0 {
```
Liamsi: Same as above.

network/net.go
ineiti: Gone with the wind

network/net.go
ineiti: gone

Liamsi:
Looks good to me (has merge conflicts now, though)

ineiti:
fixed merge conflict.

Issue #482, opened 2016-07-15
Skipchain tests fail randomly


Liamsi:
See https://github.com/dedis/cothority/pull/473#issuecomment-232897406
and https://github.com/dedis/cothority/pull/468#issuecomment-232915001

ineiti:
#483

Issue #479, opened 2016-07-14
Medco - sometimes fails


nikkolasg:
When working on the PR #468 the `TestDeterministicSwitching` is failing on Travis (not locally though). According to @ineiti , `protocols/medco` seems to fail from time to time (10%).  I disabled the test in the PR while waiting for a fix. 
To reproduce: 
* `git checkout refactor_processor`
* Enable the test in protocols/medco/deterministic_switching_test.go

Note that I only it failing on Travis and not on my machine... Looking at the code, it seems a concurrency issue but i did not dig into it.

ineiti:
Another `skip` is in `services/medco/service_test.go`

Liamsi:
We can close this (as medco was moved to a private repo).

PR #499, opened 2016-07-28
Refactor SDA - Mocking network layer

nikkolasg:
This PR is about refactoring the whole network stack in `dedis/cothority`. Our goals are:
- [ ] **[1/4]** Using only *one* address for each node, greatly simplifying every operations with ServerIdentity. Each address should contains the type of connection. Ex: `tcp://5.135.161.91:2000`. PR #567.
- [ ] Implementing a `Local` network framework using only msg passing, arrays & signaling.
- [ ] Re-organize the TCP network framework with the new interfaces definition defined.
- [ ] Moving the whole network connection management to the `network/` package and having a reduced exported API (Router) on that so SDA have a strict set of operations allowed.
- [ ] Make the rest of the code uses the new `network/` stack with allowing to choose between TCP & Local tests.


Liamsi:
> Open question: Should I go forward and change the tests also in all protocols / services ? For me it would be the best reasonable thing to do, again to test not the network layer which is already tested, but only the logic of protocols and services..

I think it would be reasonable to make this a separate PR (if you are OK with that, too).

sda/overlay_test.go
```diff
@@ -64,3 +66,254 @@ func TestOverlayDone(t *testing.T) {
...
+		t.Fatal("List should be equal to original list")
+	}
+
+	// put back the overlay into place
+	h1.RegisterProcessor(h1.overlay, SendRosterMessageID)
+	err = h1.SendRaw(h2.ServerIdentity, &RequestRoster{RosterID: el.ID})
+	if err != nil {
+		t.Fatal("Couldn't send message to h2:", err)
+	}
+	// XXX We should be able to remove this arbitrary sleeping
```
Liamsi: Please try :-)

protocols/bftcosi/bftcosi_test.go
```diff
@@ -15,7 +15,7 @@ var veriCount int
 var countMut sync.Mutex
 
 func TestMain(m *testing.M) {
-	log.Info("skipping test because of https://github.com/dedis/cothority/issues/467")
+	//log.Info("skipping test because of https://github.com/dedis/cothority/issues/467")
```
Liamsi: Huh, why remove this output?

sda/processor_test.go
```diff
@@ -64,7 +65,7 @@ func TestProcessorHost(t *testing.T) {
 		basic, ok := m.Msg.(basicMessage)
 		assert.True(t, ok)
 		assert.Equal(t, basic.Value, 10)
-	default:
+	case <-time.After(100 * time.Millisecond):
```
Liamsi: back to (arbitrary) timeouts? Are you experimenting with that or will this stay? If the latter: create a comment why 100ms are reasonable.

nikkolasg:
Finally !! Travis is happy !! With a hardcoded branch name in `install.sh` for the cosi repo.. :/
Just to compare:
before
```
ok  	github.com/dedis/cothority/lib/sda	48.556s
```
after
```
ok  	github.com/dedis/cothority/sda	19.843s
```
The difference is not so huge on my local machine however but still !

network/encoding.go
```diff
@@ -90,7 +94,7 @@ func TypeToMessageTypeID(msg Body) MessageTypeID {
 	}
 	url := NamespaceProtocolMessageType + val.Type().String()
 	u := uuid.NewV5(uuid.NamespaceURL, url)
-	log.Lvl5("Reflecting", reflect.TypeOf(msg), "to", u)
+	//log.Lvl5("Reflecting", reflect.TypeOf(msg), "to", u)
```
Liamsi: Should we remove this?

network/encoding.go
```diff
@@ -36,7 +36,11 @@ var ErrorType = MessageTypeID(uuid.Nil)
 
 // String returns the canonical string representation of the MessageTypeID
 func (mId MessageTypeID) String() string {
-	return uuid.UUID(mId).String()
+	t, ok := registry.get(mId)
+	if !ok {
+		return uuid.UUID(mId).String()
+	}
+	return t.String()
```
Liamsi: If a type is already registered, you return the String representation of `reflect.Type` and if not a `uuid`? Is this for (human) readability or does it have another motivation?

network/encoding.go
```diff
@@ -117,7 +121,7 @@ func DefaultConstructors(suite abstract.Suite) protobuf.Constructors {
 	return constructors
 }
 
-// Error returns the error that has been encountered during the unmarshaling of
+// Error returns the eror that has been encountered during the unmarshaling of
```
Liamsi: s/eror/error

network/encoding.go
nikkolasg: human readability : instead of having to decode `uuda4-230vdas-123fs-blabla-blabla`, we get `sda.TreeRequest`

network/struct.go
```diff
@@ -185,6 +185,8 @@ type Packet struct {
 	Constructors protobuf.Constructors
 	// possible error during unmarshalling so that upper layer can know it
 	err error
+	// XXX REMOVE
+	To *ServerIdentity
```
Liamsi: Looks like although it is written in two places, this field is never read. so, we could really remove it?

protocols/bftcosi/bftcosi.go
```diff
@@ -161,7 +161,7 @@ func (bft *ProtocolBFTCoSi) Start() error {
 // Dispatch listens on all channels and implements the sda.ProtocolInstance
 // interface.
 func (bft *ProtocolBFTCoSi) Dispatch() error {
-	log.Lvl2(bft.Name(), "Starts")
+	log.LLvl2(bft.Name(), "Starts")
```
Liamsi: s/log.LLvl2/log.Lvl2

network/struct.go
nikkolasg: Yep I forgot.

protocols/bftcosi/bftcosi_test.go
```diff
@@ -15,8 +15,8 @@ var veriCount int
 var countMut sync.Mutex
 
 func TestMain(m *testing.M) {
-	log.Info("skipping test because of https://github.com/dedis/cothority/issues/467")
-	//log.MainTest(m)
+	//log.Info("skipping test because of https://github.com/dedis/cothority/issues/467")
```
Liamsi: 👍 👍 👍   But don't forget to remove the comment.

protocols/medco/deterministic_switching_test.go
```diff
@@ -13,7 +13,6 @@ import (
 )
 
 func TestDeterministicSwitching(t *testing.T) {
-	t.Skip("Skipping due to error in medco: https://github.com/dedis/cothority/issues/479")
```
Liamsi: Did you also resolve this?

nikkolasg: Nope as for me it's deterministically failing I don't know why, I told @froelich to look into it so I keep it skipped

sda/export_test.go
```diff
@@ -48,25 +42,35 @@ func (o *Overlay) TokenToNode(tok *Token) (*TreeNodeInstance, bool) {
 	return tni, ok
 }
 
-func (h *Host) AbortConnections() error {
-	h.closeConnections()
-	close(h.ProcessMessagesQuit)
-	return h.host.Close()
+// AddTree registers the given Tree struct in the underlying overlay.
+// Useful for unit-testing only.
+func (h *Host) AddTree(t *Tree) {
```
Liamsi: Having public methods for testing purposes only. I guess there is no way to circumvent this?

nikkolasg: I can put them as unexported but since it is in `export_test.go` it is already only available for the tests so it does not change really anything.

network/encoding.go
Liamsi: I think 7bff8dbce29cfd690ce1d08ace5e566eed14334ea287e82e06ee2f4df925eed0 .

protocols/medco/deterministic_switching_test.go
Liamsi: Then you better keep it skipped (as you say).

sda/export_test.go
```diff
@@ -48,25 +42,35 @@ func (o *Overlay) TokenToNode(tok *Token) (*TreeNodeInstance, bool) {
...
+// Useful for unit-testing only.
+func (h *Host) AddRoster(el *Roster) {
+	h.overlay.RegisterRoster(el)
 }
 
-func (h *Host) RegisterConnection(e *network.ServerIdentity, c network.SecureConn) {
-	h.networkLock.Lock()
-	defer h.networkLock.Unlock()
-	h.connections[e.ID] = c
+func (t *TCPRouter) AbortConnections() error {
```
Liamsi: Isn't `Close` a more common name for this?

nikkolasg: No `Close` is already defined by `Router` , this was to simulate a hard network failure of a node but in fact it does not do what we want. I kept this a compatibility in `export_test.go` since I'm sure @ineiti would want to use it, but I don't see how. For more info see `TestTcpRouterReconnection`.

sda/host.go
```diff
@@ -25,54 +18,29 @@ type Host struct {
...
-
+	treesLock            sync.Mutex
 	serviceManager       *serviceManager
 	statusReporterStruct *statusReporterStruct
 }
 
-// NewHost starts a new Host that will listen on the network for incoming
-// messages. It will store the private-key.
+// NewHost returns a new Host that out of a private-key and its relating public
+// key with a default TcpRouter as Router.
```
Liamsi: Where do we use the its "relating public key" here?

nikkolasg: Everywhere ? This is the public key used by all protocols, given in any ServerIdentity relating to this Host, etc ... :D

Liamsi: I know that we use the public "everywhere". But I'm obviously missing sth. here: `pkey` stands for private key here, right? (at least it is a `abstract.Scalar`) I think the comment is a little confusing because we pass a private-key and no corresponding public key. In fact, the public key is implicitly passed through the `ServerIdentity`.

sda/export_test.go
Liamsi: OK, I see. Makes sense!

Liamsi:
One general remark: Could you locally run `go test` with a coverprofile on the (most) affected/changed packages? Just to see if we (drastically) changed the test coverage in one way or the other?

sda/local.go
```diff
@@ -243,60 +242,48 @@ func (l *LocalTest) GetServices(hosts []*Host, sid ServiceID) []Service {
...
+	priv, id := NewPrivIdentity(port)
 	return NewHost(id, priv)
 }
 
-// GenLocalHosts will create n hosts with the first one being connected to each of
-// the other nodes if connect is true.
-func GenLocalHosts(n int, connect bool, processMessages bool) []*Host {
+// NewLocalHost returns a new host using a LocalRouter (channels) to communicate.
+// At the return of this function, the router is already Run()ing in a go
+// routine.
```
Liamsi: +1 for good/useful comment (mentioning that `Run` is already started in a go-rountine).

sda/processor.go
```diff
@@ -70,10 +74,13 @@ func (d *BlockingDispatcher) RegisterProcessor(p Processor, msgType ...network.M
...
 // blocking call if the Processor is blocking !
 func (d *BlockingDispatcher) Dispatch(packet *network.Packet) error {
+	d.Lock()
 	var p Processor
 	if p = d.procs[packet.MsgType]; p == nil {
-		return errors.New("No Processor attached to this message type.")
+		d.Unlock()
+		return errors.New("No Processor attached to this message type " + packet.MsgType.String())
 	}
+	d.Unlock()
```
Liamsi: Why not `defer d.Unlock()` instead?

sda/local.go
nikkolasg: Yep. It just made me realizes that I can even remove all of these suspicious magical boolean flags on theses methods (`connect, process, register bool`...)

sda/processor.go
```diff
@@ -222,6 +229,7 @@ func (p *ServiceProcessor) SendISMOthers(el *Roster, msg network.Body) error {
 func (p *ServiceProcessor) GetReply(e *network.ServerIdentity, mt network.MessageTypeID, m network.Body) network.Body {
 	fu, ok := p.functions[mt]
 	if !ok {
+		log.Error("Error ServiceProcessor(1) Didn't register message", mt.String())
```
Liamsi: What about the (1) in "ServiceProcessor(1) "?

nikkolasg:
Cover master: `ok  	github.com/dedis/cothority/sda	14.620s	coverage: 72.5% of statements`
Cover branch: `ok  	github.com/dedis/cothority/sda	11.275s	coverage: 71.8% of statements`

sda/processor.go
nikkolasg: I wanted to unlock before the processor gets the message otherwise no other processor could be called at the same time (by different go routines for example). Only one processor could be having a message at a time.

Liamsi:
OK, thats not a big deal (although I think we could aim for slightly higher coverage, but that's another issue).

sda/processor.go
Liamsi: Thanks for the explanation!

sda/processor_test.go
```diff
@@ -171,7 +170,7 @@ type testMsg struct {
 func procMsg(e *network.ServerIdentity, msg *testMsg) (network.Body, error) {
 	// Return an error for testing
 	if msg.I == 42 {
-		return nil, errors.New("6 * 9 != 42")
+		return nil, errors.New("ServiceProcessor received unwanted value (42)")
```
Liamsi: Can you make this `fmt.Errorf("ServiceProcessor received msg=%s but expected 42", msg.I)`?

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+	"github.com/dedis/cothority/network"
+	"github.com/dedis/crypto/abstract"
+	"golang.org/x/net/context"
+)
+
+// Router is an abstraction to represent the bridge between the communication
+// layer (network, channels etc) and the logical/processing layer (overlay &
+// protocols, services etc). It is a duplex communication link (send/receive)
+// from/to other router of the same type.
+// Typically, for deployement you would use a tcpRouter so it opens tcp ports
```
Liamsi: s/deployement/deployment

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+	"github.com/dedis/cothority/log"
+	"github.com/dedis/cothority/network"
+	"github.com/dedis/crypto/abstract"
+	"golang.org/x/net/context"
+)
+
+// Router is an abstraction to represent the bridge between the communication
+// layer (network, channels etc) and the logical/processing layer (overlay &
+// protocols, services etc). It is a duplex communication link (send/receive)
+// from/to other router of the same type.
```
Liamsi: "from/to other router of the same type." other router*s* or from/to another router (if it is just one)

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+// Typically, for deployement you would use a tcpRouter so it opens tcp ports
+// and communicate through tcp connections. For testing, there is a localRouter
+// which pass all messages through channels going to other localRouter.
```
Liamsi: passes

Liamsi: another localRouter

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+// For the Router to dispatch messages to your struct, you need to register a
+// `Processor` (see the `Dispatcher` interface in processor.go).
+type Router interface {
+	// Run will start the Router:
+	//  * Accepting new connections
+	//  * Dispatching  incoming messages
+	// It is a blocking call which will return until Close() is called.
```
Liamsi: "It will block until Close()" is called or "It is a blocking call which *won't* return until Close() is called."

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+	Run()
+	// Close will stop the Router from running and will close all connections.
+	// It makes the Run() returns.
```
Liamsi: It makes the Run() method return~~s~~.

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+	// Router is a Dispatcher so you can register any Processor to it. Every
+	// messages coming to this Router will be dispatched accordingly to its
+	// registered Processors.
+	Dispatcher
+
+	// Send will send the message msg to e.
+	Send(e *network.ServerIdentity, msg network.Body) error
+
+	// XXX Feels like there's a lot of common goal for the next methods
+	// that could maybe be factored together into something simpler...
```
Liamsi: Can we describe your feelings about this in a new issue? (and remove this comment?)

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+	//  * Accepting new connections
+	//  * Dispatching  incoming messages
+	// It is a blocking call which will return until Close() is called.
+	Run()
+	// Close will stop the Router from running and will close all connections.
+	// It makes the Run() returns.
+	Close() error
+
+	// Router is a Dispatcher so you can register any Processor to it. Every
+	// messages coming to this Router will be dispatched accordingly to its
```
Liamsi: s/according/according (?)

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+	// that could maybe be factored together into something simpler...
+	Tx() uint64
+	Rx() uint64
+	StatusReporter
+	Address() string
+}
+
+// TCPRouter is a Router implementation that uses TCP connections to communicate
+// to different hosts. It manages automatically the connection to hosts, the
+// maintenance of the connections etc. It is believed to be thread-safe.
```
Liamsi: Is it thread-safe or not?

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+	return &TCPRouter{
+		Dispatcher:     NewBlockingDispatcher(),
+		workingAddress: e.First(),
+		connections:    make(map[network.ServerIdentityID]network.SecureConn),
+		host:           network.NewSecureTCPHost(pkey, e),
+		suite:          network.Suite,
+		serverIdentity: e,
+		quitProcessMsg: make(chan bool),
+		closing:        make(chan bool, 1), // buffered channel of 1 so Close() without
+		// Run() before does not fail
```
Liamsi: What does that comment refer to?

sda/router.go
```diff
@@ -0,0 +1,550 @@
...
+		c, err = t.Connect(e)
+		if err != nil {
+			return err
+		}
+	}
+
+	log.Lvlf4("%s sends to %s msg: %+v", t.serverIdentity.Addresses, e, msg)
+	var err error
+	err = c.Send(context.TODO(), msg)
+	if err != nil /*&& err != network.ErrClosed*/ {
```
Liamsi: Can we remove `/*&& err != network.ErrClosed*/`?

sda/router.go
```diff
@@ -0,0 +1,549 @@
...
+}
+
+// processMessages checks if it is one of the messages for us or dispatch it
+// to the corresponding instance.
+// Our messages are:
+// * SDAMessage - used to communicate between the Hosts
+// * RequestTreeID - ask the parent for a given tree
+// * SendTree - send the tree to the child
+// * RequestPeerListID - ask the parent for a given peerList
+// * SendPeerListID - send the tree to the child
```
Liamsi: This is a question: What do all these message types have to do with the code below? I don't see any obvious connection. Searching for `RequestTreeID` for instance doesn't seem to yield any results.

nikkolasg: Old comment. Should have been removed during the pr for `refactor_processor`. Will remove it. (it belongs to `overlay` now to treat these messages)

sda/router.go
```diff
@@ -0,0 +1,543 @@
...
+	err := t.host.Listen(fn)
+	if err != nil {
+		log.Fatal("Error listening on", t.workingAddress, ":", err)
+		return err
+	}
+	return nil
+}
+
+// processMessages is receiving all the messages coming from the network and
+// dispatch it to the Dispatcher.
```
Liamsi: s/dispatch it to the Dispatcher/dispatches them to the Dispatcher

sda/router.go
```diff
@@ -0,0 +1,543 @@
...
+	delete(t.connections, c.ServerIdentity().ID)
+	return nil
+}
+
+// Handle a connection => giving messages to the MsgChans
+func (t *TCPRouter) handleConn(c network.SecureConn) {
+	address := c.Remote()
+	for {
+		ctx := context.TODO()
+		am, err := c.Receive(ctx)
```
Liamsi: Can we name this local var `msg` or just `m` to be consistent?

sda/router.go
```diff
@@ -0,0 +1,543 @@
...
+// real physical address of the connection and the connection itself
+// it locks (and unlocks when done):  networkLock
+func (t *TCPRouter) registerConnection(c network.SecureConn) {
+	log.Lvl4(t.serverIdentity.First(), "registers", c.ServerIdentity().First())
+	t.networkLock.Lock()
+	defer t.networkLock.Unlock()
+	id := c.ServerIdentity()
+	_, okc := t.connections[id.ID]
+	if okc {
+		// TODO - we should catch this in some way
```
Liamsi: Would it make sense to return an error here (at `registerConnection`)?

nikkolasg: I'm not sure. I've put the TODO here a long time ago and it has never re appeared again so... I'm not sure it's useful. Moreover, we should before doing anything else about this, specify what happens when the same connection or identity appears again. For the moment the latest registered connection will always be the one active. I guess that's subject to discussion somewhere else... (and things will change when/if we introduce TLS in the whole story)

sda/router.go
```diff
@@ -0,0 +1,543 @@
...
+	t.networkLock.Unlock()
+	// The tcp host is supposed to take care of the connection for us
+	if err := t.host.Close(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// closeConnection closes a connection and removes it from the connections-map
+// The t.networkLock must be taken.
```
Liamsi: It seems like networkLock is used to protect several/different fields. Can you group this fields together in https://github.com/dedis/cothority/pull/499/files#diff-1ce7e0aab009980d18d8fdd037e3e602R64

Do comments like `// The t.networkLock must be taken.` help the dev using this method in any way? Should it warn the dev: "hey, I'll use this lock be careful howto use this function"?

sda/router_test.go
```diff
@@ -0,0 +1,325 @@
...
+	// XXX Actually it DOES send a FIN packet: using tcphost.Close(), it closes
+	// the listener AND all the connections (calling golang tcp connection
+	// Close() which I'm pretty sure will send a FIN packet)
+	// This test is ambiguous as it does not really simulate a network hardware
+	// failure of a node, but merely a host which does weird abort
+	// connections...
+	// One idea if we really want to simulate that is calling tcphost.Close()
+	// and at the same time, at the IP level, blocking all FIN packet.
+	// Then start a new host with the same entity etc..
+	// See also https://github.com/tylertreat/comcast
```
Liamsi: OK, it looks like we can remove this test then.

sda/router_test.go
```diff
@@ -0,0 +1,325 @@
...
+	case <-time.After(500 * time.Millisecond):
+		t.Fatal("TcpHost should have returned from Run() by now")
+	}
+}
+
+func TestTcpRouterReconnection(t *testing.T) {
+	h1 := NewMockTcpRouter(2000)
+	h2 := NewMockTcpRouter(2001)
+	/* h1 := NewLocalHost(2000)*/
+	/*h2 := NewLocalHost(2001)*/
```
Liamsi: remove this comments

sda/router_test.go
Liamsi: Or at least rename it. Because now it doesn't attempt to reconnect (`TestTcpRouterReconnection `), right?

sda/router.go
nikkolasg: OK right I've added a new lock then only for the processing of the message. Now networkLock is only used for the connections.

sda/router_test.go
nikkolasg: It still tries to reconnect after a closing so technically the name seems to fit in ...

sda/router.go
```diff
@@ -0,0 +1,543 @@
...
+	c, _ := t.connections[e.ID]
+	return c
+}
+
+// MOCKING NETWORK ROUTER
+// localRelay defines the basic functionalities such as sending and
+// receiving a message, locally. It is implemented by LocalRouter and
+// localClient so both a Router and a Client can be emulated locally without
+// opening any real connections.
+type localRelay interface {
```
Liamsi: Can't we move into a test package?

nikkolasg: I wanted so at the beginning but then we can't use mocking network outside of `sda/` anymore... which is also the whole point of this PR. Is there a way for test packages to import other test packages and so on ? If possible then yes. I haven't looked it through. The whole `local.go` file could also be put into a test file then.

sda/service.go
```diff
@@ -113,6 +114,13 @@ func RegisterNewService(name string, fn NewServiceFunc) {
 	ServiceFactory.Register(name, fn)
 }
 
+// DeleteNewService will remove the NewServiceFunc from the global store of
+// NewServiceFunc so it can't be initialized again. If the service needs to be
+// initialized again, a new call to RegisterNewService must be made.
```
Liamsi: This comment seems a little outdated or not? 
Change to: `DeleteNewService` completely removes the service associated with the passed name.

sda/service.go
```diff
@@ -135,6 +143,22 @@ func (s *serviceFactory) RegisteredServicesName() []string {
 	return names
 }
 
+// DeleteNewService will remove the NewServiceFunc from the global store of
+// NewServiceFunc so it can't be initialized again. If the service needs to be
+// initialized again, a new call to RegisterNewService must be made.
```
Liamsi: Same as above.

sda/service.go
```diff
@@ -358,28 +382,38 @@ func CreateServiceMessage(service string, r interface{}) (*InterServiceMessage,
...
+type TCPClient struct {
 	host      *network.SecureTCPHost
 	ServiceID ServiceID
 	sync.Mutex
 }
 
-// NewClient returns a random client using the service s
-func NewClient(s string) *Client {
-	return &Client{
+// NewClient returns a random client using the service s. It's a TCPClient by
```
Liamsi: what is a "random client". Maybe you mean an "arbitrary client". But I think you can just say client (without random/anything else).

sda/service.go
```diff
@@ -460,18 +495,6 @@ func (c *Client) SendToAll(dst *Roster, msg network.Body) ([]*network.Packet, er
...
-	log.Fatal("Not yet implemented")
-	return nil, nil
-}
-
-// BinaryUnmarshaler sets the different values from a byte-slice
-func (c *Client) BinaryUnmarshaler(b []byte) error {
-	log.Fatal("Not yet implemented")
-	return nil
-}
-
```
Liamsi: 👍

sda/service_test.go
```diff
@@ -126,6 +150,45 @@ func TestServiceNew(t *testing.T) {
 	waitOrFatal(ds.link, t)
 }
 
+func TestServiceChannels(t *testing.T) {
+	sc1 := &ServiceChannels{}
+	sc2 := &ServiceChannels{}
+	var count int
+	// Needed because of TestServiceNew which listens on a channel. Strange that
+	// it worked before.
+	DeleteNewService("DummyService")
```
Liamsi: Maybe, I don't really get that: If `TestServiceNew` creates a "DummyServices" it should be `TestServiceNew` which does its cleanup and calls `DeleteNewService("DummyService")` or not?

services/identity/api.go
```diff
@@ -35,17 +35,26 @@ func init() {
...
 // Identity structure holds the data necessary for a client/device to use the
 // identity-service. Each identity-skipchain is tied to a roster that is defined
-// in 'Cothority'.
+// in 'Cothority'
 type Identity struct {
 	// Client is included for easy `Send`-methods.
-	*sda.Client
+	sda.Client
+	// IdentityData holds all the data related to this identity
+	// It will be stored and load from a config file or else.
```
Liamsi: Change to: "It can be stored and loaded from a config file."

services/identity/service_test.go
```diff
@@ -6,16 +6,23 @@ import (
...
 	"github.com/dedis/cothority/network"
 	"github.com/dedis/cothority/sda"
+	"github.com/dedis/cothority/services/skipchain"
 	"github.com/dedis/crypto/config"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestMain(m *testing.M) {
-	log.Info("Skipping because of bftcosi and skipchain - #482")
+	//log.Info("Skipping because of bftcosi and skipchain - #482")
```
Liamsi: Cool! Remove the comment. Maybe we can even close #482 ...

services/skipchain/skipchain_test.go
```diff
@@ -16,7 +16,8 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	log.Info("Skipping all skipchain tests for now as it randomly fails on master. https://github.com/dedis/cothority/pull/483")
+	//j	log.Info("Skipping all skipchain tests for now as it randomly fails on master. https://github.com/dedis/cothority/pull/483")
```
Liamsi: Remove this comment

simul/cothority/cothority.go
```diff
@@ -60,13 +61,28 @@ func main() {
 	sims := make([]sda.Simulation, len(scs))
 	var rootSC *sda.SimulationConfig
 	var rootSim sda.Simulation
+	// having a waitgroup so the binary stops when all hosts are closed
+	var wg sync.WaitGroup
```
Liamsi: 👍 👍 👍

services/skipchain/skipchain_test.go
nikkolasg: I **hope** it works, we're not sure since the bug does not occur deterministically...

protocols/medco/deterministic_switching_test.go
Liamsi: I'll do extensive tests (on travis). If they all pass repeatedly we even close #479

Liamsi:
Of course it was too good to be true: https://travis-ci.org/dedis/cothority/builds/149748561#L225

Liamsi:
OK, I re-ran the travis build several times (sth. we shouldn't need to do, but currently there are several bugs which occur but only sometimes and it is not clear if these bugs occur only in the tests or have effects in real deployment). As the last run reveals a deadlock which directly seems to be related to the changes in this PR, I'll put it back to WIP (https://travis-ci.org/dedis/cothority/builds/149791908).

Liamsi:
Went back to run the tests sequentially (after running about 15 times): https://travis-ci.org/dedis/cothority/builds/149992516

nikkolasg:
>  As the last run reveals a deadlock which directly seems to be related to the changes in this PR, I'll put it back to WIP (https://travis-ci.org/dedis/cothority/builds/149791908).

Unfortunatly this error already occured during somes Travis CI tests, again not deterministically... I'm afraid no PR will ever be merged because of theses hidden bugs >< As we found out also there was even some tests which are not working locally on our machine from *master* ... A big cleanup of these bugs seems more than needed :/  But to the point: I really doubt this one comes from this PR. But again, we can't be sure :)

ineiti:
Short summary of some of the discussions we had with Ismail: how does this work together with the interfaces from `network/struct.go`? `Host`, `Conn` and their `Secure`-counterparts seem more appropriate to handle what is being done here in `Router`.

@nikkolasg , can you explain how the `Router`-interface is better than having different implementations of the `network/struct.go`-interfaces?

If we keep the `Router`-interface, would it make sense to have a `NetworkRouter` instead of a `TCPRouter` and then add `TCP`, `UDP`, `TLS` to that `NetworkRouter`, let the user chose what is most appropriate?

Or, if we don't use the `Router`-interface, should we use the `ServerIdentity` and its `Address` to reflect different possibilities of connections (again, TCP, UDP and TLS)? IMO, this would make it easy to add different sending-paths for the protocols (like PriFi wanting to use UDP instead of TCP).

Looking forward to white-board this thing with you!

ineiti:
replaced by #574

Issue #506, opened 2016-08-08
Cleanup code in general


ineiti:
- method-signature mixups
  - [x] `local.StartProtocol` and `local.CreateProtocol` have very similar functions but different signatures -> mend it
  - [x] `overlay.*Protocol*` also differs in signature -> mend it
  - [x] `network` still has some fields called `entity` instead of `serverIdentity`
  - [x] remove old `ProtocolMessages` and `Packet` references and methods

Import Makefile-changes from #499 and #508 once it's clear it's working.

Liamsi:
> interfaces with only one implementation -> remove them

Is there any particular reason to remove them? I mean even if there is just one implementation: deleting the interfaces makes it impossible to mock them (even if locally) or pick another implementation (greater flexibility with almost no cost).

ineiti:
If the interfaces are used, it is good, and we keep them.
If the interfaces just hide a single structure, they make it more complicated.
As the implementation of the 'channel-network' doesn't use the interfaces (which would be the 1 use-case I can think about for these interfaces), I see no need to keep them around.

Besides giving a nice feeling, what is the advantage of having an interface with only one implementation? In my opinion it just adds an unused layer which makes things more complicated.

Liamsi:
Pros&Cons of interfaces (even or especially with a single implementation):

Pros:
+ ability change underlying implementation without tocuhing the rest of the code
+ clear interface independent from implementation
+ ability to "mock" away the implementation for unit tests
+ robust API for future use-cases (change `SecureTCPConnection` with `SecureUPDConnection` and/or `SecurePURPSConnection`, etc.)

Cons:
- "hides" the implementation (one extra search/click) to find it
- additional complexity (?)

Did I miss something?

ineiti:
Due to popular request and promises to use the interfaces I revived `Host`, `Conn`, `SecureHost`, and `SecureConn`.

As a general rule I do think it's better to add interfaces only once there is more than one implementation.

PR #508, opened 2016-08-09
Refactor cothority 506

ineiti:
Renamed branch so it works with new `install.sh`

This replaces #507 and implements #506


Liamsi:
I ran that over 20 times on travis. Seems to work fine 👍

install.sh
```diff
@@ -0,0 +1,46 @@
...
+# https://gist.github.com/derekstavis/0526ac13cfecb5d6ffe5#file-travis-github-pull-request-integration-sh
+GITHUB_PR_URL=https://api.github.com/repos/$TRAVIS_REPO_SLUG/pulls/$TRAVIS_PULL_REQUEST
+GITHUB_PR_BODY=$(curl -s $GITHUB_PR_URL 2>/dev/null)
+if [[ $GITHUB_PR_BODY =~ \"ref\":\ *\"([a-zA-Z0-9_-]*)\" ]]; then
+  BRANCH3=${BASH_REMATCH[1]}
+fi
+
+echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, BRANCHES=$BRANCH1--$BRANCH2--$BRANCH3"
+export BRANCH=$BRANCH1
+export BRANCH=refactor_cothority_506
```
Liamsi: How do we want to handle this here? And in general?

services/identity/service_test.go
```diff
@@ -12,7 +12,7 @@ import (
 
 func TestMain(m *testing.M) {
 	log.Info("Skipping because of bftcosi and skipchain - #482")
-	log.MainTest(m)
+	//log.MainTest(m)
```
Liamsi: Is this on purpose? Can we have a comment with reference to some issue (or a skip message?)

Liamsi:
This PR is fine! Not sure how we want to handle this "getting the right branch" for cosi, though. I hardcoded the branch again for now.

ineiti:
Hmm - could you confirm my hunch that you can re-run the PR's up to the moment you add a new commit?

Liamsi:
Unfortunately, it looks random to me. Even failed on the first build sometimes. But I might have looked at the wrong build, though.

Liamsi:
I suspect that the github API gets so many requests from travis' IP range and that's why the `curl` command in our Makefile doesn't always work. 
I see only two ways how-to resolve this: 

1) Hacky-di-Hack-Way: Hardcode the branch name (for cosi) in the PR (not the push) case. The PR case doesn't harm  (won't get called) if merged into master.

2) Clean (yet certainly more unpopular) way: If there always will be a dependency from *and* to cosi and cothtority repo we should rethink our decision to seperate both in two repos. I think you cannot refactor one, without also refactoring the other it is clear sign that both projects belong in one repo instead.

ineiti:
1) I think we'll end up there...

2) I prefer having a separation of `SDA`, `cosi` and `cothority` as described in
https://docs.google.com/document/d/1sJTRgyezQAcbPzqrSoMxmAoqlG_RW8Rtw7DGmYjl1L8/edit?usp=sharing
We're mainly changing things in `SDA` which break the tests including cosi. So if we separate `SDA` and `cosi`, this breakage will only come once we merge the changes into `SDA`...

Liamsi:
> I prefer having a separation of SDA, cosi and cothority as described in
https://docs.google.com/document/d/1sJTRgyezQAcbPzqrSoMxmAoqlG_RW8Rtw7DGmYjl1L8/edit?usp=sharing

I don't see the clear benefit here. It just moves the problem (which already spent several days on) and splitting it in even more repos just adds more complexity to our daily work. I see it is useful for people to have a separate cosi repo but if we can not modify one repo without touching the other ... they should be one repo instead. People can still download/build cosi-only using the correct cmd (sth like `go get -u path/to/cosi-binary`). 
Either this or make the the dependency/dependencies only uni-directional instead of bi-directional. But let's discuss this offline together with @nikkolasg.

> 1) I think we'll end up there...

Another solution we could consider is "vendoring" all dependencies (maybe this works with brnaches, too) using `GO15VENDOREXPERIMENT `  and/or some of these tools: https://github.com/golang/go/wiki/PackageManagementTools

Liamsi:
The error message you printed (good idea) says:
```
{ "message": "API rate limit exceeded for 52.0.240.122. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)", "documentation_url": "https://developer.github.com/v3/#rate-limiting" }
```
Does this mean we could use *un*authenticated requests somehow to get a lower rate-limit (for requesting the branch name from the repo)?

ineiti:
For the moment we do _un_authenticated requests. We could use Authenticated requests, but then we'd have to put our credentials in the `install.sh`-file, which seems like a bad idea.

ineiti:
Selling this. Who wants it?

.travis.yml
```diff
@@ -7,7 +7,7 @@ install:
   - ./install.sh
 
 script:
-  - make test_fmt
+  - make test
```
Liamsi: What? We weren't even running the test prior this change :-(

Liamsi:
> We could, but then we'd have to put our credentials in the install.sh-file, which seems like a bad idea.

@nikkolasg suggested to use encrypted auth tokens to prevent this. In issue #548 we now use this and it works fine.

PR #511, opened 2016-08-10
Fix skipchain bftcosi 482

ineiti:
Starting to work on the bftcosi-error.


ineiti:
Starting to get some interesting errors when running 20 nodes in `TestCheckFailBit`:

```
E : (                  sda.(*Host).handleConn: 306) - [localhost:62509] Error with connection [::1]:62550 => Error unmarshaling message type 04d0b144-7f6c-501c-bcba-54471d6bcc47: FieldName : unknown protobuf wire-type
```
and
```
E : (                  sda.(*Host).handleConn: 306) - [localhost:64904] Error with connection [::1]:64942 => Error Received message (size=1488): runtime error: index out of range
```

sda/treenode.go
```diff
@@ -379,15 +383,20 @@ func (n *TreeNodeInstance) dispatchMsgToProtocol(sdaMsg *ProtocolMsg) error {
...
 
 	switch {
 	case n.channels[msgType] != nil:
-		log.Lvl4(n.Info(), "Dispatching to channel")
+		log.LLvl4(n.Name(), "Dispatching to channel")
 		err = n.DispatchChannel(msgs)
 	case n.handlers[msgType] != nil:
 		log.Lvl4("Dispatching to handler", n.ServerIdentity().Addresses)
-		err = n.dispatchHandler(msgs)
+		go func() {
```
ineiti: This is needed because `bftcosi` does wait in a handler on a channel to be filled with a value. But this value will be sent from another handler. So either we don't allow for more than one handler to be active, or we accept this go.

network/net.go
```diff
@@ -419,11 +419,14 @@ func (c *TCPConn) Receive(ctx context.Context) (nm Packet, e error) {
 			log.Error("Couldn't write to buffer:", err)
 		}
 		read += Size(n)
+		b = b[n:]
```
Liamsi: 👍 👍 👍  for finding and solving this one!
Things like this should be covered by unit tests (you already opened an issue for that?)

ineiti: #514

protocols/bftcosi/bftcosi.go
```diff
@@ -228,125 +237,97 @@ func (bft *ProtocolBFTCoSi) RegisterOnSignatureDone(fn func(*BFTSignature)) {
...
-}
-
 // handleCommitment collects all commitments from children and passes them
 // to the parent or starts the challenge-round if it's the root.
-func (bft *ProtocolBFTCoSi) handleCommitment(comm Commitment) error {
-	log.Lvl4(bft.Name(), "RoundType:", comm.TYPE)
-
-	var commitment abstract.Point
-	// store it and check if we have enough commitments
+func (bft *ProtocolBFTCoSi) handleCommitment(msgs []commitChan) error {
```
Liamsi: Why not make this a receiving channel in the method's signature? It would make clearer that we are actually expecting this to be called with an channel as an argument. See: https://gobyexample.com/channel-directions for examples.

protocols/bftcosi/bftcosi.go
```diff
@@ -393,29 +379,112 @@ func (bft *ProtocolBFTCoSi) handleChallengeCommit(ch *ChallengeCommit) error {
...
+		bft.challengeCommitChan <- challengeCommitChan{ChallengeCommit: *cc}
+	}
+	return nil
+}
+
+// startResponse dispatches the response to the correct round-type
+func (bft *ProtocolBFTCoSi) startResponse(t RoundType) error {
+	log.Lvl3(bft.Name(), "RoundType:", t)
+	bft.handleResponse([]responseChan{{Response: Response{TYPE: t}}})
+	return nil
```
Liamsi: This will always retun `nil` or not?

protocols/bftcosi/bftcosi.go
```diff
@@ -393,29 +379,112 @@ func (bft *ProtocolBFTCoSi) handleChallengeCommit(ch *ChallengeCommit) error {
...
+		}
+		bft.challengeCommitChan <- challengeCommitChan{ChallengeCommit: *cc}
+	}
+	return nil
+}
+
+// startResponse dispatches the response to the correct round-type
+func (bft *ProtocolBFTCoSi) startResponse(t RoundType) error {
+	log.Lvl3(bft.Name(), "RoundType:", t)
+	bft.handleResponse([]responseChan{{Response: Response{TYPE: t}}})
```
Liamsi: Not sure if this would work with https://github.com/dedis/cothority/pull/511/files#r74425842

protocols/bftcosi/bftcosi.go
ineiti: sometimes the `handle.*`-methods get called directly with a structure. So instead of having `handle` with receiving channels and `handle` with structures/slices, I prefer having only one signature-type.

protocols/bftcosi/bftcosi.go
```diff
@@ -61,18 +62,11 @@ type ProtocolBFTCoSi struct {
 	// channel for challenge during the commit phase
 	challengeCommitChan chan challengeCommitChan
 	// channel for response
-	responseChan chan responseChan
+	responseChan chan []responseChan
```
Liamsi: Couldn't this be a buffered channel instead of a channel of an array?

protocols/bftcosi/bftcosi.go
```diff
@@ -228,125 +237,97 @@ func (bft *ProtocolBFTCoSi) RegisterOnSignatureDone(fn func(*BFTSignature)) {
...
-		},
-	}
-	return bft.handleChallengeCommit(cc)
+	return nil
 }
 
 // handleChallengePrepare collects the challenge-messages
-func (bft *ProtocolBFTCoSi) handleChallengePrepare(ch *ChallengePrepare) error {
+func (bft *ProtocolBFTCoSi) handleChallengePrepare(msg challengePrepareChan) error {
+	log.Lvl3(bft.Name(), "RoundType: 0")
```
Liamsi: What is "RoundType: 0"? Can we make this log output slightly more readable (RoundType: Challenge Prepare / sth similar)

protocols/bftcosi/bftcosi.go
```diff
@@ -357,17 +338,22 @@ func (bft *ProtocolBFTCoSi) handleChallengePrepare(ch *ChallengePrepare) error {
...
 	}
-	return bft.SendToChildrenInParallel(ch)
+	return bft.SendToChildrenInParallel(&ch)
 }
 
 // handleChallengeCommit verifies the signature and checks if not more than
 // the threshold of participants refused to sign
-func (bft *ProtocolBFTCoSi) handleChallengeCommit(ch *ChallengeCommit) error {
+func (bft *ProtocolBFTCoSi) handleChallengeCommit(msg challengeCommitChan) error {
+	log.Lvl3(bft.Name(), "RoundType: 1")
```
Liamsi: same as with "RoundType: 0"

protocols/manage/broadcast.go
```diff
@@ -16,6 +18,7 @@ func init() {
 // using RegisterOnDone()
 type Broadcast struct {
 	*sda.TreeNodeInstance
+	sync.Mutex
```
Liamsi: This mutex only protects repliesLeft or did I miss something?

protocols/randhound/randhound.go
```diff
@@ -24,6 +26,7 @@ func init() {
 // sda.ProtocolInstance interface.
 type RandHound struct {
 	*sda.TreeNodeInstance
+	sync.Mutex
```
Liamsi: Is this mutex supposed to protect the whole RandHound struct rom modifications?

Liamsi:
Besides a few comments/questions: LGTM.

protocols/bftcosi/bftcosi.go
ineiti: As I wrote there - difficult.

protocols/bftcosi/bftcosi.go
ineiti: Yes, but to keep the signatures of the `start.*`-methods constant I kept the `error` here.

protocols/bftcosi/bftcosi.go
ineiti: That's the SDA-way of saying that it has to wait for all children before sending the messages.

protocols/bftcosi/bftcosi.go
ineiti: Again compatibility with `log.Lvl3(ann.TYPE)` which prints `0` and `1` instead of `Prepare` and `Commit`. Furthermore it's more easy to parse `0` and `1` in the brain than `Prepare` and `Commit` when going through logs with a lot of lines ;)

But now that it works I'll take away those messages (hoping I won't need them soon again ;)

protocols/bftcosi/bftcosi.go
Liamsi: Ah, I remember. Maybe we should change the SDA-way then. But that's not related to this issue.

ineiti:
Also included #514

ineiti:
And another one:
```
--- FAIL: TestReconnection (0.03s)
panic: runtime error: invalid memory address or nil pointer dereference
	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x38 pc=0x494b0f]
goroutine 97 [running]:
panic(0xac8a80, 0xc820010170)
	/home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/panic.go:464 +0x3ff
testing.tRunner.func1(0xc8203aa2d0)
	/home/travis/.gimme/versions/go1.6.linux.amd64/src/testing/testing.go:467 +0x22a
panic(0xac8a80, 0xc820010170)
	/home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/panic.go:426 +0x521
github.com/dedis/cothority/sda.(*Host).closeConnections(0xc820262900, 0x0, 0x0)
	/home/travis/gopath/src/github.com/dedis/cothority/sda/host.go:180 +0x2cf
github.com/dedis/cothority/sda.(*Host).Close(0xc820262900, 0x0, 0x0)
	/home/travis/gopath/src/github.com/dedis/cothority/sda/host.go:167 +0x508
panic(0xac8a80, 0xc820010170)
	/home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/panic.go:426 +0x521
github.com/dedis/cothority/sda.(*Host).SendRaw(0xc820262900, 0xc8203bcf40, 0x9b17a0, 0xc820401870, 0x0, 0x0)
	/home/travis/gopath/src/github.com/dedis/cothority/sda/host.go:224 +0x578
github.com/dedis/cothority/sda.sendrcv(0xc820262900, 0xc8202629c0, 0x0, 0x0)
	/home/travis/gopath/src/github.com/dedis/cothority/sda/host_test.go:413 +0xd1
github.com/dedis/cothority/sda.TestReconnection(0xc8203aa2d0)
	/home/travis/gopath/src/github.com/dedis/cothority/sda/host_test.go:409 +0xbed
testing.tRunner(0xc8203aa2d0, 0xe96c00)
	/home/travis/.gimme/versions/go1.6.linux.amd64/src/testing/testing.go:473 +0xdd
created by testing.RunTests
	/home/travis/.gimme/versions/go1.6.linux.amd64/src/testing/testing.go:582 +0xae3
FAIL	github.com/dedis/cothority/sda	4.994s
?   	github.com/dedis/cothority/services	[no test files]
```

ineiti:
Passed 15 Travis-tests. Removed the `TestReconnection`.

Makefile
```diff
@@ -28,10 +28,10 @@ test_lint:
 # for more than once in Travis. Change `make test` in .travis.yml
 # to `make test_multi`.
 test_multi:
-	cd network; \
+	cd services/skipchain; \
```
nikkolasg: ? Why testing skipchain only here ? either test everything (or only services) or change the name to `test_skipchain` if that's really needed (I think not).

network/encoding.go
```diff
@@ -36,7 +36,8 @@ var ErrorType = PacketTypeID(uuid.Nil)
 
 // String returns the canonical string representation of the PacketTypeID
 func (mId PacketTypeID) String() string {
-	return uuid.UUID(mId).String()
+	t, _ := registry.get(mId)
```
nikkolasg: It was for testing, so here we should be more careful, namely check the `ok` variable to see if it is present or not. If it's not present, we should just print the hexadecimal version then.

protocols/bftcosi/bftcosi.go
```diff
@@ -155,43 +151,59 @@ func (bft *ProtocolBFTCoSi) Start() error {
...
-			err = bft.startResponse(msg.Response.TYPE, &msg.Response)
-		case <-bft.doneProcessing:
-			// we are done
-			log.Lvl2(bft.Name(), "BFTCoSi Dispatches stop.")
+	bft.closingMutex.Lock()
+	if bft.closing {
+		return nil
+	}
+	// Close unused channels for the leaf nodes. This is not easily
+	// possible for the root-node.
```
nikkolasg: What does it mean `not easily possible for the root-node` ?
BTW, Golang takes care of that alone, we don't have to close necessarily a channel just because *it is not used anymore* (i.e. it's not like a file descriptor).
So do we need that ?

protocols/bftcosi/bftcosi.go
```diff
@@ -228,125 +240,94 @@ func (bft *ProtocolBFTCoSi) RegisterOnSignatureDone(fn func(*BFTSignature)) {
...
 }
 
-// Shutdown will close the dispatch-method if the protocol is stopped
-// before the normal termination.
+// Shutdown closes all channels in case we're done
 func (bft *ProtocolBFTCoSi) Shutdown() error {
-	bft.doneProcessing <- true
+	defer func() {
+		// In case the channels were already closed
+		recover()
```
nikkolasg: Again, we don't need to close channels if it's not to terminate a `select`, so can we drop this recover here and the closing a few lines below ? 
Having a `recover()` might hide other panicking stuff...

Makefile
ineiti: it's the `test_multi`-target which is used for whatever tests need to be done. It is normal for this target to change whenever somebody needs to run travis on one test only.

nikkolasg: Then instead of changing the makefile we can just use a environment variable [0] in the `.travis.yml` and set up the path using that variable, this won't affect all PR where we want to change that.

[0] https://docs.travis-ci.com/user/environment-variables/

protocols/bftcosi/bftcosi.go
```diff
@@ -228,125 +240,94 @@ func (bft *ProtocolBFTCoSi) RegisterOnSignatureDone(fn func(*BFTSignature)) {
...
-	log.Lvl4("RoundType:", t)
-	return bft.SendToChildrenInParallel(a)
-}
-
 // handleAnnouncement passes the announcement to the right CoSi struct.
-func (bft *ProtocolBFTCoSi) handleAnnouncement(ann Announce) error {
-	log.Lvl4(bft.Name(), "RoundType:", ann.TYPE)
+func (bft *ProtocolBFTCoSi) handleAnnouncement(msg announceChan) error {
+	ann := msg.Announce
+	if bft.closing {
```
nikkolasg: mhhh no need of synchronization here ? Just asking..

protocols/bftcosi/bftcosi.go
ineiti: Updated comment.

protocols/bftcosi/bftcosi.go
ineiti: Here we only close the channels. The `panic` will come for the channels that have been closed in `Dispatch` (look at the new comment there).

protocols/bftcosi/bftcosi.go
ineiti: of course - added `isClosing` and `setClosing`.

protocols/bftcosi/bftcosi.go
ineiti: OK, added a `if bft.IsLeaf` instead of the `panic`.

Makefile
ineiti: possible, but the `test_multi` I used had sometimes just these four lines, sometimes 2 or 3 times 4 lines, there were lines with `-v`, ... It's just a placeholder so I don't have to spend hours (or even days) re-writing those 4 lines.

nikkolasg: Travis Env Var would still solve the problem elegantly but for this PR then we should at least rename it something like `test_playground` or stg explicit, with an appropriate comment as you just described above.

nikkolasg:
After the lasts comments, I'm OK to merge.

ineiti:
I started travis again because of some non-trivial changes - `isClosing` and the removal of `panic`... If it holds 10 times, it should be OK.

ineiti:
OK, it did 10 rounds without error.

Issue #515, opened 2016-08-11
Test skipchain and identity


ineiti:
Now that BFTCoSi is running with (nearly) no errors, check skipchains and identity.

ineiti:
More errors:

```
E : (                       sda.(*LocalTest).CloseAll: 140) - Closing host localhost:52760 gives error close tcp [:
:]:52760: use of closed network connection
2 : (                         bftcosi.verifyRefuseBit: 330) - Refusing for myBit == 0
2 : (                         bftcosi.verifyRefuseBit: 330) - Refusing for myBit == 1
2 : (bftcosi.(*ProtocolBFTCoSi).handleResponsePrepare: 490) - &{f74b3067-c69e-4ce0-8cc6-51010d34ba61 [[localhost:52
764] [localhost:52779] [localhost:52802]] d1359fb9b91ffef7c952d10c4693930560feb4e85f51ff09d4915905c81d0cd1} Refused
 to sign
2 : (bftcosi.(*ProtocolBFTCoSi).handleResponsePrepare: 490) - &{f74b3067-c69e-4ce0-8cc6-51010d34ba61 [[localhost:52
764] [localhost:52779] [localhost:52802]] d1359fb9b91ffef7c952d10c4693930560feb4e85f51ff09d4915905c81d0cd1} Refused
 to sign
2 : (                       sda.(*Overlay).nodeDelete: 420) - Node 1602efe8d31e51dbb7b531e9e7deb79a already gone
--- PASS: TestCheckRefuseBit (2.25s)
```
...
```
--- PASS: TestCheckRefuseParallel (2.03s)
PASS
F : (                                   log.AfterTest:  64) - 1 instances of:
github.com/dedis/cothority/network.(*TCPHost).listen(0xc8200f6840, 0x63d9a0, 0xb, 0xc8203d4800, 0x0, 0x0)
        /Users/altcode/Documents/go/src/github.com/dedis/cothority/network/net.go:205 +0x87d
github.com/dedis/cothority/network.(*SecureTCPHost).Listen.func2(0xc820269730, 0xc8200f6900, 0xc820269750, 0xc8203d
4800)
        /Users/altcode/Documents/go/src/github.com/dedis/cothority/network/net.go:269 +0x74
created by github.com/dedis/cothority/network.(*SecureTCPHost).Listen
        /Users/altcode/Documents/go/src/github.com/dedis/cothority/network/net.go:275 +0x781

[...]

PR #487, opened 2016-07-20
Hpets16

ineiti:
Added docker-generation and the scripts needed to reproduce the hotpets16-demo

PR #516, opened 2016-08-11
Skipchain identity 515

ineiti:
Implements #515 - Searching errors in skipchain and identity service

Also merges #481 and #487


ineiti:
Besides two manually aborted tests (due to a calculated run-time of 3h), I can't reproduce any error. Proposing it for inclusion in the code.

ineiti:
If you merge #511, I'll merge this back into here so there will be less commits.

PR #518, opened 2016-08-16
implemented 517

ineiti:
Waiting for Travis...

### DIFF
--- a/crypto/hash.go
+++ b/crypto/hash.go
@@ -123,7 +123,8 @@ func ConvertToBinaryMarshaler(args ...interface{}) ([]encoding.BinaryMarshaler, 
 	var ret []encoding.BinaryMarshaler
 	for _, a := range args {
 		refl := reflect.ValueOf(a)
-		if refl.Kind() == reflect.Slice {
+		switch refl.Kind() {
+		case reflect.Slice, reflect.Array:
 			for b := 0; b < refl.Len(); b++ {
 				el := refl.Index(b)
 				bms, err := ConvertToBinaryMarshaler(el.Interface())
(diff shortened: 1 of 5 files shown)


crypto/hash.go
```diff
@@ -108,6 +113,32 @@ func HashArgs(hash h.Hash, args ...encoding.BinaryMarshaler) ([]byte, error) {
...
 	return HashArgs(suite.Hash(), args...)
 }
+
+// ConvertToBinaryMarshaler takes a slice of interfaces and returns
+// a slice of BinaryMarshalers.
+func ConvertToBinaryMarshaler(args ...interface{}) ([]encoding.BinaryMarshaler, error) {
+	var ret []encoding.BinaryMarshaler
+	for _, a := range args {
+		refl := reflect.ValueOf(a)
+		if refl.Kind() == reflect.Slice {
```
nikkolasg: Should have a if for Array also no ?

ineiti: Hmm - probably yes.

crypto/hash.go
```diff
@@ -108,6 +113,32 @@ func HashArgs(hash h.Hash, args ...encoding.BinaryMarshaler) ([]byte, error) {
...
 
 // HashArgsSuite makes a new hash from the suite and calls HashArgs
-func HashArgsSuite(suite abstract.Suite, args ...encoding.BinaryMarshaler) ([]byte, error) {
+func HashArgsSuite(suite abstract.Suite, args ...interface{}) ([]byte, error) {
 	return HashArgs(suite.Hash(), args...)
 }
+
+// ConvertToBinaryMarshaler takes a slice of interfaces and returns
+// a slice of BinaryMarshalers.
+func ConvertToBinaryMarshaler(args ...interface{}) ([]encoding.BinaryMarshaler, error) {
```
nikkolasg: Wondering if we should let that exported or not. I don't see any reasons why it should be exported for the moment.

ineiti: for the moment it's only exported so that I can use it in the `crypto_test`...

nikkolasg: Damn, good thing you talk about it. Can we go back to `package crypto` in test ? It's a *unit test* here, I see no reason to put it externally.

ineiti: OK - I'll do that first thing tomorrow

ineiti:
Changed the tests to be unit-tests.

nikkolasg:
OK to merge.

Issue #341, opened 2016-04-05
RandHound Refactoring


Daeinar:
Implement the new 2-round version of RandHound that uses Public Verifiable Secret Sharing (PVSS).

Daeinar:
I think it's ready to be merged. At some later point we should probably extract the PVSS functionality and put it into the crypto library.

Liamsi:
Merged in #519

Issue #321, opened 2016-03-31
RandHound Transcript Generation/Verification


Daeinar:
This is just a reminder that we do not forget to implement transcript generation and verification in RandHound later on.

Liamsi:
Merged in #519

PR #519, opened 2016-08-16
RandHound 341

Daeinar:
Addresses  #341 and #321.


protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+	for i := 0; i < n; i++ {
+		base[i] = ProofBase{g: g[i], h: h[i]}
+	}
+
+	return &Proof{suite: suite, base: base, core: core}, nil
+}
+
+// Setup initializes the proof by randomly selecting a commitment v,
+// determining the challenge c = H(xG,xH,vG,vH), and the response r = v - cx.
+func (p *Proof) Setup(scalar ...abstract.Scalar) ([]abstract.Point, []abstract.Point, []ProofCore, error) {
```
ineiti: Would it make sense to put the `[]abstract.Point` together in a structure?

protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+
+// Verify validates the proof against the given input by checking that
+// vG == rG + c(xG) and vH == rH + c(xH).
+func (p *Proof) Verify(xG []abstract.Point, xH []abstract.Point) ([]int, []int, error) {
+
+	if len(xG) != len(xH) {
+		return nil, nil, errors.New("Received unexpected number of points")
+	}
+
+	var good, bad []int
```
ineiti: The good, the bad, and the ugly.

protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+	for i := 0; i < len(p.base); i++ {
```
ineiti: `for i, core := range p.core`
and replace `p.core[i]` with `core`

protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+// Split creates PVSS shares encrypted by the public keys in X and
+// provides a NIZK encryption consistency proof for each share.
+func (pv *PVSS) Split(X []abstract.Point, secret abstract.Scalar) ([]int, []abstract.Point, []ProofCore, []byte, error) {
+
+	n := len(X)
+
+	// Create secret sharing polynomial
+	p := new(poly.PriPoly).Pick(pv.suite, pv.t, secret, random.Stream)
```
ineiti: Possibility of renaming `p`, `s` and `P` to `poly`, `shares`, `commit`, so you don't need comments.

protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+	s := new(poly.PriShares).Split(p, n)
+
+	// Create public polynomial commitments with respect to basis H
+	P := new(poly.PubPoly).Commit(p, pv.h)
+
+	// Prepare data for encryption consistency proofs ...
+	share := make([]abstract.Scalar, n)
+	H := make([]abstract.Point, n)
+	idx := make([]int, n)
+	for i := 0; i < n; i++ {
```
ineiti: `for i := range idx`

protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+// Commits reconstructs a list of commits from the given polynomials and indices.
+func (pv *PVSS) Commits(polyBin [][]byte, index []int) ([]abstract.Point, error) {
+
+	if len(polyBin) != len(index) {
+		return nil, errors.New("Inputs have different lengths")
+	}
+
+	n := len(polyBin)
+	sH := make([]abstract.Point, n)
+	for i := 0; i < n; i++ {
```
ineiti: `for i := range sH`

protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+	if len(S) < pv.t {
+		return nil, errors.New("Not enough shares to recover secret")
+	}
+
+	//log.Lvlf1("%v %v %v %v", pos, pv.t, len(pos), len(S))
+
+	pp := new(poly.PubPoly).InitNull(pv.suite, pv.t, pv.suite.Point().Base())
+	ps := new(poly.PubShares).Split(pp, n) // XXX: ackward way to init shares
+
+	for i := 0; i < len(S); i++ {
```
ineiti:
```
for i, s := range S{
  ps.SetShare(pos[i], s)
}
```

protocols/randhound/proof_test.go
```diff
@@ -0,0 +1,190 @@
...
+	h2, _ := suite.Point().Pick([]byte("H2"), random.Stream)
+
+	// 2nd secret value
+	y := suite.Scalar().Pick(random.Stream)
+
+	// Create proofs
+	g := []abstract.Point{g1, g2}
+	h := []abstract.Point{h1, h2}
+	p, err := randhound.NewProof(suite, g, h, nil)
+	if err != nil {
```
ineiti: Possibility of using
`log.ErrFatal(err)`
which `fatal`s if `err != nil`
or use
`require.Nil(t, err)`

protocols/randhound/proof_test.go
```diff
@@ -0,0 +1,190 @@
...
+		t.Fatal(err)
+	}
+
+	_, bad, err := q.Verify(xG, xH)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(bad) != 0 {
+		t.Fatal(fmt.Errorf("Some proofs failed: %v", bad))
```
ineiti: `log.Fatalf("Some proofs failed: %v", bad)`

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,1006 @@
...
+		for j, server0 := range group {
+			s0 := server0.ServerIdentityIdx
+			rh.ServerIdxToGroupNum[s0] = i
+			rh.ServerIdxToGroupIdx[s0] = j
+			g[j] = s0
+		}
+		rh.group[i] = g
+	}
+
+	// Comptue session id
```
ineiti: s/Comptue/Compute/

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,1006 @@
...
+			Threshold: rh.threshold[i],
+			Group:     index,
+			Key:       rh.key[i],
+		}
+
+		rh.mutex.Lock()
+
+		// Sign I1 and store signature in i1.Sig
+		if err := signSchnorr(rh.Suite(), rh.Private(), i1); err != nil {
+			return err
```
ineiti: mutex will be held here. Add `Unlock`

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,1006 @@
...
+				for k := 0; k < len(r2.DecShare); k++ {
+					if r2.DecShare[k].Source == src {
+						decPos = append(decPos, r2.DecShare[k].Pos)
+						decShare = append(decShare, r2.DecShare[k].Val)
+						decProof = append(decProof, r2.DecShare[k].Proof)
+					}
+				}
+			}
+
+			//log.Lvlf1("Env vs Dec (Pre-Sync):  %v %v %v %v %v %v %v", encPos, decPos, len(X), len(encShare), len(encProof), len(decShare), len(decProof))
```
ineiti: Is this still used?

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,1006 @@
...
+			if len(decPos) < len(encPos) {
+				l := len(decPos)
+				poly = poly[:l]
+				encPos = encPos[:l]
+				encShare = encShare[:l]
+				encProof = encProof[:l]
+				X = X[:l]
+			}
+
+			//log.Lvlf1("Env vs Dec (Post-Sync): %v %v %v %v %v %v %v", encPos, decPos, len(X), len(encShare), len(encProof), len(decShare), len(decProof))
```
ineiti: Still used? Can also write `log.Lvlf3` or so.

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,1006 @@
...
+
+			// Check encryption consistency proofs
+			goodEnc, badEnc, err := pvss.Verify(H, X, polyCommit, encShare, encProof)
+			if err != nil {
+				return err
+			}
+			_ = goodEnc
+			_ = badEnc
+
+			//log.Lvlf1("Enc: %v %v %v %v %v %v", goodEnc, badEnc, len(X), len(encShare), len(decShare), len(decProof))
```
ineiti: Still used?

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,1006 @@
...
+			_ = badDec
+
+			// Remove bad shares
+			for j := len(badDec) - 1; j >= 0; j-- {
+				k := badDec[j]
+				decPos = append(decPos[:k], decPos[k+1:]...)
+				decShare = append(decShare[:k], decShare[k+1:]...)
+			}
+
+			//log.Lvlf1("Dec: %v %v", goodDec, badDec)
```
ineiti: Is this still used?

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,1006 @@
...
+
+			// Recover secret and add it to the collective random point
+			ps, err := pvss.Recover(decPos, decShare, len(t.Group[i]))
+			if err != nil {
+				return err
+			}
+			rnd = rh.Suite().Point().Add(rnd, ps)
+			//log.Lvlf1("Transcript: %v %v", src, ps)
```
ineiti: Is this still used?

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,1006 @@
...
+	good, _, err := pvss.Verify(H, rh.key[grp], polyCommit, encShare, encProof)
 	if err != nil {
-		return nil, nil, err
+		return err
 	}
-	if err = binary.Write(buf, binary.LittleEndian, tm); err != nil {
-		return nil, nil, err
+
+	// Record valid encrypted shares per secret/server
+	for i := 0; i < len(good); i++ {
```
ineiti: `for i, g := range good`
and replace `good[i]` with `g`.

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,1006 @@
...
+			prng := rh.Suite().Cipher(rand)
+			secret := goodSecret[i]
+			for j := 0; j < len(secret)-rh.threshold[i]; j++ {
+				k := int(random.Uint32(prng) % uint32(len(secret)))
+				secret = append(secret[:k], secret[k+1:]...)
+			}
+			rh.chosenSecret[i] = secret
+		}
+
+		//log.Lvlf1("Grouping: %v", rh.group)
```
ineiti: Still used? Possible to use `log.Lvlf3`?

protocols/randhound/proof.go
Daeinar: I am not sure. I tried to keep it as simple as possible and I am not sure if you get any benefit of having just `[]abstract.Point` in a structure. You would probably have to put all the return values into a structure. I guess that's what you mean? Iirc I tried that a while ago and it complicated quite a few things in the other code that uses PVSS. I'll think about it again.

nikkolasg: > You would probably have to put all the return values into a structure.

Yes, that'd be nice if possible. If you remember which part posed problems, don't hesitate to drop it to us to see if something's possible.

protocols/randhound/proof.go
Daeinar: You probably mean `for i := range p.base {...}`?

Daeinar:
@ineiti: Addressed most of your comments. I did not change the format of the return values in `Setup(...)`. I guess we can do that when we resolve #624.

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,995 @@
...
+		return nil, nil, fmt.Errorf("Number of requested shards not supported")
+	}
+
+	// Compute a random permutation of [1,n-1]
+	prng := rh.Suite().Cipher(seed)
+	m := make([]int, rh.nodes-1)
+	for i := range m {
+		j := int(random.Uint64(prng) % uint64(i+1))
+		m[i] = m[j]
+		m[j] = i + 1
```
nikkolasg: I don't know where this comes from but it deviates a bit from the "Fisher Yates " shuffle then derived by Durstenfeld for CS ( see [link](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_.22inside-out.22_algorithm) ).
The part that seems missing is the `if i != j { m[i] = m[j] }`. Of course I haven't checked out the recursion proof of the wikipedia algo so I would have to spend more time to check if this impl. is correct. 
How did you came up with this implementation ?

Daeinar: It's basically an adaptation of Golang's `rand.Perm(...)` function, see here: https://golang.org/src/math/rand/rand.go?s=5071:5103#L164

Daeinar: Ah, yes that makes sense!

nikkolasg:
Ok this is a huge PR and I'm not sure to be familiar enough with all the crypto inside to be a really good reviewer for this one. I'd need to read the paper on JVSS and the one you're writing to get a full grasp. I'm planning on this but now now so on the surface, to me the code looks good :)

On a more structural level, I see you coded PVSS yourself in the same package. Would it make sense now / later to decouple that from the randhound code and to integrate that in the crypto library ? I think it would be a real good target for the crypto library, the same way we have jvss & cosi integrated in the crypto library. Of course these were introduced after we coded the protocol so it does not have to be now. What do you think ?

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,995 @@
...
+	var err error
+
+	// Set timestamp
+	rh.time = time.Now()
+
+	// Choose client randomness
+	hs := rh.Suite().Hash().Size()
+	rand := make([]byte, hs)
+	random.Stream.XORKeyStream(rand, rand)
+	rh.cliRand = rand
```
Daeinar: Ah thanks I didn't know about this function in dedis/crypto, I will of course use that instead.

protocols/randhound/randhound.go
```diff
@@ -1,251 +1,995 @@
...
+			}
+			r1 := t.R1s[src]
+
+		OUTER:
+			for j := 0; j < len(r1.EncShare); j++ {
+
+				// Check availability of corresponding R2 messages, skip if not there
+				target := r1.EncShare[j].Target
+				if _, ok := t.R2s[target]; !ok {
+					continue OUTER
```
Daeinar: Note that the r1 message changes the next time you go to `OUTER`, see https://github.com/dedis/cothority/blob/randhound_341/protocols/randhound/randhound.go#L366.

Daeinar:
@nikkolasg: Fully agree on the PVSS part. I've already added an issue on that topic a couple of days ago, see #624. I am not sure how urgent that is though and we maybe first want to focus on other things?

protocols/randhound/randhound.go
Daeinar: @nikkolasg: Okay leaving out the `OUTER` also works. ;)

crypto/hash.go
```diff
@@ -118,22 +118,21 @@ func HashArgsSuite(suite abstract.Suite, args ...interface{}) ([]byte, error) {
 }
 
 // ConvertToBinaryMarshaler takes a slice of interfaces and returns
-// a slice of BinaryMarshalers.
-func convertToBinaryMarshaler(args ...interface{}) ([]encoding.BinaryMarshaler, error) {
+// a slice of BinaryMarshalers
+func ConvertToBinaryMarshaler(args ...interface{}) ([]encoding.BinaryMarshaler, error) {
```
Liamsi: Will this be used externally (would make sense). If not, I think this should remain un-exported.

protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+		r.Mul(x, c).Sub(v[i], r)
+		p.core[i] = ProofCore{c, r, vG[i], vH[i]}
+	}
+
+	return xG, xH, p.core, nil
+}
+
+// Verify validates the proof against the given input by checking that
+// vG == rG + c(xG) and vH == rH + c(xH).
+func (p *Proof) Verify(xG []abstract.Point, xH []abstract.Point) ([]int, []int, error) {
```
Liamsi: I haven't read the paper of the algo you are implementing here. But this `Verify` function looks like something an external dev/PhD student would use. It would be great if was clearer for the caller when this `Proof` is valid or not. From looking into the implementation and the naming of the local vars (`good`, `bad`) it looks like the proof is valid iff the second returned array is empty? (OK, looking into the tests clarifies this). But you could either use named vars or if you prefer comments, you could add a short comment, when this `Proof` is a valid one or not.

protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+	polyBin, err := pubPoly.MarshalBinary()
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+
+	return idx, sX, encProof, polyBin, nil
+}
+
+// Verify checks that log_H(sH) == log_X(sX) using the given proof.
+func (pv *PVSS) Verify(H abstract.Point, X []abstract.Point, sH []abstract.Point, sX []abstract.Point, core []ProofCore) ([]int, []int, error) {
```
Liamsi: Same comment as https://github.com/dedis/cothority/pull/519/files#r82609495

protocols/randhound/randhound.go
Liamsi: I think @ineiti meant the mutex won't be released in the error case. Not sure if this is problematic but to be sure you could add a `Unlock` in this branch:
```
if err := signSchnorr(rh.Suite(), rh.Private(), i1); err != nil {
          rh.mutex.Unlock()
          return err
}
```

Liamsi:
Besides a few comments this PR looks good to me. Feel free to decide which are relevant to you (and which not).

protocols/randhound/randhound.go
Daeinar: Ah of course, that makes sense. Thanks for clarifying!

protocols/randhound/proof.go
Daeinar: Yes good point, I'll fix the comments! Since we are doing threshold crypto here, it is allowed that some proofs fail, which is also the reason for the `good` and `bad` return values.

protocols/randhound/struct.go
```diff
@@ -0,0 +1,137 @@
...
+	faulty  int       // Maximum number of Byzantine servers
+	purpose string    // Purpose of protocol run
+	time    time.Time // Timestamp of initiation
+	cliRand []byte    // Client-chosen randomness (for initial sharding)
+	sid     []byte    // Session identifier
+
+	// Group information
+	server              [][]*sda.TreeNode  // Grouped servers
+	group               [][]int            // Grouped server indices
+	threshold           []int              // Group thresholds
```
Daeinar: I am usually sticking to the singular, so that when you iterate over the variable you have `threshold[i]` instead of `thresholds[i]` which is more intuitive to me. In the above example the same holds for `server[i][j]` etc. I should probably check if there are some Golang coding conventions on that.

crypto/hash.go
Daeinar: We will move all of the stuff from `crypto/hash.go` to the crypto library at some point. So I this can be addressed later.

protocols/randhound/struct.go
Liamsi: I think the most idiomatic golang way would be to iterate using `range` instead. Like here: https://play.golang.org/p/HtZlb8Ctav
But as I said, this was just nitpicking ;-)

Daeinar: Hmm right. I just get used to that. ;)

protocols/randhound/proof.go
```diff
@@ -0,0 +1,298 @@
...
+		c := p.suite.Scalar().Pick(p.suite.Cipher(cb))
+
+		// Response
+		r := p.suite.Scalar()
+		r.Mul(x, c).Sub(v, r)
+
+		p.core[i] = ProofCore{c, r, vG, vH}
+	}
+
+	return xG, xH, p.core, nil
```
Liamsi: > Would it be better coding style to have a "getter"?

I think this depends on your preferences. As `p.core` is an array, the returned object (in any case) can be manipulated. So you could as well just make `core` a public/exported field:
See here: https://play.golang.org/p/pkZHSVzSgk

Liamsi: And the reason I suggested that isn't about the getter, it's just to make the code more readable by reducing the number of returned objects (as the caller could have access to `core` by naming it `Core` without returning it and the provided "safety" would be the same ). Mainly because the others complained about the number of args. But if you don't like it, just keep it like this. It's just a matter of style (and I'm sure many people could have vim vs emacs like discussion about it ;-)

protocols/randhound/proof_test.go
Liamsi: @Daeinar If you use `log.ErrFatal(err)` as @ineiti suggested, you don't need to check `err != nil` (because it's done in [`ErrFatal`](https://github.com/dedis/cothority/blob/08796f10dfa5b418c0deae99bcb12fac08f0d846/log/ui.go#L83)). Personally, I don't care much about the code style in the tests...

protocols/randhound/struct.go
ineiti: +1 for `range` ;)

Issue #474, opened 2016-07-13
Implement followers


ineiti:
add some commands:
`cisc follow id` - starting to follow an identity-skipchain
`cisc unfollow id` - don't follow that skipchain anymore
`cisc follow update` - update all followed skipchains

Liamsi:
Is this ready for merge? The PR is this #481, right?

ineiti:
yes, I think it's OK. Besides the BFTCoSi-error...

ineiti:
Done in #511

Issue #491, opened 2016-07-21
Add load/save methods to Context


ineiti:
Add the following methods to `onet.Context` (see `identity/service.go`)
- [ ] `Save(name string, data interface{}) error`
- [ ] `Load(name string)(interface{} error)`
- [ ] `Path()string`

Others
- [ ] remove the `path` from the arguments in `onet.NewServiceFunc`
- [ ] use `Save` and `Load` in `services/*`

PR #529, opened 2016-08-18
Test skipchain and identity

ineiti:
This implements #515 and #474 .

Specifically it does:
- implement the `follow`-argument to `cisc` to set up a follower and update the `authorized_keys`-file
- implement removing of a device on the identity-skipchain
- save and load the `skipchain`s and the `identity`es
- adding of a docker-file
- hpets16 demo-scripts
- bugfixes


app/cisc/cisc.go
```diff
@@ -97,16 +103,29 @@ func idConnect(c *cli.Context) error {
...
+func idDel(c *cli.Context) error {
+	if c.NArg() == 0 {
+		log.Fatal("Please give device to delete")
+	}
+	cfg := loadConfigOrFail(c)
+	dev := c.Args().First()
+	if _, ok := cfg.Config.Device[dev]; !ok {
+		log.Info("Didn't find", dev, "in config")
+		configList(c)
+		log.Fatal("Try with one of those")
```
nikkolasg: what's up with that ?

Liamsi:
> Rebased on master - but loosing all comments :(

Please, find a way to use github without deleting all comments. The whole review history is gone :-( In the old Pull #516 they are all gone. We need to find out if changing the base branch of a PR deletes all comments...

app/cisc/cisc.go
```diff
@@ -290,8 +324,66 @@ func sshSync(c *cli.Context) error {
...
+	idBytes, err := hex.DecodeString(c.Args().Get(1))
+	log.ErrFatal(err)
+	id := identity.ID(idBytes)
+	newID, err := identity.NewIdentityFromCothority(group.Roster, id)
+	log.ErrFatal(err)
+	if c.NArg() == 3 {
+		newID.DeviceName = c.Args().Get(2)
+	} else {
+		var err error
+		newID.DeviceName, err = os.Hostname()
```
nikkolasg: Maybe print up something saying that this is the name that cisc will use to follow that skipchain.

app/cisc/cisc.go
```diff
@@ -290,8 +324,66 @@ func sshSync(c *cli.Context) error {
...
+		log.ErrFatal(err)
+	}
+	cfg.Follow = append(cfg.Follow, newID)
+	cfg.writeAuthorizedKeys(c)
```
nikkolasg: Is there a big warning somewhere saying that using cisc for ssh key mgmt will write into your authroized_key file ? If not, there should be in my opinion, because this is not something to take lightly: `Oh, I'll just add some keys to your authorized_key file, don't worry !`  I know it's not necessarily related to this PR but still..

app/cisc/cisc.go
```diff
@@ -290,8 +324,66 @@ func sshSync(c *cli.Context) error {
...
+	if c.NArg() != 1 {
+		log.Fatal("Please give id of skipchain to unfollow")
+	}
+	cfg := loadConfigOrFail(c)
+	idBytes, err := hex.DecodeString(c.Args().First())
+	log.ErrFatal(err)
+	idDel := identity.ID(idBytes)
+	var newSlice []*identity.Identity
+	for _, id := range cfg.Follow {
+		if !bytes.Equal(id.ID, idDel) {
```
nikkolasg: Nitpick and because it's cool: https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating

app/cisc/cisc.go
```diff
@@ -290,8 +324,66 @@ func sshSync(c *cli.Context) error {
...
+	cfg.writeAuthorizedKeys(c)
+	return cfg.saveConfig(c)
+}
+func followList(c *cli.Context) error {
+	cfg := loadConfigOrFail(c)
+	for _, id := range cfg.Follow {
+		log.Infof("Following ID: %x", id.ID)
+		log.Infof("As service %s getting ssh-keys from %s:",
+			id.DeviceName,
+			id.Config.GetIntKeys("ssh", id.DeviceName))
```
nikkolasg: > "As service %s getting ssh-keys from %s:"

* What does it mean I don't get it... :s Printing `id.DeviceName` as a `service` ?
* `GetIntKeys` what does it do ? 
* TO BE DISCUSSED: maybe we should print out stuff in more machine way to be easily *parsable*, between full english sentences and bytecode :D Like an array of information separated by a well-knwon character or something. I'm not sure but there may be a lot of information and it's better looking and more easily processable if it's displayed in a ... "straight way."

app/cisc/cisc.go
ineiti: and if it's `log.Error` + `configList` + `log.Fatal`?

app/cisc/cisc.go
ineiti: I now make a `.back` and write a `log.Info` about the backup-file and the fact that the `.ssh/authorized_key` is getting rewritten.

nikkolasg:
As a general remark: you say this implements #515 but in #515 the only thing was to test if skipchain + identity services worked (i.e. bug free). Here you add a whole bunch of new features it seems.
What are the features  you implement here ? (and maybe a why?) 
Thanks.

app/cisc/cisc.go
ineiti: 1. it prints out all stored keys for all followers. I'll replace `As service` with `Follower`.
2. `Get.*Keys` handles `a:b:c`-keys with regard to the first, middle or last element. It's for handling the key/value 'database' of the identity-structure.
3. the `kvList` method (using argument `kv list`) does a more parser-friendly output. If you think it's worth it, I can do a JSON-compatible output in `kvList`. Or add a `--json` flag to the output-methods so that they print it JSON-compatible.

Liamsi:
> What are the features you implement here ? (and maybe a why?) 

Even better: split this PR into several PRs (and issues explaining what to achieve and why) which can be reviewed separately. 

Here is a blog post with some pointers: https://medium.com/@kurtisnusbaum/large-diffs-are-hurting-your-ability-to-ship-e0b2b41e8acf#.5sg825338

But I think this is for future PRs. I already reviewed this one thoroughly. I just need to find out what additional changes were done. Still have the old branch locally, so I'll do a diff against that.

services/identity/service.go
```diff
@@ -143,35 +161,45 @@ func (s *Service) ProposeFetch(si *network.ServerIdentity, cnc *ProposeFetch) (n
...
-	}
-	if _, exists := sid.Votes[v.Signer]; exists {
-		return nil, errors.New("Already voted for that block")
-	}
-	log.Lvl3(v.Signer, "voted", v.Signature)
-	if v.Signature != nil {
-		err = crypto.VerifySchnorr(network.Suite, owner.Point, hash, *v.Signature)
+
+	// Putting this in a function because of the lock which needs to be held
+	// over all calls that might return an error.
```
Liamsi: Wouldn't a simple `defer sid.Unlock`do the same job? Why does the `sid` needs to be locked in this scope but some lines later where it is also read but not modified (like here: https://github.com/dedis/cothority/pull/529/files#diff-2433496499fd4c0770b52cb82bb9d4caL181) the Lock isn't used.

ineiti: Because if there is an error, I need to be able to return from the function. As there are a lot of returns, I'd have to write `sid.Unlock` every time, because I also need it to be unlocked later. Perhaps it would be more clear to have the `Lock` before the function and the `Unlock` after the function?

ineiti:
Updated the description of the PR.

services/identity/struct.go
```diff
@@ -221,7 +221,7 @@ type ProposeVote struct {
 
 // PropagateIdentity sends a new identity to other identityServices
 type PropagateIdentity struct {
-	*storage
+	*Storage
```
Liamsi: Question: Why is this exported? I didn't see a usage outside of the `identity`package.

ineiti: Because it will be sent over the network, so it needs to be `BinaryMarshal`lable.

services/identity/service.go
Liamsi: > As there are a lot of returns, I'd have to write sid.Unlock every time, because I also need it to be unlocked later. 

No, that's fine. That is what defer is for. I just wonder why it needs to be unlocked after this scope as you read similarly from `sid` as in the scope where you use the lock (like `sid.Proposed `, `sid.Votes` etc.).

nikkolasg:
> implement removing of a device on the identity-skipchain

What are the consequences of that ? How can you enforce that a "deleted" device can't follow again ? (not subscribe again, but really can't get update on the skipchain after)

> save and load the skipchains and the identityes

Do you enforce a path (Service) ? Or can the user choose ?

Liamsi:
You said that travis still fails (like after 40 runs or so). Are there open issues for all the failure cases you've found?

ineiti:
@nikkolasg - any device can follow the skipchain and update - that's the point of it. But it cannot propose and vote for new configurations.

Saving/loading: this is done through the `service`, so whatever the service gives as `path` will be taken by the skipchain/identity.

ineiti:
@Liamsi no issue, as the error-messages strongly point to the network-layer. But I'll re-run it in parallel again with the new layer.

ineiti:
@nikkolasg for saving and loading: #491 and #542

app/cisc/cisc.go
```diff
@@ -76,9 +80,11 @@ func idCreate(c *cli.Context) error {
 	}
 	log.Info("Creating new blockchain-identity for", name)
 
-	cfg := identity.NewIdentity(group.Roster, 2, name)
+	thr := c.Int("thr")
```
Liamsi: What does `thr` stand for?

ineiti: short for `threshold`...

Liamsi: You use `majority` and `threshold` interchangeably. Can we make this more consistent? In `NewIdentity` call the param threshold (then the above is also clear).

ineiti:
Is there anything left in this PR?

app/cisc/lib.go
```diff
@@ -11,50 +11,148 @@ import (
...
+}
+
+type ciscConfig struct {
+	*identity.Identity
+	Follow []*identity.Identity
+}
+
 // loadConfig will try to load the configuration and fail if it can't load it.
-func loadConfig(c *cli.Context) *identity.Identity {
+func loadConfig(c *cli.Context) (*ciscConfig, bool) {
```
Liamsi: In this case it might be useful if the return statements were named like `(cfg *ciscConfig, empty bool)` and just use return. One needs to search for callers to find out what this `bool` means.

app/docker/cothority/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+	vim \
+	sudo
+RUN adduser --gecos "" --ingroup sudo --disabled-password dedis; \
+	perl -pi -e "s/%sudo.*/%sudo ALL=NOPASSWD: ALL/" /etc/sudoers
+USER dedis
+WORKDIR /home/dedis
+RUN echo 'export GOPATH=/home/dedis/go' >> ~/.profile; \
+   	echo 'export PATH=$PATH:/usr/lib/go-1.6/bin:~/bin:$GOPATH/bin' >> ~/.profile; \
+   	echo 'export GOPATH=/home/dedis/go' >> ~/.bashrc; \
+    echo 'export PATH=$PATH:/usr/lib/go-1.6/bin:~/bin:$GOPATH/bin' >> ~/.bashrc; \
```
Liamsi: If you like you can change this to go1.7 which (among other improvements) brings some major speedup on some (standard) crypto routines.

app/cisc/lib.go
nikkolasg: Even better... Specifications ! In the comments !

app/docker/cothority/Dockerfile
Liamsi: On the other hand, maybe in this debian version it's still 1.6? Anyways, it's not important.

Liamsi:
> Is there anything left in this PR?

Some parts of the code only made sense after you repeatedly explained them on paper I'm not sure if we could find more such parts while spending (much) more time for the review (and also re-review the original cisc & skipchain changes). 
For instance: why is the hight of the skipchain always `10` (hardcoded in several places, but not a constant, or a config parameter). How do you guys (especially @nikkolasg) want to proceed with this? I think it's a very good prototype (and in a much better shape then the `byzcoin` package/protocol for instance). I would propose to merge it and open issues if we stumble upon glitches or even just have questions about the code (we already have a question label). Do you all agree?

nikkolasg:
After correcting the last two comments then I'm OK to merge !

app/docker/cothority/Dockerfile
ineiti: Debian-lenny-backports doesn't have golang-1.7 as of Firday 26 August.

Issue #530, opened 2016-08-18
Add a sda.ServiceProcessor.RegisterMessages


ineiti:
Like for channels and handlers in the protocols, add a method that does the for-loop automatically.

PR #531, opened 2016-08-18
implemented #530

ineiti:
Small change.

### DIFF
--- a/sda/processor_test.go
+++ b/sda/processor_test.go
@@ -103,6 +103,16 @@ func TestProcessor_AddMessage(t *testing.T) {
 	}
 }
 
+func TestProcessor_RegisterMessages(t *testing.T) {
+	h1 := newHostMock(network.Suite, "127.0.0.1")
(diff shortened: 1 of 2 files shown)


sda/processor_test.go
```diff
@@ -173,6 +183,14 @@ func procMsg(si *network.ServerIdentity, msg *testMsg) (network.Body, error) {
 	return msg, nil
 }
 
+func procMsg2(si *network.ServerIdentity, msg *testMsg) (network.Body, error) {
+	// Return an error for testing
+	if msg.I != 42 {
+		return nil, errors.New("6 * 9 != 42")
```
nikkolasg: There is already this "6 * 9 != 42" somewhere else I think... I know you're a fan of 42 but **if it's a duplicate** can you change so we can find easily where does the error comes from ?

ineiti: It's just above, in `procMsg`. I'll change it.

PR #534, opened 2016-08-22
removed v0-reference and Host.workingAddress with Host.Address()

ineiti:
Sorry, no Issue for that. Only a quickfix for a wrong import which made a test working that should've failed. Now `Host` doesn't have `workingAddress` anymore, as this was a duplicate of `Host.ServerIdentity.WorkingAddress`.


nikkolasg:
Have you tested it against app/ and `dedis/cosi`? I believe there might be some stuff that are broke with these changes...

ineiti:
Both `app` and `cosi` seem to run fine - with the exception of `test_cosi.sh` in `cothority`, which should not exist, and `test_cosi.sh` which throws

```
Testing Check
panic: runtime error: index out of range

goroutine 5 [running]:
panic(0x4c4020, 0xc82000e090)
	/usr/local/Cellar/go/1.6.2/libexec/src/runtime/panic.go:481 +0x3e6
github.com/dedis/cosi/service.(*Client).SignMsg(0xc820026060, 0xc82000b8c0, 0xc8200c1e40, 0x20, 0x20, 0x0, 0x0, 0x0)
	/Users/epfl/Documents/go/src/github.com/dedis/cosi/service/api.go:28 +0x70e
main.signStatement.func1(0xc820026060, 0xc82000b8c0, 0xc8200c1e40, 0x20, 0x20, 0xc8200d8290, 0xc82005c9c0)
	/Users/epfl/dedis/cosi/client.go:196 +0x135
created by main.signStatement
	/Users/epfl/dedis/cosi/client.go:203 +0x1a6

	FAILED: starting runCl 1 check failed
```

but I think this is independent of my change here...

nikkolasg:
OK to merge.

Issue #537, opened 2016-08-23
make cothority/test_integration.sh work


ineiti:
`cothority/test_integration.sh` doesn't run because of:
- `app/test_cosi.sh` is present -> remove
- `app/test_cothority.sh` does not use `dedis/cosi` in a correct way -> fix
- `dedis/cosi` doesn't use `exit(1)` in case of an error -> make cosi `exit(1)` if a server fails

PR #539, opened 2016-08-23
all integration-tests run (with the exception of test_cisc)

ineiti:
Implements #537 - depends on https://github.com/dedis/cosi/pull/54

### DIFF
--- a/app/lib/server/server.go
+++ b/app/lib/server/server.go
@@ -228,7 +228,6 @@ func CheckServers(g *config.Group) error {
 	}
 
 	if !success {
-		log.Error("at least one failed")
 		return errors.New("At least one of the tests failed")
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


Liamsi:
General remark: What I'm missing in this PR (or in the related issue #537) is a short description what wasn't working (and what is now; after this PR and https://github.com/dedis/cosi/pull/54). One can only guess from skimming through the changes.

app/lib/server/server.go
```diff
@@ -228,6 +228,7 @@ func CheckServers(g *config.Group) error {
 	}
 
 	if !success {
+		log.Error("at least one failed")
```
nikkolasg: Why logging ? If it's already printed at the screen for each failing line + a returning error at the end of this function...

ineiti:
@Liamsi - better description now?

nikkolasg:
OK to merge.

Liamsi:
> @Liamsi - better description now?

Yep, thanks!

Issue #461, opened 2016-06-29
make test doesn't work in symbolic link


ineiti:
If you're using symbolic links to the go-directories, `go vet` and `gofmt` fail...

ineiti:
Easy solution is to do

```
echo "alias cp='cp -P'" >> ~/.bash_profile
```

if you have MacOSX. Now if you open up a new terminal and `cd` into a directory that has a symbolic link in it, it will follow that link and make `make` happy.

Should we put `cp -P .` in the Makefile to do that automatically?

PR #546, opened 2016-08-24
cd -P . so gofmt and golint work also in linked directories

ineiti:
Implements solution to #461


Liamsi:
Closed for the same reasons as https://github.com/dedis/cosi/pull/55

PR #548, opened 2016-08-25
Add coveralls reporting

Liamsi:
Add goveralls/coveralls reporting (to keep track of test code coverage).
Resolves issue #549

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ env:
 
 script:
   - make test
-  - ./coveralls.sh
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,16 @@ go:
 
 install:
   - ./install.sh
+  - go get golang.org/x/tools/cmd/cover
+  - go get github.com/mattn/goveralls
+
+env:
+  global:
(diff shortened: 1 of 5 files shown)


nikkolasg:
Same for @ineiti ,there's no issue about it,  so let's discuss about test cover in an issue so you know what to do .. :)

Liamsi:
> Same for @ineiti ,there's no issue about it, so let's discuss about test cover in an issue so you know what to do .. :)

You can discuss here: #549

.travis.yml
```diff
@@ -5,9 +5,16 @@ go:
...
+  - go get golang.org/x/tools/cmd/cover
+  - go get github.com/mattn/goveralls
+
+env:
+  global:
+    secure: Qu8PbKlf0Q5LoVeiut3+vEgghyLdZmvgrLsW/wI4XKSOL49K2LetELsZA6298OCoCaXm2BbESsoMApLglDx4h4PAOUwm/Sf28kJKRzmA0WidTR3XnIvBDnQrW7bm/9KV+de4uW6HUi7j4jp93E4y3LhBjBiq4ayVxlTHMKgQFBD02QFA7namD3LF5xjGnU99y7RNw0QNBv7BJfuOllIBJaetXQsHAcGX2pQmyZK5uQy52hXe6lVivWPdRrtDUVJQ3Yb7ez80fnLw3+5cBU21Om3L8P88QfbPZu7BCbUb6uwqruHX8a+ZqeDXVe9zpzqhvW+FckQz8MCHkLVXM6BucAS1y//s7VtYOiw6uhU4vsmP6dPbh90yS325n310P43mZ+7JX4XOV41/33CCK/zG4HxsTlsLeH0MTWjDpIBHR+UVJGySYrLD4hj1/WLTFPcycbTb9RwatUFao80KksaLG1TEH7zp14hav8M6cgZcJooabmNlqvj6UvkAOU88ZdUnjt3Bu64dhZpd/9cXQ65NRLFaQAyt3gufFUtT31azrPhomAP+JcK77CqcGDGye/Nhl2BEeio/dqeoACwwZHDPbhveCDqWEtumtV0LVUKuIyj0/EFkXByZBcnpzecIl5F4BB3ihgckQP5cS/OLKAsCztNOoIwewm+5/aCjIKh+T2M=
 
 script:
   - make test
+  - ./coveralls.sh
```
nikkolasg: Is it possible to do it at the same time as `make test` ? So it's not doing the tests twice.

Liamsi: I don't mind to run the test twice but there are two possibilities to achieve this: 
1) Run only coveralls (which need to run anyways) with race enabled and replace `test_go` with `./coveralls.sh`. 2) Modify `test_go`in the Makefile to also run coveralls.
I vote for 1) because the shell code in the makefile is a pain to modify (all the `; \`,  `$$`, etc)

nikkolasg: +1 for option 1

Liamsi: Thanks for the feedback. Done in https://github.com/dedis/cothority/pull/548/commits/a9e1eb41c5dd3369bb210b4134c74e084d2a870e

Liamsi: Changed to `-mode=atomic` because of https://github.com/golang/go/issues/12118

nikkolasg:
I'm super OK to merge, let's get that number up...

ineiti:
Do we need the `coveralls`-messages? They only seem to clutter up the comments. And I have a hard time already catching up with the comments as-is.

-> can you remove those messages?

Liamsi:
> Do we need the coveralls-messages? They only seem to clutter up the comments.

I though we could keep them in case someone doesn't want to do the extra click to visit travis and still can have a detailed view & summary of the test coverage. But maybe that's overkill. Removed in https://github.com/dedis/cothority/pull/548/commits/53dcf777ae67b48b348b133841086e73f93ddb16.

Issue #551, opened 2016-08-25
Byzcoin protocol simulation doesn't run anymore


ineiti:
- wrong name in runfiles/byzcoin.simul
- not included in `protocols/protocols.go`
- error in downloading of initial block
- don't use `wget` but go-download - and verify the md5-hash

Liamsi:
Duplicate of #550

ineiti:
Done

PR #552, opened 2016-08-25
Byzcoin protocol simulation doesn't run anymore

ineiti:
Implements #551


protocols/protocols.go
```diff
@@ -38,7 +38,7 @@ import (
 	_ "github.com/dedis/cothority/protocols/randhound"
 	// ByzCoin has some strange library which uses 'seelog' that doesn't
 	// free all go-routines
-	//_ "github.com/dedis/cothority/protocols/byzcoin"
-	//_ "github.com/dedis/cothority/protocols/byzcoin/ntree"
-	//_ "github.com/dedis/cothority/protocols/byzcoin/pbft"
+	_ "github.com/dedis/cothority/protocols/byzcoin"
```
Liamsi: There was a reason we commented these 3 lines out (see comment on top). I think we should keep it this way and add a comment in the ByzCoin simulation pointing to these lines?

ineiti: True...

Liamsi:
LGTM

Liamsi:
But maybe add a pointer to these 3 lines? So @LefKok doesn't need to ask us how-to compile it in the future?

ineiti:
I'll also add the downloading of the block-file and md5-verification.

PR #567, opened 2016-09-02
[1/7] Network refactoring - Address

nikkolasg:
This PR shows the first part of the big #499 PR regarding the use of an Address type which contains also the type of the connection.

### DIFF
--- a/app/lib/server/server.go
+++ b/app/lib/server/server.go
@@ -65,7 +65,7 @@ func InteractiveConfig(binaryName string) {
 	var hostStr string
 	var ipProvided = true
 	var portStr string
-	var serverBinding string
+	var serverBinding network.Address
 	if !strings.Contains(str, ":") {
 		str = ":" + str
 	}
@@ -87,16 +87,17 @@ func InteractiveConfig(binaryName string) {
 		portStr = port
 	}
 
-	serverBinding = hostStr + ":" + portStr
-	if net.ParseIP(hostStr) == nil {
-		log.Fatal("Invalid connection  information for", serverBinding)
+	serverBinding = network.NewTCPAddress(hostStr + ":" + portStr)
+	if !serverBinding.Valid() {
+		log.Error("Unable to validate address given", serverBinding)
+		return
 	}
 
 	log.Info("We now need to get a reachable address for other CoSi servers")
 	log.Info("and clients to contact you. This address will be put in a group definition")
 	log.Info("file that you can share and combine with others to form a Cothority roster.")
 
-	var publicAddress string
+	var publicAddress network.Address
 	var failedPublic bool
 	// if IP was not provided then let's get the public IP address
 	if !ipProvided {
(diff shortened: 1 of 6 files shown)


network/address.go
```diff
@@ -0,0 +1,130 @@
...
+	// PlainTCP represents a vanilla TCP connection
+	PlainTCP ConnType = "tcp"
+	// TLS represents a TLS encrypted connection over TCP
+	TLS = "tls"
+	// PURB represents a PURB encryption connection over TCP
+	PURB = "purb"
+	// Local represents a channel based connection type
+	Local = "local"
+	// UnvalidConnType represents a non valid connection type
+	UnvalidConnType = "wrong"
```
ineiti: `InvalidConnType`

network/address.go
```diff
@@ -0,0 +1,130 @@
...
+func (a Address) Valid() bool {
+	vals := strings.Split(string(a), typeAddressSep)
+	if len(vals) != 2 {
+		return false
+	}
+	if connType(vals[0]) == UnvalidConnType {
+		return false
+	}
+
+	if ip, _, e := net.SplitHostPort(vals[1]); e != nil {
```
ineiti: don't we have to test the port here, too?

network/address.go
```diff
@@ -0,0 +1,130 @@
...
+		return false
+	} else if ip == "localhost" {
+		// localhost is not recognized by net.ParseIP ?
+		return true
+	} else if net.ParseIP(ip) == nil {
```
ineiti: I suppose this also works for IPv6?

Liamsi: Works for both: https://golang.org/pkg/net/#ParseIP

ineiti:
I like it, looks good.

Liamsi:
> I like it, looks good.

Yep, nice work @nikkolasg

app/lib/server/server.go
```diff
@@ -319,10 +324,10 @@ func entityListToPublics(el *sda.Roster) []abstract.Point {
 	return publics
 }
 
-func isPublicIP(ip string) bool {
+func isPublicIP(ip network.Address) bool {
```
ineiti: I propose to move this to `network.Address`, as it is useful in general, no? But then we'd also have to add missing blocks - http://www.tutorialspoint.com/ipv4/ipv4_quick_guide.htm

network/address.go
nikkolasg: Should I test the port being between 0 & 65536 then ?

ineiti: The golang-specs also allow for a string in the port (like `http` or `ftp`). Shall we insist on a port-number? I would be OK we do a `0<=port<=65535`. What do you think?

Liamsi: I though `SplitHostPort` does return an error if this isn't the case. Apparently, it doesn't: https://play.golang.org/p/4_Cz1Qln-G 
So yes, if `Valid` should only return true if the structure/syntax of the address is guaranteed to be valid, we should validate that.

ineiti: But to follow the go-specs we should also allow port-names like `http`. But I prefer to limit port-names to numbers.

Liamsi: Hmm, true. I would also prefer to limit port-names to numbers. But I step back a little on this one: it isn't critical (if the port is invalid we would recognize it while establishing the connection) anyways.

PR #568, opened 2016-09-02
[2/7] Refactor network - interfaces + Local network

nikkolasg:
This PR implements the local communication part using signaling and queues..
This PR comes with the basic interface definitions which are used throughout the network library such as `Conn`, `Listener`  & `Host`

---
This change is [Reviewable](https://reviewable.io/reviews/dedis/cothority/568)

### DIFF
--- a/network/local.go
+++ b/network/local.go
@@ -360,7 +360,8 @@ func NewLocalListenerWithManager(ctx *LocalManager, addr Address) (*LocalListene
 	return l, nil
 }
 
-// Listen implements the Listener interface
+// Listen implements the Listener interface. This call blocks until Stop() is
+// called on the listener.
 func (ll *LocalListener) Listen(fn func(Conn)) error {
 	ll.Lock()
 	if ll.listening {
--- a/network/local.go
+++ b/network/local.go
@@ -11,25 +11,25 @@ import (
 )
 
 // NewLocalRouter returns a fresh router which uses local connections. It takes
-// the default context.
+// the default manager.
 func NewLocalRouter(sid *ServerIdentity) (*Router, error) {
-	return NewLocalRouterWithContext(defaultLocalContext, sid)
+	return NewLocalRouterWithManager(defaultLocalManager, sid)
 }
 
-// NewLocalRouterWithContext is the same as NewLocalRouter but takes a specific
-// LocalContext. This is useful to run parallel different local overlays.
-func NewLocalRouterWithContext(ctx *LocalContext, sid *ServerIdentity) (*Router, error) {
-	h, err := NewLocalHostWithContext(ctx, sid.Address)
+// NewLocalRouterWithmanager is the same as NewLocalRouter but takes a specific
+// Localmanager. This is useful to run parallel different local overlays.
+func NewLocalRouterWithManager(ctx *LocalManager, sid *ServerIdentity) (*Router, error) {
+	h, err := NewLocalHostWithManager(ctx, sid.Address)
 	if err != nil {
 		return nil, err
 	}
 	return NewRouter(sid, h), nil
 }
 
-// LocalContext keeps reference to all opened local connections
+// LocalManager keeps reference to all opened local connections
 // It also keeps tracks of who is "listening", so it's possible to mimics
 // Conn & Listener.
-type LocalContext struct {
+type LocalManager struct {
 	// queues maps a remote endpoint to its packet queue. It's the main
 	//stucture used to communicate.
 	queues map[endpoint]*connQueue
@@ -39,16 +39,16 @@ type LocalContext struct {
 	baseUID uint64
 }
 
-// NewLocalContext returns a fresh new context that can be used by LocalConn,
+// NewLocalManager returns a fresh new manager that can be used by LocalConn,
 // LocalListener & LocalHost.
-func NewLocalContext() *LocalContext {
-	return &LocalContext{
+func NewLocalManager() *LocalManager {
+	return &LocalManager{
 		queues:    make(map[endpoint]*connQueue),
 		listening: make(map[Address]func(Conn)),
 	}
 }
 
-var defaultLocalContext = NewLocalContext()
+var defaultLocalManager = NewLocalManager()
 
 // endpoint represents one endpoint of a connection.
 type endpoint struct {
@@ -61,14 +61,14 @@ type endpoint struct {
 }
 
 // LocalReset reset the whole map of connections + listener so it is like
-// a fresh defaultLocalContext.
+// a fresh defaultLocalManager.
 func LocalReset() {
-	defaultLocalContext = NewLocalContext()
+	defaultLocalManager = NewLocalManager()
 
 }
 
 // IsListening returns true if the remote address is listening "virtually"
-func (ccc *LocalContext) isListening(remote Address) bool {
+func (ccc *LocalManager) isListening(remote Address) bool {
 	ccc.Lock()
 	defer ccc.Unlock()
 	_, ok := ccc.listening[remote]
@@ -77,7 +77,7 @@ func (ccc *LocalContext) isListening(remote Address) bool {
 
 // setListening marks the address as being able to accept incoming connection.
 // For each incoming connection, fn will be called in a go routine.
-func (ccc *LocalContext) setListening(addr Address, fn func(Conn)) {
+func (ccc *LocalManager) setListening(addr Address, fn func(Conn)) {
 	ccc.Lock()
 	defer ccc.Unlock()
 	ccc.listening[addr] = fn
@@ -85,7 +85,7 @@ func (ccc *LocalContext) setListening(addr Address, fn func(Conn)) {
 
 // unsetListening marks the address as *not* being able to accept incoming
 // connections.
-func (ccc *LocalContext) unsetListening(addr Address) {
+func (ccc *LocalManager) unsetListening(addr Address) {
 	ccc.Lock()
 	defer ccc.Unlock()
 	delete(ccc.listening, addr)
@@ -94,7 +94,7 @@ func (ccc *LocalContext) unsetListening(addr Address) {
 // connect will check if the remote address is listening, if yes it creates
 // the two connections, and launch the listening function in a go routine.
 // It returns the outgoing connection with any error.
-func (ccc *LocalContext) connect(local, remote Address) (*LocalConn, error) {
+func (ccc *LocalManager) connect(local, remote Address) (*LocalConn, error) {
 	ccc.Lock()
 	defer ccc.Unlock()
 
@@ -124,7 +124,7 @@ func (ccc *LocalContext) connect(local, remote Address) (*LocalConn, error) {
 // send will get the connection denoted by this endpoint and will call queueMsg
 // with the packet as argument on it. It returns ErrClosed if it does not find
 // the connection.
-func (ccc *LocalContext) send(e endpoint, nm Packet) error {
+func (ccc *LocalManager) send(e endpoint, nm Packet) error {
 	ccc.Lock()
 	defer ccc.Unlock()
 	q, ok := ccc.queues[e]
@@ -138,7 +138,7 @@ func (ccc *LocalContext) send(e endpoint, nm Packet) error {
 
 // close will get the connection denoted by this endpoint and will Close it if
 // present.
-func (ccc *LocalContext) close(conn *LocalConn) {
+func (ccc *LocalManager) close(conn *LocalConn) {
 	ccc.Lock()
 	defer ccc.Unlock()
 	// delete this conn
@@ -153,7 +153,7 @@ func (ccc *LocalContext) close(conn *LocalConn) {
 }
 
 // len returns how many local connections is there
-func (ccc *LocalContext) len() int {
+func (ccc *LocalManager) len() int {
 	ccc.Lock()
 	defer ccc.Unlock()
 	return len(ccc.queues)
@@ -165,35 +165,35 @@ type LocalConn struct {
 	local  endpoint
 	remote endpoint
 
-	// connQueue is the part that is accesible from the LocalContext (i.e. is
+	// connQueue is the part that is accesible from the LocalManager (i.e. is
 	// shared). Reason why we can't directly share LocalConn is because go test
 	// -race detects as data race (while it's *protected*)
 	*connQueue
 
-	ctx *LocalContext
+	manager *LocalManager
 }
 
 // newLocalConn simply init the fields of a LocalConn but do not try to
 // connect. It should not be used as-is, most user wants to call NewLocalConn.
-func newLocalConn(ctx *LocalContext, local, remote endpoint) *LocalConn {
+func newLocalConn(ctx *LocalManager, local, remote endpoint) *LocalConn {
 	return &LocalConn{
 		remote:    remote,
 		local:     local,
 		connQueue: newConnQueue(),
-		ctx:       ctx,
+		manager:   ctx,
 	}
 }
 
 // NewLocalConn returns a new channel connection from local to remote
 // Mimics the behavior of NewTCPConn => tries connecting right away.
-// It uses the default local context.
+// It uses the default local manager.
 func NewLocalConn(local, remote Address) (*LocalConn, error) {
-	return NewLocalConnWithContext(defaultLocalContext, local, remote)
+	return NewLocalConnWithManager(defaultLocalManager, local, remote)
 }
 
-// NewLocalConnWithContext is similar to NewLocalConn but takes a specific
-// LocalContext.
-func NewLocalConnWithContext(ctx *LocalContext, local, remote Address) (*LocalConn, error) {
+// NewLocalConnWithManager is similar to NewLocalConn but takes a specific
+// LocalManager.
+func NewLocalConnWithManager(ctx *LocalManager, local, remote Address) (*LocalConn, error) {
 	return ctx.connect(local, remote)
 }
 
@@ -212,7 +212,7 @@ func (lc *LocalConn) Send(ctx context.Context, msg Body) error {
 		MsgType: typ,
 		Msg:     body,
 	}
-	return lc.ctx.send(lc.remote, nm)
+	return lc.manager.send(lc.remote, nm)
 }
 
 // Receive implements the Conn interface.
@@ -234,7 +234,7 @@ func (lc *LocalConn) Remote() Address {
 func (lc *LocalConn) Close() error {
 	lc.connQueue.Close()
 	// close the remote conn also
-	lc.ctx.close(lc)
+	lc.manager.close(lc)
 	return nil
 }
 
@@ -313,21 +313,21 @@ type LocalListener struct {
 	// quit is used to stop the listening routine
 	quit chan bool
 
-	ctx *LocalContext
+	manager *LocalManager
 }
 
 // NewLocalListener returns a fresh LocalListener which implements the Listener
 // interface.
 func NewLocalListener(addr Address) (*LocalListener, error) {
-	return NewLocalListenerWithContext(defaultLocalContext, addr)
+	return NewLocalListenerWithManager(defaultLocalManager, addr)
 }
 
-// NewLocalListenerWithContext is similar to NewLocalListener but taking a
-// specific LocalContext to use to communicate.
-func NewLocalListenerWithContext(ctx *LocalContext, addr Address) (*LocalListener, error) {
+// NewLocalListenerWithManager is similar to NewLocalListener but taking a
+// specific LocalManager to use to communicate.
+func NewLocalListenerWithManager(ctx *LocalManager, addr Address) (*LocalListener, error) {
 	l := &LocalListener{
-		quit: make(chan bool),
-		ctx:  ctx,
+		quit:    make(chan bool),
+		manager: ctx,
 	}
 	return l, l.bind(addr)
 }
@@ -338,7 +338,7 @@ func (ll *LocalListener) bind(addr Address) error {
 	if addr.ConnType() != Local {
 		return errors.New("Wrong address type for local listener")
 	}
-	if ll.ctx.isListening(addr) {
+	if ll.manager.isListening(addr) {
 		return fmt.Errorf("%s is already listening: can't listen again", addr)
 	}
 	ll.addr = addr
@@ -349,7 +349,7 @@ func (ll *LocalListener) bind(addr Address) error {
 func (ll *LocalListener) Listen(fn func(Conn)) error {
 	ll.Lock()
 	ll.quit = make(chan bool)
-	ll.ctx.setListening(ll.addr, fn)
+	ll.manager.setListening(ll.addr, fn)
 	ll.listening = true
 	ll.Unlock()
 
@@ -361,7 +361,7 @@ func (ll *LocalListener) Listen(fn func(Conn)) error {
 func (ll *LocalListener) Stop() error {
 	ll.Lock()
 	defer ll.Unlock()
-	ll.ctx.unsetListening(ll.addr)
+	ll.manager.unsetListening(ll.addr)
 	if ll.listening {
 		close(ll.quit)
 	}
@@ -390,24 +390,24 @@ func (ll *LocalListener) Listening() bool {
 type LocalHost struct {
 	addr Address
 	*LocalListener
-	ctx *LocalContext
+	ctx *LocalManager
 }
 
 // NewLocalHost returns a fresh Host using Local communication that will listen
 // on the given addr.
 func NewLocalHost(addr Address) (*LocalHost, error) {
-	return NewLocalHostWithContext(defaultLocalContext, addr)
+	return NewLocalHostWithManager(defaultLocalManager, addr)
 }
 
-// NewLocalHostWithContext is similar to NewLocalHost but specify which
-// LocalContext to use for communicating.
-func NewLocalHostWithContext(ctx *LocalContext, addr Address) (*LocalHost, error) {
+// NewLocalHostWithManager is similar to NewLocalHost but specify which
+// LocalManager to use for communicating.
+func NewLocalHostWithManager(ctx *LocalManager, addr Address) (*LocalHost, error) {
 	lh := &LocalHost{
 		addr: addr,
 		ctx:  ctx,
 	}
 	var err error
-	lh.LocalListener, err = NewLocalListenerWithContext(ctx, addr)
+	lh.LocalListener, err = NewLocalListenerWithManager(ctx, addr)
 	return lh, err
 
 }
@@ -419,7 +419,7 @@ func (lh *LocalHost) Connect(addr Address) (Conn, error) {
 	}
 	var finalErr error
 	for i := 0; i < MaxRetryConnect; i++ {
-		c, err := NewLocalConnWithContext(lh.ctx, lh.addr, addr)
+		c, err := NewLocalConnWithManager(lh.ctx, lh.addr, addr)
 		if err == nil {
 			return c, nil
 		}
@@ -432,14 +432,14 @@ func (lh *LocalHost) Connect(addr Address) (Conn, error) {
 
 // NewLocalClient returns Client that uses the Local communication layer.
 func NewLocalClient() *Client {
-	return NewLocalClientWithContext(defaultLocalContext)
+	return NewLocalClientWithManager(defaultLocalManager)
 }
 
-// NewLocalClientWithContext is similar to NewLocalClient but takes a specific
-// LocalContext to communicate.
-func NewLocalClientWithContext(ctx *LocalContext) *Client {
+// NewLocalClientWithManager is similar to NewLocalClient but takes a specific
+// LocalManager to communicate.
+func NewLocalClientWithManager(ctx *LocalManager) *Client {
 	fn := func(own, remote *ServerIdentity) (Conn, error) {
-		return NewLocalConnWithContext(ctx, own.Address, remote.Address)
+		return NewLocalConnWithManager(ctx, own.Address, remote.Address)
 	}
 	return newClient(fn)
 
--- a/network/local.go
+++ b/network/local.go
@@ -469,14 +469,3 @@ func NewLocalClientWithManager(ctx *LocalManager) *Client {
 func NewLocalAddress(addr string) Address {
 	return NewAddress(Local, addr)
 }
-
-/*// GetStatus implements the Host interface*/
-//func (l *chanHost) GetStatus() Status {
-//m := make(map[string]string)
-//m["Connections"] = strings.Join(l.conns.Get(), "\n")
-//m["Host"] = l.Address()
-//m["Total"] = strconv.Itoa(l.conns.Len())
-//m["Packets_Received"] = strconv.FormatUint(0, 10)
-//m["Packets_Sent"] = strconv.FormatUint(0, 10)
-//return m
-//}
--- a/network/local.go
+++ b/network/local.go
@@ -3,10 +3,11 @@ package network
 import (
 	"errors"
 	"fmt"
-	"golang.org/x/net/context"
 	"reflect"
 	"sync"
 	"time"
+
+	"golang.org/x/net/context"
 )
 
 // NewLocalRouter returns a fresh router which uses local connections. It takes
@@ -157,7 +158,8 @@ func (ccc *LocalContext) Len() int {
 	return len(ccc.queues)
 }
 
-// LocalConn is a connection that send and receive messages through channels
+// LocalConn is a connection that send and receive messages to other
+// connections locally
 type LocalConn struct {
 	local  endpoint
 	remote endpoint
--- a/network/local.go
+++ b/network/local.go
@@ -68,54 +68,54 @@ func LocalReset() {
 }
 
 // isListening returns true if the remote address is listening "virtually"
-func (ccc *LocalManager) isListening(remote Address) bool {
-	ccc.Lock()
-	defer ccc.Unlock()
-	_, ok := ccc.listening[remote]
+func (lm *LocalManager) isListening(remote Address) bool {
+	lm.Lock()
+	defer lm.Unlock()
+	_, ok := lm.listening[remote]
 	return ok
 }
 
 // setListening marks the address as being able to accept incoming connection.
 // For each incoming connection, fn will be called in a go routine.
-func (ccc *LocalManager) setListening(addr Address, fn func(Conn)) {
-	ccc.Lock()
-	defer ccc.Unlock()
-	ccc.listening[addr] = fn
+func (lm *LocalManager) setListening(addr Address, fn func(Conn)) {
+	lm.Lock()
+	defer lm.Unlock()
+	lm.listening[addr] = fn
 }
 
 // unsetListening marks the address as *not* being able to accept incoming
 // connections.
-func (ccc *LocalManager) unsetListening(addr Address) {
-	ccc.Lock()
-	defer ccc.Unlock()
-	delete(ccc.listening, addr)
+func (lm *LocalManager) unsetListening(addr Address) {
+	lm.Lock()
+	defer lm.Unlock()
+	delete(lm.listening, addr)
 }
 
 // connect will check if the remote address is listening, if yes it creates
 // the two connections, and launch the listening function in a go routine.
 // It returns the outgoing connection with any error.
-func (ccc *LocalManager) connect(local, remote Address) (*LocalConn, error) {
-	ccc.Lock()
-	defer ccc.Unlock()
+func (lm *LocalManager) connect(local, remote Address) (*LocalConn, error) {
+	lm.Lock()
+	defer lm.Unlock()
 
-	fn, ok := ccc.listening[remote]
+	fn, ok := lm.listening[remote]
 	if !ok {
 		return nil, fmt.Errorf("%s can't connect to %s: it's not listening", local, remote)
 	}
 
-	outEndpoint := endpoint{local, ccc.baseUID}
-	ccc.baseUID++
+	outEndpoint := endpoint{local, lm.baseUID}
+	lm.baseUID++
 
-	incEndpoint := endpoint{remote, ccc.baseUID}
-	ccc.baseUID++
+	incEndpoint := endpoint{remote, lm.baseUID}
+	lm.baseUID++
 
-	outgoing := newLocalConn(ccc, outEndpoint, incEndpoint)
-	incoming := newLocalConn(ccc, incEndpoint, outEndpoint)
+	outgoing := newLocalConn(lm, outEndpoint, incEndpoint)
+	incoming := newLocalConn(lm, incEndpoint, outEndpoint)
 
 	// outgoing knows how to store packet into the incoming's queue
-	ccc.queues[outEndpoint] = outgoing.connQueue
+	lm.queues[outEndpoint] = outgoing.connQueue
 	// incoming knows how to store packet into the outgoing's queue
-	ccc.queues[incEndpoint] = incoming.connQueue
+	lm.queues[incEndpoint] = incoming.connQueue
 
 	go fn(incoming)
 	return outgoing, nil
@@ -124,10 +124,10 @@ func (ccc *LocalManager) connect(local, remote Address) (*LocalConn, error) {
 // send will get the connection denoted by this endpoint and will call queueMsg
 // with the packet as argument on it. It returns ErrClosed if it does not find
 // the connection.
-func (ccc *LocalManager) send(e endpoint, nm Packet) error {
-	ccc.Lock()
-	defer ccc.Unlock()
-	q, ok := ccc.queues[e]
+func (lm *LocalManager) send(e endpoint, nm Packet) error {
+	lm.Lock()
+	defer lm.Unlock()
+	q, ok := lm.queues[e]
 	if !ok {
 		return ErrClosed
 	}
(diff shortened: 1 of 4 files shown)


ineiti:
I changed the base so that only the changes to 1/7 are visible.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+
+// NewLocalRouter returns a fresh router which uses local connections. It takes
+// the default context.
+func NewLocalRouter(sid *ServerIdentity) (*Router, error) {
+	return NewLocalRouterWithContext(defaultLocalContext, sid)
+}
+
+// NewLocalRouterWithContext is the same as NewLocalRouter but takes a specific
+// LocalContext. This is useful to run parallel different local overlays.
+func NewLocalRouterWithContext(ctx *LocalContext, sid *ServerIdentity) (*Router, error) {
```
ineiti: Please convince me again why we need `Context` here. Is it really used?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return NewRouter(sid, h), nil
+}
+
+// LocalContext keeps reference to all opened local connections
+// It also keeps tracks of who is "listening", so it's possible to mimics
+// Conn & Listener.
+type LocalContext struct {
```
ineiti: `LocalQueues`? Because `Context` is something go-specific...

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+)
+
+// NewLocalRouter returns a fresh router which uses local connections. It takes
+// the default context.
+func NewLocalRouter(sid *ServerIdentity) (*Router, error) {
+	return NewLocalRouterWithContext(defaultLocalContext, sid)
+}
+
+// NewLocalRouterWithContext is the same as NewLocalRouter but takes a specific
+// LocalContext. This is useful to run parallel different local overlays.
```
ineiti: I don't understand the second part of that comment. Can you elaborate, please?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+// NewLocalContext returns a fresh new context that can be used by LocalConn,
+// LocalListener & LocalHost.
+func NewLocalContext() *LocalContext {
+	return &LocalContext{
+		queues:    make(map[endpoint]*connQueue),
+		listening: make(map[Address]func(Conn)),
+	}
+}
+
+var defaultLocalContext = NewLocalContext()
```
ineiti: I'm a bit torn with this global variable here. Remembering what you had to do to `sda.LocalTest` I ask myself whether this should not be put in `sda.LocalTest`?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+
+// LocalReset reset the whole map of connections + listener so it is like
+// a fresh defaultLocalContext.
+func LocalReset() {
+	defaultLocalContext = NewLocalContext()
+
+}
+
+// IsListening returns true if the remote address is listening "virtually"
+func (ccc *LocalContext) IsListening(remote Address) bool {
```
ineiti: Why is it `ccc`? I'd prefer `lc` or `lq` if we go with `LocalQueues`.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+
+	outgoing := newLocalConn(ccc, outEndpoint, incEndpoint)
+	incoming := newLocalConn(ccc, incEndpoint, outEndpoint)
+
+	// outgoing knows how to store packet into the incoming's queue
+	ccc.queues[outEndpoint] = outgoing.connQueue
+	// incoming knows how to store packet into the outgoing's queue
+	ccc.queues[incEndpoint] = incoming.connQueue
+
+	go fn(incoming)
```
ineiti: This makes one go-routine per incoming connection. Does this work OK if we have like 1000 local connections?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+		return ErrClosed
+	}
+
+	q.Push(nm)
+	return nil
+}
+
+// Close will get the connection denoted by this endpoint and will Close it if
+// present.
+func (ccc *LocalContext) Close(conn *LocalConn) {
```
ineiti: Should it return an error if the connection is already closed?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+}
+
+// Len returns how many local connections is there
+func (ccc *LocalContext) Len() int {
+	ccc.Lock()
+	defer ccc.Unlock()
+	return len(ccc.queues)
+}
+
+// LocalConn is a connection that send and receive messages through channels
```
ineiti: I thought you don't use channels?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+type LocalConn struct {
+	local  endpoint
+	remote endpoint
+
+	// connQueue is the part that is accesible from the LocalContext (i.e. is
+	// shared). Reason why we can't directly share LocalConn is because go test
+	// -race detects as data race (while it's *protected*)
+	*connQueue
+
+	ctx *LocalContext
```
ineiti: As `Context` is golang-specific and `LocalContext` is not a `Context` I really urge for changing its name.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+		val = val.Elem()
+	}
+	body = val.Interface()
+
+	var typ = TypeFromData(body)
+	nm := Packet{
+		MsgType: typ,
+		Msg:     body,
+	}
+	return cc.ctx.Send(cc.remote, nm)
```
ineiti: This is confusing me - `ctx` from method-argument is a `Context` and `ctx` from `cc` (why not `lc`?) is a `LocalContext`.

network/local.go
ineiti: I see that `LocalContext` is not at all a context here.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+}
+
+// Receive implements the Conn interface.
+func (cc *LocalConn) Receive(ctx context.Context) (Packet, error) {
```
ineiti: Future will tell me, but I'll ask anyway: is `Context` used anywhere?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+func (cc *LocalConn) Close() error {
+	cc.connQueue.Close()
+	// close the remote conn also
+	cc.ctx.Close(cc)
+	return nil
+}
+
+// Rx implements the Conn interface
+func (cc *LocalConn) Rx() uint64 {
+	return 0
```
ineiti: Please add a counter for the actual sent and received bytes.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+func (cc *LocalConn) Tx() uint64 {
+	return 0
+}
+
+// Type implements the Conn interface
+func (cc *LocalConn) Type() ConnType {
+	return Local
+}
+
+type connQueue struct {
```
ineiti: Needed: comments on `connQueu` and its methods.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+	defer c.L.Unlock()
+	c.isClosed = true
+	c.Signal()
+}
+
+// LocalListener is a Listener that uses LocalConn to communicate. It tries to
+// behave as much as possible as a real golang net.Listener but using LocalConn
+// as the underlying communication layer.
+type LocalListener struct {
+	// addr is the addr we're listening to + mut
```
ineiti: What is `mut` here?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+	isClosed bool
+}
+
+func newConnQueue() *connQueue {
+	return &connQueue{
+		Cond: sync.NewCond(&sync.Mutex{}),
+	}
+}
+
+func (c *connQueue) Push(p Packet) {
```
ineiti: use `cq` instead of `c`?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+	ll.ctx.Listening(ll.addr, fn)
+	ll.listening = true
+	ll.Unlock()
+
+	<-ll.quit
+	return nil
+}
+
+// Stop implements the Listener interface
+func (ll *LocalListener) Stop() error {
```
ineiti: Shouldn't this return an error if the connection is already closed?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+		quit: make(chan bool),
+		ctx:  ctx,
+	}
+	return l, l.bind(addr)
+}
+
+func (ll *LocalListener) bind(addr Address) error {
+	ll.Lock()
+	defer ll.Unlock()
+	if addr.ConnType() != Local {
```
ineiti: I would put this test in `NewLocalListenerWithContext` and fail as early as possible.

ineiti: Correct myself: it is called in `NewLocalListenerWithContext`, so that's OK.

ineiti: Why don't merge `bind` into `NewLocalListenerWithContext`?
And I think you can remove the lock here - the caller doesn't have the `LocalListener` yet, so he cannot do anything with it.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+	if ll.ctx.IsListening(addr) {
+		return fmt.Errorf("%s is already listening: can't listen again", addr)
+	}
+	ll.addr = addr
+	return nil
+}
+
+// Listen implements the Listener interface
+func (ll *LocalListener) Listen(fn func(Conn)) error {
```
ineiti: I'd add that it is blocking in the comment.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+// incoming connections.
+func (ll *LocalListener) Listening() bool {
+	ll.Lock()
+	defer ll.Unlock()
+	return ll.listening
+}
+
+// LocalHost is a Host implementation using LocalConn and LocalListener as
+// the underlying means of communication. It is a implementation of the Host
+// interface.
```
ineiti: Thinking out loud: can we say programmatically that it is an implementation of the `Host`-interface by adding `Host` to the struct?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+// NewLocalHostWithContext is similar to NewLocalHost but specify which
+// LocalContext to use for communicating.
+func NewLocalHostWithContext(ctx *LocalContext, addr Address) (*LocalHost, error) {
+	lh := &LocalHost{
+		addr: addr,
+		ctx:  ctx,
+	}
+	var err error
+	lh.LocalListener, err = NewLocalListenerWithContext(ctx, addr)
+	return lh, err
```
ineiti: Is `lh` usable if `err != nil`? If it is usable, leave as is. Else, return `nil` if there is an error.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+
+}
+
+// Connect implements the Host interface.
+func (lh *LocalHost) Connect(addr Address) (Conn, error) {
+	if addr.ConnType() != Local {
+		return nil, errors.New("Can't connect to non-Local address")
+	}
+	var finalErr error
+	for i := 0; i < MaxRetryConnect; i++ {
```
ineiti: Can we have a failure in local-connections?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+	return newClient(fn)
+
+}
+
+// NewLocalAddress returns an Address of type Local with the given raw addr.
+func NewLocalAddress(addr string) Address {
+	return NewAddress(Local, addr)
+}
+
+/*// GetStatus implements the Host interface*/
```
ineiti: Remove if it's not used anymore.

network/network.go
```diff
@@ -0,0 +1,65 @@
...
+	// It should not be a Golang type.
+	Send(ctx context.Context, obj Body) error
+	// Receive any message through the connection. It is a blocking call that
+	// returns either when a message arrived or when Close() has been called, or
+	// when a network error occured.
+	Receive(ctx context.Context) (Packet, error)
+	// Close will close the connection. Implementations must take care that
+	// Close() makes Receive() returns with an error, and any subsequent Send()
+	// will return with an error.
+	Close() error
```
ineiti: What happens when calling `Close` on an unlistened address or an already closed address?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+	c.queue = append(c.queue, p)
+	c.Signal()
+}
+
+func (c *connQueue) Pop() (Packet, error) {
+	c.L.Lock()
+	defer c.L.Unlock()
+	for len(c.queue) == 0 {
+		if c.isClosed {
```
ineiti: can it happen that the queue is still full but that it is closed anyway?

ineiti:
General nitpicks:
- local names of structures (`ccc` for `LocalContext` throws me off, can we generalize on the first letters in general?)
- coveralls doesn't show the coverage of the local.go, but it says it's 83%, which is better than the overall code, but still too low IMO.
- shall we keep the home-made queues or go with channels?

ineiti:
I cannot even do
```
cd network
go test -run TestLocal
```
is this normal?

network/local.go
nikkolasg: `LocalEnvironment` `LocalSetting` ? It does not only store queues, but do everything : create connections, pass messages around, setup listening functions etc..

network/local.go
nikkolasg: > This is useful to run parallel different local overlays

In case multiple tests are to be ran in parallel, where each one listens and connect to the same set of addresses, each tests must create their own LocalContext, so each Local listeners / conn of each tests don't interfer with the other Local listeners / conn of the other tests.

Is that OK ? (I'll put that in the comment if it's OK for you, otherwise the comments won't be shown)

app/lib/server/server.go
```diff
@@ -319,10 +324,10 @@ func entityListToPublics(el *sda.Roster) []abstract.Point {
 	return publics
 }
 
-func isPublicIP(ip string) bool {
+func isPublicIP(ip network.Address) bool {
```
Liamsi: Seems to be implemented here: https://github.com/dedis/cothority/pull/567/commits/9125dfdb38993916d664ae38e305a89e5c48bb7a

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+func (ccc *LocalContext) IsListening(remote Address) bool {
+	ccc.Lock()
+	defer ccc.Unlock()
+	_, ok := ccc.listening[remote]
+	return ok
+}
+
+// Listening put the address as "listening" mode. If a user connects to this
+// addr, this function will be called.
+func (ccc *LocalContext) Listening(addr Address, fn func(Conn)) {
```
Liamsi: It's to easy to confuse `Listening` and `IsListening`. Why not call this one `Listen`?

Liamsi: I see, `Listen` won't be the best name, because it doesn't start the listening or anything else. `RegisterListenCB`?

network/local.go
nikkolasg: There **is** a *LocalContext in `LocalTest` indeed (present in #570) so all tests will have their own `LocalContext` but you should not be required to do so if you don't need an explicit one.

network/local.go
nikkolasg: Once decided on the name of `LocalContext` I wil change.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+
+// IsListening returns true if the remote address is listening "virtually"
+func (ccc *LocalContext) IsListening(remote Address) bool {
+	ccc.Lock()
+	defer ccc.Unlock()
+	_, ok := ccc.listening[remote]
+	return ok
+}
+
+// Listening put the address as "listening" mode. If a user connects to this
```
Liamsi: This comment doesn't help the dev using the code. To understand what you mean by "listening" mode, he/she needs to look into the underlying implementation.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+// addr, this function will be called.
+func (ccc *LocalContext) Listening(addr Address, fn func(Conn)) {
+	ccc.Lock()
+	defer ccc.Unlock()
+	ccc.listening[addr] = fn
+}
+
+// StopListening remove the address from the "listening" mode
```
Liamsi: Similar to above: The comment and the name of this public method sound like this actively stops sth. from listening (which is not case). `ResetOnListenCB`?

network/local.go
ineiti: What does @Liamsi think about this global variable? @nikkolasg is somewhat true in that we would like to make it as simple as possible. But then again, having a global variable here...

network/local.go
Liamsi: This will depend on the machine where the tests are running: http://stackoverflow.com/a/8534711/1315573
1000 shouln0t be a problem though.

app/lib/server/server.go
ineiti: So we can remove this one?

network/local.go
nikkolasg: This is exactly how it's supposed to be used in Golang, one go routine per connections. It's exactly the same for TCP listener. 
So the question is whether we want to allow unlimited number of go routine here or not. I think it should not be an issue here because it's local <=> test. So if someone wants to blow up his computer  with more than 1 billions connections to itself, let him go for it... ;)

network/local.go
ineiti: Corrected English. IIRC, `each` takes singular and is 3rd person, so `listen_s_` and `connect_s_`, and `each test` without s. But `other` takes plural, so `other listener_s_`. Any native English speakers?

network/local.go
ineiti: fine for me then.

network/local.go
nikkolasg: If it's a feature that we want to have, I prefer having it at the Connection level than here. 
Here, with two connections `c1` and `c2`, `c1.Close` calls `localContext.Close(c1)` calls `c2.Close` which calls `localContext.Close(c2)` which terminate s (the `if !ok { return }` part). So a close on a connection directly closes its remote counterpart. 
So `if the connection is already closed` means in our `LocalContext` that we are closing the remote counterpart already, so it's not an error. 
But we can definitely impl that features in the `LocalConn` itself, would that be OK ?

network/local.go
Liamsi: I wan't sure what
>  Remembering what you had to do to sda.LocalTest

meant. That's why I didn't really comment here. If this `defaultLocalContext` isn't changed by different go-rountines etc I have absolutely nothing against it. But it doesn't look like that... If all tests use the same defaultContext instance couldn't that lead to a problem? I'll need to read the tests, too before I have a clear stance here.

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+		finalErr = err
+		time.Sleep(WaitRetry)
+	}
+	return nil, finalErr
+
+}
+
+// NewLocalClient returns Client that uses the Local communication layer.
+func NewLocalClient() *Client {
+	return NewLocalClientWithContext(defaultLocalContext)
```
Liamsi: Is `NewLocalClientWithContext` used somewhere with a different ctx than the global `defaultLocalContext`?

network/local.go
```diff
@@ -0,0 +1,459 @@
...
+}
+
+// LocalConn is a connection that send and receive messages through channels
+type LocalConn struct {
+	local  endpoint
+	remote endpoint
+
+	// connQueue is the part that is accesible from the LocalContext (i.e. is
+	// shared). Reason why we can't directly share LocalConn is because go test
+	// -race detects as data race (while it's *protected*)
```
Liamsi: That's really strange. If you are 100% sure that go detects a race although there is non, please report a bug report to golang!

network/local.go
Liamsi: Yep, also found this confusing. The only explicit channel seems to be https://github.com/dedis/cothority/pull/568/files#diff-758a279d81fc93cf992b979af141f431R311

network/local.go
nikkolasg: I'm not sure it's a bug. 
`LocalContext` has two connections `c1` and `c2`. `c1` is `Send()`ing to `c2` in one goroutine(g1), and *at the same time* , `c2` is receiving from `c1` in another goroutine (g2). In g1, `c1` will call `localcontext.Send()` which will call `c2.Push`. So g1 is *modifying* `c2`'s BUFFER, and  g2 is *reading* `c2`'s BUFFER. Even though all operations to the buffer are protected by mutexes, the *access to the connections's buffer* is not (g1 access `c2` and g2 too). 
So I'm not sure it's a bug or if I do strange things.

network/local.go
nikkolasg: Yes in `sda.LocalTest` specifically !

network/local.go
nikkolasg: Please ideas, suggestions !

network/local.go
Liamsi: I don't care if you call this `c`or `cq` (as long as it is consistent through the whole code). Another question regarding this line, though: `Push` will be used outside of this package?

network/local.go
nikkolasg: I understand your point but it's not only a registration of a callback, it's also setting the fact that this address is now currently listening, and ready to accept incoming connections, and you can know that information by calling `IsListening`, and make this address stop the listening process by calling `StopListening`. Either I put more on the comment, or either we must find 3 good names to replace that. On top of my head I have 
* `IsListening` stays the same
* `Listening` => `SetListening`
* `UnsetListening` ?

network/network.go
nikkolasg: > What happens when calling Close on an unlistened address or an already closed address?

This is about `Conn` not `Address` so I'm afraid I don't get your question. 
If your question is: `What happens when calling Close on an unlistened address or an already closed Conn?`, the answer is ... undefined ! 
Should we want to put a specific implementation rules that returns an error ? Is that necessary ?
Even though I'm not sure it's necessary, I'd be OK to do that, if you agree.

app/lib/server/server.go
nikkolasg: It's already gone in `refactor_mocking_address` but the comments still sticks here I guess.

network/local.go
nikkolasg: Again, each `sda.LocalTest` have **their own** `LocalContext** so all sda tests which are using `LocalTest` are fine (I'd say roughly >90% 95% of sda tests use `LocalTest`).

network/local.go
nikkolasg: Sorry old comment, will remove.

network/local.go
ineiti: Who else uses `LocalContext`?

Liamsi: > Again, each sda.LocalTest have their own LocalContext so all sda tests which are using LocalTest are fine (I'd say roughly >90% 95% of sda tests use LocalTest).

This sounds like 5-10% aren't fine?

network/local.go
Liamsi: > , it's also setting the fact that this address is now currently listening

So, by putting some callback into the `listening` map the `LocalContext` is set into a "I'm OK to accept connections and listen" mode? So it is merely an initialization that needs to happen before it can be used to actually listen? If yes, I propose: `InitListening`, `IsListening` and `ResetListening`?

network/local.go
nikkolasg: > This sounds like 5-10% aren't fine?

By fine you mean they can't be run in parallel then yes. But it's not the only global variable we have in our sda tests (registered services, registered protocols etc). 
Seeing all kinds of reactions around `LocalContext`, I propose something:
* I make *all* sda tests using `LocalTest` so they can't all be ran in parallel.
* I make all functions of `LocalContext` **unexported** so we can't do anything with a `LocalContext`. I should already have done that anyway.
* I keep the global var as for me it's does not really make senses to force any user to create and pass around an explicit `LocalContext` while he does not need to. There's no un-mutex-protected access and since **all methods will be unexported**, he can't do much with it anyway.

network/local.go
nikkolasg: For the moment not, but reading it seems a good idea (It provides simple control over timeouts and cancelation and ensures that critical values like security credentials transit Go programs properly.)
I think we should all read on it, get to know the subject better and *then* decide if we want to keep that (because we'll mostly use it later) or just drop it because we don't see any ways in which we can use that.

Liamsi: IIRC originally we wanted to use the context to pass data/state from/to different layers of the framework. We don't do that at all for the moment. Not sure if we should just remove the context as currently it isn't really used (I guess yes).

network/local.go
Liamsi: > I make all functions of LocalContext unexported so we can't do anything with a LocalContext. I should already have done that anyway.

If this is possible (and not complicated) I really like this idea!

nikkolasg: .. and it's done :)

network/local.go
nikkolasg: > Push will be used outside of this package?

Nope and everything is now back to unexported.

network/local.go
nikkolasg: Corrected.

network/local.go
nikkolasg: All right. So I'm doing:
* Make `Listen` returns an error if it is already listening
* Make `Stop()` if it is already stopped
* Making `LocalConn.Close()` returns an error if it is already closed

What do you mean by `if nobody is listening` ?

network/local.go
nikkolasg: It's already in the comment of the interface in `network.go`. I won't copy all the comments for each impl ...

network/local.go
nikkolasg: I don't think so, Golang has this particularity of having kinda like the Ruby duck typing system but at compile time. So defining a struct don't tell you what interfaces it implements. Only by looking at which methods does this struct implements, you see that it has all the methods `Host` is defining **so** it is a `Host`.

network/local.go
```diff
@@ -11,25 +11,25 @@ import (
...
 func NewLocalRouter(sid *ServerIdentity) (*Router, error) {
-	return NewLocalRouterWithContext(defaultLocalContext, sid)
+	return NewLocalRouterWithManager(defaultLocalManager, sid)
 }
 
-// NewLocalRouterWithContext is the same as NewLocalRouter but takes a specific
-// LocalContext. This is useful to run parallel different local overlays.
-func NewLocalRouterWithContext(ctx *LocalContext, sid *ServerIdentity) (*Router, error) {
-	h, err := NewLocalHostWithContext(ctx, sid.Address)
+// NewLocalRouterWithmanager is the same as NewLocalRouter but takes a specific
```
Liamsi: (To save you some time later) This might not pass the linter: Comment should start with NewLocalRouterWithManager

network/local.go
```diff
@@ -11,25 +11,25 @@ import (
...
+// Localmanager. This is useful to run parallel different local overlays.
+func NewLocalRouterWithManager(ctx *LocalManager, sid *ServerIdentity) (*Router, error) {
```
Liamsi: Change the name of `ctx` to `lm` to be consistent with other parts of the code

network/local.go
nikkolasg: Not failure but it's the same story with TCP: If we launch the Listening part in a go routine, we don't know when the go routine is actually launched so sometimes it fails if we just try to connect once.

network/local.go
```diff
@@ -313,21 +334,21 @@ type LocalListener struct {
...
-	return NewLocalListenerWithContext(defaultLocalContext, addr)
+	return NewLocalListenerWithManager(defaultLocalManager, addr)
 }
 
-// NewLocalListenerWithContext is similar to NewLocalListener but taking a
-// specific LocalContext to use to communicate.
-func NewLocalListenerWithContext(ctx *LocalContext, addr Address) (*LocalListener, error) {
+// NewLocalListenerWithManager is similar to NewLocalListener but taking a
+// specific LocalManager to use to communicate.
+func NewLocalListenerWithManager(ctx *LocalManager, addr Address) (*LocalListener, error) {
```
Liamsi: s/ctx/lm

network/local.go
nikkolasg: That's actually left for a reason I forgot to mention... :)
What to do with regards to Status? I can't make them implement the interface because the interface uses something sda specific (`type Status map[string]string` is in sda package). So for the moment I dropped them but now we don't have access to the statistics of the network layer (which is not great). Any plans / solutions ?
On top of my head, I would remove the specific `Status` type and then let `Router` / `Host` / `Conn` / `Listener` implements `GetStatus() map[string]string`.

network/local.go
ineiti: `if nobody is listening` -> if `Listen` hasn't been called and just `Stop` is called

network/local.go
ineiti: Personally I don't mind have a 'This blocks until the connection is Stopped'.

network/local.go
ineiti: I thought we said that `Conode` will do the status-implementation and get the values directly from the `Router`. Does that work?

ineiti:
Starts to look OK. I'll continue my channel-implementation, as it shows me some errors I'm not sure are due to the channels but which might show some deeper bugs. But I don't know yet for sure ;)

network/local.go
nikkolasg: Ok I'll make this in the #570 PR (because it involves the `Router` which is not there at the moment) and remove this comment here.

nikkolasg:
> coveralls doesn't show the coverage of the local.go, but it says it's 83%, which is better than the overall code, but still too low IMO.

We did not specify clearly what is our code coverage expectation on PRs so ... what would be a good coverage ?

> shall we keep the home-made queues or go with channels?

The main advantages is that the buffer is "unlimited" this way so it mimics the TCP stack, and *all* the tests are passing for the moment on my laptop. Shall Travis fails for unknown reason and it seems impossible to debug, I'd switch to an alternate version of Local but for the moment I don't see any incentives to change that. Also, I already spend quite some time on it; it seems to me it'd be quite a loss of time to re-code something new with the tests and so on & to throw this version away because "it's not a channel".

network/local.go
nikkolasg: Yep this is also done. I was not clear enough: `Make Stop() return an error if it is already stopped <=> not listening`

Liamsi:
> We did not specify clearly what is our code coverage expectation on PRs so ... what would be a good coverage ?

I think this rule of thumb is reasonable: In every PR the code coverage shall not decrease (maybe sth like +/-0.5% is OK though). We should aim for ~100% and have 80% as lower boundary. Code containing only `func main` and not much more is OK if it is tested by integration tests and manually instead of unit tests (for these *few* packages may have lower code coverage). 

For this PR specifically: It would make sense to see which lines aren't covered and why.

nikkolasg:
> For this PR specifically: It would make sense to see which lines aren't covered and why.

```
git checkout refactor_mocking ## the big main branch where all the code is here <-> compiling
cd network/
go test -v -coverprofile=network.out -covermode=atomic 
go tool cover -html=network.out
```
And then you get a nice html page showing you the coverage for each file & each line :)

network/struct.go
```diff
@@ -279,6 +279,7 @@ func GlobalBind(address string) (string, error) {
 	return "0.0.0.0:" + addr[1], nil
 }
 
+<<<<<<< HEAD
```
Liamsi: This shouldn't be here, right? Same for `=======` some lines below

network/struct.go
```diff
@@ -301,4 +302,42 @@ func handleError(err error) error {
...
+	defer c.Unlock()
+	c.rx += delta
+}
+
+// UpdateRx adds delta to the tx counter
+func (c *counterSafe) updateTx(delta uint64) {
+	c.Lock()
+	defer c.Unlock()
+	c.tx += delta
+>>>>>>> 429e587... more coverage (>95%) + tx/rx impl
```
Liamsi: Same as above

network/local.go
```diff
@@ -199,25 +201,27 @@ func NewLocalConnWithManager(ctx *LocalManager, local, remote Address) (*LocalCo
...
-	if val.Kind() == reflect.Ptr {
-		val = val.Elem()
-	}
-	body = val.Interface()
-
-	var typ = TypeFromData(body)
-	nm := Packet{
-		MsgType: typ,
-		Msg:     body,
+	buff, err := MarshalRegisteredType(msg)
```
Liamsi: +1 for this simplifications!

network/struct.go
```diff
@@ -301,4 +302,42 @@ func handleError(err error) error {
...
+}
+
+// UpdateRx adds delta to the rx counter
+func (c *counterSafe) updateRx(delta uint64) {
+	c.Lock()
+	defer c.Unlock()
+	c.rx += delta
+}
+
+// UpdateRx adds delta to the tx counter
```
Liamsi: outdated comment

network/local.go
nikkolasg: I'm not really a fan but it is because we want to have a working `Tx()` and `Rx()` method so I got to marshal the message now ..  (before it was not needed to marshal).

Liamsi: For really sending the message you need to marshal it anyways, or not? So does this happen twice now or only here?

network/local.go
```diff
@@ -0,0 +1,466 @@
...
+
+// NewLocalRouter returns a fresh router which uses local connections. It takes
+// the default manager.
+func NewLocalRouter(sid *ServerIdentity) (*Router, error) {
+	return NewLocalRouterWithManager(defaultLocalManager, sid)
+}
+
+// NewLocalRouterWithmanager is the same as NewLocalRouter but takes a specific
+// Localmanager. This is useful to run parallel different local overlays.
+func NewLocalRouterWithManager(ctx *LocalManager, sid *ServerIdentity) (*Router, error) {
```
ineiti: `ctx` -> `lm`?

network/local.go
```diff
@@ -0,0 +1,466 @@
...
+
+	// counter to keep track of how many bytes read / written this connection
+	// has seen.
+	counterSafe
+	manager *LocalManager
+}
+
+// newLocalConn simply init the fields of a LocalConn but do not try to
+// connect. It should not be used as-is, most user wants to call NewLocalConn.
+func newLocalConn(ctx *LocalManager, local, remote endpoint) *LocalConn {
```
ineiti: `ctx` -> `lm` and all further occurences?

network/network.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/dedis/cothority/monitor"
+)
+
+// Conn is the basic interface to represent any communication mean
+// between two host.
+type Conn interface {
+	// Send a message through the connection.
+	// obj should be a POINTER to the actual struct to send, or an interface.
+	// It should not be a Golang type.
+	Send(ctx context.Context, obj Body) error
```
ineiti: Can @Liamsi sign off on using this `Context` that is not used? Else I propose we remove it.

network/network.go
```diff
@@ -0,0 +1,68 @@
...
+	Close() error
+
+	// Type returns the type of this connection
+	Type() ConnType
+	// Gives the address of the remote endpoint
+	Remote() Address
+	// Returns the local address and port
+	Local() Address
+	// XXX Can we remove that ?
+	monitor.CounterIO
```
ineiti: Out of my head I think we'd need to put it into `Conode`, no?

network/network.go
Liamsi: If we don't want this flexibility: "Context type, which carries deadlines, cancelation signals, and other request-scoped values across API boundaries and between processes." 
(which we currently don't seem to use at all), I'm OK with removing it.

network/network.go
nikkolasg: I just put that as `Tx()` & `Rx()` methods so it still implements the monitor.CounterIO interface but not explicitely now. They're usefuls anyway.

network/network.go
nikkolasg: Ok... I'm still hesitating about that because I'm afraid we **should** use this `Context`, rather than say "if we don't use it, then lets drop it"...
but since you're the majority, I abide :)

ineiti: I propose we remove it, and as soon as we have a use-case, we add it in. Or you make up a use-case and we keep it ;)

network/local.go
```diff
@@ -0,0 +1,487 @@
...
+var defaultLocalManager = NewLocalManager()
+
+// endpoint represents one endpoint of a connection.
+type endpoint struct {
+	// addr is the Address of this endpoint.
+	addr Address
+	// uid is a unique identifier of the remote endpoint
+	// it's unique  for each direction:
+	// 127.0.0.1:2000 -> 127.0.0.1:2000 => 14
+	// 127.0.0.1:2000 <- 127.0.0.1:2000 => 15
```
ineiti: Shouldn't this be
`// 127.0.0.1:2000 -> 127.0.0.1:2001 => 14`
?

network/local.go
```diff
@@ -0,0 +1,487 @@
...
+	_, ok := lm.listening[remote]
+	return ok
+}
+
+// setListening marks the address as being able to accept incoming connections.
+// For each incoming connection, fn will be called in a go routine.
+func (lm *LocalManager) setListening(addr Address, fn func(Conn)) {
+	lm.Lock()
+	defer lm.Unlock()
+	lm.listening[addr] = fn
```
ineiti: Is it only me who prefers
```
lm.Lock()
lm.listening[addr] = fn
lm.Unlock()
```
here?

ineiti:
Channels again:
1. every time we implement something that is standard, we _might_ make mistakes
2. when I tried to replace the queues with channels, it was
    1. shorter (which is most of the time better, IMO)
    2. buggy - which _might_ say that there is something not correct going on, even with the queues

We can leave it for now, but I might come back with a channel implementation - it does not change a lot, but it seems to point out some problems in the code.

It's OK for me to merge.

network/local.go
nikkolasg: No  since `it's unique  for each direction`.

ineiti: Yes, but here you connect from one port to itself...

network/local.go
nikkolasg: I like the defer thing, since you can directly check if  the mutex is "correctly" released: `it locks and make sure at the end it is unlocked`. Ok here this is one line above / below, but still. And being consistent on one way of doing one thing is also good I think.

ineiti: As I said: for longer methods or multiple returns I think `defer` is OK, but here for 3 lines it takes more space than it helps. IMO.

network/local.go
nikkolasg: Yes it's to handle one very specific case regarding Local conns: you can listen on `127.0.0.1:2000`, and when you want to connect to it, you must pass one `from Address` and one `to Address`. Nothing rejects you from selecting the same address.

ineiti: I'm not sure the connector's endpoint-port can be 2000 in this case. Confusing to me anyway with a connection from 2000 to 2000. 

I'd put in port for the connector's endpoint that will more probably show up, like `32109` or so.

network/local.go
nikkolasg: let's get down to the nitpick's hole... 
If you nitpick here, that means you will also do it when we again have three lines with a mutex, which can happen sometimes. So in order to not have to have a debate on this each time, we must find a *rule*. IMO, it seems the two choices we have here are:
* (1) Always use `defer`
* (2) When it's only three lines like this, always put the unlock at the last line.

I prefer (1) because it's consistent. It does *NOT* take more space ( come on it's one word, not even one line) & for whoever is reading this code, he should know about defer, if it's confusing for him, then he should not read this code first. And **more importantly**, there's much less chances of doing it right than doing it wrong with `defer`; it should become a good habit `each time there is a Lock, I put a defer Unlock`

So voting time ( @Liamsi this is for you )

network/local.go
nikkolasg: +1

network/local.go
ineiti: If it's confusing for me, I can stop doing code-review and go home? Coole!

Liamsi: Wow, long discussion about this. Here is my vote:
For sake of consistency: I would prefer the `defer` version here. I think we should only use the non-`defer` version if there is "no other" choice (most of the time there is).

network/local.go
```diff
@@ -0,0 +1,487 @@
...
+
+// NewLocalConn returns a new channel connection from local to remote.
+// It mimics the behavior of NewTCPConn and tries to connect right away.
+// It uses the default local manager.
+func NewLocalConn(local, remote Address) (*LocalConn, error) {
+	return NewLocalConnWithManager(defaultLocalManager, local, remote)
+}
+
+// NewLocalConnWithManager is similar to NewLocalConn but takes a specific
+// LocalManager.
```
Liamsi: Could we have a comment about which go-rountines this starts? It looks like an innocent constructor from outside but starts at least [one](https://github.com/dedis/cothority/pull/568/files/dcf609773e5958bd6e867c776df98b73881cd0c2#diff-758a279d81fc93cf992b979af141f431R123) go-rountine

network/local.go
```diff
@@ -0,0 +1,487 @@
...
+		remote:    remote,
+		local:     local,
+		connQueue: newConnQueue(),
+		manager:   lm,
+	}
+}
+
+// NewLocalConn returns a new channel connection from local to remote.
+// It mimics the behavior of NewTCPConn and tries to connect right away.
+// It uses the default local manager.
```
Liamsi: Same as https://github.com/dedis/cothority/pull/568/files/dcf609773e5958bd6e867c776df98b73881cd0c2#r80501285

network/local.go
nikkolasg: Not really. It starts the listening go routine from the other endpoint, so *like tcp*: if you have a tcp listener, and then tries to connect to it, a new go routine will be launched. Not from the connect-er perspective but from the connect-ee.

I could put a comment here but I'm not sure it would be less confusing than anything else... (as an example, is my comment above clear ?)

Liamsi: Not sure I got that. But maybe it's OK to leave it as it is. I just expected some struct to be constructed and was surprised to find that it calls `connect` which does much more.

nikkolasg:
If everybody is happy, then merge it !

PR #569, opened 2016-09-02
[3/7] Refactor network - TCP layer

nikkolasg:
This implements the TCP part of the network package, defining all interfaces in `network.go`. It is like the previous PR but using TCP connections.


ineiti:
Changed base to `refactor_mocking_local`

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+	if err != nil {
+		return EmptyApplicationPacket, fmt.Errorf("Error unmarshaling message type %s: %s", am.MsgType.String(), err.Error())
+	}
+	am.From = c.Remote()
+	return am, nil
+}
+
+// receive is responsible for getting first the size of the message, then the
+// whole message. It returns the raw message as slice of bytes.
+func (c *TCPConn) receive() ([]byte, error) {
```
ineiti: IntelliJ doesn't like having twice the same name in upper/lowercase.

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+
+	// set the size read
+	c.updateRx(uint64(read))
+	return buffer.Bytes(), nil
+}
+
+// how many bytes do we write at once on the socket
+// 1400 seems a safe choice regarding the size of a ethernet packet.
+// https://stackoverflow.com/questions/2613734/maximum-packet-size-for-a-tcp-connection
+const maxChunkSize Size = 1400
```
ineiti: Do we really need this? Can't we just send the whole packet and let TCP handle the splicing up?

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+	var b []byte
+	b, err = am.MarshalBinary()
+	if err != nil {
+		return fmt.Errorf("Error marshaling  message: %s", err.Error())
+	}
+	return c.send(b)
+}
+
+// send takes care of sending this slice of bytes FULLY to the connection
+func (c *TCPConn) send(b []byte) error {
```
ineiti: Same as `Receive`/`receive`. Can we take `Receive`/`receiveRaw`, `Send`/`sendRaw`?

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+	// First write the size
+	packetSize := Size(len(b))
+	if err := binary.Write(c.conn, globalOrder, packetSize); err != nil {
+		return err
+	}
+	// Then send everything through the connection
+	// Send chunk by chunk
+	var sent Size
+	for sent < packetSize {
```
ineiti: We need a test for this to make sure the packets are correctly spliced up.

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+	defer c.receiveMutex.Unlock()
+	// First read the size
+	var total Size
+	if err := binary.Read(c.conn, globalOrder, &total); err != nil {
+		return nil, handleError(err)
+	}
+	b := make([]byte, total)
+	var read Size
+	var buffer bytes.Buffer
+	for read < total {
```
ineiti: We need a test for this to make sure the packets get correctly spliced up.

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+func (c *TCPConn) Type() ConnType {
+	return PlainTCP
+}
+
+// Close ... closes the connection
+func (c *TCPConn) Close() error {
+	c.closedMut.Lock()
+	defer c.closedMut.Unlock()
+	if c.closed == true {
+		return nil
```
ineiti: Is closing a closed connection not an error?

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+	c.closed = true
+	if err != nil {
+		return handleError(err)
+	}
+	return nil
+}
+
+// handleError produces the higher layer error depending on the type
+// so user of the package can know what is the cause of the problem
+func handleError(err error) error {
```
ineiti: Is this still useful?

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+	return ErrUnknown
+}
+
+// TCPListener is the underlying implementation of
+// Host using Tcp as a communication channel
+type TCPListener struct {
+	// the underlying golang/net listener
+	listener net.Listener
+	// the close channel used to indicate to the listener we want to quit
+	quit chan bool
```
ineiti: why is this not in the `listener`-field?

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+	// quitListener is a channel to indicate to the closing function that the
+	// listener has actually really quit
+	quitListener  chan bool
```
ineiti: Same - why is this not in the listener-field?

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+// address (useful if you set it to port :0).
+func NewTCPListener(addr Address) (*TCPListener, error) {
+	l := &TCPListener{
+		quit:         make(chan bool),
+		quitListener: make(chan bool),
+	}
+	return l, l.bind(addr)
+}
+
+func (t *TCPListener) bind(addr Address) error {
```
ineiti: Why is this separated from `NewTCPListener`? I would merge it with the above. Except if it is called somewhere else.

network/tcp.go
```diff
@@ -0,0 +1,422 @@
...
+	receiver := func(tc *TCPConn) {
+		go fn(tc)
+	}
+	return t.listen(receiver)
+}
+
+// listen is the private function that takes a function that takes a TCPConn.
+// That way we can control what to do of the TCPConn before returning it to the
+// function given by the user. fn is called in the same routine.
+func (t *TCPListener) listen(fn func(*TCPConn)) error {
```
ineiti: Why is that separate from `Listen`?

network/tcp.go
nikkolasg: Proposition please !
This "little" receive is made thinking about future type of Conn such as TLS or PURB that will work over TCP but will do something over the received slice of bytes (decrypting or whatever).
Sooo...
`tcpConn.receive_` ? It's not great but I'm tired :D

network/tcp.go
nikkolasg: I've been mostly copying up things over for the TCP part so I did not really dig into the code but I can't remember why we were doing that in the first place ... and you're right, it seems probably a better choice to leave it out to the TCP stack...

network/tcp.go
nikkolasg: ... or let TCP handles the slicing as you suggested above ;)

network/tcp.go
nikkolasg: yes (used throughout this file).

network/tcp.go
nikkolasg: ?? You need to elaborate on that I don't get the question... 
`quit chan bool` is used in the listening routine to check when there is an error if we should quit or not the routine (because we closed this channel in the `Close` method)

network/tcp.go
nikkolasg: Again, this is in prevision of future Conn that relies on TCP. Within PURBConn, I want to be able to access the `TCPConn` , and not an opaque `Conn` interface.

network/tcp.go
ineiti: Is it not simpler to return the underlying golang-network-error?

network/tcp.go
ineiti: Sorry - I thought we could add `quit` and `quitListener` to `net.Listener`. But I think the golang-people will refuse our changes.

network/tcp.go
ineiti: I can't follow you - what do you want to do with PURBs that cannot be done if the methods `Listen` and `listen` get merged?

network/tcp.go
```diff
@@ -0,0 +1,395 @@
...
+// Listening implements the Listener interface
+func (t *TCPListener) Listening() bool {
+	t.listeningLock.Lock()
+	defer t.listeningLock.Unlock()
+	return t.listening
+}
+
+// TCPHost implements the Host interface using TCP connections.
+type TCPHost struct {
+	addr Address
```
ineiti: Should we replace `addr` with `address`?

network/tcp_test.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	log.Lvl3("Closing h3")
+	err = h3.Stop()
+	if err != nil {
+		// try closing the underlying connection manually and fail
+		t.Fatal("Couldn't Stop()", h3)
+	}
+}
+
+/*func TestTCPHostReconnection(t *testing.T) {*/
```
ineiti: Will this be used later on?

ineiti:
I would like to have a test for `TCPConn.receiveRaw` that takes into account sliced receiving an tests whether this happens in a correct way.

Also have a look at #581 for comment changes.

network/tcp.go
nikkolasg: Not really because in golang-network errors are defined in here https://golang.org/pkg/net/#OpError . And it can contains any kind of errors so I tried to make the most common ones in variables so we can do equality like ` if err == ErrClosed` or `if err == ErrEOF` and such, instead of looking at the string the error outputs...

network/tcp.go
nikkolasg: I'm pretty sure they will refuse the way we do it, but maybe we should put an request feature or something because I guess we're not alone in this case. Having a `Close` method on the listener interface would be really nice.

network/tcp.go
nikkolasg: `Conn.receiveRaw` returns an object. `TCPConn.receiveRaw` returns a slice of byte. With access to TCPConn, a PurbConn can take the slice of bytes iwith `receiveRaw`  and then decrypt it so its `Receive()` returns an object.

network/tcp.go
ineiti: It's old, I know. But I think it's the right time to re-think parts of it...

nikkolasg: So please, argue another position & propose ! 
For me it's simpler so we can react to specific events by a simple equality.

network/tcp.go
nikkolasg: wow you're usually the one with the short variable. Golang also is a fan of short variable and I find `addr` pretty specific in that context ...

network/tcp.go
ineiti: Where do you use this? In how many places?

How much more difficult would it be to use only golang-net-errors?

network/tcp.go
ineiti: ;) Let it be then.

network/tcp.go
nikkolasg: Look where I use `ErrClosed`:
```
network/local.go|129 col 10| return ErrClosed
network/local.go|238 col 10| return ErrClosed
network/local.go|287 col 16| return nil, ErrClosed
network/local.go|292 col 15| return nil, ErrClosed
network/router.go|181 col 14| if err == ErrClosed || err == ErrEOF || err == ErrTemp {
network/tcp.go|192 col 10| return ErrClosed
network/tcp.go|207 col 10| return ErrClosed
network/tcp.go|319 col 27| if handleError(err) != ErrClosed {
network/local_test.go|100 col 18| assert.Equal(t, ErrClosed, err)
network/local_test.go|101 col 18| assert.Equal(t, ErrClosed, outgoing.Close())
network/local_test.go|276 col 12| if err != ErrClosed {
network/local_test.go|279 col 18| assert.Equal(t, ErrClosed, outgoing.Close())
```
So it's used quite a lot already. 
I would not ask the question regarding the difficulty, but rather what are the pros/cons of each side ? 
Pros:
* Unified error handling: Having a LocalConn which can return errors that have **the same meaning** (and **are** indeed the same) as with TCPConn, so router use them independently, I find that rather useful.
* Simpler error handling at higher level: 
`if strings.Contains(err.Error(), "Closed") ` vs `if err == ErrClosed` ... 
The first one is error prone and not clear.
Cons:
* We can't have the exact error string. However I coded theses errors while looking at the source code of golang net so I tried to catch most of them. If one is not known, then it can returns this error.

nikkolasg:
* Coverage at 91.5%
* Test of receiveRaw done in `TestTCPConnReceiveRaw` : tell me if it looks correct to you or not.. I think it tests the intended behavior.

network/tcp_test.go
```diff
@@ -0,0 +1,488 @@
...
+		}
+	}
+	// prepare the msg
+	msg := &BigMsg{Array: make([]byte, 7893)}
+	_ = RegisterPacketType(BigMsg{})
+	buff, err := MarshalRegisteredType(msg)
+	assert.Nil(t, err)
+
+	fn := func(c net.Conn) {
+		checking = func() bool {
```
ineiti: Why is this re-defined here?

network/tcp_test.go
```diff
@@ -0,0 +1,488 @@
...
+			currentChunk += maxChunk
+		}
+		slices = append(slices, buff[currentChunk:])
+		// send the size first
+		binary.Write(c, globalOrder, Size(len(buff)))
+		// then send pieces and check if the other side already returned or not
+		for i, slice := range slices[:len(slices)-1] {
+			t.Logf("Will write slice %d/%d...", i, len(slices))
+			if n, err := c.Write(slice); err != nil || n != len(slice) {
+				t.Fatal("Whut?")
```
ineiti: More explicit text please.

network/tcp_test.go
```diff
@@ -0,0 +1,488 @@
...
+			}
+			t.Logf(" OK\n")
```
ineiti: `log.Lvl1("OK")`

network/tcp_test.go
```diff
@@ -0,0 +1,488 @@
...
+			if !checking() {
+				t.Fatal("Already returned even if not finished")
+			}
+			time.Sleep(5 * time.Millisecond)
+		}
+		// the last one should make the other end return
+		t.Logf("Will write last piece...")
```
ineiti: `log.Lvl1`

network/tcp_test.go
```diff
@@ -0,0 +1,488 @@
...
+		if n, err := c.Write(slices[len(slices)-1]); n != len(slices[len(slices)-1]) || err != nil {
+			t.Fatal("could not send the last piece")
+		}
+		t.Logf(" OK\n")
```
ineiti: `log.Lvl1`

network/tcp_test.go
```diff
@@ -0,0 +1,488 @@
...
+		check <- true
+	}
+
+	go func() {
+		ln, err := net.Listen("tcp", "127.0.0.1:0")
+		assert.Nil(t, err)
+		addr <- ln.Addr().String()
+		c, err := ln.Accept()
+		assert.Nil(t, err)
+		// do the thing
```
ineiti: not helping - what thing?

network/tcp_test.go
```diff
@@ -0,0 +1,488 @@
...
+		for currentChunk+maxChunk < len(buff) {
+			slices = append(slices, buff[currentChunk:currentChunk+maxChunk])
+			currentChunk += maxChunk
+		}
+		slices = append(slices, buff[currentChunk:])
+		// send the size first
+		binary.Write(c, globalOrder, Size(len(buff)))
+		// then send pieces and check if the other side already returned or not
+		for i, slice := range slices[:len(slices)-1] {
+			t.Logf("Will write slice %d/%d...", i, len(slices))
```
ineiti: `log.Lvlf1`

network/tcp_test.go
```diff
@@ -0,0 +1,488 @@
...
+
+	de := dummyErr{true, true}
+	assert.Equal(t, ErrTemp, handleError(&de))
+	de.temporary = false
+	assert.Equal(t, ErrTimeout, handleError(&de))
+	de.timeout = false
+	assert.Equal(t, ErrUnknown, handleError(&de))
+}
+
+/*func TestTCPHostReconnection(t *testing.T) {*/
```
ineiti: Was this code that is supposed to be re-enabled later?

ineiti:
General remark: `assert.Equal` should be replaced by `require.Equal` for most of the tests here. `assert.Equal` will continue if the test fails, while `require.Equal` will stop the test and go to the next test.

network/tcp_test.go
nikkolasg: .. for no reason !

app/lib/config/config.go
```diff
@@ -1,288 +0,0 @@
-package config
```
ineiti: Why are two files from `app/lib/config` deleted in this PR?

nikkolasg: I don't know I messed up somewhere between all the branches > <

nikkolasg: They are not supposed to be in this PR, but later.

nikkolasg:
@ineiti all of your comments are corrected + `require` is in place. Still don't know why that does not make your comments disappear...

ineiti:
Do you know why the files of `app/lib/config` disappear?

network/struct.go
```diff
@@ -17,80 +17,80 @@ import (
...
 )
 
-// MaxRetryConnect defines how many times should we try to connect
+// MaxRetryConnect defines how many times we should try to connect.
 const MaxRetryConnect = 10
 
-// MaxIdentityExchange is the maximum time waited for an exchange of
-// the identity to happen.
+// MaxIdentityExchange is the timeout for an identityExchange.
 const MaxIdentityExchange = 5 * time.Second
```
Liamsi: Looks like this isn't used (but maybe in one of the following PRs which depend on this one)

network/struct.go
```diff
@@ -17,80 +17,80 @@ import (
...
+// been shut down).
 var ErrEOF = errors.New("EOF")
 
-// ErrCanceled means something went wrong with the sending or receiving
+// ErrCanceled means something went wrong in the sending or receiving part.
 var ErrCanceled = errors.New("Operation Canceled")
 
-// ErrTemp is a temporary error
+// ErrTemp is a temporary error, recovery possible.
 var ErrTemp = errors.New("Temporary Error")
```
Liamsi: General remark / not directly related to this PR: when does such a temporary error occur. Is it neccessary? 
Also, seems like these constants replicate these ones: https://golang.org/pkg/io/#pkg-variables
(No change in this PR necessary)

network/struct.go
```diff
@@ -17,80 +17,80 @@ import (
...
+	// Mutex on peers.
 	peersMut sync.Mutex
-	// its listeners
+	// The listener attached to this host.
 	listener net.Listener
-	// the close channel used to indicate to the listener we want to quit
+	// the close channel used to indicate to the listener we want to quit.
 	quit chan bool
 	// quitListener is a channel to indicate to the closing function that the
 	// listener has actually really quit
```
Liamsi: +1 for polishing the comments!

network/tcp.go
```diff
@@ -0,0 +1,411 @@
...
+	return r, nil
+}
+
+// TCPConn implements the Conn interface using plain, unencrypted TCP.
+type TCPConn struct {
+	// The name of the endpoint we are connected to.
+	endpoint Address
+
+	// The connection used
+	conn net.Conn
```
Liamsi: If this is explicitly a TCP connection, why not use `conn net.TCPConn` instead of the interface?

network/tcp.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	c := TCPConn{
+		endpoint: addr,
+		conn:     conn,
+	}
+	return &c, err
+}
+
+// Receive calls the receive routine to get the bytes from the connection then
+// it decodes the buffer.
```
Liamsi: Change to: 
```
// Receive receives the raw bytes from the connection and
// tries decoding them into a `Packet`.
```
(To my opinion a caller from outside shouldn't care about the "receive routine")

network/tcp.go
Liamsi: Definitely! This method should be fully covered and all edge-cases/errors should have small tests...

Liamsi: (If this isn't already the case, otherwise sorry for the noise; I'll continue reading ;-)

network/tcp.go
```diff
@@ -0,0 +1,411 @@
...
+	if err != nil {
+		return fmt.Errorf("Error marshaling  message: %s", err.Error())
+	}
+	return c.sendRaw(b)
+}
+
+// sendRaw writes the number of bytes of the message to the network then the
+// whole message b in slices of size maxChunkSize.
+// In case of an error it aborts and returns error.
+func (c *TCPConn) sendRaw(b []byte) error {
```
Liamsi: Same as above: Should be fully tested (also everything that could go wrong). There should be a `TestsendRaw` even if it is an unexported method. It's too important...

network/struct.go
nikkolasg: I don't know actually when these occurs, I just looked at the golang source and saw these errors so  I picked the most important ones. And they do not replicate the one that you linked to, i.e. there's no `ErrClosed` for example. These errors are more specific to the network operations, and not juste "generic" IO operations.

Liamsi: I would remove TempErr then. I mean, we introduce these errors/vars, but we don't know when they occur? Doesn't sound very reasonable ;-)
`ErrClosedPipe` is golang's `ErrClosed`I would guess. 
But yeah, my nitpicking is orthogonal to this PR.

network/tcp.go
Liamsi: They return the interface type because they have to (`net.Listen` is a general function that can create all sorts of connections); here, the opposite is the case: We explicitly want a TCPConn (and that's what `net.Listen("tcp",address)` implicitly returns, too).

Liamsi: This also is only nitpicking and since this isn't exported it's OK to leave it as it is. I just like to be more explicit. Keeping it a `net.Conn` has the advantage of being able to mock this with another implementation (for testing purposes).

sda/service.go
```diff
@@ -260,31 +260,34 @@ func (s *serviceManager) Process(data *network.Packet) {
...
-// TODO
-func (s *serviceManager) AvailableServices() []string {
-	panic("not implemented")
+// AvailableServices returns a list of all services available to the serviceManager.
+// If no services are instantiated, it returns an empty list.
+func (s *serviceManager) AvailableServices() (ret []string) {
+	for id := range s.services {
+		ret = append(ret, ServiceFactory.Name(id))
+	}
+	return
```
Liamsi: This code is from another branch, right?

network/struct.go
nikkolasg: Please look at the golang doc:
+  https://golang.org/pkg/net/#UnknownNetworkError.Temporary
+ https://golang.org/pkg/net/#AddrError.Temporary
They have absolutely no comment on them that's why I said "I don't know". Should I remove them, we remove the possiblity of getting them.  But it's true that we should know when they appear, I just mimicked the golang net stack. I'll remove `TempErr` then.

sda/service.go
nikkolasg: erf ... I'm loosing my mind with all these branches to maintain concurrently -_-"
Yes.

network/struct.go
Liamsi: Thanks for the clarification! I'm still not sure if it's a good idea that we mimic them after all... But yeah, if this currently isn't used (or only in the tests), you can remove it. Because we do not seem to handle it/work with it.

Liamsi: BTW: There is no TempError but errors return if they are temporary or not. I guess to signal: Here it makes sense to try again later, or no chance, sth. is really wrong.
https://golang.org/src/net/net.go?s=15732:15777#L360

Liamsi: Sorry wrong link: https://golang.org/src/net/net.go?s=#L360

sda/service.go
Liamsi: Yeah, it's confusing...  But no need to change anything.

network/tcp_test.go
```diff
@@ -19,6 +19,125 @@ type BigMsg struct {
...
+			}
+			continue
+		}
+		// - 4 is for the size, uint32_t
+		if test.bytesExpected != test.conn.writtenBytes-4 {
+			t.Error(i, "Wrong number of bytes? ", test.bytesExpected, test.conn.writtenBytes)
+		}
+	}
+}
+
```
Liamsi: +1

nikkolasg:
Ok I made all changes from @Liamsi exept one with the `net.TCPConn` since I'm mocking the connection to test the `sendRaw` and can't mock if I'm using the `net.TCPConn` , only the `net.Conn`

network/tcp_test.go
```diff
@@ -0,0 +1,540 @@
...
+func (f *fakeConn) SetDeadline(t time.Time) error {
+	return nil
+}
+
+func (f *fakeConn) SetReadDeadline(t time.Time) error {
+	return nil
+}
+func (f *fakeConn) SetWriteDeadline(t time.Time) error {
```
ineiti: I wonder whether it would work to embed `TCPConn` into `fakeConn` and then don't re-define those messages, but only the ones you want to change? Would make it more readable.

nikkolasg: Nice one, it's possible and done.

Issue #522, opened 2016-08-17
Re-write the network-layer using a Router


ineiti:
Instead of having the `sda/Host` handle connections, put all connection-handling into a `network/Router` who interacts with `Listeners` and has its own `Connections`. According to the document in

https://docs.google.com/document/d/1Z6nCL76JPMMZLmpqaijPgvjHo5DxBWtFjj6db_VV_SI/edit?usp=sharing

ineiti:
This has been successfully implemented.

PR #570, opened 2016-09-02
[4/7] Refactor network - Router + SDA/Network separation

nikkolasg:
This PR implements the central struct of the network package `Router` and shows the separation of the network code from `sda/` to `network/`. Implements the idea of #522 

**TODO**
[ ] Implement an way to get info from `Router` such as `Usage() RouterUsage` where `RouterUsage` struct contains all info we need.


nikkolasg:
92.1 % coverage

nikkolasg:
Ok from now on, I think you can run `go test ` inside the `network/` package and it should work.
Now is 90.3% of coverage. I'll stop at this number for the moment.. :D

network/router.go
```diff
@@ -0,0 +1,394 @@
...
+	r.connsMut.Unlock()
+
+	// wait for all handleConn to finish
+	r.wg.Wait()
+	return nil
+}
+
+// Closed returns true if the router is closed (or is closing). For a router
+// to be closed means that a call to Stop() must have been made.
+func (r *Router) Closed() bool {
```
ineiti: If `Closed` is used only once (as IntelliJ says), integrate it in the caller.

network/router.go
```diff
@@ -0,0 +1,394 @@
...
+	r.connsMut.Lock()
+	defer r.connsMut.Unlock()
+	_, okc := r.connections[remote.ID]
+	if okc {
+		log.Lvl5("Connection already registered. Appending new connection to same identity.")
+	}
+	r.connections[remote.ID] = append(r.connections[remote.ID], c)
+}
+
+func (r *Router) reset() {
```
ineiti: This function looks strange - what does it do?

network/router.go
```diff
@@ -0,0 +1,394 @@
...
+	r.closedMut.Lock()
+	r.isClosed = false
+	r.closedMut.Unlock()
+}
+
+func (r *Router) launchHandleRoutine(dst *ServerIdentity, c Conn) {
```
ineiti: Shouldn't we include this into `handleConn`?

network/router.go
```diff
@@ -0,0 +1,394 @@
...
+	r.wg.Add(1)
+	go r.handleConn(dst, c)
+}
+
+// Close shuts down all network connections and returns once all processing go
+// routines are done.
+func (r *Router) stopHandling() error {
```
ineiti: Why is this not in `Stop`? It's used only once.

network/router.go
```diff
@@ -0,0 +1,394 @@
...
+// to be closed means that a call to Stop() must have been made.
+func (r *Router) Closed() bool {
+	r.closedMut.Lock()
+	defer r.closedMut.Unlock()
+	return r.isClosed
+}
+
+// close set the isClosed variable to true, any subsequent call to Closed()
+// will return true.
+func (r *Router) close() {
```
ineiti: Only used once, according to IntelliJ - can you include it in `stopHandling`?

network/router.go
```diff
@@ -0,0 +1,394 @@
...
+	log.Lvl5("Message sent")
+	return nil
+}
+
+func (r *Router) connect(si *ServerIdentity) (Conn, error) {
+	c, err := r.host.Connect(si.Address)
+	if err != nil {
+		return nil, err
+	}
+	if err := r.negotiateOpen(si, c); err != nil {
```
ineiti: Is this OK to have the `negotiateOpen` in the router instead of the `r.host.Connect`? I think it's the hosts business of how to do the `Connect`, no?

network/router.go
nikkolasg: I had quite a hard time figuring out how to make sure every go routines finishes when calling `Stop()`: I use a `WaitGroup`, and the `wg.Add(1)` *must* be done before launching the go routine (see [godoc](https://golang.org/pkg/sync/#WaitGroup). And this is launched at two places so .. :)

network/router.go
nikkolasg: I've put it in `Stop()`. It makes sure we can re-use the router again after a `Stop()` using `Start()`, although I'm not sure it's really necessary.

network/router.go
nikkolasg: I'm not sure about this one, I guess it can still be useful to know if the `Router` is closed or not *externally* from the network package ? And also, it makes the code much more clean because I use this `Closed` in a `if` condition, so I must open the lock before the condition and unlock it at two places (inside the `if` and in the `else`) because I don't want this mutex to be locked during the full lifetime of the function.

network/router.go
nikkolasg: After-discussion-conclusion: let's keep it in here, but remove the sending of the identity from the listener perspective, since the connect-er knows already the listener identity since he is connecting to it.

ineiti: Which means there is no symmetric `negotiateOpen` anymore but a `waitForClientID` and `sendClientID`?

network/router.go
```diff
@@ -0,0 +1,394 @@
...
+				// remote connection closed
+				log.Lvl3(r.address, "handleConn with closed connection: stop (dst=", remote.Address, ")")
+				return
+			}
+			// weird error let's try again
+			log.Lvl3(r.id, "Error with connection", address, "=>", err)
+			continue
+		}
+
+		if err := r.Dispatch(&packet); err != nil {
```
ineiti: Aesthetic/vote: replace the `continue` with an `else` here?

network/router.go
nikkolasg: yes

network/router.go
```diff
@@ -0,0 +1,394 @@
...
+// Stop will stop the listening routine, and stop any routine of handling
+// connections. Calling r.Start(), then r.Stop() then r.Start() again leads to
+// an undefined behaviour. Callers should most of the time re-create a fresh
+// Router.
+func (r *Router) Stop() error {
+	var err error
+	if r.host.Listening() {
+		err = r.host.Stop()
+	}
+	err2 := r.stopHandling()
```
ineiti: I propose to include `stopHandling`, `reset` and `close` in here, as they are only used once.

I'd also remove the `closedMut` and use `connsMut` instead.

nikkolasg: It's done it's done sorry did not push it soon enough.

network/router.go
```diff
@@ -0,0 +1,361 @@
...
+// connections. Calling r.Start(), then r.Stop() then r.Start() again leads to
+// an undefined behaviour. Callers should most of the time re-create a fresh
+// Router.
+func (r *Router) Stop() error {
+	var err error
+	if r.host.Listening() {
+		err = r.host.Stop()
+	}
+	// set the isClosed to true
+	r.closedMut.Lock()
```
ineiti: As I wrote in a comment that now (correctly) vanished: why not use `connsMut` instead of `closedMut`?

network/router.go
nikkolasg: It's maybe only me it seems, but instead of having hierarchies of `if` inside `if` etc where the "normal" behavior happens *inside* the if, I prefer that the `if` checks if something is wrong, and the normal behavior stays at the most possible top level indentation so it's way more easy to follow. 
Same story here, I could put the `r.Dispatch` inside an `else` but then,  the `Dispatch` which is the **normal behavior** is not emphasized, rather put alongside a big `if` which checks if something is wrong.
Most of the times it's like that in the wild, for example, the first example to me seems cleaner than the second:
```
f, err := os.Open("filename.ext")
if err != nil {
    log.Fatal(err)
}
// do something with the open *File f
// then something else
```
vs:
```
f, err := os.Open("filename.ext")
if err != nil {
    log.Fatal(err)
} else {
    // do something with the open *File f
}
// then something else
```
Yeah sorry again, I put up a large comment ... :D

network/router.go
ineiti: I see what you're saying about the `if` and looked for a nicer way, but didn't find any... So we can keep it.

network/router.go
ineiti: Or would it even be OK to use a 'global' mutex for `closedMut` and `connsMut`, embedded in `Router`?

network/client.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func newClient(c func(own, remote *ServerIdentity) (Conn, error)) *Client {
+	return &Client{c}
+}
+
+var baseID uint64
+var baseIDLock sync.Mutex
+
+var timeoutResponse = 10 * time.Second
```
Liamsi: Change to 
`const TimeoutResponse = 10 * time.Second` 
so it's transparent to the user/developer  from outside.

Liamsi: Ah, I see you chnage it in the tests (which makes sense)... Sorry for the noise.

network/client.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Send will send the message to the destination service and return the
+// reply.
+// The error-handling is done using the ErrorRet structure which can be returned
+// in place of the standard reply. This method will catch that and return
+// the appropriate error as a network.Packet.
+func (cl *Client) Send(dst *ServerIdentity, msg Body) (*Packet, error) {
```
Liamsi: Add a comment about that sending will timeout after 10s.

network/router.go
```diff
@@ -0,0 +1,343 @@
...
+	connections map[ServerIdentityID][]Conn
+	connsMut    sync.Mutex
+
+	// boolean flag indicating that the router is already clos{ing,ed}
+	isClosed  bool
+	closedMut sync.Mutex
+
+	// we wait that all handleConn routines are done
+	wg    sync.WaitGroup
+	wgMut sync.Mutex
```
ineiti: Not used here - if it's not used in later branches, delete.

network/router.go
```diff
@@ -0,0 +1,343 @@
...
+	if nm.MsgType != ServerIdentityType {
+		return nil, fmt.Errorf("Received wrong type during negotiation %s", nm.MsgType.String())
+	}
+	// Set the ServerIdentity for this connection
+	e := nm.Msg.(ServerIdentity)
+	return &e, nil
+
+}
+
+func sendServerIdentity(own *ServerIdentity, c Conn) error {
```
ineiti: Do we really need this or can we write that in the `Client` directly? It's one line that will be removed later probably.

network/router.go
```diff
@@ -0,0 +1,343 @@
...
+// and then registers the connection.
+func (r *Router) sendServerIdentity(c Conn) error {
+	err := sendServerIdentity(r.id, c)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func receiveServerIdentity(c Conn) (*ServerIdentity, error) {
```
ineiti: Merge that with `Router.receiveServerIdentity`.

network/router.go
```diff
@@ -0,0 +1,343 @@
...
+
+// Listening returns true if this router is listening or not.
+func (r *Router) Listening() bool {
+	return r.host.Listening()
+}
+
+// receiveServerIdentity takes a fresh new conn issued by the listener and
+// wait for the server identities of the remote party. It returns
+// the ServerIdentity of the remote party and register the connection.
+func (r *Router) receiveServerIdentity(c Conn) (*ServerIdentity, error) {
```
ineiti: I would even merge that with `Router.Start` - but if you think it makes the method grow too big, then leave it here.

network/router.go
```diff
@@ -0,0 +1,343 @@
...
+		return nil, err
+	}
+	log.Lvl4(r.address, "Identity received from ", dst.Address)
+	r.registerConnection(dst, c)
+	return dst, nil
+}
+
+// sendIdentity takes a fresh issued new Conn. It sends the router's identity
+// and then registers the connection.
+func (r *Router) sendServerIdentity(c Conn) error {
```
ineiti: Merge this with the `Router.connect`-method. It only adds a new method but offers nothing else...

network/router.go
```diff
@@ -0,0 +1,343 @@
...
+	err := sendServerIdentity(r.id, c)
+	if err != nil {
```
ineiti: Doesn't matter, as I'd like this merged in `Router.connect` anyway, but the lines can be shorter like:

```
return c.Send(r.id)
```

network/router.go
nikkolasg: I've put everything under `connsMut` ...

network/router.go
nikkolasg: Yeah I think it's a bit too big, so I kept it, but I removed the `SendServerIdentity`.

network/client.go
```diff
@@ -0,0 +1,112 @@
+package network
```
ineiti: Why is this in `network` and not in `sda`?

network/client.go
```diff
@@ -0,0 +1,112 @@
...
+	// lower the chance of having a collision in the Router.
+	baseIDLock.Lock()
+	id := baseID
+	baseID++
+	baseIDLock.Unlock()
+	sid := NewServerIdentity(kp.Public, NewAddress(dst.Address.ConnType(), "client:"+strconv.FormatUint(id, 10)))
+
+	var c Conn
+	var err error
+	c, err = cl.connector(sid, dst)
```
ineiti: `c, err := cl.connector(sid, dst)` ?

network/client.go
nikkolasg: Since we emulate network, we must emulate clients also, so it have to have access to the local network stack... And more basically, it's related to network operations so it should be there IMO :)

network/client.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}()
+
+	select {
+	case resp := <-msgCh:
+		return &resp, nil
+	case err := <-errCh:
+		return nil, err
+	case <-time.After(timeoutResponse):
+		return &Packet{}, errors.New("Timeout on sending message")
```
ineiti: Why is this not returning `nil`?

nikkolasg: Because I copied the sda code without thoroughly looking at the code for the client part ... :) Will be done.

network/client.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	status, ok := em.Msg.(StatusRet)
+	if !ok {
+		return nil
+	}
+	statusStr := status.Status
+	return errors.New("Remote-error: " + statusStr)
+}
+
+func init() {
```
ineiti: Inits to the power! Move to top of file.

network/client.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+}
+
+// StatusRet is used when a status is returned - mostly an error
+type StatusRet struct {
+	Status string
+}
+
+// StatusOK is used when there is no error but nothing to return
+var StatusOK = &StatusRet{""}
```
ineiti: Is this used in the service-apis?

nikkolasg: Not sure but it's a good thing to have in my opinion, if it's possible to register a function using the `ServiceProcessor` using this as a return value. I mean, it will surely come the time when a Service does not need to answer to a client (for example an update or something).

network/dispatch.go
```diff
@@ -0,0 +1,129 @@
...
+// BlockingDispatcher is implemented, which is a blocking dispatcher.
+// Later, one can easily imagine to have a dispatcher with one worker in a
+// goroutine or a fully fledged producer/consumers pattern in go routines.
+// Each Processor that wants to receive all messages of a specific
+// type must register itself to the dispatcher using `RegisterProcessor()`.
+// The network layer must call `Dispatch()` each time it receives a message, so
+// the dispatcher is able to dispatch correctly to the right Processor for
+// further analysis.
+type Dispatcher interface {
+	// RegisterProcessor is called by a Processor so it can receive all packets
```
Liamsi: I haven't really read further, but: Is this really called by a processor? The signature looks different!

nikkolasg: ? yes, if you have a struct which implements the `Processor` interface, then it must registers itself to a `Dispatcher` so it can get the message it wants.

Liamsi: But the method is called by a `Dispatcher` then and not by a `Processor`.  I'll continue reading. Maybe, I just don't get it yet.

network/dispatch.go
```diff
@@ -0,0 +1,129 @@
...
+// Dispatch will directly call the right processor's method Process. It's a
+// blocking call if the Processor is blocking !
+func (d *BlockingDispatcher) Dispatch(packet *Packet) error {
+	d.Lock()
+	var p Processor
+	if p = d.procs[packet.MsgType]; p == nil {
+		d.Unlock()
+		return errors.New("No Processor attached to this message type " + packet.MsgType.String())
+	}
+	d.Unlock()
```
Liamsi: For consistency: should be `defer d.Unlock()` on top

network/router.go
```diff
@@ -0,0 +1,309 @@
...
+	r.address = h.Address()
+	r.host.Listening()
+	h.Listening()
+	return r
+}
+
+// Start will start the listening routine of the underlying Host. It is a
+// blocking call until the listening is done (by calling r.Stop()).
+func (r *Router) Start() {
+	// The function given to the listener  does the exchange of ServerIdentity
```
Liamsi: outdated comment, or not?

network/router.go
```diff
@@ -0,0 +1,309 @@
...
+	// and pass the connection along to the router.
+	err := r.host.Listen(func(c Conn) {
+		//dst, err := r.exchangeServerIdentity(c)
```
Liamsi: remove commented code

network/router.go
```diff
@@ -0,0 +1,309 @@
...
+		dst, err := r.receiveServerIdentity(c)
+		if err != nil {
+			log.Error("ExchangeServerIdentity failed:", err)
```
Liamsi: Should be: `Receiving ServerIdentity failed:`

network/router.go
Liamsi: Although I dislike the `continue` key-word, I agree with @nikkolasg on this one. Option one in his example looks more like idiomatic go-code.

network/router.go
```diff
@@ -0,0 +1,309 @@
...
+		}
+		// and release one on the waitgroup
+		r.wg.Done()
+	}()
+	address := c.Remote()
+	log.Lvl3(r.address, "Handling new connection to ", remote.Address)
+	for {
+		packet, err := c.Receive()
+		// So the receiver can know about the error
+		packet.SetError(err)
```
Liamsi: Why bother and set the packet error here? If `err != nil` that packet instance will never be used/dispatched, or not?

network/router.go
ineiti: Dang, been downvoted again ;)

network/dispatch.go
```diff
@@ -0,0 +1,129 @@
...
+
+// RoutineDispatcher is a Dispatcher that will dispatch messages to Processor
+// in a go routine. RoutineDispatcher creates one go routine per messages it
+// receives.
+type RoutineDispatcher struct {
+	*BlockingDispatcher
+}
+
+// NewRoutineDispatcher returns a fresh RoutineDispatcher
+func NewRoutineDispatcher() *RoutineDispatcher {
```
ineiti: Why don't we use `RoutineDispatcher` in the sda?

network/dispatch_test.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type basicMessage struct {
+	Value int
+}
+
+var basicMessageType = RegisterPacketType(&basicMessage{})
+
+func TestBlockingDispatcher(t *testing.T) {
+	defer log.AfterTest(t)
```
ineiti: can be removed, as `log.MainTest` is called.

network/dispatch_test.go
```diff
@@ -0,0 +1,111 @@
...
+		Msg:     basicMessage{10},
+		MsgType: basicMessageType})
+
+	if !found {
+		t.Error("ProcessorFunc should have set to true")
+	}
+}
+
+func TestRoutineDispatcher(t *testing.T) {
+	defer log.AfterTest(t)
```
ineiti: can be removed, as log.MainTest is called.

network/local.go
```diff
@@ -7,17 +7,14 @@ import (
 	"time"
 )
 
-// NewLocalRouter returns a fresh router which uses only local queues. It uses
-// the default local manager.
-// If you need multiple independent local-queues, use NewLocalRouterWithManager.
-// In case of an error it is returned together with a nil-Router.
+// NewLocalRouter returns a fresh router which uses local connections. It takes
```
ineiti: Why did my comment-changes get lost?

ineiti: It seems to me this is an old version of the file...

network/router.go
```diff
@@ -0,0 +1,309 @@
...
+func (r *Router) Listening() bool {
+	return r.host.Listening()
+}
+
+// receiveServerIdentity takes a fresh new conn issued by the listener and
+// wait for the server identities of the remote party. It returns
+// the ServerIdentity of the remote party and register the connection.
+func (r *Router) receiveServerIdentity(c Conn) (*ServerIdentity, error) {
+	// Receive the other ServerIdentity
+	nm, err := c.Receive()
```
ineiti: Does this timeout? If not, should we add a timeout here?

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+// Test the automatic connection upon request
+func TestRouterAutoConnectionTCP(t *testing.T) {
+	testRouterAutoConnection(t, NewTestRouterTCP)
+}
+func TestRouterAutoConnectionLocal(t *testing.T) {
+	testRouterAutoConnection(t, NewTestRouterLocal)
+}
+
+func testRouterAutoConnection(t *testing.T, fac routerFactory) {
+	//	log.TestOutput(true, 5)
```
ineiti: Remove unused comment.

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+	fmt.Println("Before fac()")
```
ineiti: Replace `fmt.Println` with `log.Lvl1`.

network/router_test.go
ineiti: Remove unused comment.

network/router_test.go
ineiti: And all further occurrences.

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+		t.Fatal("Should not be able to send")
+	}
+	fmt.Println("after send() #1")
+	h2, err := fac(2008)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = h1.Send(h2.id, nil)
+	if err == nil {
```
ineiti: I still prefer `log.ErrFatal(h1.Send(h2.id, nil))` or `log.ErrFatal(err)`

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+	// Receive the message
+	msg := <-proc.relay
+	if msg.I != 12 {
+		t.Fatal("Simple message got distorted")
+	}
+
+	h12 := h1.connection(h2.id.ID)
+	h21 := h2.connection(h1.id.ID)
+	if h12 == nil {
+		t.Error("h1 has no connection to h2")
```
ineiti: is this an `Error` or a `Fatal`?

ineiti: Lines 120-124 could also be written as:
```
require.NotNil(h12)
require.NotNil(h21)
```

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+		time.Sleep(250 * time.Millisecond)
+	}()
+
+	proc := &simpleMessageProc{t, make(chan SimpleMessage)}
+	h1.RegisterProcessor(proc, SimpleMessageType)
+	h2.RegisterProcessor(proc, SimpleMessageType)
+
+	msgSimple := &SimpleMessage{3}
+	err := h1.Send(h2.id, msgSimple)
+	if err != nil {
```
ineiti: `require.Nil(err, "Couldn't send from h2 -> h1")`

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+		t.Fatal("Couldn't send from h2 -> h1:", err)
+	}
+	decoded := <-proc.relay
+	if decoded.I != 3 {
```
ineiti: `require.Equal(3, decoded.I, "Received message from h2 -> h1 is wrong")`

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+	// a go routine). Since we try to avoid random timeout, let's send a msg
+	// from host2 -> host1.
+	assert.Nil(t, h2.Send(h1.id, msgSimple))
+	decoded = <-proc.relay
+	assert.Equal(t, 3, decoded.I)
+
+	written := h1.Tx()
+	read := h2.Rx()
+	if written == 0 || read == 0 || written != read {
+		t.Logf("Tx = %d, Rx = %d", written, read)
```
ineiti: `log.Lvlf1`

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+	} else if p.actual > p.expected {
+		p.t.Fatal("Too many response ??")
+	}
+	msg := pack.Msg.(SimpleMessage)
+	p.r.Send(pack.ServerIdentity, &msg)
+}
+
+// Makes a big mesh where every host send and receive to every other hosts
+func testRouterLotsOfConn(t *testing.T, fac routerFactory) {
+	log.TestOutput(true, 2)
```
ineiti: remove

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+	nbrRouter := 2
+	// create all the routers
+	routers := make([]*Router, nbrRouter)
+	var wg1 sync.WaitGroup
+	wg1.Add(nbrRouter)
+	for i := 0; i < nbrRouter; i++ {
+		go func(j int) {
+			r, err := fac(2000 + j)
+			if err != nil {
```
ineiti: `log.ErrFatal(err)`

network/dispatch.go
nikkolasg: No no, ex:
```
d := NewBlockingDispatcher()
p := MyProcessor()
d.RegisterProcessor(MyMessageType, p)
```
Ok while writing I see what you mean, but saying that the Dispatcher calls the method is also ambiguous (because the dispatcher does not know the processor necessarily, usually, it's the processor who has access to the Dispatcher so it can register itself to it) so I'll just remove that part of the comment.

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+			// expect nbrRouter - 1 messages
+			proc := newNSquareProc(t, r, nbrRouter-1, &wg2)
+			r.RegisterProcessor(proc, SimpleMessageType)
+			for k := 0; k < nbrRouter; k++ {
+				if k == j {
+					// don't send to yourself
+					continue
+				}
+				// send to everyone else
+				if err := r.Send(routers[k].id, &SimpleMessage{3}); err != nil {
```
ineiti: `log.ErrFatal(r.Send(routers[k].id, &SimpleMessage{3}))`

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+					t.Fatal(err)
+				}
+			}
+		}(i)
+	}
+	wg2.Wait()
+	for i := 0; i < nbrRouter; i++ {
+		r := routers[i]
+		if err := r.Stop(); err != nil {
```
ineiti: `log.ErrFatal(r.Stop())`

network/dispatch.go
nikkolasg: I'm not sure here. The reason is I want to unlock **before** the call to `p.Process()` happens, so other messages can still be dispatched while the the processor is processing the message. If I do a defer, the unlock will only occurs **after** the processing part...

network/router_test.go
```diff
@@ -0,0 +1,365 @@
...
+	// stop everything
+	log.Lvl4("Closing connections")
+	if err := router2.Stop(); err != nil {
+		t.Fatal("Couldn't close host", err)
+	}
+	if err := router1.Stop(); err != nil {
+		t.Fatal("Couldn't close host", err)
+	}
+	<-done
+}
```
ineiti: Lots of more possibilities for `log.ErrFatal` or `require.Nil` or `require.NotNil`. In general, since I discovered `require`, I think most of our `assert`s should be `require`, as we cannot continue for most of the `assert`s we have.

network/dispatch.go
nikkolasg: Because spawning an infinite number of go routine for incoming messages seems not a good idea ... ?

network/local.go
nikkolasg: Everything comments ? It's possible, I maybe messed up because for the `refactor_mocking_local` branch , I mostly worked on `refactor_mocking` (so I could build & test) and then merg after into `refactor_mocking_local` . Since your comments have been merged directly on `refactor_mocking_local` maybe I lost a version ?
I'm happy to arrive at this point of the series of PR where I can build directly on it, it was really a nightmare .. ><

nikkolasg: I'll re-write the comments.

network/dispatch.go
ineiti: Oups, upvoted too fast. You could put it in a getter, as it's used twice, then return an error if the processor doesn't exist.

network/dispatch.go
ineiti: Aww, come on. go-routines are lightweight. And I'm sure `Conode` and `Overlay` are re-entrant, no?

network/local.go
ineiti: There are also some other changes we did in 3/7 - a short overview of the file in `mocking_local` looks good, I'll copy that to here. Except:

```
// 127.0.0.1:2000 -> 127.0.0.1:7869 => 14
// 127.0.0.1:7869 <- 127.0.0.1:2000 => 15
```

Have a good look at it and tell me what's wrong ;)

sda/local.go
```diff
@@ -26,8 +26,20 @@ type LocalTest struct {
...
 	// All single nodes
 	Nodes []*TreeNodeInstance
+	// are we running tcp or local layer
+	mode string
+	// the context for the local connections
+	// it enables to have multiple local test running simultaneously
+	ctx *network.LocalManager
 }
 
+var (
```
Liamsi: Shouldn't this be `const (` instead of `var (`

network/router.go
```diff
@@ -0,0 +1,306 @@
...
+		packet.From = address
+		packet.ServerIdentity = remote
+
+		// whether the router is closed
+		if r.Closed() {
+			// signal we are done with this go routine.
+			return
+		}
+
+		if err != nil {
```
Liamsi: Now, this could immediately go under `packet, err := c.Receive()` Or is it essential to do the `if r.Closed()` before?

network/dispatch.go
nikkolasg: `go-routines are lightweight` : we can have one go-routine per connections, but one go routine **per messages**, I'm really not sure... On the other hand, a pool of go-routine sounds completely reasonable!  And yes, `Conode` and `Overlay` are thread safe.

network/router.go
Liamsi: I mean:
```golang
packet, err := c.Receive()
if err != nil {
    // ...
}
if r.Closed() {
    // ..
```

network/router_test.go
nikkolasg: But it's not the same behavior. `log.ErrFatal` will fatal and `t.Fatal` calls `t.FailNow` which does:
`FailNow marks the function as having failed and stops its execution. Execution will continue at the next test or benchmark.`

network/router_test.go
nikkolasg: Why ? It's something that is not supposed to happen, why not using the testing output which is exactly made for that ?

network/router_test.go
nikkolasg: Same as above, do we care or not if it's not the same behavior ?

sda/host.go
```diff
@@ -1,21 +1,14 @@
 package sda
```
nikkolasg: In the next PR.

network/dispatch.go
Liamsi: Now, I understand what you mean. Thanks for the clarification. Yeah, maybe removing the "is called by"-part creates less confusion.

network/router.go
nikkolasg: Not essential but basically, if the router is closed I don't care at all about the error and/or message I just want to quit.

network/router_test.go
nikkolasg: > as we cannot continue for most of the asserts we have

Let's not make big rules for that please because good units tests IMO should only need assert instead of require... but `require` is cool when we need it definitely !!

network/router.go
Liamsi: Or can't you know if the connection is closed, before? Almost everywhere else the error-handling is done immediately after we got the error. I don't see a good reason to not be consistent here.

nikkolasg: Because a `receive` is blocking, so when I close all connections, the `receive` will return and after we check if we are closed or not. If I put the `Closed` before, then some goroutines might leak since the `receive` will never return.

network/dispatch.go
ineiti: My second comment was half-joking. But for later it might be OK to do some go-routines. But I think that a go-routine-pool is overkill and makes more overhead than it saves.

network/router_test.go
ineiti: If the tests fail, it's over anyway, no? So the sooner we have a red Travis, the better. Or use `require.NotNil`.

network/router_test.go
ineiti: So we don't mix different logging-mechanisms.

network/router_test.go
ineiti: Let's put that up for a longer discussion - most of the tests I see don't are `assert` but `require`.

ineiti:
Still waiting for old `network/local.go`, then I'm OK, too.

PR #572, opened 2016-09-02
[6/7] Refactor network - Identity changes

nikkolasg:
This PR shows some changes done to the Identity service + struct. Before, `Identity` embedded a `*sda.Client`, but when calling `Identity.Save`, it also marshalled the `*sda.Client`.  The first impl. changed `sda.Client` to an interface so it could not be marshalled, so I made theses changes. With the current version, we could roll back to what was first present, but I think it still makes sense to only marshal and unmarshal what we need only.


ineiti:
Is this compatible with the latest `services/identity`-changes in `master`?

nikkolasg:
> Is this compatible with the latest services/identity-changes in master?

After doing a quick check with the master version, it seems that:
* the master version had changes between, in the service.go
* this version brings changes in the api.go 

What do you mean exactly by compatible ? To me this PR still makes sense because the problem it try to resolves (marshalling a sda.Client) is still present in the master version. Whether we'll be able to merge easily both version, that I don't know but that's why git exists.

PR #571, opened 2016-09-02
[5/7] Refactor network - SDA changes

nikkolasg:
This PR shows the changes in SDA; mainly:
* changes name of method (`SendRaw` => `Send` etc)
* uses of Address
* changes most of the tests so they use the `LocalTest` structs
* removed most of the custom timeouts

**TODO**
[ ] Make **all** sda tests using `sda.LocalTest`

Issue #170, opened 2016-01-19
Use mocking for network connections


Liamsi:
Currently, every unit test, that uses network communication also tests the whole network stack.  
We already spent a lot of time in trying to figure out, why tests run locally but fail on travis (see #162, #163, #157, etc). One simple solution to that would be to mock the network connection. We should just test if `Open()`, `Close()`, `Listen()`, etc. are called in an appropriate order.
Tasks:

1) Either use an simple mocking framework (could be golangs [default mocking framework](https://github.com/golang/mock/tree/master/sample) ?) or write a Mock just for the networking layer
2) Make some tests use the mocked network library (instead opening and closing ports on localhost), especially tests which open, close and re-open the same ports are good candidates (see: https://github.com/golang/go/issues/11872#issuecomment-126183460)

ineiti:
We still get occasional improvement of the code by watching what works and what not. So I'll still let it be in there.

ineiti:
Closing, as network is well debugged now and doesn't need to be mocked anymore.

ineiti:
I still think we're merely hiding problems of the network-stack by using mocking here. Our network-stack is broken, so if we put another layer on top of it, it is still broken.

ineiti:
This will be done in #522

PR #574, opened 2016-09-02
[MAIN] Refactor network - base PR

nikkolasg:
This PR is about refactoring the whole network stack in `dedis/cothority`. This is the new place of #499 and the issue #170 + #522 .
Here's a list of relatively small PRs (I tried what I could...). Each one is to be merged first in order so commits from a previous PRs disappear when it gets merged and the following PR only has its change.
All of the below PRs **do not pass Travis**; when they all gets merged, this one should pass Travis of course.
- [x] [**[1/7]**](https://github.com/dedis/cothority/pull/567) Using only *one* address for each node, greatly simplifying every operations with ServerIdentity. Each address should contains the type of connection. Ex: `tcp://5.135.161.91:2000`.
- [x] [**[2/7]**](https://github.com/dedis/cothority/pull/568) Implementing a `Local` network framework using only msg passing, arrays & signaling.
- [x] [**[3/7]**](https://github.com/dedis/cothority/pull/569) Re-organize the TCP network framework with the new interfaces definition defined.
- [x] [**[4/7]**](https://github.com/dedis/cothority/pull/570) Moving the whole network connection management to the `network/` package and having a reduced exported API (Router) on that so SDA have a strict set of operations allowed.
- [x] [**[5/7]**](https://github.com/dedis/cothority/pull/571) Showing all the changes needed in SDA + tests changes (no timeouts, using `LocalTest` etc)
- [x] [**[6/7]**](https://github.com/dedis/cothority/pull/572) Some  identity changes needed at the first version of this refactoring but  still makes sense.
- [x] [**[7/7]**](https://github.com/dedis/cothority/pull/572) Make the rest of the code uses the new network stack 

For the future, I will try to follow this [guide](https://graysonkoonce.com/stacked-pull-requests-keeping-github-diffs-small/). The reason I could not do this here, is because I started by doing some *big* changes that broke the building + testing process right at the beginning, so I could not split thoses changes in separate *travis-passing* PRs.


network/address.go
```diff
@@ -0,0 +1,156 @@
...
+			return ct
+		}
+	}
+	return InvalidConnType
+}
+
+// ConnType return the connection type from this address
+// It returns an unvalid connection type if the address is not valid or if the
+// connection type is not known.
+func (a Address) ConnType() ConnType {
```
ineiti: Why doesn't this return an error?

network/address.go
```diff
@@ -0,0 +1,156 @@
...
+	Local = "local"
+	// InvalidConnType represents a non valid connection type
+	InvalidConnType = "wrong"
+)
+
+// typeAddressSep is the separator between the type of connection and the actual
+// ip address.
+const typeAddressSep = "://"
+
+func connType(t string) ConnType {
```
ineiti: Why doesn't this return an error?

network/address.go
```diff
@@ -0,0 +1,156 @@
...
+		return false
+	}
+
+	ip, port, e := net.SplitHostPort(vals[1])
+	if e != nil {
+		return false
+	}
+
+	p, err := strconv.Atoi(port)
+	if err != nil || p < 0 || p > 65536 {
```
ineiti: `if err != nil || p < 0 || p > 65535 {`

network/address.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	h, _, e := net.SplitHostPort(a.NetworkAddress())
+	if e != nil {
+		return ""
+	}
+	return h
+}
+
+// Port will return the port part of the Address
+func (a Address) Port() string {
```
ineiti: Why is this not an `int`?

network/address.go
```diff
@@ -0,0 +1,156 @@
...
+		"(^172\\.3[0-1]\\.)|(^192\\.168\\.)|(^169\\.254)", a.NetworkAddress())
+	if err != nil {
+		return false
+	}
+	return !private && a.Valid()
+}
+
+// NewAddress takes a connection type and the raw address. It returns a
+// correctly formatted address, which will be of type t.
+func NewAddress(t ConnType, network string) Address {
```
ineiti: Why is this `network`? Shouldn't it be `addressPort` or something?

network/address.go
nikkolasg: It returns InvalidConnType in case it's unknown.

network/address.go
nikkolasg: Why should it be ? 
Both approaches are arguable. In Golang net package, they return a string, so I just used that.

network/address.go
ineiti: True - I saw that. I just wondered if it's the most logical thing to do. Personally I would go for an error here.

network/address.go
ineiti: In golang-net, a port can be a string, whereas we make sure its `0..65535`...

network/address.go
nikkolasg: Well in the `Address` there is now the `ConnType` concatenated with the `NetworkAddress` so to keep it consistent I just call that `network` as being the second part of an `Address`

ineiti: As a former sysadmin when I see `network` I think of `8.8.8.8/32` or something.

network/address.go
nikkolasg: I'm not sure. It depends on our uses cases. If we go for returning an error, then on every operation of the `Address` we have to return an error also (because if ConnType is not valid we should not return anything OK in every other operations). It makes using this whole file a lot more cumbersome than it has to be to my feeling.
That's why there is a `Valid()` method for. If the address is `Valid()` (when we read it from file / user input / whatever) then we can relax and call all the methods without checking each time if something's wrong while we know the address is correct after the first time we checked it. It's not a new operation each time, it's operating with the same data. If there was some operation generating some new data each time (like read from a randomness source) then YES, ok we have to make sure, but here it's not the case.

Printing, or switching over different connections type
1) Using an `InvalidConnType`:
```
if  !a.Valid() {
    fmt.Println("aie something's wrong")
   os.Exit(1)
}
fmt.Printf("Address Conntype = %s",a.ConnType())
... // other method
switch a.ConnType() {
    case PlainTCP:
         // do TCP
    case Local:
         // do Local
    case InvalidConnType:
         // do something !
}
```
2) Using errors:
```
if  !a.Valid() {
    fmt.Println("aie something's wrong")
   os.Exit(1)
}
ct,err := a.ConnType()
if err != nil {
   fmt.Println("Something's stil wrong??")
}
fmt.Printf("Address ConnType: %s", ct)
.... // other method
ct,err := a.ConnType() 
if err != nil {
fmt.Printf("Aie still something wrong")
}
switch ct { 
    case PlainTCP:
         // do TCP
    case Local:
         // do Local
    case InvalidConnType:
         // do something !
}
```

sda/service_test.go
```diff
@@ -641,25 +607,25 @@ type simpleService struct {
...
+	req := pm.(SimpleRequest)
 	tree := req.ServerIdentities.GenerateBinaryTree()
 	tni := s.ctx.NewTreeNodeInstance(tree, tree.Root, "BackForth")
 	proto, err := newBackForthProtocolRoot(tni, req.Val, func(n int) {
-		if err := s.ctx.SendRaw(si, &simpleResponse{
+		if err := s.ctx.SendRaw(si, &SimpleResponse{
 			Val: n,
 		}); err != nil {
-			log.Error(err)
+			log.Print(err)
```
ineiti: Why is this not `log.Error`?

simul/cothority/cothority.go
```diff
@@ -76,15 +92,15 @@ func main() {
...
 			rootSim = sim
 			rootSC = sc
 		}
 	}
 	if rootSim != nil {
-		// If this cothority has the root-host, it will start the simulation
-		log.Lvl2("Starting protocol", simul, "on host", rootSC.Host.ServerIdentity.Addresses)
+		// If this cothority has the root-conode, it will start the simulation
+		log.Lvl2("Starting protocol", simul, "on conode", rootSC.Conode.ServerIdentity.Address)
 		//log.Lvl5("Tree is", rootSC.Tree.Dump())
```
ineiti: Delete or uncomment.

protocols/example/handlers/example_test.go
```diff
@@ -18,7 +18,7 @@ func TestMain(m *testing.M) {
 func TestNode(t *testing.T) {
 	local := sda.NewLocalTest()
 	nbrNodes := 2
-	_, _, tree := local.GenTree(nbrNodes, false, true, true)
+	_, _, tree := local.GenTree(nbrNodes, true)
 	//log.Lvl3(tree.Dump())
```
ineiti: Delete or uncomment

app/lib/server/server.go
```diff
@@ -257,8 +270,10 @@ func checkList(list *sda.Roster, descs []string) error {
...
 
-// signStatement can be used to sign the contents passed in the io.Reader
-// (pass an io.File or use an strings.NewReader for strings)
+// signStatement signs the contents passed in the io.Reader
+// (pass an io.File or use an strings.NewReader for strings). It uses
+// the roster el to create the collective signature.
+// In case the signature fails, an error is returned.
 func signStatement(read io.Reader, el *sda.Roster) (*s.SignatureResponse,
 	error) {
 	//publics := entityListToPublics(el)
```
ineiti: This and further three lines: uncomment or delete.

ineiti:
LGTM - some small fixes. Also
`./test_integration.sh`
and deterlab-simulation work.

Verify also #629

PR #578, opened 2016-09-06
Comments

ineiti:
Improved comments (IMHO). English, return values and explanations. If you have further improvements, merge directly into this branch.

### DIFF
--- a/app/lib/config/config.go
+++ b/app/lib/config/config.go
@@ -29,14 +29,16 @@ func init() {
 	out = os.Stdout
 }
 
-// CothoritydConfig is the cothority daemon config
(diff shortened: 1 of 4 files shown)


nikkolasg:
On a generale note: why the comments for the `lib` part are here ? You could just make another PR directly on master. 
But anyhow, OK to merge.

ineiti:
I'm just taking all files you touched - yes, some comments are already old. But it's faster to go through the whole files than only take the lines you changed.

ineiti:
I'm following your unwritten advice and not change all comments - it takes in fact longer ;)

PR #580, opened 2016-09-06
Comments

ineiti:
Changes to the comments - Linus' new way of doing code-review ;)

### DIFF
--- a/network/local.go
+++ b/network/local.go
@@ -264,7 +264,7 @@ func (lc *LocalConn) Type() ConnType {
 }
 
 // connQueue manages the message queue of a LocalConn.
-// Messages are pushed and retrieved in a FIFO.
+// Messages are pushed and retrieved in a FIFO-queue.
 // All operations are thread-safe.
 // The messages are marshalled and stored in the queue as a slice of bytes.
 type connQueue struct {
--- a/network/local.go
+++ b/network/local.go
@@ -10,14 +10,17 @@ import (
 	"golang.org/x/net/context"
 )
 
-// NewLocalRouter returns a fresh router which uses local connections. It takes
-// the default manager.
+// NewLocalRouter returns a fresh router which uses only local queues. It uses
+// the default local manager.
+// If you need multiple independant local-queues, use NewLocalRouterWithManager.
+// In case of an error it is returned together with a nil-Router.
 func NewLocalRouter(sid *ServerIdentity) (*Router, error) {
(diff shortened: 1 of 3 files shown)


network/local.go
```diff
@@ -26,16 +29,18 @@ func NewLocalRouterWithManager(ctx *LocalManager, sid *ServerIdentity) (*Router,
...
 type LocalManager struct {
 	// queues maps a remote endpoint to its packet queue. It's the main
-	//stucture used to communicate.
+	// stucture used to communicate.
 	queues map[endpoint]*connQueue
 	sync.Mutex
+	// The listening-functions used when a new connection-request arrives.
 	listening map[Address]func(Conn)
 
+	// Unique ID of this Local Manager.
```
nikkolasg: No it's not. It's more like a counter that we increment for each connection it creates/handle. So maybe rename the field to `counter` or stg.

nikkolasg:
I think some comments on `connQueue` have to be changed so it reflects now that we are marshalling messages and passing only slice of bytes (no `Packet`)

ineiti:
I propose we keep the `packet`-references, as they still _represent_ packets.

network/local.go
```diff
@@ -25,17 +28,19 @@ func NewLocalRouterWithManager(lm *LocalManager, sid *ServerIdentity) (*Router,
...
 	// queues maps a remote endpoint to its packet queue. It's the main
-	//stucture used to communicate.
+	// structure used to communicate.
 	queues map[endpoint]*connQueue
 	sync.Mutex
+	// The listening-functions used when a new connection-request arrives.
 	listening map[Address]func(Conn)
 
-	baseUID uint64
+	// connection-counter for giving unique IDs for each connection.
```
nikkolasg: `... giving unique IDs to each connection` ? for vs to ?

network/local.go
```diff
@@ -251,9 +265,10 @@ func (lc *LocalConn) Type() ConnType {
 	return Local
 }
 
-// connQueue is a struct that managers the message queue of a LocalConn.
-// Messages are pushed and retrieved FIFO style.
+// connQueue manages the message queue of a LocalConn.
+// Messages are pushed and retrieved in a FIFO.
```
nikkolasg: FIFO is not an object, but a method of ordering elements. A FIFO queue is an object. `...in a FIFO queue`

nikkolasg:
Cool. Merging it.

PR #581, opened 2016-09-06
Comment edits from refactor_network_tcp

ineiti:
Changes in comments helping me to read and understand the code ;)

### DIFF
--- a/network/tcp.go
+++ b/network/tcp.go
@@ -288,7 +288,7 @@ func (t *TCPListener) bind(addr Address) error {
 
 // Listen starts to listen for incoming connections and calls fn for every
 // connection-request it receives.
-// In case of an error it returns it.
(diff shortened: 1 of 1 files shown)


network/tcp.go
```diff
@@ -277,7 +286,9 @@ func (t *TCPListener) bind(addr Address) error {
 	return nil
 }
 
-// Listen implements the Listener interface
+// Listen starts to listen for incoming connections and calls fn for every
+// connection-request it receives.
+// In case of an error it returns it.
```
nikkolasg: `The call returns in case of an error` ? Just `it returns it` is not nice looking ;)

nikkolasg:
Nice OK to merge after my comment.

ineiti:
Should be all done.

Issue #585, opened 2016-09-07
Test-sh for setting up android


ineiti:
Add a bash-script in `app/cisc` that sets up a cothority and saves the roster to a file.

ineiti:
Implemented in #588

PR #587, opened 2016-09-07
added script for setting up cothorities

ineiti:
First implementation of #585 for easy testing with `cisc`.


ineiti:
Closing, replaced by #588

PR #596, opened 2016-09-14
Swupdate timestamper

nikkolasg:
Introduce the timestamper operations for the swupdate service:
* `timestamp` method which compute the merkle tree of all latest skipblock's hashes & timestamp it
  + each conode will verify that this merkle tree represents their own local view
* Each time a package is created / updated, `timestamp()` is called
* `TimestampProof(packageName)` returns the proof of inclusion for the latest skipblock's hash representing the package's skipchain.


nikkolasg:
This can be closed as the features are already integrated in `plotting_595`

PR #597, opened 2016-09-15
"Flood" simulation

Liamsi:
@nikkolasg this is what worked for me 
(10-20k client request simulation running after doing:
```
sudo sysctl -w kern.maxfilesperproc=122888 #maybe there is a more reasonable number here?
sudo sysctl -w kern.maxfiles=122288 
ulimit -n 122888
sudo sysctl -w kern.ipc.somaxconn=2048
```
Run with 
`go build; ./simul -debug 1 -platform localhost runfiles/swup_flood_first.toml`
or 
`go build; ./simul -debug 1 -platform localhost runfiles/swup_flood_latest.toml`


nikkolasg:
Ok I merge it in swupdate_3 even if there's no real changes... That means we don't need `swupdate_time` anymore.

PR #604, opened 2016-09-23
Nsdi changes 602

ineiti:
First chunk of merging back into master:
- added a `Copy`-method for copying files
- one more merkle-tree-test in `crypto` (tests are always good)


app/lib/config/config.go
```diff
@@ -288,3 +289,27 @@ func InputYN(def bool, args ...interface{}) bool {
 	}
 	return strings.ToLower(string(Input(defStr, args...)[0])) == "y"
 }
+
+// Copy makes a copy of a local file with the same file-mode-bits set.
```
nikkolasg: and `If dst is a directory, it will copy the src file into dst`, or `it mimics the posix "cp" command` or something else saying the same thing :D

crypto/proof_test.go
```diff
@@ -29,6 +30,26 @@ func TestPath(t *testing.T) {
 	}
 }
 
+// can we also store sth. else in the leaves (not only hashIds)? The tree
```
nikkolasg: `HashID` is only a type alias to a slice of bytes so to me a slice of bytes is pretty agnostic already ;) Most of the time you don't even need to cast to `HashID` golang does it for you.

monitor/measure.go
```diff
@@ -185,6 +189,59 @@ func (cm *CounterIOMeasure) Record() {
...
+		docker: docker,
+	}
+}
+
+func (dm *DockerMeasure) Update() {
+	if len(dm.docker) == 0 {
+		// Search all docker images
+	}
+	for i, d := range dm.docker {
+		cpuFile := "/sys/fs/cgroup/cpuacct/docker/" + d + "/cpuacct.stat"
```
nikkolasg: Not necessarily in the comment but could you share your knowledge of this ? What's this file etc ?  I'm curious about that..
*Edit*: ok no worries I've found this quite useful : https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Resource_Management_Guide/ch01.html

network/net.go
```diff
@@ -304,14 +304,15 @@ func (st *SecureTCPHost) Open(si *ServerIdentity) (SecureConn, error) {
 	for _, addr := range si.Addresses {
 		// try to connect with this name
 		log.Lvl4("Trying address", addr)
-		c, err := st.TCPHost.openTCPConn(addr)
+		c, err := st.TCPHost.Open(addr)
+		//c, err := st.TCPHost.openTCPConn(addr)
```
nikkolasg: let's remove that buggy comment ;)

protocols/swupdate/Readme.md
```diff
@@ -0,0 +1,7 @@
+This is basically a complete copy of the cosi protocol (created by 
+@nikkolasg) with some nice additional features (possibility to pass the 
+message directly in the Announcement and the possibility to validate 
+this data in a general way). 
+This shouldn't be merged into master as it is. All changes (which are 
+useful) should most probably directly go into the main cosi
```
nikkolasg: +1. That should be in our "next steps" list (somewhere in our minds). So let's remove this protocol from this PR. I make an issue on `dedis/cosi`.

sda/service.go
```diff
@@ -434,11 +434,14 @@ func (c *Client) Send(dst *network.ServerIdentity, msg network.Body) (*network.P
 		// Catch an eventual error
 		err := ErrMsg(&response, nil)
 		if err != nil {
+			log.Lvl4("Closing connection to", dst)
 			return nil, err
 		}
+		log.Lvl4("Closing connection to", dst)
 		return &response, nil
-	case <-time.After(time.Second * 10):
+	case <-time.After(time.Minute * 30):
```
nikkolasg: wow wow wow. You also changed a value in a separate PR in client. let's put someting reasonable please :)

simul/cothority/cothority.go
```diff
@@ -82,6 +77,15 @@ func main() {
 			rootSC = sc
 		}
 	}
+	if monitorAddress != "" {
+		log.Lvl2("Connecting to monitor", monitorAddress)
+		if err := monitor.ConnectSink(monitorAddress); err != nil {
+			log.Fatal("Couldn't connect monitor to sink:", err)
+			//}
+			//} else {
+			//	log.Fatal("No monitorAddress in", hostAddress)
```
nikkolasg: outdated comments here.

simul/matplotlib/NSDI_2016/swup_verification.csv
```diff
@@ -0,0 +1,6 @@
...
+6,5,4,6000,1,10,10,10,3600,4,0.002139,0.002139,0.002139,0.002139,NaN,0.009862,0.009862,0.009862,0.009862,NaN,0.204168,0.204168,0.204168,0.204168,NaN,0.005538,0.005538,0.005538,0.005538,NaN,0.087215,0.087215,0.087215,0.087215,NaN,0.086892,0.086892,0.086892,0.086892,NaN,128422.000000,128422.000000,128422.000000,128422.000000,NaN,289660.000000,289660.000000,289660.000000,289660.000000,NaN,123558.000000,133237.000000,128397.500000,256795.000000,6844.086535,8701.000000,293020.000000,150860.500000,301721.000000,201043.892920,0.000000,0.000005,0.000001,0.000010,0.000002,0.000000,0.000011,0.000002,0.000024,0.000003,0.000000,0.000001,0.000000,0.000004,0.000000,0.016384,0.016384,0.016384,0.016384,NaN,0.084461,0.084461,0.084461,0.084461,NaN,0.129318,0.129318,0.129318,0.129318,NaN,0.005542,0.005542,0.005542,0.005542,NaN,0.087340,0.087340,0.087340,0.087340,NaN,0.086980,0.086980,0.086980,0.086980,NaN,0.000000,0.004062,0.001139,0.013670,0.001489,0.003751,0.018578,0.007356,0.088270,0.004066,0.004046,0.006605,0.004493,0.053912,0.000771
```
nikkolasg: I'm not sure I agree with these numbers... :D

simul/platform/cliutils.go
```diff
@@ -23,14 +27,38 @@ func Scp(username, host, file, dest string) error {
 	return cmd.Run()
 }
 
+// Copy makes a copy of a local file with the same file-mode-bits set.
+func Copy(dst, src string) error {
```
nikkolasg: Wait this is already in app/lib/<something>, so if we need it at least twice, then we should have a `fs` package or `utils` package at the top level where we could put such functions.

simul/platform/mininet.go
```diff
@@ -0,0 +1,373 @@
+// Mininet is the platform-implementation that uses the MiniNet-framework
```
nikkolasg: Yeah !

ineiti:
Closing - bad checkout...

PR #605, opened 2016-09-23
including small protocol changes

ineiti:
Changes include:
- less verbosity in bftcosi-test
- propagate timeout in manage-protocol and more debug if the timeout is reached
- add some documentation to the ntree-protocol

### DIFF
--- a/protocols/manage/propagate.go
+++ b/protocols/manage/propagate.go
@@ -121,7 +121,8 @@ func (p *Propagate) Dispatch() error {
 		p.Unlock()
 		select {
 		case msg := <-p.ChannelSD:
-			log.Lvl3(p.ServerIdentity(), "Got data from", msg.ServerIdentity)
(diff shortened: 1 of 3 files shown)


protocols/manage/propagate.go
```diff
@@ -121,7 +121,8 @@ func (p *Propagate) Dispatch() error {
 		p.Unlock()
 		select {
 		case msg := <-p.ChannelSD:
-			log.Lvl3(p.ServerIdentity(), "Got data from", msg.ServerIdentity)
+			log.Lvl3(p.ServerIdentity(), "Got data from", msg.ServerIdentity, "and setting timeout to", msg.Msec)
+			p.sd.Msec = msg.Msec
```
nikkolasg: what does this line do ? (And why did it work before without it ? or did it ?)

ineiti: It did only work if the timeout was OK with 1sec. If a new propagation-protocol has been started with a bigger timeout, all children nodes kept the 1sec. timeout.

So this makes it work as required.

PR #607, opened 2016-09-23
added skipchain and identity-changes

ineiti:
- a `SecureTCPHost` opens a connection so that it can be closed again
- putting the time-out in identity-service from 1sec to 10sec
- adding a test for `GetUpdateChain` in `skipchain`
- registering services uses a slice so that the instantiations get done chronologically
- verifier-functions in SkipChain get now registered directly to the instance of the service


services/identity/service.go
```diff
@@ -110,7 +110,7 @@ func (s *Service) CreateIdentity(si *network.ServerIdentity, ai *CreateIdentity)
 
 	roster := ids.Root.Roster
 	replies, err := manage.PropagateStartAndWait(s.Context, roster,
-		&PropagateIdentity{ids}, 1000, s.Propagate)
+		&PropagateIdentity{ids}, 10000, s.Propagate)
```
nikkolasg: Instead of `10000` or `1000`, it should be easier to change at a single place using an unexported (or exported?) global variable (or with getter/setter). Even better, we could specify a value to the service (`service.SetTimeout()`) and the service would use the default unexported global variable by default otherwise.

services/identity/service.go
```diff
@@ -150,7 +150,7 @@ func (s *Service) ProposeSend(si *network.ServerIdentity, p *ProposeSend) (netwo
 	}
 	roster := sid.Root.Roster
 	replies, err := manage.PropagateStartAndWait(s.Context, roster,
-		p, 1000, s.Propagate)
+		p, 10000, s.Propagate)
```
nikkolasg: same comment as above.

services/skipchain/msgs.go
```diff
@@ -33,6 +36,21 @@ func init() {
 // deny a SkipBlock.
 type VerifierID uuid.UUID
 
+var verifiers map[VerifierID]bftcosi.VerificationFunction
+var verifiersMutex sync.Mutex
+
+// RegisterVerification stores the verification in a map and will
+// call it whenever a verification needs to be done.
+func VerificationRegistration(v VerifierID, f bftcosi.VerificationFunction) error {
```
nikkolasg: Inconsistency between names here `RegisterVerification` vs `VerificationRegistration`. (I have a preference for the former)

services/skipchain/skipchain.go
```diff
@@ -397,7 +397,7 @@ func (s *Service) startPropagation(blocks []*SkipBlock) error {
 			roster = sb.Roster
 		}
 		replies, err := manage.PropagateStartAndWait(s.Context, roster,
-			block, 1000, s.PropagateSkipBlock)
+			block, 10000, s.PropagateSkipBlock)
```
nikkolasg: same comment as for the identity service.

services/skipchain/skipchain.go
```diff
@@ -443,30 +443,36 @@ func (s *Service) bftVerify(msg []byte, data []byte) bool {
 				return true
 			}
 		}
+	default:
+		f, ok := verifiers[sb.VerifierID]
```
nikkolasg: *Discussion*: Maybe it's not possible directly or with too much work I don't know but since we introduce the registration of verification functions, it seems best to 
+ delete the two above `case` 
+ put their logic into separate verification functions 
+ register these verification functions in a `init()` or so
+ get rid of this switch case 

That way all is "unified" according to this new way of using verification functions (and in a few months we won't be like `whut ? Why is it not used in a registration function ... ??` ).
 What do you think ? From just what I see now it does not seem to be too much work but @ineiti you know better this code so what's your thoughts on this ?

services/identity/service.go
ineiti: It's very specific for that service and shouldn't change during runtime, so I'll put in a constant.

services/skipchain/msgs.go
ineiti: Me too.

services/skipchain/skipchain.go
ineiti: Same reasoning as above. I'll put in a constant.

services/skipchain/skipchain.go
ineiti: I left it here because it uses fields of the structure, but it makes sense what you say. It will be a bit clumsy, but probably nicer...

ineiti:
All changes implemented - waiting for travis.

services/skipchain/skipchain.go
```diff
@@ -411,6 +447,7 @@ func (s *Service) startPropagation(blocks []*SkipBlock) error {
 // bftVerify takes a message and verifies it's valid
 func (s *Service) bftVerify(msg []byte, data []byte) bool {
 	log.Lvlf4("%s verifying block %x", s.ServerIdentity(), msg)
+	// XXX What's this variable for ?
 	s.testVerify = true
```
nikkolasg: Sorry I wrote it in the code; What's `testVerify` for ?

services/skipchain/msgs.go
```diff
@@ -33,18 +36,23 @@ func init() {
...
 type VerifierID uuid.UUID
 
+// SkipBlockVerifier is function that should return whether this skipblock is
+// accepted or not. This function is used during a BFTCosi round, but wrapped
+// around so it accepts a block.
+type SkipBlockVerifier func(msg []byte, s *SkipBlock) bool
+
+// RegisterVerification stores the verification in a map and will
+// call it whenever a verification needs to be done.
+func RegisterVerification(c *sda.Context, v VerifierID, f SkipBlockVerifier) error {
```
nikkolasg: Should be deleted I guess.

ineiti: I thought keeping it, so that `Identity` or any other skipchain can call `skipchain.RegisterVerification` and give its `sda.Context`, so that it gets stored correctly.

services/skipchain/skipchain.go
ineiti: `// testVerify is set to true if a verification happened - only for testing`

PR #608, opened 2016-09-23
smaller set of changes

ineiti:
- `Copy` for copying files

### DIFF
--- a/app/lib/config/config.go
+++ b/app/lib/config/config.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"path"
 	"strings"
 
 	"os/user"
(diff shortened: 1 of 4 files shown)


ineiti:
adding test for copy so that coverall is happy...

Liamsi:
> adding test for copy so that coverall is happy...

Although it is nice to add a test

crypto/proof_test.go
```diff
@@ -29,6 +30,26 @@ func TestPath(t *testing.T) {
...
+	}
+
+	root, proofs := ProofTree(newHash, leaves)
+	for i := range proofs {
+		if proofs[i].Check(newHash, root, leaves[i]) == false {
+			t.Error("check failed at leaf", i)
+		}
+	}
+}
+
```
Liamsi: This seem to be changes I did on the first timestamp branch. I think it's totally fine to loose some of the history but I don't think this test isn't really neccessary (I forgot to delete it)

nikkolasg:
> Although it is nice to add a test maybe we should configure coveralls not to fail if the coverage decreases?

NOPE ! Please let's try to have high standards on master when we can. If really we need to merge something with a decreasing coverage, we can still do it anyway and we'll have a warning as a bonus.

Liamsi:
OK, fair enough. Coveralls still complains about the coverage decreasing.

app/lib/config/config.go
```diff
@@ -288,3 +289,27 @@ func InputYN(def bool, args ...interface{}) bool {
 	}
 	return strings.ToLower(string(Input(defStr, args...)[0])) == "y"
 }
+
+// Copy makes a copy of a local file with the same file-mode-bits set.
+func Copy(dst, src string) error {
```
ineiti: I just searched, and it's neither in `master`, nor in `nsdi_skipchain_602`.

PR #610, opened 2016-09-26
Fix coveralls / travis failure reporting

### LINKED ISSUE #609: Coveralls doesn't always show failing tests
https://travis-ci.org/dedis/cothority/builds/162707149

Also, we might want to move the coveralls report to `after_success` so the tests won't fail even if coveralls isn't reachable etc.

Liamsi:
Should resolve #609


Liamsi:
For some reason travis doesn't start at all for this PR ?! I'll create a new branch with the same changes and open a new one.

PR #614, opened 2016-09-27
Refactor network router comments

ineiti:
Some updated comments for the 4th part of our sequel.


network/client.go
```diff
@@ -14,12 +14,7 @@ func init() {
 	RegisterPacketType(&StatusRet{})
 }
 
-// Client is an interface which is a simpler version of a Router. The main use
-// for a Client is to directly Send something and get a result back. It is used
-// intensively by Services to have a easy external API.
-// Two implementations are done: TcpClient to use for applications and
-// deployement,and localClient to use for local testing alongside with
-// LocalRouter.
+// Client is the external APIs used for services.
```
nikkolasg: Euhhh.... what's wrong with my comments ? I mean I'm ok for english faults and adding informations but *reducing* informations ... not so !

network/router.go
```diff
@@ -9,40 +9,40 @@ import (
...
 //   between ServerIdentity <-> address
-// * manage the reconnections of non-working Conn,
+//   * managing the re-connections of non-working Conn
 // Most caller should use the creation function like NewTCPRouter(...),
 // NewLocalRouter(...) then use the Host such as:
-// `router.Start() // will listen for incoming Conn and block`
-// `router.Stop() // will stop the listening and the managing of all Conn`
+//
+//   router.Start() // will listen for incoming Conn and block
+//   router.Stop() // will stop the listening and the managing of all Conn
```
nikkolasg: thanks for the formatting > <

network/router.go
```diff
@@ -183,22 +186,20 @@ func (r *Router) handleConn(remote *ServerIdentity, c Conn) {
...
 			return
 		}
 
 		if err != nil {
-			// something went wrong on this connection
 			log.Lvlf4("%+v got error (%+s) while receiving message", r.id.String(), err)
 
 			if err == ErrClosed || err == ErrEOF {
+				// Connection got closed.
 				// remote connection closed
```
nikkolasg: Either one or the other but both seems redundant !

network/client.go
ineiti: You wrote it's an interface - which it's not. And I didn't see the two implementations, which makes sense, as it's not an interface. So I supposed it was a comment from an earlier version that was an interface and was implemented twice.

PR #615, opened 2016-09-28
Refactor mocking local - [3/7]

ineiti:
I don't understand why this is still here - did I make a mistake when merging?


ineiti:
wrong number - already merged

PR #621, opened 2016-09-30
Different keys for simulation

nikkolasg:
For the moment, all keys are equal since to a pointer issue (interface!). This PR fixes that.

### DIFF
--- a/sda/simulation.go
+++ b/sda/simulation.go
@@ -254,8 +254,8 @@ func (s *SimulationBFTree) CreateRoster(sc *SimulationConfig, addresses []string
 		} else {
 			address += strconv.Itoa(port + c/nbrAddr)
 		}
-		entities[c] = network.NewServerIdentity(key.Public, address)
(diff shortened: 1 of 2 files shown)


ineiti:
Once the `Clone`-method works, there is a much more easy fix to that. Just add `Clone` to the assignments:

```
		entities[c] = network.NewServerIdentity(key.Public.Clone(), address)
		sc.PrivateKeys[entities[c].Addresses[0]] = key.Secret.Clone()
```

ineiti:
As this is quite a serious bug, we should perhaps do something like:

```
	if hosts > 1 {
		if sc.PrivateKeys[entities[0].Addresses[0]].Equal(
			sc.PrivateKeys[entities[1].Addresses[0]]) {
			log.Fatal("Please update dedis/crypto with\n" +
				"go get -u github.com/dedis/crypto")
		}
	}
```

nikkolasg:
Deleting since @ineiti already made another PR for this.

PR #628, opened 2016-10-04
AGPL-license

### LINKED ISSUE #627: AGPL license
Change the license to AGPL and insert a dual-licensing reference.

ineiti:
resolves #627 
We'll wait for Bryan's final OK before merging that, but please comment already now.


AUTHORS
```diff
@@ -0,0 +1,30 @@
...
+Author: Evan Visher <evan.visher@gmail.com>
+Summer 2016 intern - is he DEDIS, as he signed a contract?
+
+Author: Mahdi Zamani <mahdi.zamani@yale.edu>
+PostDoc student from Yale - working on Prifi
+
+Author: david froelicher <david.froelicher@epfl.ch>
+PhD student from JP - working on MedCo
+
+Author: nalos <Command not found>
```
nikkolasg: lol forget that, that's me.

CLAC
```diff
@@ -0,0 +1,157 @@
...
+   must have a Contributor License Agreement (CLA) on file that has been
+   signed by each Contributor, indicating agreement to the license terms
+   below. This license is for your protection as a Contributor as well
+   as the protection of the Lab and its users; it does not change
+   your rights to use your own Contributions for any other purpose.
+
+   This version of the Agreement allows an entity (the "Corporation") to
+   submit Contributions to the Lab, to authorize Contributions 
+   submitted by its designated employees to the Lab, and to grant 
+   copyright and patent licenses thereto.
```
nikkolasg: ? Well I'm lost, you'll have to explain me next week maybe but here it basically says that companies have a right to patent cothority ?

CONTRIBUTION
```diff
@@ -0,0 +1,26 @@
...
+files), so we actively seek contributions (help on the Cothority users list,
+documentation, source code, ideas, …).  All contributions help us make a
+better product.
+
+All contributions without an explicit copyright statement or a CLAI/CLAC (see below)
+are assumed to be covered under a AGPL 2-Clause license as described in the
+file LICENSE.
+
+Developers who have contributed significant changes to the Cothority code must sign
+a Contributor License Agreement (CLAI), eventually also a CLAC if they work for
```
nikkolasg: Maybe it's already too late for that, but joinging on @Liamsi comment from our last HR meeting that this has been not really transparent, but... Don't you think that requiring one or even two signatures is too much, and on the contrary is more a "show stopper" than a incentive to contribute ? I'm just saying. But I'm truely unfamiliar with this whole patent/license world, but a lot of open source project don't require that, why do we do so ?

CLAC
ineiti: No, this is in the case an employee from another corporation wants to contribute. Then his corporation has to sign off that this employee is allowed to contribute. When the corporation signs off on the `CLAC`, it will grant its copyright and patent licenses to the `Lab`, which is DEDIS.

If you, Nicolas, want to contribute in your work-time to a project that needs a `CLA`, then EPFL needs to sign off, too, because you're under a working contract and cannot give your copyright (which belongs to EPFL) to somebody else.

CONTRIBUTION
ineiti: That's a good point. For the moment we only have contributions from other labs - will they accept to sign the CLA[IC]? 

Now that we have an actual text we can ask people what they think about. Please write down your concerns, so we can discuss them at the next meeting.

nikkolasg: perfect. thanks.

CLAC
nikkolasg: so at the end, DeDiS is *always* in detention of the copyright of the cothority code, is that right ?

Liamsi:
I just started reading, but I'm already a little worried that the bureaucracy added by the Contributor License Agreement will make @bford's original vision of cothority nodes running a **dynamic** set of protocols, or validation methods impossible or at least very, very unlikely:
  
Currently, if you want to add another protocol/validation method for the data to be signed, you would need to change the code, wait for the review, change the code again, wait for approval of the PR, and with the CLA, even sign a contract/real document. I think, no individual has an incentive to do so. Also, I'm not sure the organizations that @bford original had in mind (@EFForg, @mozilla, @torproject, universities, or the CCC) would be willing to do so. 

If @bford changed his mind, and suddenly only is more interested in corporate partners adding protocols, and (commercial and a few academic) institutions who are willing to sign a CLA, I personally think this PR is fine.
(See slide 77/78 for @bford's vision on how cothority was supposed to work: (hosted at @Daeinar's [home page](https://zerobyte.io/files/talks/2015-12-27-collective-authorities-32c3.pdf) and presented at the 32c3)

Coming back to the "dynamic" part (which I thought was our common goal): If we ever want people running cothority nodes, which can pick up new protocols/validation functions while they are running, we would need to make this possible, without people needing to change the code (and going through a long process, including a CLA signature). I have a few ideas on how to make this possible (but maybe they are too ambitious). One approach is similar to what bitcoin currently does with it's stack-based scripting language. As far as I remember, Bryan mentioned sth. similar some time ago. I'll try sketch my idea out and send it to dev-dedis. 
An easier approach would be to make it as easy as possible for NGOs and researchers (who don't have any commercial interests) to participate in the cothority code.

LICENSE.AGPL
```diff
@@ -0,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
Liamsi: To also say something positive, after reading about the AGPL, it looks like a good idea.

bford: Hi Ismail,

Many open-source software organizations ask contributors to sign their copyrights to the primary maintaining organization, for multiple reasons - this is fairly standard practice.  For example, even the Free Software Foundation does it:

https://www.gnu.org/licenses/why-assign

I have personally signed FSF copyright-release forms for the (fairly small) contributions I've made to GNU software in the (somewhat distant) past.  

One reason for this (probably the main reason the FSF does it) is to have better legal leverage in the event of litigation: e.g., if some company tries to "steal" the free software and sell some variation of it as proprietary software without releasing the source or otherwise complying with the terms of the open-source license.  

Another common reason is that the company holding the copyright can "sell exceptions" to the terms of the GPL; this is the traditional (fairly successful) funding model that Qt/Trolltech used for example before it was bought by Nokia and Nokia imploded.  The FSF talks about this, too - they don't like it all that much (understandably given the fairly purist positions they usually take), but they call this practice "acceptable, though only barely". :)

https://www.fsf.org/blogs/rms/assigning-copyright

Regardless, all this only applies to files and modules that were primarily written in DEDIS (and may incorporate small contributions from non-DEDIS/EPFL people); there is no reason this needs to prevent us from incorporating and building on open-source code copyrighted by other organizations and released with compatible open-source licenses, nor does this prevent other organizations from taking DEDIS code and building on it, modifying it, and using it more-or-less in whatever ways they want.  That's what the open-source license is designed to allow.  Note that an explicit exception is made to the contributor release rule for code that is explicitly documented to be copyrighted by someone else, so this is all just about the "default" copyright that applies to the code that was written mostly by DEDIS and for which DEDIS/EPFL is the primary copyright holder.

Does that help?

CLAC
ineiti: The code is under the AGPL3: 
- If you change the code and use and/or distribute it, even in the cloud, you have to publish the changes.
Regarding our repo:
- If you want your code to be included in our repo, you need to sign off the `CLA[IC]`.
Commercial use:
- If you want to change the code and not publish the changes, or if you want us to work on the code and keep the result for yourself, you get a special, paying license.

LICENSE.AGPL
Liamsi: Thanks a lot for your lengthy reply! 

>nor does this prevent other organizations from taking DEDIS code and building on it, modifying it, and using it more-or-less in whatever ways they want. That's what the open-source license is designed to allow. 
[...]
> Does that help?

Yes, that clarifies a lot! The text actually sounded a little different to me with this regard. So I'm definitely OK with this changes now. 

I'm still a little concerned about the static nature of the protocols supported by the current cothority code and the vision that it's possible for node to receive a message like "hey, do you want to run this protocol/validation function on this particular type of data, with this set of particular nodes" - "OK, never heard of this protocol, but sure, why not". This currently isn't possible without going through this repo, DEDIS and the aforementioned CLA and recompiling/restarting a running node. 

But maybe this concern is orthogonal  to the licensing part (?) and we should just make it possible to run different 'protocols' without modifying the code? Would be happy to discuss this further with you soon.

Thanks again for your clarifications.

CLAC
Liamsi: Thanks for the clarifications, too @ineiti!

LICENSE.AGPL
bford: Yes, it would be nice if the code can run different protocols without modifying the code, but that's an orthogonal (technical) issue we should discuss separately.

If the problem you see there is that all the protobuf protocols currently supported are "wired together" via nested interlocking message/struct definitions using only small-integer codes to distinguish them, we already discussed that.  To reiterate what my memory of the plan was, for simplicity and efficiency we would just use the direct protobuf nested-message/struct approach for "native" or "intended-to-be-stable" protocols, but then (eventually) have an explicit "extension message/protocol" option that uses ASCII strings instead of small-ints to identify arbitrary extension protocols, which would ideally be pluggable without modifying the existing code.

But to be clear, even then, this will not mean that "only DEDIS" can add new "native/stable" protocols to the protobuf definitions.  It will simply mean that anyone who wants to add new native/stable protocols will need to modify the code (which anyone can do), whereas adding "experimental" protocols via the ASCI-string-identified extension mechanism will be a bit easier (and encouraged for experimental protocols anyway).  Does this make sense?  Or am I barking up the wrong tree (i.e., discussing the wrong problem)?

Liamsi: Sorry if my comment wasn't really clear. One problem I saw is that less people might actually participate and modify the code in the official repo (for instance for adding new protocols) because of the CLA (and people would rather fork and build their own modified cothority instead; not sure if this is good or bad). But, I guess, I was barking up the wrong tree with my concerns (learned a new idiom ;-). 

>Yes, it would be nice if the code can run different protocols without modifying the code, but that's an orthogonal (technical) issue we should discuss separately.

Would be happy to discuss this issue further (maybe offline). I think we all (?) weren't aware that the "extension message/protocol" that uses ASCII as you described it would allow new protocols, ideally, without modifying the code. I'm not sure this is possible with the current architecture.

Thanks again!

nikkolasg:
Oh one thing that just popped off my mind: What should we do about CoSi ? Everything is OK if we move it back to cothority *after* putting this license ? I think so, as long as we don't put another explicit license on CoSi. Just asking because I don't know the answer although it seems to be OK.

LICENSE.AGPL
nikkolasg: > Would be happy to discuss this issue further (maybe offline). I think we all (?) weren't aware that the "extension message/protocol" that uses ASCII as you described it would allow new protocols, ideally, without modifying the code. I'm not sure this is possible with the current architecture.

As @bford said, there was a meeting when we discussed the RFC which led to a discussion to the wire protocol (including this above idea). I don't recall if you were here or not ( @Liamsi ). Anyway, we'll start soon to have meetings about this whole subject soon so we'll be able to expand and discuss anything related there.

Liamsi: @nikkolasg Not sure which exact meeting you are referring to, but yes I can remember @bford proposing an optional (ASCII) field for additional (experimental) protocols. But as far as I understood it, this was only about the *data* that will be send around for that protocol and not about potential validation functionality on this data. Were you aware that this means that ideally one can add protocols (messages and methods that validate/act on this data) without modifying the underlying (golang) code? But I agree, let's discuss the details in the meetings 👍

nikkolasg: Okay there has been a mixed up of problems. @bford talks about this problem which we discusse in the meeting:
> If the problem you see there is that all the protobuf protocols currently supported are "wired together" via nested interlocking message/struct definitions using only small-integer codes to distinguish them

@Liamsi is talking about the problem of having many different validations for BFT-CoSi, CoSi alike protocols without the need of coding them from scratch each time, but simply have this pluggable mechanism for a validation function. 
Both problems seems disjoint to me so to answer your last comment, I don't think using the extension field  will solve the second problem. Indeed scripting language for that would be nice. Such an idea popped off months ago but since we did not need it, we focused on more urgent things at the time, we need to see if that makes more sense now...

Liamsi: This was exactly what I was trying to say: With the current code/architecture having different validations for (BFT-)CoSi without modifying the code isn't possible. Thanks for the clarification.

AUTHORS
```diff
@@ -0,0 +1,27 @@
+# Directly affiliated with DEDIS
+Bryan Ford <bryan.ford@epfl.ch>
+Kokoris Kogias Eleftherios <eleftherios.kokoriskogias@epfl.ch>
+Ismail Khoffi <ismail.khoffi@epfl.ch>
+Linus Gasser <linus.gasser@epfl.ch>
+Nicolas Gailly <nicolas.gailly@epfl.ch>
+Nikitin Kirill <kirill.nikitin@epfl.ch>
+Philipp Jovanovic <philipp.Jovanovic@epfl.ch>
+
```
Daeinar: Nitpick: Shouldn't those names be ordered alphabetically either by <first name> <last name> <email> or <last name> <first name> <email>? Currently it's not consistent.

AUTHORS
```diff
@@ -0,0 +1,27 @@
...
+# NON-DEDIS - probably need a CAA or CLA
```
Daeinar: I probably wouldn't merge the second part of this file still saying "probably need a CCA or CLA". Also: Aren't David and Patricia automatically covered by this license agreement since they are affiliated with EPFL?

CLAC
```diff
@@ -0,0 +1,157 @@
+                       DEDIS / EPFL / Lausanne / CH
+  Software Grant and Corporate Contributor License Agreement ("Agreement")
+                      http://www.apache.org/licenses/
+
+   Thank you for your interest in software from DEDIS (Lab).
```
Daeinar: Instead of speaking of "the Lab" everywhere wouldn't it be more precise when replacing it by "DEDIS" or by "the DEDIS Lab"?

Daeinar: Okay just checked the reference from the Apache Foundation and they used "Foundation" everywhere. So I guess that's where this all is coming from and then probably okay. :)

Daeinar:
Okay after reading through the discussion above and putting technical issues aside, this all seems fairly reasonable to me. So +1 from my side.

AUTHORS
Liamsi: Maybe this was because they aren't directly affiliated with the DEDIS lab (which seems to be the copyright holder). Also, I wonder, wasn't their code "moved"? But I guess this doesn't change anything.

Daeinar: Okay, I thought EPFL is also a copyright holder in a broader sense. That's definitely something we should clarify.

Liamsi: I'm not sure. I though, as DEDIS is part of EPFL this shouldn't pose a problem, right?

Daeinar: Yeah that's what I would expect as well but nevertheless we should clarify it.

Liamsi:
After clarifying these concerns https://github.com/dedis/cothority/pull/628#discussion_r83631551 raised by @Daeinar: Also +1 and thanks for taking care of this.

AUTHORS
```diff
@@ -0,0 +1,19 @@
...
+Kokoris Kogias Eleftherios <eleftherios.kokoriskogias@epfl.ch>
+Ismail Khoffi <ismail.khoffi@epfl.ch>
+Nicolas Gailly <nicolas.gailly@epfl.ch>
+Linus Gasser <linus.gasser@epfl.ch>
+Philipp Jovanovic <philipp.Jovanovic@epfl.ch>
+
+# NON-EPFL/DEDIS - signed a CLA / CLAI:
+
+Christian Mouchet <christian.mouchet@epfl.ch>
+Evan Visher <evan.visher@gmail.com>
```
nikkolasg: does evan knows that by signing he would have his private address put on the list of AUTHORS ?

ineiti: `git log | egrep "Author:.*(Evan|Christian|Mahdi)" | sort | uniq`

nikkolasg: ?? I'm just saying all the others have their educational address. evan has his gmail address which is supposedly his private one. so just to make sure he knows his private one will be public.

ineiti: His address is already public. I only copied it from the git-log to the `AUTHORS`-file. So I don't see the issue.

nikkolasg:
Ok to merge, I'll let you the pleasure ;)

Issue #599, opened 2016-09-23
scmgr - skipchain manager


ineiti:
# SkipChain Manager
- Goal: setting up root- and config-skipchains
- Given: a set of running conodes
- Actors:
  - C - Conodes
  - A - Administrators
  - U - Users
## Setting up root
- A1..n identifies with C1..n to get Kpub1..n and Kpriv1..n
  - Kpriv1..n are stored offline and only used to sign requests
  - Kpub1..n are stored in C1..n
- A1 creates a list of A1..n
- A2..n sign off on the list
- For changes, 2/3 or 90% or whatever have to sign off
## Adding config-nodes
- Ai adds config-conodes
- 2/3 of A1..n!=i sign off on the config-conodes
## How to get a config-sc
- U asks root-sc for n nodes

ineiti:
Closed by #872

PR #630, opened 2016-10-05
Scmgr 599

ineiti:
Implements #599


ineiti:
Started implementation in skipchain_835_2

PR #635, opened 2016-10-06
Store protocols in Conode

ineiti:
Implements 632

### DIFF
--- a/sda/conode.go
+++ b/sda/conode.go
@@ -54,7 +54,7 @@ func NewConodeWithRouter(e *network.ServerIdentity, pkey abstract.Scalar, r *net
 	c.overlay = NewOverlay(c)
 	c.serviceManager = newServiceManager(c, c.overlay)
 	c.statusReporterStruct.RegisterStatusReporter("Status", c)
-	for name, inst := range protocols.Instantiators {
+	for name, inst := range protocols.instantiators {
 		log.Lvl4("Registering global protocol", name)
 		c.ProtocolRegister(name, inst)
 	}
@@ -104,7 +104,7 @@ func (c *Conode) ProtocolRegister(name string, protocol NewProtocol) ProtocolID 
 
 // ProtocolInstantiate instantiate a protocol from its ID
 func (c *Conode) ProtocolInstantiate(protoID ProtocolID, tni *TreeNodeInstance) (ProtocolInstance, error) {
-	fn, ok := c.protocols.Instantiators[c.protocols.ProtocolIDToName(protoID)]
+	fn, ok := c.protocols.instantiators[c.protocols.ProtocolIDToName(protoID)]
 	if !ok {
 		return nil, errors.New("No protocol constructor with this ID")
 	}
--- a/sda/conode.go
+++ b/sda/conode.go
@@ -31,8 +31,7 @@ type Conode struct {
 	statusReporterStruct *statusReporterStruct
 	// protocols holds a map of all available protocols and how to create an
 	// instance of it
-	protocols     map[ProtocolID]NewProtocol
-	protocolNames map[ProtocolID]string
+	protocols *ProtocolStorage
 }
 
 // NewConode returns a new Host that out of a private-key and its relating public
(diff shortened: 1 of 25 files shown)


sda/conode.go
```diff
@@ -101,3 +97,26 @@ func (c *Conode) Address() network.Address {
 func (c *Conode) GetService(name string) Service {
 	return c.serviceManager.Service(name)
 }
+
+// ProtocolRegisterName will sign up a new protocol to this Conode.
+// It returns the ID of the protocol.
+func (c *Conode) ProtocolRegisterName(name string, protocol NewProtocol) ProtocolID {
```
nikkolasg: I'm a bit hesitant to have two `ProtocolRegisterName` in sda, one in `sda.Conode` and one a the top lvl of `sda`. It seems they do the same thing while really the former has more power than the latter since it can overwrite the already registered `NewProtocol`. What do you think about `OverwriteNewProtocol` ? Whatever the name is, I'm not comfortable as using the same name.

nikkolasg: And also while I think of it, it could be nice if it would be possible to reduce code duplication - the code below is the same as in the top level method.

sda/conode.go
```diff
@@ -101,3 +97,26 @@ func (c *Conode) Address() network.Address {
...
+		return u
+	}
+	c.protocolNames[u] = name
+	c.protocols[u] = protocol
+	log.Lvl4("Registered", name, "to", u)
+	return u
+}
+
+// ProtocolInstantiate instantiate a protocol from its ID
+func (c *Conode) ProtocolInstantiate(protoID ProtocolID, tni *TreeNodeInstance) (ProtocolInstance, error) {
```
nikkolasg: If the regular top lvl `ProtocolInstantiate` is not used anymore, I'd remove it - the `sda.ProtocolInstantiate`.

sda/conode_test.go
```diff
@@ -0,0 +1,30 @@
+package sda
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConode_ProtocolRegisterName(t *testing.T) {
```
nikkolasg: If you could provide tests for insantiating that'd be nice. It's two small functions so we could aim for high coverage for them.

sda/conode.go
ineiti: The protocols can register themselves only in the `init()`-method, so we need a global registration (renaming it using `GlobalProtocolRegisterName`).

sda/conode.go
ineiti: It doesn't exist anymore in this branch. Do I miss something?

sda/conode_test.go
ineiti: `conode.go` and `protocol.go` are 100% covered - I feel like I should get a gold medal ;)

ineiti: Travis says: no gold medal...

sda/protocol.go
```diff
@@ -37,59 +38,62 @@ type ProtocolInstance interface {
...
 
-// ProtocolNameToID returns the ProtocolID corresponding to the given name
-func ProtocolNameToID(name string) ProtocolID {
-	url := network.NamespaceURL + "protocolname/" + name
-	return ProtocolID(uuid.NewV3(uuid.NamespaceURL, url))
+// ProtocolStorage holds all protocols either globally or per-Conode.
+type ProtocolStorage struct {
+	// Instantiators maps the name of the protocols to the `NewProtocol`-
+	// methods.
+	Instantiators map[string]NewProtocol
```
nikkolasg: Is there a reason why it's public ? Otherwise I would prefer this being unexported, otherwise it's possible to mess up things pretty bad... :D

sda/protocol.go
```diff
@@ -37,59 +38,62 @@ type ProtocolInstance interface {
...
+func (ps *ProtocolStorage) ProtocolExists(protoID ProtocolID) bool {
+	_, ok := ps.Instantiators[ps.ProtocolIDToName(protoID)]
 	return ok
 }
 
-// String returns canonical string representation of the ID
-func (pid ProtocolID) String() string {
-	return uuid.UUID(pid).String()
+// Register takes a name and a NewProtocol and stores it in the structure.
+// If the protocol already exists, a warning is printed.
```
nikkolasg: `... and the NewProtocol is *not* overwritten `

ineiti: Shouldn't it return an `error`?

sda/protocol.go
Liamsi: It looks like the `ProtocolStorage` is only used internally (in the `sda` package). I also think it should be private (as well as `NewProtocolStorage`) if there isn't a good reason for it to be public.

Liamsi:
Besides the potentially unnecessary public structs and methods. This LGTM!

ineiti:
OK - should pass all the @Liamsi - @nikkolasg -tests ;)

PR #638, opened 2016-10-07
Add TLS-host

ineiti:
This adds a TLS-host. Some other changes:
- renamed `NewConode` and `NewConodeWithRouter` to `NewConodeTCP` and `NewConode`
- passing `ServerIdentity` instead of `Address` in connection- and listening- handlers.
- [x] - move `TLSKC.Key` from `ServerIdentity` to `TLSHost`
- [x] - check why it doesn't work with IP-addresses (127.0.0.1 instead of 'localhost').
- [ ] - clean up `CommonName` / `IP` / `DNS`
- [ ] - enter Country, Org, ... on keyboard
- [ ] - also allow `PlainTCP` in a `TLSHost` - does that really make sense?


app/lib/server/server.go
```diff
@@ -144,10 +146,17 @@ func InteractiveConfig(binaryName string) {
 
 	// create the keys
 	privStr, pubStr := createKeyPair()
+	ips, err := net.LookupIP(publicAddress.Host())
+	log.ErrFatal(err)
+	cert, key, err := network.NewCertKey(network.NewTLSCert(big.NewInt(1), "ch", "epfl", "dedis",
```
nikkolasg: I know you will hate my comment but... I think it would GREAT if users 
+ have the possibility of setting their own value for the fields instead of "ch" and "epfl"
+ or give a path to an already PEM encoded certificate + key

It'd be really nice if the first option at least would be possible. The second one can probably wait.
What do you think ?

app/lib/server/server.go
```diff
@@ -180,7 +189,8 @@ func InteractiveConfig(binaryName string) {
 		log.Fatal("Impossible to parse public key:", err)
 	}
 
-	server := config.NewServerToml(network.Suite, public, publicAddress)
+	log.Print(publicAddress)
+	server := config.NewServerToml(network.Suite, public, publicAddress, cert)
```
nikkolasg: just to make it ultra crystal clear: by making this taking a cert we are going in the direction of "everyone" uses TLS. It's good but we then don't worry about backward compatibility. Which for the moment I don't want to since it changes so much. Just to saying the implicit out loud... :D

network/client.go
```diff
@@ -44,6 +44,7 @@ func (cl *Client) Send(dst *ServerIdentity, msg Body) (*Packet, error) {
 	baseIDLock.Unlock()
 	sid := NewServerIdentity(kp.Public, NewAddress(dst.Address.ConnType(),
 		"client:"+strconv.FormatUint(id, 10)))
+	sid.Cert = dst.Cert
```
nikkolasg: ? Is it necessary ? The authentication is done only one-way right ? (i.e. the connectOR knows the certificate of the connectEE but not the inverse)

network/router.go
```diff
@@ -22,7 +22,7 @@ import (
 //   router.Stop() // will stop the listening and the managing of all Conn
 type Router struct {
 	// id is our own ServerIdentity
-	id *ServerIdentity
+	ServerIdentity *ServerIdentity
```
nikkolasg: why do you export it ? Is it necessary ? I don't think it's meant to change...
On another matter, we should rethink our rule "use pointers always" because I had some problems with ServerIdentity pointers, when you change the address (for listening on port 0 for example). One day or another, it will bite us...

network/struct.go
```diff
@@ -107,6 +109,7 @@ func NewServerIdentity(public abstract.Point, address Address) *ServerIdentity {
 		Public:  public,
 		Address: address,
 		ID:      ServerIdentityID(uuid.NewV5(uuid.NamespaceURL, url)),
+		Cert:    []byte{},
```
nikkolasg: Mhhhhh. Instead of putting a default value here, and after calling `si.Cert = tls.Cert`, it could be better to have a `NewTLSServerIdentity` taking the cert directly and setting the Cert to nil in this case (or a flag saying this SI does not have any cert in it).

network/tls.go
```diff
@@ -0,0 +1,171 @@
...
+	h.TLSListener, err = NewTLSListener(si, key)
+	return h, err
+}
+
+// Connect establishes a connection to the remote host given in addr.
+func (h *TLSHost) Connect(si *ServerIdentity) (Conn, error) {
+	log.Lvl3(h.addr, "connecting to", si.Address)
+	addr := si.Address
+	switch addr.ConnType() {
+	case TLS:
```
nikkolasg: Here it should support ALSO the PlainTCP` type.

network/tls.go
```diff
@@ -0,0 +1,171 @@
...
+	if addr.ConnType() != TLS {
+		return nil, errors.New("TLSListener can't listen on non-TLS address")
+	}
+	t := &TLSListener{
+		TCPListener: &TCPListener{
+			quit:         make(chan bool),
+			quitListener: make(chan bool),
+		},
+	}
+	global, _ := GlobalBind(addr.NetworkAddress())
```
nikkolasg: I was planning to re-use the TCPConn to upgrade to TLSConn so to avoid duplication of code. You can do it like explained [here](https://stackoverflow.com/questions/13110713/upgrade-a-connection-to-tls-in-go)

network/tls.go
```diff
@@ -0,0 +1,171 @@
...
+	t.addr = t.listener.Addr()
+	log.Lvl3("Got address", t.addr)
+	return t, nil
+}
+
+// Listen starts to listen for incoming connections and calls fn for every
+// connection-request it receives.
+// If the connection is closed, an error will be returned.
+func (t *TLSListener) Listen(fn func(Conn)) error {
+	receiver := func(tc Conn) {
```
nikkolasg: The same reasoning works here, that was the reason I had `TCPConn` before, to avoid doing a cast from `Conn` to `TCPConn` to get the underlying `net.Conn`

network/tls.go
```diff
@@ -0,0 +1,171 @@
...
+		tls := TLSConn{TCPConn: tc.(*TCPConn)}
```
nikkolasg: The same reasoning works here, that was the reason I had `TCPConn` before, to avoid doing a cast from `Conn` to `TCPConn` to get the underlying `net.Conn`. If we can avoid cast where possible that's best.

network/tls.go
```diff
@@ -0,0 +1,171 @@
...
+	addr := si.Address
+	netAddr := addr.NetworkAddress()
+	config, err := si.Cert.ConfigClient()
+	if err != nil {
+		return nil, err
+	}
+	var conn *tls.Conn
+	for i := 0; i < MaxRetryConnect; i++ {
+		log.Lvl3("Connecting to", netAddr)
+		conn, err = tls.Dial("tcp", netAddr, config)
```
nikkolasg: Same thing here, you can just upgrade your connection, and avoid all the duplication.

network/tls.go
```diff
@@ -0,0 +1,171 @@
...
+		if err == nil {
+			log.Lvl3("client: connected to:", conn.RemoteAddr())
+
+			state := conn.ConnectionState()
+			for _, v := range state.PeerCertificates {
+				b, _ := x509.MarshalPKIXPublicKey(v.PublicKey)
+				log.Lvlf3("Connection by %s with key %x",
```
nikkolasg: Is this for testing / debugging purpose ? I'm not sure we need to see each certificate for each user connection.. And btw, out of curiosity, what is `PeerCertificates` ? The cert +  the intermediates ? Why is there an array instead of a single entity as in `ServerIdentity` ?

network/tls.go
```diff
@@ -0,0 +1,171 @@
...
+	return t.listen(receiver)
+}
+
+// TLSConn implements the Conn interface using TLS.
+type TLSConn struct {
+	*TCPConn
+}
+
+// NewTLSConn will open a TLSConn to the given address.
+// In case of an error it returns a nil TLSConn and the error.
```
nikkolasg: I would just put `it returns nil` :  a `nil TLSConn` is strange.

network/tls_test.go
```diff
@@ -0,0 +1,67 @@
...
+		[]net.IP{})
+	_, _, err := NewCertKey(c1, 2048)
+	log.ErrFatal(err)
+	_, _, err = NewCertKey(c2, 2048)
+	log.ErrFatal(err)
+}
+
+func TestNewTLSRouter(t *testing.T) {
+	si1 := NewTestServerIdentity("tls://localhost:2000")
+	si2 := NewTestServerIdentity("tls://localhost:2001")
```
nikkolasg: NICEEEEEEE

network/tlskc.go
```diff
@@ -0,0 +1,99 @@
...
+			CommonName:         "*",
+		},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(validYear, 0, 0),
+		SubjectKeyId:          subjectKeyID,
+		BasicConstraintsValid: true,
+		IsCA:        true,
+		IPAddresses: ips,
+		ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		KeyUsage: x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign |
```
nikkolasg: Mhhh, since you coded it, would you be ok to just make us (or just me) a quick crash course on all these parameters ? What they mean and what are the choices you chose and why ? It can still be quick :D And while I think of it, even better... put that in the doc (comment) !

app/lib/config/config.go
```diff
@@ -33,6 +33,8 @@ type CothoritydConfig struct {
 	Public  string
 	Private string
 	Address network.Address
+	TLSCert string
+	TLSKey  string
```
Liamsi: Shouldn't both be bytes instead?

network/tlskc.go
```diff
@@ -0,0 +1,99 @@
...
+			x509.KeyUsageKeyAgreement,
+	}
+}
+
+// NewTLSKC returns a TLSKC-structure with a PEM-encoded certificate and key.
+func NewCertKey(ca *x509.Certificate, keyLen int) (cert TLSCertPEM, key TLSKeyPEM, err error) {
+	if keyLen < 2048 {
+		log.Warn("Small key-length:", keyLen)
+	}
+	priv, _ := rsa.GenerateKey(rand.Reader, keyLen)
```
nikkolasg: **DISCLAIMER** This does not have to integrated right now, but if not, then we should make an issue about it.
Would it be possible to generate elliptic curves keys instead of an old RSA ? We all know all the benefits of switching to elliptic curve crypto and since we are in a security lab, I figure it should be possible. I've quickly glanced at the RFC and it's supported, I don't know which suites yet (P256 I'd guess): https://tools.ietf.org/search/rfc4492#page-27

app/lib/config/config.go
```diff
@@ -135,6 +143,7 @@ type ServerToml struct {
 	Address     network.Address
 	Public      string
 	Description string
+	Cert        string
```
Liamsi: Shouldn't this be bytes?

network/tlskc.go
nikkolasg: Addition: Goland supports the ECDSA key ([link](https://golang.org/pkg/crypto/x509/#CreateCertificate)):
> All keys types that are implemented via crypto.Signer are supported (This includes *rsa.PublicKey and *ecdsa.PublicKey.)

network/local.go
```diff
@@ -460,13 +460,14 @@ func NewLocalHostWithManager(lm *LocalManager, addr Address) (*LocalHost, error)
 // Connect sets up a connection to addr. It retries up to
 // MaxRetryConnect while waiting between each try.
 // In case of an error, it will return a nil Conn.
-func (lh *LocalHost) Connect(addr Address) (Conn, error) {
-	if addr.ConnType() != Local {
+// TODO
```
Liamsi: TODO what?

network/tlskc.go
```diff
@@ -0,0 +1,99 @@
...
+	key = pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(priv)})
+	return
+}
+
+func secureConfig(c *tls.Config) *tls.Config {
+	c.CipherSuites = []uint16{tls.TLS_RSA_WITH_AES_128_CBC_SHA,
+		tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+		tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
+		tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
```
nikkolasg: Have you tested all supported suites ? I think we should support only the one we generate keys from since we don't have a typical "client -> server" architecture, but I'm newbie in the TLS field. What do you think ?

Liamsi:
Linter fails: https://travis-ci.org/dedis/cothority/builds/166438573#L183

network/tlskc.go
```diff
@@ -0,0 +1,99 @@
...
+	cert = pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE",
+		Bytes: x509cert})
+	key = pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(priv)})
+	return
+}
+
+func secureConfig(c *tls.Config) *tls.Config {
+	c.CipherSuites = []uint16{tls.TLS_RSA_WITH_AES_128_CBC_SHA,
+		tls.TLS_RSA_WITH_AES_256_CBC_SHA,
```
Liamsi: From the above it looks like we only use and support `RSA`?

network/router.go
nikkolasg: +1

app/lib/server/server.go
Liamsi: Nodes will not only be ran by the org/domain dedis.epfl.ch ...  For tests it's OK to hardcode this but not for actual code.

app/lib/config/config.go
Liamsi: Or, if it is difficult to use byte arrays in toml files: It would be better if (at least) the private key could be stored in a separate file (in a user configurable directory). Then, this would be OK to be left a string.

ineiti: these are the PEM-encoded data, so string makes more sense here. 

I could also store them as `[]byte` and add a `PEM`-method and a NewTLSKeyFromPEM`...  But that seems to be overkill to me. Even though I always have to encode/decode in PEM.

app/lib/config/config.go
ineiti: same thing - `Cert` is PEM-encoded.

app/lib/config/config.go
ineiti: And with regard to the private key - this is the configuration-file of the server, so I think it's OK to store the private key in there. The Conode-private key is just 3 lines above...

app/lib/server/server.go
ineiti: I'll never hate you, or only for a short time - Love is the source of all live ;)

Do you think it's better to add it as arguments to `setup` or add more user-input?

With regard to already existing PEM encoded certificate + key - where would those come from? Apache/Postfix? Letsencrypt?

app/lib/config/config.go
Liamsi: > these are the PEM-encoded data, so string makes more sense here.

OK. I was (and still am a little) surprised because everywhere `TLSCert` is used you need to cast it to `[]byte`... Why is that? But of course, if it is just a base64 encoded string, it needs to be saved as a string 👍

app/lib/server/server.go
ineiti: Yes, every new conode will use TLS. I could add a switch `-notls` to `setup` which would make a PlainTCP connection.

network/client.go
ineiti: It makes it more easy to handle the `ServerIdentity` - if it gets stored in a configuration-file and used by another client, the cert is already there.

network/router.go
ineiti: Yes, it's used in `sda/conode.go:44`.

network/tlskc.go
ineiti: no, I didn't test them. I copied those lines from 

http://www.hydrogen18.com/blog/your-own-pki-tls-golang.html

I don't even know whether the lines `TLS_ECDHE` make sense if we have an RSA-certificate.

network/router.go
ineiti: IMO it is related to this issue: `Send` and `Connect` only took `Address` before, but that is not enough with TLS, we need the `Cert` which is in `ServerIdentity`.

network/tlskc.go
```diff
@@ -0,0 +1,99 @@
...
+func NewTLSCert(serial *big.Int, country, org, orgUnit string,
+	validYear int, subjectKeyID []byte,
+	ips []net.IP) *x509.Certificate {
+	return &x509.Certificate{
+		SerialNumber: serial,
+		Subject: pkix.Name{
+			Country:            []string{country},
+			Organization:       []string{org},
+			OrganizationalUnit: []string{orgUnit},
+			CommonName:         "*",
```
ineiti: Good question - we'll be connecting using IP-addresses anyway, except with the simulations, where it uses 'localhost'. We can remove the `CommonName` and add a `DNSNames`. Or set that later?

network/tlskc.go
ineiti: exactly - I'll try to use `EC` and remove the `CipherSuites` that are not used.

app/lib/server/server.go
ineiti: like above - there will be either arguments or more input lines.

nikkolasg: > Do you think it's better to add it as arguments to setup or add more user-input?

I would say more user input, like when you create a certificate using openssl.

> With regard to already existing PEM encoded certificate + key - where would those come from? Apache/Postfix? Letsencrypt?

I did not think of that but indeed that would be neat if certificates could be directly tied into the global "PKI".

network/tlskc.go
Liamsi: >I don't even know whether the lines TLS_ECDHE make sense if we have an RSA-certificate.

I don't think so. If you want to support all these suites we would need to let the user chose which method they want to use; like the golang devs did here: https://golang.org/src/crypto/tls/generate_cert.go

network/router.go
Liamsi: I was only referring to the renaming. It's good to do such refactoring if you see it, just open a separate PR so it's easier to review the changes here.

network/tlskc.go
Liamsi: Both should be possible for the user to set: `DNSNames` and `IPAddresses`

Liamsi: (And the user needs to chose at least one)

network/struct.go
nikkolasg: Ok I've looked into protobuf and if you define the field as : 
``` type myStruct struct {
...
Cert []byte `protobuf:"opt"`
...
}
```
It becomes optional so protobuf does not complain anymore.
It makes much sense since TLS is *still* optional !
The thing with having a `NewTLSServerIdentity` is it would avoid doing all these assigments `si.Cert = cert` right after creating a `ServerIdentity`. I'm proposing one way to get rid of that, if you find another one that's also good.

app/lib/server/server.go
ineiti: One thing I don't like about user-input is that it's difficult to script it. Of course you can use `echo -e "arg1\narg2\narg3"`, but then you cannot change the order of the inputs anymore because you would break the scripts.

app/lib/server/server.go
ineiti: ok, no `-notls` then.

network/client.go
ineiti: Of course, you're right, sorry - been in another part of the code. Removed.

network/struct.go
ineiti: I changed `network.TLS(Cert|Key)PEM` to `type string`, so this problem does not show up anymore.
If you think a `NewTLSServerIdentity` makes sense, I'll add one. But to me a `NewServerIdentity` followed by a `si.Cert = cert` is OK.

network/tls.go
ineiti: Copy/paste cruft - but might be OK for debugging - no?

network/tlskc.go
ineiti: Oups - got me. I'll get wise and add comments to it.

network/router.go
ineiti: Got in a separate PR - #648 - that is merged in the meantime.

app/lib/server/server.go
nikkolasg: Well, the tests scripts are not the end goal, but rather the user don't you think ? It's our job to make the two work together. 
+ user input: the user is being asked explicitely for each each so it's kinda clear
+ arguments: well, if there were only 2 inputs why not, but there's more into a certificate so having a very loooooong line of options when generating the setup... I'm not sure. And we already do user input, don't you think it's strange to ask for options and inputs at the same time for different things ?

network/struct.go
nikkolasg:
```
sid := NewTLSServerIdentity(public,address,cert)
```
vs 
```
sid := NewServerIdentity(public, address)
sid.Cert = cert
```

@Liamsi will maybe have to be the deciding party but to me it's clear which version is shorter / cleaner and have less chance of making weird bug if we forget this `sid.Cert = cert` line somewhere...
After the refactoring PRs, it's my turn to nit pick the code of others now .. ! :D

app/lib/server/server.go
nikkolasg: Well our goal is more the end user than our tests, I think it's our job to make the two work together. But anyway, we already started with user inputs for the addresses, don't you think it's a bit weird to have both user inputs value + user options at the same time for a one-time operations ? I do..  Moreover, a cert a many fields and having a looooong command line with many options seems rather unelegant also 
```
./cothorityd setup -DNSName "blabla" -owner "me" -address "here" -org "none for the moment" -Unit "blabla" ....
...
## USER INPUT FOR IP ADDRESS ##
```

I prefer having our test being not elegant that our command line. what do you think ?

network/tls.go
nikkolasg: I really don't think so. Looking at the doc of golang TLS does not seem to imply any more roundtrip and I'd be surprised if it did, it would be really inneficient (and golang is already used in many production ready projects using TLS so..)

network/tlskc.go
ineiti: OK, as you might have guessed, `ECDHE` is for the Diffie-Hellman exchange, and can be used with both `RSA` or `ECDSA` keys.  So only the `RSA` and `ECDSA` need to match. Makes sense ;)

network/tls.go
nikkolasg: Then a lvl4 or 5 should be enough if we need to fall into this rabbit hole, 3 is already much crowded.

network/struct.go
ineiti: "You have heard: bug for bug, nitpick for nitpick, but I say: do not resist commenters. If they comment on a bad line, show them your worst..."

Anyway - added a `NewTLSServerIdentity`, even though it is used twice for the moment.

network/tls.go
ineiti: Do we really want to have a `TLSHost` accept `PlainTCP` connections? I know we said so, but it seems insecure and an easy attack, no?

ineiti:
Replaced by https://github.com/dedis/onet/pull/101

Issue #228, opened 2016-03-01
Implement mininet-platform


ineiti:
Write a mininet-platform to be used on the EPFL-cluster.

As the EPFL-cluster needs to be set up from scratch, add scripts to do the setup automatically.

ineiti:
New PR in #443

ineiti:
Done

PR #640, opened 2016-10-12
Implement mininet-platform 228

ineiti:
Implements #228 
Added:
- [x] forward output of conode from server to host
- [x] call conodes with same `-debug` as the simulator
- [x] debug problems with archlinux
- [x] propagate `DEBUG_TIME` to `start.py`
- [x] merge back things from snp17-run


simul/platform/cliutils_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"bytes"
+
+	"os"
+
+	"path"
+
+	"github.com/dedis/cothority/log"
+)
+
+func TestCopy(t *testing.T) {
```
Liamsi: Same comment as for `Copy`.

simul/platform/mininet.go
```diff
@@ -0,0 +1,372 @@
...
+	m.External = m.HostIPs[0]
+	m.ProxyAddress = "localhost"
+	m.MonitorPort = pc.MonitorPort
+	m.Debug = pc.Debug
+	m.Delay = 0
+	m.Bandwidth = 1000
+
+	// Clean the MiniNet-dir, create it and change into it
+	os.RemoveAll(m.buildDir)
+	os.Mkdir(m.buildDir, 0700)
```
Liamsi: For both methods there is no error-handling on purpose?

simul/platform/mininet/start.py
```diff
@@ -0,0 +1,243 @@
...
+    net = Mininet(topo=topo, link=TCLink)
+    net.start()
+
+    for host in net.hosts[1:]:
+        host.startCothority()
+
+    # Also set setLogLevel('info') if you want to use this, else
+    # there is no correct reporting on commands.
+    # CLI(net)
+    while not os.path.exists(logdone):
```
nikkolasg: I'm sure you already searched for it but is there not a way to do like in the `simul` package, to wait until the cothority is done ? Using the utility `wait` for example.

simul/platform/mininet.go
```diff
@@ -0,0 +1,372 @@
...
+		log.Lvl3("Cleaning up server", h)
+		_, err = SSHRun(m.Login, m.External, "pkill -9 -f start.py")
+		if err != nil {
+			log.Lvl2("Error while cleaning up:", err)
+		}
+	}
+	//err = SSHRunStdout(m.Login, m.Host, "ps aux")
+	//if err != nil {
+	//	log.Lvl3(err)
+	//}
```
Liamsi: Remove comment?

simul/platform/mininet/install_mininet.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+# Package installation
+echo 'export LC_ALL="en_US.UTF-8"' >> /etc/environment
+. /etc/environment
+apt-get update
+apt-get install -y screen rsync software-properties-common git vim cifs-utils pv htop mtr \
```
Liamsi: Since 8.0 it can do BFTCoSi !

ineiti: and blockchains...

simul/platform/mininet/setup_servers.sh
```diff
@@ -0,0 +1,58 @@
...
+	done
+	sleep 2
+done
+
+echo -e "\nAll servers are done installing - copying ssh-keys"
+
+rm -f server_list
+for s in $SERVERS; do
+	login=root@$s
+	cat $KEYS | ssh $login "cat - >> .ssh/authorized_keys"
```
ineiti: I start to set up `ssh-copy-id` between the host-machine and the servers, and this will ask for a password for every connection. In order not to have that prompt for every connection between the servers, I copy the public-keys with this line. Saves (n^2 - n) password-entries.

simul/platform/mininet/setup_servers.sh
```diff
@@ -0,0 +1,58 @@
...
+	ip=$( host $s | sed -e "s/.* //" )
+	ssh root@$SERVER_GW "ssh-keyscan -t ssh-ed25519 $s >> .ssh/known_hosts 2> /dev/null"
```
ineiti: copies the fingerprint of the remote server, so that you don't get the "remote fingerprint is - do you trust that server?"

sda/overlay.go
```diff
@@ -177,7 +177,7 @@ func (o *Overlay) TransmitMsg(sdaMsg *ProtocolMsg) error {
 	done := o.instancesInfo[sdaMsg.To.ID()]
 	o.instancesLock.Unlock()
 	if done {
-		log.Error("Message for TreeNodeInstance that is already finished")
+		//log.Error("Message for TreeNodeInstance that is already finished")
```
ineiti: `Lvl2` is fine for me. But `log.Error` with the JVSS-protocol is confusing.

simul/platform/cliutils.go
```diff
@@ -23,14 +27,38 @@ func Scp(username, host, file, dest string) error {
...
+		return err
+	}
+	defer fDst.Close()
+	_, err = io.Copy(fDst, fSrc)
+	return err
+}
+
 // Rsync copies files or directories to the remote host. If the DebugVisible
 // is > 1, the rsync-operation is displayed on screen.
 func Rsync(username, host, file, dest string) error {
```
ineiti: `Rsync` is used only here, so I'd leave it (contrary to `Copy`).

simul/platform/mininet.go
```diff
@@ -0,0 +1,372 @@
...
+	m.ProxyAddress = "localhost"
+	m.MonitorPort = pc.MonitorPort
+	m.Debug = pc.Debug
+	m.Delay = 0
+	m.Bandwidth = 1000
+
+	// Clean the MiniNet-dir, create it and change into it
+	os.RemoveAll(m.buildDir)
+	os.Mkdir(m.buildDir, 0700)
+	sda.WriteTomlConfig(*m, m.mininetDir+"/mininet.toml", m.mininetDir)
```
ineiti: You think these things will be used outside of `simul`?

simul/platform/mininet.go
```diff
@@ -0,0 +1,372 @@
...
+		}
+		m.started = false
+	}
+	return nil
+}
+
+// Returns the servers to use for mininet.
+func (m *MiniNet) parseServers() error {
+	hosts, err := ioutil.ReadFile(path.Join(m.mininetDir, "server_list"))
+	if err != nil {
```
ineiti: I'd rather avoid user-inputs. Deterlab is special, because you don't change the configuration. Whereas mininet (at least with iccluster) will need a new set of servers every simulation... But a better error-message is the least I can do.

simul/platform/mininet.go
```diff
@@ -0,0 +1,372 @@
...
+		return err
+	}
+	m.HostIPs = []string{}
+	for _, hostRaw := range strings.Split(string(hosts), "\n") {
+		h := strings.Replace(hostRaw, " ", "", -1)
+		if len(h) > 0 {
+			ips, err := net.LookupIP(h)
+			if err != nil {
+				// Hope it's an IP-address
```
ineiti: Oups - that was wrong. The 'hope' was for what's happening if `err != nil`, so if `LookupIP` doesn't work. I thought that this is happening if we give IP-addresses to `LookupIP`. But that's not the case. So I abandon all hope and just return the error.

simul/platform/mininet.go
```diff
@@ -0,0 +1,372 @@
...
+// getHostList returns the hosts that will be used for the communication
+// and the list for `start.py`.
+func (m *MiniNet) getHostList(rc RunConfig) (hosts []string, list string, err error) {
+	hosts = []string{}
+	list = ""
+	physicalServers := len(m.HostIPs)
+	nets := make([]*net.IPNet, physicalServers)
+	ips := make([]net.IP, physicalServers)
+	for n := range nets {
+		ips[n], nets[n], err = net.ParseCIDR(fmt.Sprintf("10.%d.0.0/16", n+1))
```
ineiti: I put some comments at the beginning and in the file - does it make more sense now?

simul/platform/mininet/start.py
```diff
@@ -0,0 +1,243 @@
...
+            self.cmd('/usr/sbin/sshd -D &')
+
+    def startCothority(self):
+        socat="socat -v - udp-sendto:%s:%d" % (self.gw, socatPort)
+
+        args = "-debug %s -address %s:2000 -simul %s" % (debugging, self.IP(), self.simul)
+        if True:
+            args += " -monitor %s:10000" % global_root
+        ldone = ""
+        if self.IP().endswith(".0.2"):
```
ineiti: Added a comment. ".0.1" is the 'BaseRouter', and ".0.2" is the first conode, then comes ".0.3", ..., ".0.255", ".1.0", ...

simul/platform/mininet/start.py
nikkolasg: You can give a PID to wait on to `wait` so you can just collect all PIDs and wait them in a loop.  But anyway, yeah there's not huge benefits so it's OK :)

simul/platform/mininet/start.py
nikkolasg: Cool thx.

simul/platform/mininet.go
ineiti: `RemoveAll` can fail if the directory is not there yet. But `MkDir` just got a `log.ErrFatal`.

simul/platform/mininet.go
```diff
@@ -0,0 +1,392 @@
...
+	DebugTime bool
+	// Whether to show color debugging-messages
+	DebugColor bool
+	// The number of seconds to wait for closing the connection
+	RunWait int
+	// Delay in ms of the network connection
+	Delay int
+	// Bandwidth in Mbps of the network connection
+	Bandwidth int
+}
```
Liamsi: This struct is almost identical to the deterlab one: https://github.com/dedis/cothority/blob/master/simul/platform/deterlab.go#L38

I wonder if we should refactor out a common struct instead of we are currently doing: basically copy&paste plus modifying comments and adding/removing very few fields. 
This might be out of scope of this PR... Do you think it's worth it to open an issue for that (or add it to an already existing more general refactoring issue)?

simul/platform/mininet.go
```diff
@@ -0,0 +1,392 @@
...
+	log.ErrFatal(os.Mkdir(m.buildDir, 0700))
+	sda.WriteTomlConfig(*m, m.mininetDir+"/mininet.toml", m.mininetDir)
+
+	if m.Simulation == "" {
+		log.Fatal("No simulation defined in runconfig")
+	}
+
+	// Setting up channel
+	m.sshMininet = make(chan string)
+}
```
Liamsi: Similar comment as above: https://github.com/dedis/cothority/blob/master/simul/platform/deterlab.go#L91
Here the code duplication is far less, but I still think we could do better.

simul/platform/platform.go
```diff
@@ -44,6 +44,7 @@ type Config struct {
 
 var deterlab = "deterlab"
 var localhost = "localhost"
+var mininet = "mininet"
```
Liamsi: Maybe, change to 
```
const (
	deterlab  = "deterlab"
	localhost = "localhost"
	mininet   = "mininet"
)
```

simul/runfiles/mininet_cosi.toml
```diff
@@ -0,0 +1,12 @@
+Simulation = "CoSimul"
+Servers = 16
+Bf = 8
+Rounds = 10
+CloseWait = 6000
```
Liamsi: You renamed this [above](https://github.com/dedis/cothority/pull/640/files#diff-7c46770087ffa7cc7eefbe702ec35654L84). Either change it back to the original name in the golang code (`CloseWait`) or consistently rename it in all toml files.

simul/platform/mininet.go
ineiti: The differences are in the detail - different computer-types, how they are copied and started, ...

It's probably not worth it to put those together - even though I'd like to add a third, more 'general' type that only needs ssh and runs one node per server.

simul/runfiles/mininet_cosi.toml
ineiti: ;)

simul/platform/mininet.go
Liamsi: Here, is a diff between both (deterlab/minitnet): https://www.diffchecker.com/KDSLPbbh
I'm OK with leaving it as it is.

ineiti:
Continued in https://github.com/dedis/onet/pull/100

PR #641, opened 2016-10-12
Quickfix: Change default behavior of dbg library to monochrome output

### LINKED ISSUE #617: Monochrome debugging output
Switch debugging output per default to monochrome. Only if DEBUG_COLOR is set, use colors.

Liamsi:
Mini PR; should resolve #617

### DIFF
--- a/log/lvl.go
+++ b/log/lvl.go
@@ -116,8 +116,9 @@ var debugVisible = 1
 // If showTime is true, it will print the time for each line of debug-output.
 var showTime = false
 
-// If useColors is true, debug-output will be colored.
-var useColors = true
+// If useColors is true, debug-output will be colored (defaults to monochrome
+// output).
+var useColors = false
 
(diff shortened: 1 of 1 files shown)


ineiti:
- [ ] - change comment of `DEBUG_COLOR`
- [ ] - discuss whether the first `log.Lvl` should print out "if you want color, use DEBUG_COLOR=true"

nikkolasg:
> discuss whether the first log.Lvl should print out "if you want color, use DEBUG_COLOR=true"

I'm not a fan of printing that unless asked with `--help`. It sure should at least be included in the option help text of `-debug` and in the README.

Liamsi:
> I'm not a fan of printing that unless asked with --help. It sure should at least be included in the option help text of -debug and in the README.

I agree, shouldn't print without the `--help` flag. The flag `debug-color` already comes with help text.

Issue #619, opened 2016-09-29
Cothority-stats


ineiti:
Bastian Nanchen's project on creating a status-webpage for the running cothority-nodes. Its goal is to present in a nice way the running cothority nodes and their status, as well as interact with the nodes for collective signing and/or timestamping.

PR #642, opened 2016-10-12
Cothority stats

ineiti:
Implements #619


ineiti:
Important parts have been merged.

PR #643, opened 2016-10-12
using numpy-arrays

ineiti:
small change in stats.py that allows usage of read .csv-file with numpy - ported from `nsdi_base`.

### DIFF
--- a/simul/matplotlib/stats.py
+++ b/simul/matplotlib/stats.py
@@ -2,7 +2,7 @@
 
 import csv
 import unittest
-
+import numpy as np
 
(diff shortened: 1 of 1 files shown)


Liamsi:
A description why this is neccessary would have been nice. But as you explained it to me now, I think this is ready to be merged.

nikkolasg:
If you already tested it for generating graphs and such, then I'm OK to merge also.

Liamsi:
> If you already tested it for generating graphs and such, then I'm OK to merge also.

Merging it. As this comes from the nsdi branch, I guess we already used it to create the plots for the papers.

PR #644, opened 2016-10-13
unified creation of tcp/local conode for local

nikkolasg:
This requests bring a small fix to the sda Local API. Namely, it enables on to create only *one* Conode without specifying the type (local/tcp). The `Local` is doing that for us. I needed that during coding the `randhoundco` protocol.

### DIFF
--- a/sda/local.go
+++ b/sda/local.go
@@ -364,15 +364,16 @@ func (l *LocalTest) NewConode(port int) *Conode {
 	default:
 		conode = l.NewLocalConode(port)
 	}
-	l.Conodes[conode.ServerIdentity.ID] = conode
-	l.Overlays[conode.ServerIdentity.ID] = conode.overlay
-	l.Services[conode.ServerIdentity.ID] = conode.serviceManager.services
 	return conode
 }
 
+// NewTCPConode returns a new TCP Conode attached to this LocalTest.
 func (l *LocalTest) NewTCPConode() *Conode {
 	conode := NewTCPConode(0)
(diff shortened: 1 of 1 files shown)


nikkolasg:
Why coveralls is not there .. ?

ineiti:
Do we need a test for this?
Can you remind me how you set `mode` of the `LocalTest`?

Liamsi:
Regarding the test: as there wasn't one before, I won't root for it. 
Similar to @ineiti with regard to `mode`: I also don't know how to set it and was surprised but as this doesn't change the logic and I assume this was well reviewed and understood at some point, I guess it's OK. 
LGTM.

nikkolasg:
So the mode is by default `Local`, you can change the default to `TCP` in the constructor of `LocalTest` which is `NewLocalTest()`. 
If you need explicitely a `TCP` layer, then you can call `NewTCPTest()` which returns a `LocalTest` using explicitely TCP connections.

Regarding the test... Since `LocalTest` is only for test and I have other stuff to do, I'm lazy to do a test for a test. BUT my moral obligation towards democracy makes me abide to the majority, so if you deem it's useful and necessary, I'll code it.

Liamsi:
> Regarding the test... Since LocalTest is only for test and I have other stuff to do, I'm lazy to do a test for a test. BUT my moral obligation towards democracy makes me abide to the majority, so if you deem it's useful and necessary, I'll code it.

As I said, I won't root for it. In other words: similar to what @ineiti wrote it just was my first thought, but, I don't think it's necessary!

Thanks for the clarifcation about `mode`.

PR #645, opened 2016-10-13
Create a new protocol from a TreeNodeInstance

nikkolasg:
This PR brings a small feature long-wanted: the ability to isntantiate a protocol from an already running protocol.


ineiti:
Looks good - I'd put the test in `treenode_test.go` instead of `protocol.go`.

sda/treenode.go
```diff
@@ -605,6 +605,14 @@ func (n *TreeNodeInstance) SendToChildrenInParallel(msg interface{}) error {
 	return collectErrors("Error while sending to %s: %s\n", errs)
 }
 
+// CreateProtocol makes SDA instantiates a new protocol of name "name" and
+// returns it with any error that might have happened during the creation. This
+// protocol is only handled by SDA, no service are "attached" to it.
+func (n *TreeNodeInstance) CreateProtocol(name string, t *Tree) (ProtocolInstance, error) {
+	pi, err := n.overlay.CreateProtocolSDA(name, t)
+	return pi, err
+}
```
Liamsi: Cool, that's it?! Why did we wait so long for that.

nikkolasg: Because we did not really need it that much. lazy programmers.

ineiti: There was a discussion whether this makes sense or not - because only `service`s should instantiate protocols.

Liamsi:
LGTM. Will merge as soon as travis/coveralls is done.

Liamsi:
> LGTM. Will merge as soon as travis/coveralls is done.
About this...
https://travis-ci.org/dedis/cothority/builds/167329857#L251
Looks like there still some deadlock occurring rarely? Or could this be related to this pull (doesn't look like this to me).

Liamsi:
As the build was restarted and the deadlock/race doesn't occur every time the failing output is gone for now. @nikkolasg knows that the test in question was `testRouterLotsOfConnLocal`.

PR #648, opened 2016-10-14
Network 647

### LINKED ISSUE #647: Network changes
A proposal for the network-library - pre-work for TLS implementation.
- `NewConode` takes a router
- `NewConodeTCP` directly instantiates a TCP-router
- Use `ServerIdentity` for `Connect`
- `coveralls.sh` change to allow for local `make test`

ineiti:
resolves #647


sda/conode.go
```diff
@@ -34,18 +34,10 @@ type Conode struct {
...
-	return NewConodeWithRouter(e, pkey, r)
-}
-
-// NewConodeWithRouter returns a fresh Host with a given Router.
-func NewConodeWithRouter(e *network.ServerIdentity, pkey abstract.Scalar, r *network.Router) *Conode {
+// NewConode returns a fresh Host with a given Router.
+func NewConode(r *network.Router, pkey abstract.Scalar) *Conode {
 	c := &Conode{
-		ServerIdentity:       e,
+		ServerIdentity:       r.ServerIdentity,
```
nikkolasg: since now `Router` has an `ServerIdentity` field and is embedded into `Conode` I guess you don't need that field anymore, calling `conode.ServerIdentity` should work.

sda/service.go
```diff
@@ -386,15 +386,6 @@ func NewClient(s string) *Client {
 	}
 }
 
-// NewLocalClient is for test mainly, it uses the local comunication network
-// offered by the network package.
-func NewLocalClient(s string) *Client {
```
nikkolasg: why do you delete the local client ?

ineiti: There is a `NewLocalClient` in `LocalTest` which seems to be enough.

PR #657, opened 2016-10-19
return correct IPv6 unspecified address in Address.Host

ineiti:
When using `SplitHostPort`, the IPv6 unspecified address is returned as "::", whereas other IPv6 addresses are returned in "[]". This patch unifies the behaviour, so that `Address.Host` can be used in all circumstances.

### DIFF
--- a/network/address.go
+++ b/network/address.go
@@ -152,7 +152,8 @@ func (a Address) Port() string {
 func (a Address) Public() bool {
 	private, err := regexp.MatchString("(^127\\.)|(^10\\.)|"+
 		"(^172\\.1[6-9]\\.)|(^172\\.2[0-9]\\.)|"+
-		"(^172\\.3[0-1]\\.)|(^192\\.168\\.)|(^169\\.254)", a.NetworkAddress())
+		"(^172\\.3[0-1]\\.)|(^192\\.168\\.)|(^169\\.254)|"+
+		"(^\\[::\\])", a.NetworkAddress())
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


nikkolasg:
OK to merge once travis is happy ;)

network/address_test.go
```diff
@@ -51,12 +53,13 @@ func TestAddress(t *testing.T) {
...
-		assert.Equal(t, str.Address, add.NetworkAddress())
-		assert.Equal(t, str.Host, add.Host())
-		assert.Equal(t, str.Port, add.Port())
-		assert.Equal(t, str.Public, add.Public())
+		require.Equal(t, str.Valid, add.Valid(), "Address (%d) %s", i, str.Value)
+		require.Equal(t, str.Type, add.ConnType(), "Address (%d) %s", i, str.Value)
+		require.Equal(t, str.Address, add.NetworkAddress())
+		require.Equal(t, str.Host, add.Host())
+		require.Equal(t, str.Port, add.Port())
+		require.Equal(t, str.Public, add.Public())
```
nikkolasg: Ok I don't mind so much but here is a great example where `assert` is exactly what we need because the things that are tested after a failing assertion are independant from the failing assertion. In short, we can see all errors at once instead of failing over the first one.

nikkolasg: So I'm waiting on a short feedback from you and then I'll merge (so I'm sure you saw my comment :D )

Liamsi:
Looks good to me. Merging

network/address.go
```diff
@@ -122,6 +122,10 @@ func (a Address) Host() string {
 	if e != nil {
 		return ""
 	}
+	// IPv6 global address has to be in brackets.
```
Liamsi: Minor nitpick: This comment is confusing. Should be rather "unspecified address" instead of "global adress"

PR #658, opened 2016-10-20
Add ProtocolRegister

### LINKED ISSUE #656: Add ProtocolRegister to Context
Now that each conode has its own set of Protocols, the `ProtocolRegister` method should also be made available to services. This makes it possible for services to register their own protocol which are only valid for one conode and not stored in the global protocol-registration. We need this for simulations where more than one conode is started in one process, and the services use a protocol that depends on service-internal-variables.

ineiti:
resolves #656 and implements the method in `services/identity/service.go` as an example.

### DIFF
--- a/log/lvl.go
+++ b/log/lvl.go
@@ -64,6 +64,7 @@ var stdOut io.Writer
 var stdErr io.Writer
 
 func init() {
+	ParseEnv()
 	stdOut = os.Stdout
 	stdErr = os.Stderr
 }
--- a/log/lvl.go
+++ b/log/lvl.go
@@ -64,7 +64,6 @@ var stdOut io.Writer
 var stdErr io.Writer
 
 func init() {
-	ParseEnv()
 	stdOut = os.Stdout
 	stdErr = os.Stderr
 }
--- a/protocols/manage/propagate.go
+++ b/protocols/manage/propagate.go
@@ -28,8 +28,8 @@ type Propagate struct {
 		PropagateReply
 	}
 
-	received int
-	subtree  int
+	received     int
+	subtreeCount int
 	sync.Mutex
 }
 
@@ -64,7 +64,7 @@ func NewPropagationFunc(c *sda.Context, name string, f PropagationStore) (Propag
 			sd:               &PropagateSendData{[]byte{}, 1000},
 			TreeNodeInstance: n,
 			received:         0,
-			subtree:          n.TreeNode().SubtreeCount(),
+			subtreeCount:     n.TreeNode().SubtreeCount(),
 			onData:           f,
 		}
 		for _, h := range []interface{}{&p.ChannelSD, &p.ChannelReply} {
@@ -78,7 +78,7 @@ func NewPropagationFunc(c *sda.Context, name string, f PropagationStore) (Propag
 		name, pid)
 	return func(el *sda.Roster, msg network.Body, msec int) (int, error) {
 		tree := el.GenerateNaryTreeWithRoot(8, c.ServerIdentity())
-		log.Lvl3("Starting to propagate", reflect.TypeOf(msg))
+		log.Lvl3(el.List[0].Address, "Starting to propagate", reflect.TypeOf(msg))
 		pi, err := c.CreateProtocolSDA(name, tree)
 		if err != nil {
 			return -1, err
@@ -147,11 +147,11 @@ func (p *Propagate) Dispatch() error {
 			}
 		case <-p.ChannelReply:
 			p.received++
-			log.Lvl4(p.ServerIdentity(), "received:", p.received, p.subtree)
+			log.Lvl4(p.ServerIdentity(), "received:", p.received, p.subtreeCount)
 			if !p.IsRoot() {
 				p.SendToParent(&PropagateReply{})
 			}
-			if p.received == p.subtree {
+			if p.received == p.subtreeCount {
 				process = false
 			}
 		case <-time.After(timeout):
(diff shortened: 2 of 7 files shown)


protocols/manage/propagate.go
```diff
@@ -173,7 +179,7 @@ func (p *Propagate) RegisterOnDone(fn func(int)) {
 // RegisterOnData takes a function that will be called once all connections
 // are set up. The argument to the function is the number of children that
 // sent OK after the propagation
```
nikkolasg: I know it's not from this PR but I don't get the comment with respect to the function name + implementation. Is not the given function will be called when the protocol receives the data ?

protocols/manage/propagate.go
```diff
@@ -51,22 +47,48 @@ type PropagateReply struct {
...
+
+// NewPropagationFunc registers a new protocol name with the context c and will
+// set f as handler for every new instance of that protocol.
+func NewPropagationFunc(c *sda.Context, name string, f PropagationStore) (PropagationFunc, error) {
+	pid, err := c.ProtocolRegister(name, func(n *sda.TreeNodeInstance) (sda.ProtocolInstance, error) {
+		p := &Propagate{
+			sd:               &PropagateSendData{[]byte{}, 1000},
+			TreeNodeInstance: n,
+			received:         0,
+			subtree:          n.TreeNode().SubtreeCount(),
```
nikkolasg: Is this needed ? Does it go over all over the tree in order to compute that value ?

protocols/manage/propagate.go
```diff
@@ -51,22 +47,48 @@ type PropagateReply struct {
...
+				return nil, err
+			}
+		}
+		return p, nil
+	})
+	log.Lvl3("Registering new propagation for", c.ServerIdentity(),
+		name, pid)
+	return func(el *sda.Roster, msg network.Body, msec int) (int, error) {
+		tree := el.GenerateNaryTreeWithRoot(8, c.ServerIdentity())
+		log.Lvl3("Starting to propagate", reflect.TypeOf(msg))
```
nikkolasg: can you add the name of "who" is starting to propagate this value ?

sda/context.go
```diff
@@ -57,6 +57,12 @@ func (c *Context) CreateProtocolSDA(name string, t *Tree) (ProtocolInstance, err
 	return pi, err
 }
 
+// ProtocolRegister will sign up a new protocol to this Conode.
+// It returns the ID of the protocol.
+func (c *Context) ProtocolRegister(name string, protocol NewProtocol) (ProtocolID, error) {
```
nikkolasg: I think a "one sentence" comment on why do we provide this method, since we already have 2 ways in total to register a protocol, would be nice. I think it'll be mostly useful for us, because we forget... :)

sda/local.go
```diff
@@ -398,6 +398,11 @@ func (l *LocalTest) NewLocalConode(port int) *Conode {
 
 }
 
+// NewContext creates a context from a ServerIdentity.
+func (l *LocalTest) NewContext(si *network.ServerIdentity, servID ServiceID, manager *serviceManager) *Context {
```
nikkolasg: mhhh I'm not a fan of providing Yet Another public method to `LocalTest`. Would it be not possible to have `func (l *LocalTest) GetContext(c *sda.Conode,servID ServiceID)` ? 
A Context without its service is useless.

services/identity/service.go
```diff
@@ -373,6 +388,17 @@ func newIdentityService(c *sda.Context, path string) sda.Service {
 		skipchain:        skipchain.NewClient(),
 		path:             path,
 	}
+	var err error
+	s.propagateIdentity, err =
+		manage.NewPropagationFunc(c, "IdentityPropagateID", s.propagateIdentityHandler)
+	s.propagateSkipBlock, err =
+		manage.NewPropagationFunc(c, "IdentityPropagateSB", s.propagateSkipBlockHandler)
+	s.propagateConfig, err =
+		manage.NewPropagationFunc(c, "IdentityPropagateConf", s.propagateConfigHandler)
```
nikkolasg: Nice !

protocols/manage/propagate.go
ineiti: yes, it is needed to know when all the data has propagated. Changed the name to `subtreeCount`.

sda/local.go
ineiti: There might be a need to add a `serviceManager` later. For the moment it's only used in a test, to avoid adding a method to `Conode` that returns a context. But I can remove `NewContext` and add a `Context` method to `Conode` if you prefer.

log/lvl.go
```diff
@@ -64,6 +64,7 @@ var stdOut io.Writer
 var stdErr io.Writer
 
 func init() {
+	ParseEnv()
```
Liamsi: Is this new call related to this issue (make ProtocolRegister available to service #656)?

sda/local.go
Liamsi: Hmm, looks like one needs to add a yet another public method anyways. But I agree that `NewContext` isn't the best name, here (sounds more like a 'stand-alone constructor' than a method on `LocalTest`). `Context` (preferred) or `GetContext` are better.

services/identity/service.go
```diff
@@ -373,6 +388,17 @@ func newIdentityService(c *sda.Context, path string) sda.Service {
...
+	if err != nil {
```
Liamsi: Theses errors only get caught if the last call of `manage.NewPropagationFunc` returns an error. Otherwise the `err` variable gets overwritten by subsequent calls above.

protocols/manage/propagate.go
```diff
@@ -51,22 +47,48 @@ type PropagateReply struct {
...
+		name, pid)
+	return func(el *sda.Roster, msg network.Body, msec int) (int, error) {
+		tree := el.GenerateNaryTreeWithRoot(8, c.ServerIdentity())
+		log.Lvl3(el.List[0].Address, "Starting to propagate", reflect.TypeOf(msg))
+		pi, err := c.CreateProtocolSDA(name, tree)
+		if err != nil {
+			return -1, err
+		}
+		return propagateStartAndWait(pi, msg, msec, f)
+	}, err
```
Liamsi: Maybe I'm missing something, but it is difficult for me to see how all these changes in the propagate protocol are related/necessary because of #656. Can you explain it here or preferably in the issue? 
Currently, it just says: 
> Now that each conode has its own set of Protocols, the ProtocolRegister method should also be made available to services.

sda/local.go
ineiti: Naming... It _is_ a constructor of a `Context`, as we're not returning an existing one, but rather we create a new context. The _real_ context that we would need is stored in the service, and it's not stored anywhere else, so we cannot retrieve it.

log/lvl.go
ineiti: removed

protocols/manage/propagate.go
ineiti: Updated comments on the issue and the PR.

sda/local.go
ineiti: Removed the `NewContext` altogether and using an interface in the `NewPropagationFunc` now, which is probably cleaner anyway.

Issue #659, opened 2016-10-27
RandHound Refactoring II


Daeinar:
Add another round trip to the 2-round version of RandHound to include collective signing of the client commitment to prevent equivocation attacks.

jeffallen:
Randhound is no longer part of cothority. :(

PR #660, opened 2016-10-27
RandHound Refactoring II

Daeinar:
See #659.


protocols/randhound/randhound.go
```diff
@@ -844,6 +857,14 @@ func (rh *RandHound) handleR2(r2 WR2) error {
 	}
 
 	if proceed && !rh.SecretReady {
+
+		// XXX: dummy operation that simulates aggregation of the server responses in the 3-round version of RandHound
```
nikkolasg: Since we're onto that path, maybe adding a `time.Sleep(RTT)` should definitely make a better approx.

Daeinar:
Deleting this PR. Will do a fresh restart.

PR #663, opened 2016-11-01
Network - Local using channels

nikkolasg:
After all, if it makes Travis happy, why not using channels... I still prefer my `sync.Cond` solution but Travis is sad, so I'm sad. 
Let's see what Travis thinks of it.

### DIFF
--- a/network/local.go
+++ b/network/local.go
@@ -281,9 +281,11 @@ type connQueue struct {
 	sync.Mutex
 }
 
+const LOCAL_MAX_BUFFER = 200
+
 func newConnQueue() *connQueue {
 	return &connQueue{
-		queue: make(chan []byte, 200),
+		queue: make(chan []byte, LOCAL_MAX_BUFFER),
 	}
 }
 
--- a/network/local.go
+++ b/network/local.go
@@ -309,10 +309,11 @@ func (c *connQueue) push(buff []byte) {
 // for a packet.
 func (c *connQueue) pop() ([]byte, error) {
 	c.Lock()
-	defer c.Unlock()
 	if c.closed {
+		c.Unlock()
 		return nil, ErrClosed
 	}
+	c.Unlock()
 	return <-c.queue, nil
 }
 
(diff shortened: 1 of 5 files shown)


network/local.go
```diff
@@ -277,27 +276,27 @@ func (lc *LocalConn) Type() ConnType {
...
-	queue  [][]byte
+	queue  chan []byte
 	closed bool
+	sync.Mutex
 }
 
 func newConnQueue() *connQueue {
 	return &connQueue{
-		Cond: sync.NewCond(&sync.Mutex{}),
+		queue: make(chan []byte, 200),
```
ineiti: You would want to have that either as a constant or as a parameter.

network/local.go
```diff
@@ -277,27 +276,27 @@ func (lc *LocalConn) Type() ConnType {
...
 	}
 }
 
 // push inserts a packet in the queue.
 // push won't work if the connQueue is already closed and silently return.
 func (c *connQueue) push(buff []byte) {
-	c.L.Lock()
-	defer c.L.Unlock()
+	c.Lock()
```
ineiti: I really think embedding mutexes is cool.

ineiti:
I remember having thrown away even more code with regard to signalling that a new message is available. Did I dream that one up?

nikkolasg:
Damn, and now a TLS error:
`fatal: unable to access 'https://gopkg.in/codegangsta/cli.v1/': gnutls_handshake() failed: A TLS fatal alert has been received.
`

ineiti:
I think the tls-error is not ours... so a restart should be fine.

ineiti:
Generally - do you want to push this now, or do you want to use this to find the error that still bugs us?

network/local.go
Liamsi: Me too.

nikkolasg:
Just to make things explicit: This PR passes all the time I've tried for the network/ tests now.
I'm just trying to debug some other datarace that occurs from time to time (and I think it's non related to this PR since it occurs at others PR also).

PR #682, opened 2016-11-16
Checking 681

### LINKED ISSUE #681: Niceify check-function
The check-function has some aesthetical shortcomings:
- result is printed on next line
- it stops on first error

ineiti:
Closes #681

### DIFF
--- a/app/lib/server/server.go
+++ b/app/lib/server/server.go
@@ -224,15 +224,16 @@ func CheckServers(g *config.Group) error {
 		// Then check pairs of servers
 		for i, first := range g.Roster.List {
 			for _, second := range g.Roster.List[i+1:] {
+				log.Lvl3("Testing connection between", first, second)
 				desc := []string{"none", "none"}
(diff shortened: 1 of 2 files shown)


app/lib/server/server.go
```diff
@@ -255,7 +256,7 @@ func checkList(list *sda.Roster, descs []string) error {
 	}
 	log.Lvl3("Sending message to: " + serverStr)
 	msg := "verification"
-	log.Info("Checking server(s) ", serverStr, ": ")
+	fmt.Printf("Checking server(s) %s: ", serverStr)
```
nikkolasg: why going back  to fmt ?

nikkolasg: What about having a `log.Infof` ? (that does not add a "\n" unless you explicitely put)

ineiti: All `log.*f`-methods add a "\n", so this would be the exception.

nikkolasg:
Looks good to me. Merging.

PR #685, opened 2016-11-18
logger

### LINKED ISSUE #684: Write log-output to buffer
For some tests we need to check the error-output or other from the log. Some methods already exist to put this into a buffer. This issue asks to make those methods public.

ineiti:
Closes #684

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -60,22 +60,28 @@ func init() {
 var bufStdOut bytes.Buffer
 var bufStdErr bytes.Buffer
 
-func StdToBuf() {
+// OutputToBuf is called for sending all the log.*-outputs to internal buffers
+// that can be used for checking what the logger would've written. This is
+// mostly used for tests.
+func OutputToBuf() {
 	stdOut = &bufStdOut
 	stdErr = &bufStdErr
 }
 
-func StdToOs() {
+// OutputToOs redirects the output of the log.*-outputs again to the os.
+func OutputToOs() {
 	stdOut = os.Stdout
 	stdErr = os.Stderr
 }
 
+// GetStdOut returns all log.*-outputs to StdOut since the last call.
 func GetStdOut() string {
 	ret := bufStdOut.String()
 	bufStdOut.Reset()
 	return ret
 }
 
+// GetStdErr returns all log.*-outputs to StdErr since the last call.
 func GetStdErr() string {
 	ret := bufStdErr.String()
 	bufStdErr.Reset()
--- a/log/log.go
+++ b/log/log.go
@@ -43,7 +43,6 @@
 package log
 
 import (
-	"bufio"
 	"bytes"
 	"io"
 	"os"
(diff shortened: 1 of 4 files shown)


log/lvl_test.go
```diff
@@ -15,16 +15,16 @@ func init() {
 
 func TestTime(t *testing.T) {
 	SetDebugVisible(1)
-	getStdOut()
+	GetStdOut()
```
nikkolasg: ? Why is this line required/useful ?

log/log.go
```diff
@@ -0,0 +1,90 @@
...
+	stdOut = os.Stdout
+	stdErr = os.Stderr
+}
+
+var bufStdOut bytes.Buffer
+var wrStdOut *bufio.Writer
+var bufStdErr bytes.Buffer
+var wrStdErr *bufio.Writer
+
+func StdToBuf() {
```
nikkolasg: Mhhh API wise, if we want the log to go to another place than STDOUT, then it might make more sense to give the opportunity to the user to give the buffer it self ? For example, you might wanna log it to an external file or even to a connection.

At the same time, since the log package has no "instantiation" required (i.e. no `log = NewLog()`), having that in a thread safe manner might be a hassle...

log/log.go
```diff
@@ -0,0 +1,90 @@
...
+	wrStdOut = bufio.NewWriter(&bufStdOut)
+	stdOut = wrStdOut
+	wrStdErr = bufio.NewWriter(&bufStdErr)
+	stdErr = wrStdErr
```
nikkolasg: I don't understand why you need both a `Writer` and a `bytes.Buffer` since the latter implements the former ? I think you don't have to create a wrapper around, although it's friday night and I'm not sure :D

log/log.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func StdToOs() {
+	stdOut = os.Stdout
+	stdErr = os.Stderr
+}
+
+func GetStdOut() string {
```
nikkolasg: * `Get` or no `Get` ... I tend for the latter, you ? (i don't really care, just wanna know if it's a decision or just an unintentional act)  
* StdOut is for *standard* output, but here it can also be the buffer output, so ... `OutputString()` & `ErrorString()` ? I don't know, these are not great names but I'm not a fan of `Stdout` for "output stream" and `stderr` for "error stream".

What do you think ?

log/log.go
```diff
@@ -0,0 +1,90 @@
...
+	wrStdOut.Flush()
+	ret := bufStdOut.String()
+	bufStdOut.Reset()
+	return ret
+}
+
+func GetStdErr() string {
```
nikkolasg: Moreover, a comment should be there, also explaining that it flushes the "queue" so calling twice `GetXXX` won't result in the same output.

log/log.go
ineiti: I'll say that's issue #689 ;)

log/lvl_test.go
ineiti: this tests the output of the logger, so we need to flush it with a `Get`.

log/log.go
ineiti: `stdout` and `stderr` are age-old unix names - so I don't dare touching those ;)

I like the `Get` here. It's shorter than `GetAndFlush`, and for me it's better than just `StdOut`...

nikkolasg: Ok thanks, you made me read the wikipedia on stdout and stderr and I finally got it. My confusion was that I thought we refer to *stdout* only for the *standard* output stream (i.e. the default/normal/common assigned output, by the terminal emulator or whatever). In fact *stdout* refers to **any** of the output stream the process uses.
So sorry, and I agree here then :)

PR #687, opened 2016-11-18
checking for return-value

### LINKED ISSUE #631: Return error in protocol-handler
`TreeNodeInstance.dispatchHandler` doesn't return errors from the handler.

### LINKED ISSUE #686: Check return-value of protocol-handlers
Protocol-handlers should return an error, which is checked in issue #631.

ineiti:
Closes #686 
Closes #631 
- [x] write test-function


nikkolasg:
Please see my comments on the issue https://github.com/dedis/cothority/issues/686#issuecomment-261926046

sda/treenode.go
```diff
@@ -278,20 +284,25 @@ func (n *TreeNodeInstance) dispatchHandler(msgSlice []*ProtocolMsg) error {
...
-			log.Lvl4("Dispatching to", n.ServerIdentity().Address)
+			log.Lvl4("Dispatching", msg, "to", n.ServerIdentity().Address)
 			m := n.reflectCreate(to, msg)
-			f.Call([]reflect.Value{m})
+			errV = f.Call([]reflect.Value{m})[0]
 		}
 	}
+	if !errV.IsNil() {
+		err := errV.Interface().(error)
+		log.Error(n.Name(), "Dispatching", mt, "yields error", err)
```
ineiti: Good point. I also moved the error-output to `dispatchMsgReader`, as this one checks the errors of both channels and handlers - and which also ignored errors ;)

ineiti:
Merged with #688 because too much granularity and because I started to mix them up...

log/log.go
```diff
@@ -64,25 +64,33 @@ var bufStdErr bytes.Buffer
...
 func GetStdOut() string {
+	debugMut.Lock()
+	defer debugMut.Unlock()
 	ret := bufStdOut.String()
 	bufStdOut.Reset()
 	return ret
 }
 
 // GetStdErr returns all log.*-outputs to StdErr since the last call.
 func GetStdErr() string {
```
nikkolasg: Sorry One more final thought: The idea is to be able to detect if an error message or a general message has been logged in case of a specific event right ?
Would it be difficult to provide some kind of helper method in order to do that ? `log.OutputContains(text)`  which returns true if the text is contained in the log (i.e. when *it is buffered*) or false if it's not OR if the log is not directed to the buffer.
What do you think ?
Otherwise I'm OK to merge after this discussion.

ineiti: something like `strings.Contains(GetStdErr(), "My Pony-Error")?

https://golang.org/pkg/strings/#Contains

nikkolasg: This PR is exactly to do what you wrote so instead of having
```
require.True(t, strings.Contains(log.GetStdERrr(),"My Pony Error"))
```
we can have
```
require.True(t, log.Contains("My Pony Error"))
```
Since this is the only use case (now) for this PR, I think it makes sense... PLUS it's a bit more aesthetic!
What's your thoughts ?

ineiti: Something like the last commit? I still think it's overkill...

nikkolasg: Ok ok I won't insist more, you decide if you want to not have this method it's fine. There's one counter argument in favor of your approach is that `ContainsStdOut` do not say that it flushes the buffer also, and that might be an unwanted behavior.
So it's fine if you retract the last commit or not for me then, just tell me and I merge.

PR #688, opened 2016-11-18
Handler error 631

### LINKED ISSUE #631: Return error in protocol-handler

ineiti:
Closes #631 - depends on #687 and #686


ineiti:
Too much granularity - merged to #687

PR #695, opened 2016-11-24
cothorityd has also check-argument now

### LINKED ISSUE #694: app/{cothorityd,cosi} check
`cothorityd` should also get a `check` flag.

Furthermore the `check`-flag misbehaves. On the raspberry-Pi in the lab it doesn't search for the last node in the list, and tests fail for nodes that are available from other machines.

The `setup`-flag of `cothorityd` should write the public address and not the listening address.

ineiti:
closes #694

### DIFF
--- a/app/cothorityd/cothorityd.go
+++ b/app/cothorityd/cothorityd.go
@@ -77,6 +77,18 @@ func main() {
 			},
 			Flags: serverFlags,
 		},
+		{
+			Name:    "check",
+			Aliases: []string{"c"},
+			Usage:   "Check if the servers in the group definition are up and running",
+			Action:  checkConfig,
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  "g",
+					Usage: "Cothority group definition file",
+				},
+			},
+		},
 	}
 	cliApp.Flags = serverFlags
 	// default action
@@ -85,8 +97,8 @@ func main() {
 		return nil
 	}
 
-	cliApp.Run(os.Args)
-
+	err := cliApp.Run(os.Args)
+	log.ErrFatal(err)
 }
 
 func runServer(ctx *cli.Context) {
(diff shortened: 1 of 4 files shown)


app/cothorityd/cothorityd.go
```diff
@@ -96,3 +108,10 @@ func runServer(ctx *cli.Context) {
 
 	server.RunServer(config)
 }
+
+// checkConfig contacts all servers and verifies if it receives a valid
+// signature from each.
+func checkConfig(c *cli.Context) error {
```
nikkolasg: It's not for now but later we should find a way to integrate directly this command in "one-line" that includes the `cli.Command` and the given handler.

ineiti: Yes, I thought of merging it directly into the `lib/server`, but then the "-g"-flag would be hardcoded, so I didn't.

protocols/cosi/cosi.go
```diff
@@ -224,7 +224,7 @@ func (c *CoSi) startChallenge() error {
 	out := &Challenge{
 		Chall: challenge,
 	}
-	log.Lvlf3("%s Starting Chal=%+v (message = %s)", c.Name(), challenge, string(c.Message))
+	log.Lvlf3("%s Starting Chal=%+v (message = %x)", c.Name(), challenge, c.Message)
```
ineiti: IIRC `%x` _is_ hexa.

nikkolasg: Oups sorry right, now I remember that was the change. :) thanks.

ineiti: Oh, you mean the challenge, too? I think the challenge already comes in hexa:

```
3 : (              cosi.(*CoSi).startChallenge: 227) - tcp://192.33.210.8:7770 Starting Chal=0e238a77d1214f0ac161b96b06368febc10001b6dbf0fc68665b77323f6e8ab0 (message = A??y?ҋ?~??v??]?V~fW?X?KΑ?)
```

Issue #287, opened 2016-03-17
Refactor the packet-format


ineiti:
For the moment the `Host/Overlay`-packets are embedded as a binary blob in the `lib/net`-packets, which can make it difficult to redo the same in another language.

Is there a way to send both packets one after the other? Or another solution?

nikkolasg:
This issue should be resuscitated in regard with #680 . 
I think this issue should aim to have a clear definition of each packets used by Cothority and not having something like:
```
type OverlayMessage struct {
 // overlay fields 
 // ...
 // protocol message
 message []byte
}
```
because then message is completely a byte array without any meaning and does not use any of the advantages of protobuf (extensions format, compatibility etc).
Clear to me means:
```
type CoSiPaket struct {
   // Cosi messages
   Announcement
   Commitment
   Challenge
   Response
   // overlay messages / fields
   OverlayPacket
}
type OverlayPacket struct {
   // overlay fields ...
}
```

nikkolasg:
This issue is currently being resolved (PRs will come). Here's a check-list of things to do in order to be able to describe a clean wire protocol format for the RFC (#680) :
- [ ] Change the packets sent through the networks according the scheme described above.
- [ ] Change the uuids used to distinguish the different messages at the network level into something simpler such as `uint16` or `uint64` etc. This would be really neat to have so we don't have to compare 16 bytes just to know the type of the message.
- [ ] Generate the proto files definitions for all packets used by CoSi.

**Optional** (but at the same time **essential** to make sure we did not mess up):
- [ ] Implement a alternative implementation of CoSi in another language

nikkolasg:
# Wire protocol ID related problems

One problem that keeps occuring when designing the library is how to represent a
packet, a certain structure, or simply like a "hash" which is not necessarily known 
at run time. In the cothority code, this problems happens in the network library
and the sda library.

# Network library

## Packets

For the moment, each packet we send through the network is preceded by the size
of the encoded packet and then it's "PacketTypeID" which is an UUID v5
(https://www.ietf.org/rfc/rfc4122.txt).

In details, to generate the id of a specific Golang struct, we generate a UUID
v5 with:
* the namespace being the Namespace URL
  (https://stackoverflow.com/questions/7724903/where-do-uuid-namespaces-come-from)

* the "value" (the value to hash) being:
    `https://dedis.epfl.ch/ || protocolType/ || reflect.TypeOf(packet).String()`

Two potentials problems with this approach:
[...]

PR #696, opened 2016-11-25
[Wire-Protocol 1/3 ] SDA Changes

nikkolasg:
This first PR introduces a way for protocols to have a clean wire format protocol with the least amount of changes. Look for #287 issue.
First this PR introduces the notion `OverlayMessage` which contains everything the `Overlay` needs to know in order to function properly. Protocols that wants to have their full own packet going into the network must have a field of `OverlayMessage` in their packet.

Then there is the notion of `ProtocolIO` (Please suggest other names if you got any) which fulfill two purposes:
 -  `Wrap` a protocol *inner* message into the protocol's bigger message with the overlay information `OverlayMessage`
 - `Unwrap` a protocol bigger message to give back to the `Overlay` the inner protocol message as well as the `OverlayMessage`

Sorry as there are also some --cleaning-up-- changes that add a bit to the PR size.

### DIFF
--- a/sda/protocol.go
+++ b/sda/protocol.go
@@ -102,7 +102,7 @@ func GlobalProtocolRegister(name string, protocol NewProtocol) (ProtocolID, erro
 	return protocols.Register(name, protocol)
 }
 
-// ProtocolIO is an interface that allows one protocol to completely define its
+// MessageProxy is an interface that allows one protocol to completely define its
 // wire protocol format while still using the Overlay.
 // Cothority sends different messages dynamically as slices of bytes, whereas
 // Google proposes to use union-types:
@@ -111,9 +111,9 @@ func GlobalProtocolRegister(name string, protocol NewProtocol) (ProtocolID, erro
 // the dynamic cothority-messages. Implementations must provide methods to
 // pass from the 'union-types' to 'cothority-dynamic-messages' with the Wrap
 // and Unwrap method.
-// A default one is provided with defaultProtocolIO so the regular wire-format
+// A default one is provided with defaultMessageProxy so the regular wire-format
 // protocol can still be used.
-type ProtocolIO interface {
+type MessageProxy interface {
 	// Wrap takes a message and the overlay information and returns the message
 	// that has to be sent directly to the network alongside with any error that
 	// happened.
@@ -126,39 +126,39 @@ type ProtocolIO interface {
 	Unwrap(msg interface{}) (interface{}, *OverlayMessage, error)
 	// PacketType returns the packet type ID that this Protocol expects from the
 	// network. This is needed in order for the Overlay to receive those
-	// messages and dispatch them to the correct ProtocolIO.
+	// messages and dispatch them to the correct MessageProxy.
 	PacketType() network.PacketTypeID
-	// Name returns the name associated with this ProtocolIO. When creating a
-	// protocol, if one use a name used by a ProtocolIO, this ProtocolIO will be
+	// Name returns the name associated with this MessageProxy. When creating a
+	// protocol, if one use a name used by a MessageProxy, this MessageProxy will be
 	// used to Wrap and Unwrap messages.
 	Name() string
 }
 
-// NewProtocolIO is a function typedef to instantiate a new ProtocolIO.
-type NewProtocolIO func() ProtocolIO
+// NewMessageProxy is a function typedef to instantiate a new MessageProxy.
+type NewMessageProxy func() MessageProxy
 
-type protocolIOFactoryStruct struct {
-	factories []NewProtocolIO
+type messageProxyFactoryStruct struct {
+	factories []NewMessageProxy
 }
 
-var protocolIOFactory = protocolIOFactoryStruct{}
+var messageProxyFactory = messageProxyFactoryStruct{}
 
-// RegisterProtocolIO saves a new NewProtocolIO under its name.
-// When a Conode is instantiated, all ProtocolIOs will be generated and stored
+// RegisterMessageProxy saves a new NewMessageProxy under its name.
+// When a Conode is instantiated, all MessageProxys will be generated and stored
 // for this Conode.
-func RegisterProtocolIO(n NewProtocolIO) {
-	protocolIOFactory.factories = append(protocolIOFactory.factories, n)
+func RegisterMessageProxy(n NewMessageProxy) {
+	messageProxyFactory.factories = append(messageProxyFactory.factories, n)
 }
 
-// protocolIOStore contains all created ProtocolIOs. It contains the default
-// ProtocolIO used by the Overlay for backwards-compatibility.
-type protocolIOStore struct {
+// messageProxyStore contains all created MessageProxys. It contains the default
+// MessageProxy used by the Overlay for backwards-compatibility.
+type messageProxyStore struct {
 	sync.Mutex
-	protos    []ProtocolIO
-	defaultIO ProtocolIO
+	protos    []MessageProxy
+	defaultIO MessageProxy
 }
 
-func (p *protocolIOStore) getByName(name string) ProtocolIO {
+func (p *messageProxyStore) getByName(name string) MessageProxy {
 	p.Lock()
 	defer p.Unlock()
 	for _, pio := range p.protos {
@@ -169,7 +169,7 @@ func (p *protocolIOStore) getByName(name string) ProtocolIO {
 	return p.defaultIO
 }
 
-func (p *protocolIOStore) getByPacketType(t network.PacketTypeID) ProtocolIO {
+func (p *messageProxyStore) getByPacketType(t network.PacketTypeID) MessageProxy {
 	p.Lock()
 	defer p.Unlock()
 	for _, pio := range p.protos {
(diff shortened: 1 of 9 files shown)


ineiti:
`ProtocolIO`-proposition-brainstorming:

- `WireProtocol` - this is why we have it
- `WireFormatHandler` - without the `T`...
- `MessageHandler`
- `MessageMorph`

more propositions on Monday...

ineiti:
I'd like to use `branch_name_issue#` as branch-name. Do you think this is a good idea? If there are multiple issues, the biggest one should be used. Here it would be `refactor_packet_sda_287`.

nikkolasg:
Yep for the branch name I agree, I always forgot to put the issue number, you're right. should I push it on a new branch name and redo the PRs ? Did you already made some comments or it does not matter if we loose them ?

nikkolasg:
Something with `Morph` in the name sounds fun .. :D

sda/protocol.go
```diff
@@ -100,3 +101,99 @@ func ProtocolNameToID(name string) ProtocolID {
...
+// protocolIOStore contains all created ProtocolIO and is generally used by the
+// Overlay. It contains the default ProtocolIO used by the Overlay in order to
+// still function properly in case the old wire-format protocol is used.
+type protocolIOStore struct {
+	sync.Mutex
+	protos []ProtocolIO
+	names  map[string]int
+	types  map[network.PacketTypeID]int
+	// the one that gets used in case no ProtocolIO is defined
+	defaultIO ProtocolIO
```
ineiti: Why is this not a `ProtocolIO` stored under `default`? With this solution we have two places where `ProtocolIO`s are stored.

sda/protocol.go
```diff
@@ -100,3 +101,99 @@ func ProtocolNameToID(name string) ProtocolID {
...
+
+func (p *protocolIOStore) getByName(name string) ProtocolIO {
+	p.Lock()
+	defer p.Unlock()
+	idx, ok := p.names[name]
+	if !ok || idx >= len(p.protos) || p.protos[idx] == nil {
+		return p.defaultIO
+	}
+	return p.protos[idx]
+}
```
ineiti: If we don't use the `names` and `types` map but add a `Name()`-method to the `ProtocolIO`, this would be:
```
for p in range p.protos{
    if p.Name() == name{
        return p
    }
}
// If the default-protocol is the first one
return p.protos[0]
// else
return p.defaultIO
```

sda/protocol.go
```diff
@@ -100,3 +101,99 @@ func ProtocolNameToID(name string) ProtocolID {
...
+
+func (p *protocolIOStore) getByPacketType(t network.PacketTypeID) ProtocolIO {
+	p.Lock()
+	defer p.Unlock()
+	idx, ok := p.types[t]
+	if !ok || idx >= len(p.protos) || p.protos[idx] == nil {
+		return p.defaultIO
+	}
+	return p.protos[idx]
+}
```
ineiti: Same thing here:
```
for p in range p.protos{
    if p.PacketType() == t{
        return p
    }
}
// If the default-protocol is the first one
return p.protos[0]
// else
return p.defaultIO
```

sda/protocol.go
```diff
@@ -100,3 +101,99 @@ func ProtocolNameToID(name string) ProtocolID {
...
+		io := newIO()
+		pstore.protos = append(pstore.protos, io)
+		pstore.names[name] = len(pstore.protos) - 1
+		pstore.types[io.PacketType()] = len(pstore.protos) - 1
+		disp.RegisterProcessor(proc, io.PacketType())
+		log.Lvl2("Instantiating ProtocolIO", name, "at position", len(pstore.protos))
+	}
+	// also add the default one
+	return pstore
+}
```
ineiti: And this would also get more easy to write if we don't use the `names` and `types` map.

sda/protocol.go
```diff
@@ -100,3 +101,96 @@ func ProtocolNameToID(name string) ProtocolID {
...
+func RegisterProtocolIO(n NewProtocolIO) {
+	protocolIOFactory.factories = append(protocolIOFactory.factories, n)
+}
+
+// protocolIOStore contains all created ProtocolIOs. It contains the default
+// ProtocolIO used by the Overlay for backwards-compatibility.
+type protocolIOStore struct {
+	sync.Mutex
+	protos    []ProtocolIO
+	defaultIO ProtocolIO
```
nikkolasg: Here's the slice solution that we discussed then. But I think keeping the defaultIO explicit rather than using a "trick" of using the "0" position in a slice is still better. 
```
for _,p := range protos {
....
}
return defaultIO
```
vs 
```
for _, p := range protos[1:] {
...
}

return protos[0]
```

sda/protocol.go
```diff
@@ -100,3 +101,96 @@ func ProtocolNameToID(name string) ProtocolID {
...
+	// Unwrap takes the message coming from the network and returns the
+	// inner message that is going to be dispatched to the ProtocolInstance, the
+	// OverlayMessage needed by the Overlay to function correctly and then any
+	// error that might have occurred.
+	Unwrap(msg interface{}) (interface{}, *OverlayMessage, error)
+	// PacketType returns the packet type ID that this Protocol expects from the
+	// network. This is needed in order for the Overlay to receive those
+	// messages and dispatch them to the correct ProtocolIO.
+	PacketType() network.PacketTypeID
+
```
ineiti: Too much space (or not enough in front of the other functions).

sda/protocol.go
ineiti: OK, no prob. It already looks much cleaner without the `map`s.

sda/protocol_test.go
```diff
@@ -139,3 +140,133 @@ func TestProtocolAutomaticInstantiation(t *testing.T) {
...
+	*TreeNode
+	SimpleMessage
+}
+
+func (t TestProtocolInstance) handleSimpleMessage(h SimpleMessageHandler) error {
+	var ok = true
+	if h.SimpleMessage.I != 12 {
+		ok = false
+	}
+	chanTestProtoInstance <- ok
```
ineiti: `chanTestProtoInstance <- h.SimpleMessage.I != 12`

sda/protocol_test.go
```diff
@@ -139,3 +140,133 @@ func TestProtocolAutomaticInstantiation(t *testing.T) {
...
+var chanTestProtoInstance = make(chan bool)
+
+// ProtocolInstance part
+type TestProtocolInstance struct {
+	*TreeNodeInstance
+}
+
+func newTestProtocolInstance(n *TreeNodeInstance) (ProtocolInstance, error) {
+	pi := new(TestProtocolInstance)
+	pi.TreeNodeInstance = n
```
ineiti: `pi := &TestProtocolInstance{n}`

sda/protocol_test.go
```diff
@@ -139,3 +140,133 @@ func TestProtocolAutomaticInstantiation(t *testing.T) {
...
+	defer local.CloseAll()
+
+	RegisterProtocolIO(NewTestProtocolIO)
+	GlobalProtocolRegister(testProtoIOName, newTestProtocolInstance)
+	h, _, tree := local.GenTree(2, true)
+
+	go func() {
+		// first time to wrap
+		res := <-chanProtoIOFeedback
+		assert.Equal(t, "", res)
```
ineiti: `assert` vs. `require`: does it make sense to go on if the first assert fails?

sda/protocol_test.go
```diff
@@ -139,3 +140,133 @@ func TestProtocolAutomaticInstantiation(t *testing.T) {
...
 	<-chanH2
 }
+
+func TestProtocolIOFactory(t *testing.T) {
+	defer eraseAllProtocolIO()
+	RegisterProtocolIO(NewTestProtocolIOChan)
+	assert.True(t, len(protocolIOFactory.factories) == 1)
+	// register twice
+	RegisterProtocolIO(NewTestProtocolIOChan)
+	// TODO check log error output
```
ineiti: Will this come?

sda/protocol_test.go
```diff
@@ -139,3 +140,133 @@ func TestProtocolAutomaticInstantiation(t *testing.T) {
 	// entity list from h1
 	<-chanH2
 }
+
+func TestProtocolIOFactory(t *testing.T) {
+	defer eraseAllProtocolIO()
+	RegisterProtocolIO(NewTestProtocolIOChan)
+	assert.True(t, len(protocolIOFactory.factories) == 1)
+	// register twice
```
ineiti: Where is the second registration?

ineiti:
No need to open a new branch, this one's fine.
- [ ] Small fixes
- [ ] decide if `type NewProtocolIO func() *ProtocolIOInst` makes sense
- [ ] decide on name for `ProtocolIO`
- [ ] make Travis happy

ineiti:
I still like the `WireFormatHandler`...

sda/protocol_test.go
nikkolasg: There's none actually. ^^

sda/protocol_test.go
nikkolasg: true. the wait on the channel below won't ever get called otherwise.

nikkolasg:
> type NewProtocolIO func() *ProtocolIOInst

Here defining another struct with the name and packet id seems useless from the user point of view, and only useful "for us" because we now use arrays internally...

nikkolasg:
> decide on name for ProtocolIO

I just thought of something: what about `Decorator` ? like `MessageDecorator` ? 
If not, maybe even go back to `MessageWrapper`. 
See https://en.wikipedia.org/wiki/Decorator_pattern

PR #697, opened 2016-11-25
[Wire-Protocol 2/3 ] CoSi Changes

nikkolasg:
This second PR implements the `ProtocolIO` for the CoSi protocol; namely it creates **the packet**  `ProtocolPacket` for CoSi which contains everything:
- `Announcement`
- `Commitment`
- `Challenge`
- `Response`
- `OverlayMessage`

See #287

### DIFF
--- a/protocols/cosi/packets.go
+++ b/protocols/cosi/packets.go
@@ -9,7 +9,7 @@ import (
 )
 
 func init() {
-	sda.RegisterProtocolIO(Name, func() sda.ProtocolIO {
+	sda.RegisterProtocolIO(func() sda.ProtocolIO {
 		return new(ProtocolIO)
 	})
 }
@@ -98,6 +98,10 @@ func (p *ProtocolIO) Unwrap(msg interface{}) (interface{}, *sda.OverlayMessage, 
 // the ProtocolPacket.
 func (p *ProtocolIO) PacketType() network.PacketTypeID {
 	return ProtocolPacketID
+}
+
+func (p *ProtocolIO) Name() string {
+	return Name
 }
 
 // Announcement is broadcasted message initiated and signed by proposer.
--- a/protocols/cosi/packets.go
+++ b/protocols/cosi/packets.go
@@ -33,7 +33,7 @@ var ProtocolPacketID = network.RegisterPacketType(ProtocolPacket{})
 type ProtocolPacket struct {
 	Phase uint32
 
-	Info *sda.OverlayMessage
+	OverlayMessage *sda.OverlayMessage
 
 	Ann  *Announcement
 	Comm *Commitment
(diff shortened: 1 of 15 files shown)


protocols/cosi/packets.go
```diff
@@ -1,14 +1,104 @@
...
+
+// ProtocolPacketID is the network.PacketTypeID of the CoSi ProtocolPacket
+var ProtocolPacketID = network.RegisterPacketType(ProtocolPacket{})
+
+// ProtocolPacket is the main message for the CoSi protocol which includes
+// every information that the CoSi protocol might need.
+type ProtocolPacket struct {
+	Phase uint32
+
+	Info *sda.OverlayMessage
```
ineiti: `Overlay` instead of `Info`?

protocols/cosi/packets.go
```diff
@@ -1,14 +1,104 @@
...
+}
+
+// ProtocolIO implements the sda.ProtocolIO interface for the CoSi protocol.
+type ProtocolIO struct{}
 
-// The main messages used by CoSi
+// Wrap implements the sda.ProtocolIO interface by wrapping up any of the
+// four-step messages into a ProtooclPacket.
+func (p *ProtocolIO) Wrap(msg interface{}, info *sda.OverlayMessage) (interface{}, error) {
+	var packet = new(ProtocolPacket)
```
ineiti: `overlay` instead of `info`?

services/cosi/cosi.go
```diff
@@ -48,13 +48,13 @@ type SignatureResponse struct {
 // SignatureRequest treats external request to this service.
 func (cs *CoSi) SignatureRequest(si *network.ServerIdentity, req *SignatureRequest) (network.Body, error) {
 	tree := req.Roster.GenerateBinaryTree()
-	tni := cs.NewTreeNodeInstance(tree, tree.Root, protocol.Name)
-	pi, err := protocol.NewCoSi(tni)
+	tni := cs.NewTreeNodeInstance(tree, tree.Root, cosi.Name)
+	pi, err := cosi.NewCoSi(tni)
```
ineiti: Perhaps now that it's called `cosi` we should rename `NewCosi`, so it's more like `cosi.NewProtocol`?

nikkolasg: Yep agreed.

protocols/cosi/packets.go
nikkolasg: Mhhh, I'd like the name to represent that it is only a packet, a name, a set of information received, and not a structure which can *do* things such as `Overlay`, so `OverlayMessage` maybe ?

ineiti: +1 for `OverlayMessage`

PR #700, opened 2016-11-29
Test ism 699

### LINKED ISSUE #699: Remove `InterServiceMessages` and add an example in `service_test.go`
After discussing with @nikkolasg we decided to remove `InterServiceMessages` altogether and have services registering directly with the service-manager for new messages.

ineiti:
closes #699 

We never implemented the `InterServiceMessage`. This PR makes this possible.

### DIFF
--- a/sda/processor.go
+++ b/sda/processor.go
@@ -118,6 +118,19 @@ func (p *ServiceProcessor) ProcessClientRequest(si *network.ServerIdentity,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


nikkolasg:
Do we still need this feature ? I think it's already present but not under a new name, but under `network.Processor`.
I mean a `Service` is also a `network.Processor` and a `Context` is a `network.Dispatcher` so normally, a `Service` can register the message it expects to the `Context` and it wil receive those...

ineiti:
removing interservicemessages.

PR #703, opened 2016-11-29
comments

ineiti:
Comments...

### DIFF
--- a/protocols/cosi/packets.go
+++ b/protocols/cosi/packets.go
@@ -44,8 +44,7 @@ type ProtocolPacket struct {
 // ProtocolIO implements the sda.ProtocolIO interface for the CoSi protocol.
 type ProtocolIO struct{}
 
-// Wrap implements the sda.ProtocolIO interface by wrapping up any of the
-// four-step messages into a ProtooclPacket.
+// Wrap takes a dynamic OverlayMessage and returns a ProtocolPacket.
 func (p *ProtocolIO) Wrap(msg interface{}, info *sda.OverlayMessage) (interface{}, error) {
 	var packet = new(ProtocolPacket)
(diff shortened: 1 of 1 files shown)


protocols/cosi/packets.go
```diff
@@ -44,8 +44,7 @@ type ProtocolPacket struct {
 // ProtocolIO implements the sda.ProtocolIO interface for the CoSi protocol.
 type ProtocolIO struct{}
 
-// Wrap implements the sda.ProtocolIO interface by wrapping up any of the
-// four-step messages into a ProtooclPacket.
+// Wrap takes a dynamic OverlayMessage and returns a ProtocolPacket.
```
nikkolasg: what is a "dynamic" message ? I would not put dynamic unless we have a clear definition of what it is exactly. Same for the other occurences.

protocols/cosi/packets.go
```diff
@@ -94,33 +92,31 @@ func (p *ProtocolIO) Unwrap(msg interface{}) (interface{}, *sda.OverlayMessage,
...
 }
 
-// Response with which every node replies with.
+// Response of all nodes, aggregated over all children.
 type Response struct {
 	Resp abstract.Scalar
 }
 
-//Theses are pairs of TreeNode + the actual message we want to listen on.
+// Overlay-structures to retrieve the sending TreeNode.
```
nikkolasg: Singular

protocols/cosi/packets.go
ineiti: It's in the ProtocolIO description.

protocols/cosi/packets.go
ineiti: there are four structures that follow.

PR #705, opened 2016-11-29
and there was colors

### LINKED ISSUE #704: Logger doesn't parse environment-variables anymore
`log.ParseEnv()` should be called in `log.init()`, so that those who like nice colors can use their variables...

ineiti:
closes #704


nikkolasg:
That's a real nice PR.

PR #712, opened 2016-12-01
adding description and removing unused method

### LINKED ISSUE #711: Add Description to ServerIdentity
For stats and nicer output it is useful to have a `Description` in `ServerIdentity`.

ineiti:
closes #711

### DIFF
--- a/app/lib/config/config.go
+++ b/app/lib/config/config.go
@@ -17,7 +17,6 @@ import (
 	"github.com/dedis/cothority/network"
 	"github.com/dedis/cothority/sda"
 	"github.com/dedis/crypto/abstract"
-	"github.com/dedis/crypto/config"
 )
 
 var in *bufio.Reader
@@ -30,9 +29,10 @@ func init() {
 
 // CothoritydConfig is the configuration structure of the cothority daemon.
 type CothoritydConfig struct {
-	Public  string
-	Private string
-	Address network.Address
+	Public      string
+	Private     string
+	Address     network.Address
+	Description string
 }
 
 // Save will save this CothoritydConfig to the given file name. It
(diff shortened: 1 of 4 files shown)


nikkolasg:
I'm not sure what's the best thing to do here:
1) On one hand, it's true that having the description in output such as `./cosi sign` would be nice. Also when the days come we have our status web page, this would be very relevant.

2) On the other hand, I'm very not sure it's a good idea to put that in our `ServerIdentity`. In my taste, `ServerIdentity` is sent way too much in our messages (in each Tree, and Roster sent actually) and sending a description along with it seems a waste of space/bandwidth, and also useless. We don't need the description in a Tree for example.

The way I see it currently, in "my ideal world", we would have two separate version of a `Serveridentity`, one which is human readable and contains whatever information needed (such as "Maintainer name" etc) and one that only contains the address and the public key (and/or certificate) which is meant only to pass in protocol/service communications.

So all of that just to say I'm not sure we should rush to make this one happen before thinking what exactly do we want...
What do you think ;) ?

ineiti:
So having a structure of `ServerDescription` with
- comment
- maintainer/contact
- PGP-key
- whatever

An a similar, but a bit unrelated note: could we do something to directly merge this with a list of services this conode accepts? Not everybody will want to run a full 'reprobuild'-service or perhaps later a 'ddos'-service... Could we make a common structure? Or would this be a different?

nikkolasg:
> could we do something to directly merge this with a list of services this conode accepts?

Yep that's a good idea too I think and it would make sense to have a list of string first to describe the name of the service they're using.

 In a more general view, I think we're aiming for something like a complete "CoNode Profile". We could imagine that theses profiles are stored into a database or DHT and we would refer to one with an "Profile ID" which we could then embed into a `ServerIdentity`.
Anyway I'm going to far ... :D

nikkolasg:
But this issue specifically, what do you think we should do ? 
Personnally I think we should not going too fast with that and put that on hold for the moment until we know where we're going with the status page thing etc, but if you feel it's okay and want to have your work merged, I understand also. What do you think ?

ineiti:
If you're OK, I'd put it in as-is and not create a supplementary structure for description and other stuff.

nikkolasg:
Ok then we'll get down to this issue when we integrate TLS and different Keys/protocols etc.

PR #715, opened 2016-12-01
Websocket 3

### LINKED ISSUE #477: Remove first argument in serviceMessages
All `ServiceMessage`s take a `ServerIdentity` as a first argument. But as the clients are anonymous, this `ServerIdentity` has no use and can be removed.

ineiti:
closes #668 
closes #477


sda/service.go
```diff
@@ -10,9 +10,13 @@ import (
 
 	"sync"
 
+	"reflect"
+
 	"github.com/dedis/cothority/log"
 	"github.com/dedis/cothority/network"
+	"github.com/dedis/protobuf"
 	"github.com/satori/go.uuid"
+	"golang.org/x/net/websocket"
```
nikkolasg: We should try to aim for the Gorilla version which has much more features and is more maintained also: 
http://www.gorillatoolkit.org/pkg/websocket
https://github.com/gorilla/websocket

(see even the comments from the Golang dev. about the gorilla library [here](https://godoc.org/golang.org/x/net/websocket) )

sda/service.go
```diff
@@ -318,84 +303,69 @@ func (s *serviceManager) serviceByID(id ServiceID) (Service, bool) {
...
-	serviceReq := &ClientRequest{
-		Service: c.ServiceID,
-		Data:    b,
+// SendReflect wraps protobuf.(En|De)code to the Client.Send-function.
+func (c *Client) SendProtobuf(dst *network.ServerIdentity, msg interface{}, ret interface{}) error {
+	buf, err := protobuf.Encode(msg)
+	if err != nil {
+		return err
+	}
+	path := strings.Split(reflect.TypeOf(msg).String(), ".")[1]
```
nikkolasg: Would be better to put that as a one-line function to make sure we're doing the same everytime. (it's already present twice I believe)

sda/service.go
```diff
@@ -318,84 +303,69 @@ func (s *serviceManager) serviceByID(id ServiceID) (Service, bool) {
...
+	}
+	log.Lvlf4("Received %x", rcv)
+	return rcv, nil
+}
 
-	serviceReq := &ClientRequest{
-		Service: c.ServiceID,
-		Data:    b,
+// SendReflect wraps protobuf.(En|De)code to the Client.Send-function.
+func (c *Client) SendProtobuf(dst *network.ServerIdentity, msg interface{}, ret interface{}) error {
```
nikkolasg: Comments saying what's the input / output + what id does would be nice.

sda/websocket.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/dedis/cothority/log"
+	"github.com/dedis/cothority/network"
+	"golang.org/x/net/websocket"
+	"gopkg.in/tylerb/graceful.v1"
+)
+
+const readSize = 1024
+
+// WebSocket handles incoming client-requests using the WebService
```
nikkolasg: so webservice does not exists after all! :D

sda/websocket.go
```diff
@@ -0,0 +1,97 @@
...
+// protocol for JavaScript-compatibility.
+type WebSocket struct {
+	ServerIdentity *network.ServerIdentity
```
nikkolasg: Why exported?

sda/websocket.go
```diff
@@ -0,0 +1,97 @@
...
+	w.mux = http.NewServeMux()
+	w.server = &graceful.Server{
+		Timeout: 100 * time.Millisecond,
+		Server: &http.Server{
+			Addr:    webHost,
+			Handler: w.mux,
+		},
+	}
+	log.Lvl1("Starting to listen on", webHost)
+	go w.server.ListenAndServe()
```
nikkolasg: I would put the separate function that starts the websocket server  in `conode.Start()`. 
Like you, I have things that I do like to see, and being able to add `handlers` while the server is already running is one of those things. Imagine a client try to contact a conode exactly just before the URL is registered etc etc. It just seems simpler to register everything beforehand, even print those paths, and then start it.

Although the documentation does not say whether it's safe to add handlers at "run" time, all examples one could find on internet simply register everything before calling `http.ListenAndServe`. I'd advise to use the best practices as always.

sda/websocket.go
```diff
@@ -0,0 +1,97 @@
...
+	return nil
+}
+
+// Register a message-handler for a service.
+func (w *WebSocket) RegisterMessageHandler(service, path string) error {
+	log.Lvlf3("Registering websocket for ws://hostname/%s/%s", service, path)
+	h := func(ws *websocket.Conn) {
+		var buf []byte
+		for {
+			tmp := make([]byte, readSize)
```
nikkolasg: It seems there's already a way to do these things in the library: https://godoc.org/golang.org/x/net/websocket#Codec
One just needs to define the marshaling and unmarshaling function.
It would avoid us to define custom `readSize` and it autmatically fills up the buffer if it did not read enough.

sda/websocket.go
```diff
@@ -0,0 +1,97 @@
...
+type WebSocket struct {
+	ServerIdentity *network.ServerIdentity
+	services       map[string]Service
+	server         *graceful.Server
+	mux            *http.ServeMux
+}
+
+// NewWebSocket opens a webservice-listener one port above the given
+// ServerIdentity.
+func NewWebSocket(si *network.ServerIdentity) *WebSocket {
```
nikkolasg: A comment saying it will listen on port+1 !

sda/websocket.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		reply, errCode := w.services[service].ProcessClientRequest(path, buf)
+		if errCode == 0 {
+			_, err := ws.Write(reply)
+			if err != nil {
+				log.Error(err)
+				return
+			}
+		}
+		ws.WriteClose(errCode)
```
nikkolasg: Do we want that ? 
As a general rule, I've always practiced the "if you open it, you close it". Same story, in mobile environment, it's always better to keep things going I think. Why not returning the error code ?

sda/service.go
ineiti: only once now.

sda/processor.go
```diff
@@ -98,56 +108,47 @@ func (p *ServiceProcessor) Process(packet *network.Packet) {
...
-		network.DefaultConstructors(network.Suite))
-	if err != nil {
-		log.Error("Err unmarshal client request:" + err.Error())
-		return
-	}
-	reply := p.GetReply(si, mt, m)
-	if err := p.SendRaw(si, reply); err != nil {
-		log.Error(err)
-	}
+func (i *ServiceProcessor) NewProtocol(tn *TreeNodeInstance, conf *GenericConfig) (ProtocolInstance, error) {
```
ineiti: To avoid
`// NewProtocol creates a protocol for stat, as you can see it is simultanously absolutely useless and regrettably necessary.`

sda/service.go
```diff
@@ -318,84 +303,69 @@ func (s *serviceManager) serviceByID(id ServiceID) (Service, bool) {
...
-	}
-	return &ClientRequest{
-		Service: sid,
-		Data:    buff,
-	}, nil
-}
-
 // Client is a struct used to communicate with a remote Service running on a
 // sda.Conode
 type Client struct {
```
ineiti: I didn't want to have a map of open connections, so I keep only one open at any time. If the client connects to different conodes, the old connection will be closed and a new opened.

sda/websocket.go
ineiti: Deleted as it's not used after the initialisation.

ineiti:
With regard to the return error codes and messages: I think it is good to have both codes and msgs. Codes are easier to test against in a program, and messages can be displayed to the user. So I prefer to keep both.

PR #718, opened 2016-12-05
Fix wireprotocol for simulation

nikkolasg:
The new wire protocol design is not compatible for simulations because one simulation uses different "network" packages, one for each process, so there are packets which are non registered so it's not possible to Unmarshal them.

This PR fixes that by simply registering the packets a protocol use at init time.

### DIFF
--- a/network/encoding.go
+++ b/network/encoding.go
@@ -41,7 +41,7 @@ func (pId PacketTypeID) String() string {
 	if ok {
 		return t.String()
 	}
-	return fmt.Sprintf("%x", uuid.UUID(pId))
+	return uuid.UUID(pId).String()
 }
 
 // Equal returns true if pId is equal to t
@@ -77,7 +77,6 @@ func RegisterPacketUUID(mt PacketTypeID, rt reflect.Type) PacketTypeID {
 		return mt
 	}
 	registry.put(mt, rt)
-
 	return mt
 }
 
@@ -186,7 +185,7 @@ var marshalLock sync.Mutex
 
 // MarshalRegisteredType will marshal a struct with its respective type into a
 // slice of bytes. That slice of bytes can be then decoded in
-// UnmarshalRegisteredType.
+// UnmarshalRegisteredType. data MUST BE a pointer to the message.
 func MarshalRegisteredType(data Body) ([]byte, error) {
 	marshalLock.Lock()
 	defer marshalLock.Unlock()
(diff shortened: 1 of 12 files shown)


network/encoding.go
```diff
@@ -186,7 +185,7 @@ var marshalLock sync.Mutex
 
 // MarshalRegisteredType will marshal a struct with its respective type into a
 // slice of bytes. That slice of bytes can be then decoded in
-// UnmarshalRegisteredType.
+// UnmarshalRegisteredType. data MUST BE a pointer to the message.
```
ineiti: Please don't capitalize comments.

protocols/bftcosi/packets.go
```diff
@@ -4,10 +4,24 @@ import (
...
+func init() {
+	for _, i := range []interface{}{
+		BFTSignature{},
+		Announce{},
+		Commitment{},
+		ChallengePrepare{},
+		ChallengeCommit{},
+		Response{},
+	} {
+		network.RegisterPacketType(i)
```
ineiti: Should we add a `network.RegisterPacketTypes`? Like there are other `Register*s`-functions?

sda/overlay.go
```diff
@@ -730,7 +730,7 @@ func (d *defaultProtoIO) Unwrap(msg interface{}) (interface{}, *OverlayMessage,
 	case ProtocolMsg:
 		sdaMsg := inner
 		var err error
-		_, msg, err := network.UnmarshalRegistered(sdaMsg.MsgSlice)
+		_, unmar, err := network.UnmarshalRegistered(sdaMsg.MsgSlice)
```
ineiti: I prefer `msg` over `unmar`...

nikkolasg: I've named it `protoMsg` then, is it ok? I don't like shadowing a variable and `msg` is already the name of an argument.

protocols/bftcosi/packets.go
nikkolasg: Yes I agree it'd useful. But on the other hand we already have many ways to register a packet (with pointers, with uuids, and decode returns pointer or values etc) and I'd prefer to make an issue about `Cleaning network.RegisterXXX methods` and adding this new method inside the issue. 
Is it ok with you ?

ineiti: OK, open the issue and I'll close the comments.

nikkolasg: here it is: #722

PR #720, opened 2016-12-07
Fix network connections - max file descriptors

nikkolasg:
This PR is intended to solve the exhaustion of file descriptor of connections.

### DIFF
--- a/network/router.go
+++ b/network/router.go
@@ -218,11 +218,14 @@ func (r *Router) handleConn(remote *ServerIdentity, c Conn) {
 		}
 
 		if err != nil {
-			log.Lvlf4("%+v got error (%+s) while receiving message", r.ServerIdentity.String(), err)
+			if err == ErrTimeout {
+				log.Lvlf5("%s drops %s connection: timeout", r.ServerIdentity.Address, remote.Address)
+				return
+			}
 
-			if err == ErrClosed || err == ErrEOF || err == ErrTimeout {
+			if err == ErrClosed || err == ErrEOF {
 				// Connection got closed.
-				log.Lvl3(r.address, "handleConn with closed connection: stop (dst=", remote.Address, ")")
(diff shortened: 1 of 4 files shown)


network/router.go
```diff
@@ -197,7 +220,7 @@ func (r *Router) handleConn(remote *ServerIdentity, c Conn) {
 		if err != nil {
 			log.Lvlf4("%+v got error (%+s) while receiving message", r.ServerIdentity.String(), err)
 
-			if err == ErrClosed || err == ErrEOF {
+			if err == ErrClosed || err == ErrEOF || err == ErrTimeout {
```
ineiti: Now I have a lot of `Error: timeout` in my logs. Can we put that in a separate `if`-clause and not output an error for the `ErrTimeout`?

network/tcp.go
```diff
@@ -14,6 +14,10 @@ import (
 	"github.com/dedis/cothority/log"
 )
 
+// a connection will return an io.EOF after readTimeout if nothing have been
```
ineiti: s/have/has/

network/tcp.go
```diff
@@ -100,11 +104,14 @@ func (c *TCPConn) Receive() (nm Packet, e error) {
 func (c *TCPConn) receiveRaw() ([]byte, error) {
 	c.receiveMutex.Lock()
 	defer c.receiveMutex.Unlock()
+	c.conn.SetReadDeadline(time.Now().Add(readTimeout))
```
ineiti: Isn't this superfluous as we have line 114?

nikkolasg: It's a bit *deceiving* because it's **not** a timeout. Each time we read, we have to set up the new deadline in *absolute* time. In fact I'm doing it wrong, I'm supposed to put that deadline just above the line where we read in the for loop. One commit coming along.

PR #726, opened 2016-12-08
Websocket sda - 1/3

### LINKED ISSUE #477: Remove first argument in serviceMessages

ineiti:
First part of #668 and #477
Please merge first 3/3

closes #668
closes #477


log/lvl.go
```diff
@@ -16,6 +16,19 @@ import (
 	"github.com/daviddengcn/go-colortext"
 )
 
+<<<<<<< HEAD
```
nikkolasg: You probably don't want to keep that ^^

sda/conode.go
```diff
@@ -88,7 +88,11 @@ func (c *Conode) GetStatus() Status {
 	m["Uptime"] = time.Now().Sub(c.started).String()
 	uname, _ := exec.Command("uname", "-a").Output()
 	m["System"] = strings.TrimRight(string(uname), "\n\r")
+<<<<<<< HEAD
```
nikkolasg: neither that ;)

sda/webservice.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"reflect"
+
+	"github.com/dedis/cothority/log"
+	"github.com/dedis/cothority/network"
+	"github.com/dedis/protobuf"
+	"golang.org/x/net/websocket"
```
nikkolasg: I believe this file is an old one and is not supposed to be there anymore right ? Still using golang websocket implementation.

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+
+	"github.com/dedis/cothority/log"
+	"github.com/dedis/cothority/network"
+	"github.com/dedis/protobuf"
+	"github.com/gorilla/websocket"
+	"gopkg.in/tylerb/graceful.v1"
+)
+
+// WebSocket handles incoming client-requests using the WebSocket
+// protocol for JavaScript-compatibility. When making a new WebSocket,
```
nikkolasg: We are implementing a websocket server because we need a external way for client to communicate with a cothority node. It just so happens that websocket is the most inter-operable way to communicate with language such as javascript but I would not put it as the main reason. 
Sorry, nitpick as always.

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+	log.Lvl3("Starting to listen on", w.server.Server.Addr)
+	go func() {
+		w.server.ListenAndServe()
+		w.startstop <- false
+	}()
+	w.startstop <- true
+}
+
+// RegisterService saves the service as being able to handle messages.
+func (w *WebSocket) RegisterService(service string, s Service) error {
```
nikkolasg: `service` is the name of the service `s` to register

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+func (t wsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	u := websocket.Upgrader{
+		ReadBufferSize:  1024,
+		WriteBufferSize: 1024,
+		CheckOrigin: func(r *http.Request) bool {
+			return true
+		},
+		EnableCompression: true,
+	}
+	ws, err := u.Upgrade(w, r, http.Header{"Set-Cookie": {"sessionID=1234"}})
```
nikkolasg: ? What does it mean ? It is not suposed to be "random" or stg ? I did not check the specs.

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+	defer ws.Close()
+	var ce ClientError
+	// Loop as long as we don't return an error.
+	for ce == nil {
+		mt, buf, err := ws.ReadMessage()
+		if err != nil {
+			log.Error(err)
+			return
+		}
+		s := t.ws.services[t.service]
```
nikkolasg: would it not be better to directly pass a pointer to the service instead of checking the string in the map ? which I believe may cause race conditions ?

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+	log.Lvl3("Stopping", w.server.Server.Addr)
+	<-w.startstop
+	w.server.Stop(100 * time.Millisecond)
+	<-w.startstop
+	w.started = false
+}
+
+// Client is a struct used to communicate with a remote Service running on a
+// sda.Conode
```
nikkolasg: And it's ONE client for ONE cothority node.

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// Send will marshal the message into a ClientRequest message and send it.
+func (c *Client) Send(dst *network.ServerIdentity, path string, buf []byte) ([]byte, ClientError) {
+	c.Lock()
+	defer c.Unlock()
+	if c.keep &&
+		(c.si != nil && !c.si.ID.Equal(dst.ID)) {
+		// We already have a connection, but to another ServerIdentity
+		c.Close()
```
nikkolasg: Should you not return directly ?

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+	if c.si == nil {
+		// Open connection to service.
+		url, err := getWebHost(dst)
+		if err != nil {
+			return nil, NewClientError(err)
+		}
+		log.Lvlf4("Sending %x to %s/%s/%s", buf, url, c.service, path)
+		d := &websocket.Dialer{}
+		// Re-try to connect in case the websocket is just about to start
+		for a := 0; a < 10; a++ {
```
nikkolasg: Put 10 in a constant somewhere.

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+			c.conn, _, err = d.Dial(fmt.Sprintf("ws://%s/%s/%s", url, c.service, path),
+				http.Header{})
+			if err == nil {
+				break
```
nikkolasg: same here, why not returning ?

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+	if c.si != nil && c.conn != nil {
+		c.si = nil
+		return c.conn.Close()
+	}
+	return nil
+}
+
+// Pass the request to the websocket.
+type wsHandler struct {
+	ws      *WebSocket
```
nikkolasg: Do you really need to keep any reference to the websocket ? If not, I'd advise removing the reference. The service can be directly given in as a pointer to the handler. What am I missing ?

services/status/status.go
```diff
@@ -34,6 +34,21 @@ type Response struct {
 	Msg map[string]sda.Status
 }
 
+// SignRequest is interpreted as a request to sign the Hash using the
+// nodes in NodeList.
+type SignRequest struct {
```
nikkolasg: ? What does a SignRequest has anything to do with the status service ?

services/websocket/service.go
```diff
@@ -0,0 +1,200 @@
+package websocket
+
+/*
+ */
```
nikkolasg: A small description would be nice of the service.

services/websocket/service.go
```diff
@@ -0,0 +1,200 @@
...
+// package.
+const ServiceName = "WebSocket"
+
+func init() {
+	sda.RegisterNewService(ServiceName, newService)
+	network.RegisterPacketType(&WSStatus{})
+	network.RegisterPacketType(&Module{})
+}
+
+// Service is our template-service
```
nikkolasg: what is a `template-service` ?

services/websocket/service.go
```diff
@@ -0,0 +1,200 @@
...
+
+	"fmt"
+
+	"encoding/binary"
+
+	"github.com/dedis/cothority/log"
+	"github.com/dedis/cothority/network"
+	"github.com/dedis/cothority/sda"
+	"github.com/dedis/cothority/services/status"
+	"golang.org/x/net/websocket"
```
nikkolasg: Ok I believe this file also must be removed, still using the old websocket thing.

services/websocket/struct.go
```diff
@@ -0,0 +1,5 @@
+package websocket
+
+type Ping struct {
```
nikkolasg: to remove also?

services/websocket/service_test.go
```diff
@@ -0,0 +1,136 @@
+package websocket
```
nikkolasg: to remove also ?

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+			break
+		}
+	}
+	ws.WriteControl(websocket.CloseMessage,
+		websocket.FormatCloseMessage(ce.ErrorCode(), ce.ErrorMsg()),
+		time.Now().Add(time.Second))
+}
+
+// ClientError allows for returning error-codes and error-messages.
+type ClientError interface {
```
nikkolasg: Why is it an interface ? How can I define and use my owns ? I don't see anyway to use it, can you provide a small comment on "how to" ?

sda/websocket.go
```diff
@@ -0,0 +1,366 @@
...
+// RegisterService saves the service as being able to handle messages.
+func (w *WebSocket) RegisterService(service string, s Service) error {
+	w.services[service] = s
+	return nil
+}
+
+// RegisterMessageHandler for a service. Requests from a client to
+// "ws://service/path" will be forwarded to the corresponding
+// ProcessClientRequest of the 'service'.
+func (w *WebSocket) RegisterMessageHandler(service, path string) error {
```
nikkolasg: So in order to use the websocket thing we need to 1) `RegisterService` and then 2) `RegisterMessageHandler` ? 
Why the 1) step is required ? What did I miss ?

ineiti:
+1 for worst PR this month.

sda/websocket.go
ineiti: Chronologically, in the `newCoSiService`-method, the `RegisterMessageHandler` is called, then the `RegisterService` once the `newCoSiService` is done.

When a message comes in, the first part of the path is used to look up the service and find the corresponding `ProcessClientRequest`, and the second part of the path indicates what message-handler needs to be used.

But I think I can get rid of the `RegisterService`-method.

ineiti: After playing around I got rid of the `RegisterMessageHandler`-method. When calling `RegisterService`, all connections to `/service_name/*` are passed to `ProcessClientRequest`. Then it's up to `ProcessClientRequest` to further parse. The `ServiceProcessor:RegisterMessage` makes its own table to dispatch the messages.

sda/websocket.go
ineiti: No, one client per service.

sda/websocket.go
ineiti: I think it's too complicated to have multiple `Client`s for multiple connections.

sda/websocket.go
ineiti: it's in the for-loop that finishes once a connection has been established or if too many trials finished.

sda/websocket.go
ineiti: It's an interface so it can be tested to be `nil` and can pass as `error`. `NewClientError` and `NewClientErrorCode` can be used to instantiate a structure that implements this interface.

services/status/status.go
ineiti: leftover from testing with the website. removed.

services/websocket/service.go
ineiti: leftover

services/websocket/service.go
ineiti: leftover

services/websocket/service.go
ineiti: leftover

services/websocket/service_test.go
ineiti: leftover

services/websocket/struct.go
ineiti: leftover

sda/websocket.go
nikkolasg: I thought it's becaues it's one client for one conode, but if it's not that, then yeah indeed it's fine :)

ineiti:
Now the websocket only handles the different services and sends all requests for one service to `ProcessClientRequest`. The `ServiceProcessor` has a `PCR`-implementation that dispatches the messages to the correct method.

sda/processor.go
```diff
@@ -4,81 +4,76 @@ import (
...
+// WebSocket will then forward requests to "ws://service_name/struct_name"
+// to the given function f, which must be of the following form:
+// func(msg interface{})(ret interface{}, err ClientError)
 //
-// XXX Name should be changed but need to change also in dedis/cosi
+//  * msg is a pointer to a structure to the message sent.
+//  * ret is a pointer to a struct of the return-message.
+//  * err is a Client-error and can return nil or a ClientError that holds
+//	an error-id and an error-msg.
 func (p *ServiceProcessor) RegisterMessage(f interface{}) error {
```
nikkolasg: Plus saying that the `Struct_name` does not contains the package name.

sda/websocket.go
```diff
@@ -0,0 +1,362 @@
...
+type wsHandler struct {
+	serviceName string
+	service     Service
+}
+
+// Wrapper-function so that http.Requests get 'upgraded' to websockets
+// and handled correctly.
+func (t wsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	u := websocket.Upgrader{
+		ReadBufferSize:  1024,
```
nikkolasg: Just wondering if it's needed, are the default values not good ?

sda/processor.go
```diff
@@ -98,56 +93,49 @@ func (p *ServiceProcessor) Process(packet *network.Packet) {
...
-		return
-	}
-	reply := p.GetReply(si, mt, m)
-	if err := p.SendRaw(si, reply); err != nil {
-		log.Error(err)
-	}
+// NewProtocol is a stub for services that don't want to intervene in the
+// protocol-handling.
+func (p *ServiceProcessor) NewProtocol(tn *TreeNodeInstance, conf *GenericConfig) (ProtocolInstance, error) {
+	return nil, nil
```
ineiti: OK - done. Later on I'd like to merge `RegisterProtocolSDA` and `RegisterProtocolService` and put a standard handler in the `ServiceProcessor.

sda/websocket.go
ineiti: was a copy/paste - default values work also. Removed.

sda/conode.go
```diff
@@ -71,11 +84,22 @@ func (c *Conode) GetStatus() Status {
 	a := ServiceFactory.RegisteredServiceNames()
 	sort.Strings(a)
 	m["Available_Services"] = strings.Join(a, ",")
+	m["TX_bytes"] = strconv.FormatUint(c.Router.Tx(), 10)
+	m["RX_bytes"] = strconv.FormatUint(c.Router.Rx(), 10)
+	m["Uptime"] = time.Now().Sub(c.started).String()
+	m["System"] = fmt.Sprintf("%s/%s/%s", runtime.GOOS, runtime.GOARCH,
+		runtime.Version())
+	m["Version"] = "0.9.1"
```
nikkolasg: did you not put a `const version` string somewhere in the code ? Maybe I dreamt it but I thought I saw something like that hanging around.

sda/websocket.go
```diff
@@ -0,0 +1,365 @@
...
+		Server: &http.Server{
+			Addr:    webHost,
+			Handler: w.mux,
+		},
+	}
+	return w
+}
+
+// Start listening on the port.
+func (w *WebSocket) Start() {
```
nikkolasg: `Start` can / should be private ? For the moment, since we still "run" everything, we might as well put it in private.

sda/websocket.go
```diff
@@ -0,0 +1,365 @@
...
+	go func() {
+		w.server.ListenAndServe()
+		w.startstop <- false
+	}()
+	w.startstop <- true
+}
+
+// RegisterService stores a service to the given path. All requests to that
+// path and it's sub-endpoints will be forwarded to ProcessClientRequest.
+func (w *WebSocket) RegisterService(service string, s Service) error {
```
nikkolasg: `RegisterService` can / should be private ? After discussion it does not seem to be used outside sda.

sda/websocket.go
```diff
@@ -0,0 +1,365 @@
...
+	h := &wsHandler{
+		service:     s,
+		serviceName: service,
+	}
+	w.mux.Handle(fmt.Sprintf("/%s/", service), h)
+	return nil
+}
+
+// Stop the websocket and free the port.
+func (w *WebSocket) Stop() {
```
nikkolasg: Same for `Stop` , can be private.

sda/websocket.go
```diff
@@ -0,0 +1,365 @@
...
+	log.Lvl3("Stopping", w.server.Server.Addr)
+	<-w.startstop
+	w.server.Stop(100 * time.Millisecond)
+	<-w.startstop
+	w.started = false
+}
+
+// Client is a struct used to communicate with a remote Service running on a
+// sda.Conode. Using Send it can connect to multiple remote Conodes.
+type Client struct {
```
nikkolasg: Since `Client` might be the only public part of this file, we might as well put it first or at least next to the `ClientError` stuff. It justs seems strange to see in the order `Websocket, Client, wshandler, clientError`.

services/cosi/cosi.go
```diff
@@ -41,24 +40,27 @@ type SignatureRequest struct {
...
+func (cs *CoSi) SignatureRequest(req *SignatureRequest) (network.Body, sda.ClientError) {
+	if req.Roster.ID == sda.RosterID(uuid.Nil) {
+		req.Roster.ID = sda.RosterID(uuid.NewV4())
+	}
 	tree := req.Roster.GenerateBinaryTree()
 	tni := cs.NewTreeNodeInstance(tree, tree.Root, cosi.Name)
 	pi, err := cosi.NewProtocol(tni)
 	if err != nil {
-		return nil, errors.New("Couldn't make new protocol: " + err.Error())
+		return nil, sda.NewClientErrorCode(4100, "Couldn't make new protocol: "+err.Error())
```
nikkolasg: Mhhhh. It seems more longuish than it should don't you think ? 
`sda.WebError(...)` or `sda.ClientError(...)` ? No need for a `Code` part neither for the `New` in my opinion. What do you think ?

ineiti: There are two ways to make a `ClientError`: `NewClientError(error)` and `NewClientErrorCode(int, string)`

sda/websocket.go
ineiti: I can move it up.

PR #730, opened 2016-12-09
Websocket services - 2/3

ineiti:
All changes to the services - please merge 3/3 first.


services/crypto.proto
```diff
@@ -0,0 +1,6 @@
+message Roster {
```
nikkolasg: Multiple comments on the protofiles:
 + the ServerIdentity must also be generated
 + I'm not sure about putting that as "crypto.proto", it's only a list of server, there's not really any crypto.
 + I wonder what's the best place to put that also. Should all proto files should be placed under the package where they are defined (in that case SDA) ? Or all in a single package ? I'm not sure about defining Roster in the Service package, which are un-related. but again, I don't know the "best practices"... What do you think ?

services/identity/service.go
```diff
@@ -66,40 +65,35 @@ type Storage struct {
...
-		return nil, err
+	if cerr != nil {
+		return nil, cerr
 	}
 
 	roster := ids.Root.Roster
 	replies, err := s.propagateIdentity(roster, &PropagateIdentity{ids}, propagateTimeout)
 	if err != nil {
-		return nil, err
+		return nil, sda.NewClientErrorCode(4100, err.Error())
```
nikkolasg: For this occurence and all others, is it possible to put the variable name ? I thought you had defined such constants, instead of the plain error number. Or is it too ugly ?

services/identity/service_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	log.MainTest(m)
+	log.MainTest(m, 2)
```
nikkolasg: to remove

services/identity/service.go
```diff
@@ -229,7 +223,7 @@ func (s *Service) ProposeVote(si *network.ServerIdentity, v *ProposeVote) (netwo
 		}
 		_, err = s.propagateSkipBlock(sid.Root.Roster, usb, propagateTimeout)
 		if err != nil {
-			return nil, err
+			return nil, sda.NewClientErrorCode(4200, err.Error())
```
nikkolasg: same for here, between 4200 and 4100, would be nice to have clear variable name if possible.

services/status/status.proto
```diff
@@ -0,0 +1,11 @@
+message Request{
+
```
nikkolasg: Can we later extend an empty protobuf packet ? Maybe later we'd want to have a "status request" to exactly define what info we want or something in the same spirit. Just asking it's not a big deal.

ineiti: extension is always possible.

services/crypto.proto
ineiti: - Added `ServerIdentity`
- changed name to `network.proto`
- as long as it's only used for services, I think the `cothority/services` is a good place for the `network.proto`-file

services/identity/service.go
ineiti: only in the cleanup-branch

services/skipchain/skipchain.go
```diff
@@ -407,21 +401,21 @@ func (s *Service) addForwardLinks(newest *SkipBlock) ([]*SkipBlock, error) {
...
+				return sda.NewClientErrorCode(4100, "Didn't find Roster nor parent")
 			}
 			roster = sb.Roster
 		}
-		replies, err := s.Propagate(roster, block, propagateTimeout)
-		if err != nil {
-			return err
+		replies, e := s.Propagate(roster, block, propagateTimeout)
+		if e != nil {
+			return sda.NewClientErrorCode(4100, e.Error())
```
nikkolasg: Hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -407,21 +401,21 @@ func (s *Service) addForwardLinks(newest *SkipBlock) ([]*SkipBlock, error) {
...
+func (s *Service) startPropagation(blocks []*SkipBlock) sda.ClientError {
 	log.Lvlf3("Starting to propagate for service %x", s.Context.ServerIdentity().ID[0:8])
 	for _, block := range blocks {
 		roster := block.Roster
 		if roster == nil {
 			// Suppose it's a dataSkipBlock
 			sb, ok := s.getSkipBlockByID(block.ParentBlockID)
 			if !ok {
-				return errors.New("Didn't find Roster nor parent")
+				return sda.NewClientErrorCode(4100, "Didn't find Roster nor parent")
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -326,17 +320,17 @@ func (s *Service) startBFTSignature(block *SkipBlock) error {
 	// Start the protocol
 	tree := el.GenerateNaryTreeWithRoot(2, s.ServerIdentity())
 
-	node, err := s.CreateProtocolSDA(skipchainBFT, tree)
-	if err != nil {
-		return errors.New("Couldn't create new node: " + err.Error())
+	node, e := s.CreateProtocolSDA(skipchainBFT, tree)
+	if e != nil {
+		return sda.NewClientErrorCode(4100, "Couldn't create new node: "+e.Error())
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -263,17 +257,17 @@ func (s *Service) signNewSkipBlock(latest, newest *SkipBlock) (*SkipBlock, *Skip
 	if newest != nil && newest.Roster == nil {
 		log.Lvl3("Got a data-block")
 		if newest.ParentBlockID.IsNull() {
-			return nil, nil, errors.New("Data skipblock without parent")
+			return nil, nil, sda.NewClientErrorCode(4100, "Data skipblock without parent")
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -263,17 +257,17 @@ func (s *Service) signNewSkipBlock(latest, newest *SkipBlock) (*SkipBlock, *Skip
...
 		}
 		parent, ok := s.getSkipBlockByID(newest.ParentBlockID)
 		if !ok {
-			return nil, nil, errors.New("Didn't find parent block")
+			return nil, nil, sda.NewClientErrorCode(4100, "Didn't find parent block")
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -263,17 +257,17 @@ func (s *Service) signNewSkipBlock(latest, newest *SkipBlock) (*SkipBlock, *Skip
...
 		}
 		newest.Roster = parent.Roster
 	}
 	// Now verify if it's a valid block
 	if err := s.verifyNewSkipBlock(latest, newest); err != nil {
-		return nil, nil, errors.New("Verification of newest SkipBlock failed: " + err.Error())
+		return nil, nil, sda.NewClientErrorCode(4100, "Verification of newest SkipBlock failed: "+err.Error())
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -171,24 +172,24 @@ func (s *Service) GetUpdateChain(si *network.ServerIdentity, latestKnown *GetUpd
...
 	childID := scsb.ChildID
 	parent, ok := s.getSkipBlockByID(parentID)
 	if !ok {
-		return nil, errors.New("couldn't find skipblock")
+		return nil, sda.NewClientErrorCode(4100, "couldn't find skipblock")
 	}
 	child, ok := s.getSkipBlockByID(childID)
 	if !ok {
-		return nil, errors.New("couldn't find skipblock")
+		return nil, sda.NewClientErrorCode(4100, "couldn't find skipblock")
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -171,24 +172,24 @@ func (s *Service) GetUpdateChain(si *network.ServerIdentity, latestKnown *GetUpd
...
 // SetChildrenSkipBlock creates a new SkipChain if that 'service' doesn't exist
 // yet.
-func (s *Service) SetChildrenSkipBlock(si *network.ServerIdentity, scsb *SetChildrenSkipBlock) (network.Body, error) {
+func (s *Service) SetChildrenSkipBlock(scsb *SetChildrenSkipBlock) (network.Body, sda.ClientError) {
 	parentID := scsb.ParentID
 	childID := scsb.ChildID
 	parent, ok := s.getSkipBlockByID(parentID)
 	if !ok {
-		return nil, errors.New("couldn't find skipblock")
+		return nil, sda.NewClientErrorCode(4100, "couldn't find skipblock")
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -147,10 +148,10 @@ func (s *Service) ProposeSkipBlock(si *network.ServerIdentity, psbd *ProposeSkip
 // skipchain from the latest block the caller knows of to the actual latest
 // SkipBlock.
 // Somehow comparable to search in SkipLists.
-func (s *Service) GetUpdateChain(si *network.ServerIdentity, latestKnown *GetUpdateChain) (network.Body, error) {
+func (s *Service) GetUpdateChain(latestKnown *GetUpdateChain) (network.Body, sda.ClientError) {
 	block, ok := s.getSkipBlockByID(latestKnown.LatestID)
 	if !ok {
-		return nil, errors.New("Couldn't find latest skipblock")
+		return nil, sda.NewClientErrorCode(4100, "Couldn't find latest skipblock")
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -122,17 +122,18 @@ func (s *Service) ProposeSkipBlock(si *network.ServerIdentity, psbd *ProposeSkip
...
 	}
 	prop.AggregateResp = el.Aggregate
 
 	prop.updateHash()
 
+	var err error
 	prev, prop, err = s.signNewSkipBlock(prev, prop)
 	if err != nil {
-		return nil, errors.New("Verification error: " + err.Error())
+		return nil, sda.NewClientErrorCode(4100, "Verification error: "+err.Error())
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -107,10 +107,10 @@ func (s *Service) ProposeSkipBlock(si *network.ServerIdentity, psbd *ProposeSkip
 		// are correctly set up
 		prop.Index = 0
 		if prop.MaximumHeight == 0 {
-			return nil, errors.New("Set a maximumHeight > 0")
+			return nil, sda.NewClientErrorCode(4100, "Set a maximumHeight > 0")
 		}
 		if prop.BaseHeight == 0 {
-			return nil, errors.New("Set a baseHeight > 0")
+			return nil, sda.NewClientErrorCode(4100, "Set a baseHeight > 0")
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -107,10 +107,10 @@ func (s *Service) ProposeSkipBlock(si *network.ServerIdentity, psbd *ProposeSkip
 		// are correctly set up
 		prop.Index = 0
 		if prop.MaximumHeight == 0 {
-			return nil, errors.New("Set a maximumHeight > 0")
+			return nil, sda.NewClientErrorCode(4100, "Set a maximumHeight > 0")
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -96,7 +96,7 @@ func (s *Service) ProposeSkipBlock(si *network.ServerIdentity, psbd *ProposeSkip
 				var ok bool
 				pointer, ok = s.getSkipBlockByID(pointer.BackLinkIds[0])
 				if !ok {
-					return nil, errors.New("Didn't find convenient SkipBlock for height " +
+					return nil, sda.NewClientErrorCode(4100, "Didn't find convenient SkipBlock for height "+
```
nikkolasg: hardcoded error code

services/skipchain/skipchain.go
```diff
@@ -73,7 +73,7 @@ func (s *Service) ProposeSkipBlock(si *network.ServerIdentity, psbd *ProposeSkip
 		var ok bool
 		prev, ok = s.getSkipBlockByID(psbd.LatestID)
 		if !ok {
-			return nil, errors.New("Didn't find latest block")
+			return nil, sda.NewClientErrorCode(4100, "Didn't find latest block")
```
nikkolasg: hardcoded error code

services/network.proto
```diff
@@ -0,0 +1,13 @@
+message ServerIdentity{
+    required bytes public = 1;
+    required bytes id = 2;
```
nikkolasg: Should ID be required ? Because it can be deterministically generated out of the public and address, I'm not sure. The protobuf library puts it as required if it's not a pointer or it does not have the `protobuf:"opt"` struct tag field.

services/network.proto
```diff
@@ -0,0 +1,13 @@
...
+    required string address = 3;
+    required string description = 4;
```
nikkolasg: Should it be required ? Can it be empty ?

services/identity/service_test.go
nikkolasg: Still to remove ?

services/identity/service.go
```diff
@@ -179,35 +173,35 @@ func (s *Service) ProposeVote(si *network.ServerIdentity, v *ProposeVote) (netwo
...
 	if err != nil {
-		return nil, err
+		return nil, sda.NewClientErrorCode(4200, err.Error())
 	}
 
 	// Propagate the vote
 	_, err = s.propagateConfig(sid.Root.Roster, v, propagateTimeout)
 	if err != nil {
-		return nil, err
+		return nil, sda.NewClientErrorCode(4200, err.Error())
```
nikkolasg: Same here, will it be removed or is it on purpose ? (the hardcoded error code)

services/identity/service.go
nikkolasg: Still there

services/identity/api.go
```diff
@@ -151,85 +150,90 @@ func (i *Identity) GetProposed() *Config {
...
 	}
 	hash, err := i.Proposed.Hash()
 	if err != nil {
-		return err
+		return sda.NewClientErrorCode(4100, err.Error())
 	}
 	sig, err := crypto.SignSchnorr(network.Suite, i.Private, hash)
 	if err != nil {
-		return err
+		return sda.NewClientErrorCode(4100, err.Error())
```
nikkolasg: Here and above hardcoded error code

services/cosi/cosi.go
```diff
@@ -41,24 +40,30 @@ type SignatureRequest struct {
 
 // SignatureResponse is what the Cosi service will reply to clients.
 type SignatureResponse struct {
-	Sum       []byte
+	Hash      []byte
 	Signature []byte
+	Aggregate abstract.Point
```
nikkolasg: Why sending back the aggregate public key ? The client is supposed to know the aggregate key already since he is the one giving the Roster to the service. And it's strange that a service gives back the key under which to verify the signature, it defeats the whole purpose of the signature ... ?

services/identity/service.go
ineiti: cleanup_2...

services/network.proto
ineiti: I'd put it as required.

services/network.proto
ineiti: required, but can be empty.

services/cosi/cosi.go
ineiti: I put it in because I want the web-interface to be able to select any number of nodes and then request a signature using those. So instead of calculating the aggregate key in JavaScript, I calculate it in go and send it back. But yes, that's bad, I should learn how to calculate aggregate public keys in javascript...

services/status/status.proto
nikkolasg: Seems like we have to manually define an extension so it's available to others. https://developers.google.com/protocol-buffers/docs/proto#extensions

Anyway, it's probably not important right now so it's ok.

services/network.proto
nikkolasg: Does it "simply" work if you put it as optional then ? I don't see any point in having something required when it's actually optional. As a general guideline , we should be careful to put the least possible fields

PR #732, opened 2016-12-12
Websocket cleanup 2 - 3/3

ineiti:
Here go the error-messages.


ineiti:
Fixed a last error with regard to non-garbage-collected `Client`s. 2 hours to write the test ;)

PR #740, opened 2016-12-15
Check 724

### LINKED ISSUE #724: (cosi|cothorityd) check improvements
when running `check` and a conode fails, 
now: it is still tested as a pair against the other conodes
wanted: is not tested anymore

now: `cothorityd check -g group.toml`
after: `cothorityd check group.toml`

now: `check` verifies all pairs per default
after: `check` verifies individual servers and once the group in total, but adds a `--detail`-flag to do the pairwise-test

ineiti:
closes #724

### DIFF
--- a/cothority.go
+++ b/cothority.go
@@ -98,6 +98,10 @@ func main() {
 		},
 	}
 	cliApp.Flags = serverFlags
+	cliApp.Before = func(c *cli.Context) error {
+		log.SetDebugVisible(c.Int("debug"))
+		return nil
+	}
 	// default action
 	cliApp.Action = func(c *cli.Context) error {
 		runServer(c)
@@ -110,7 +114,6 @@ func main() {
 
 func runServer(ctx *cli.Context) {
 	// first check the options
-	log.SetDebugVisible(ctx.Int("debug"))
 	config := ctx.String("config")
 
 	server.RunServer(config)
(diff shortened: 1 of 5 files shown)


cothority.go
```diff
@@ -80,15 +80,20 @@ func main() {
...
+			ArgsUsage: "Cothority group definition file",
+			Action:    checkConfig,
 			Flags: []cli.Flag{
 				cli.StringFlag{
 					Name:  "g",
 					Usage: "Cothority group definition file",
 				},
+				cli.BoolFlag{
+					Name:  "detail,l",
+					Usage: "Show detail in signing",
```
nikkolasg: More detail for detail ^^  Namely, it will try pair wise servers together or stg like that.

cothority.go
```diff
@@ -115,5 +120,8 @@ func runServer(ctx *cli.Context) {
 // signature from each.
 func checkConfig(c *cli.Context) error {
 	tomlFileName := c.String("g")
-	return check.Config(tomlFileName)
+	if c.NArg() > 0 {
+		tomlFileName = c.Args().First()
+	}
+	return check.Config(tomlFileName, c.Bool("detail"))
```
nikkolasg: so the verification process is not in any lib or so ? It's "included" in the cosi, and the cothority depends on it ? mmhhh... We might want to do something else for later.

ineiti: yes, it's part of cosi, as it's signing. So it _is_ in a library, but the library is in the cosi-directory.

PR #743, opened 2016-12-19
Documentation

ineiti:
Implements the onet-documentation. While documenting, I clean up a little...
- move onet/app/lib/cothorityd.sh to libcothority/cothority.sh
- move flags and commands from cothority.go to libcothority/commands.go


nikkolasg:
I'm not sure about `libcothority`:
+ it's a very very generic name that can mean anything, sounds like our "trash" package where we put things we don't know where they are supposed to go ;)
+ We now have somewhat 3 different libraries to take into account when creating a new app:  
  - "github.com/dedis/cothority/cosi/check" 
  - "github.com/dedis/onet/app/server"
  - "github.com/dedis/cothority/libcothority"

The last one seems to be a wrapper around the first two but at the end of the day, we still have 3 or 4 different libraries *just for creating an app*.  It seems way too much and that means their purpose is not clear (one library is just filling one helper to create an app, another library is giving another helper etc).
Instead of creating a new package we should think on how to merge existing packages together, and add functionality like `cothority/libapp` or just `cothority/app` whatever the best name seems to be. I'm beginning to think that `onet/app` should maybe be moved in `cothority/app` (or any name) and merge those 3-4 mentioned above libraries into one package.

ineiti:
I'll reply once I put it into Ready4Merge ;)

ineiti:
replaced by #752

PR #746, opened 2016-12-20
CreateProtocol - everywhere

nikkolasg:
This PR is the cothority side of https://github.com/dedis/onet/pull/36 . 
Namely, it changes all calls to `CreateProtocolSDA/Onet` to `CreateProtocol` and gives a `NilServiceID` if the protocol is to be handled only by Onet.


byzcoin/byzcoin_simulation.go
```diff
@@ -91,7 +91,7 @@ func (e *Simulation) Run(onetConf *onet.SimulationConfig) error {
 	log.Lvl2("Simulation starting with: Rounds=", e.Rounds)
 	server := NewByzCoinServer(e.Blocksize, e.TimeoutMs, e.Fail)
 
-	pi, err := onetConf.Overlay.CreateProtocolOnet("Broadcast", onetConf.Tree)
+	pi, err := onetConf.Overlay.CreateProtocol("Broadcast", onetConf.Tree, onet.NilServiceID)
```
ineiti: Can't we have a simple `nil` here?

PR #752, opened 2017-01-04
adjusting READMEs to new framework

ineiti:


README.md
```diff
@@ -1,172 +1,156 @@
...
 
-[CoSi](http://dedis.cs.yale.edu/dissent/papers/witness-abs) is a 
-protocol for scalable collective signing, which enables an authority or 
-leader to request that statements be publicly validated and (co-signed) 
-by a decentralized group of witnesses. 
-Each run of the protocol yields a single digital signature with size and 
-verification cost comparable to an individual signature, but compactly
-attests that both the leader and perhaps many witnesses observed and 
-agreed to sign the statement.
+To configure your conode you need to *open two consecutive ports* (e.g., 6879 and 6880) on your machine, then execute
```
nikkolasg: Maybe a very short explanation of why we need two ports? 
`One port is used to communicate with other conodes, and the second one is used by external client applications wishing to interacts with a conode over websockets`

Daeinar: Good idea, I'll add that.

PR #771, opened 2017-01-10
following onet changes 1/2

### LINKED ISSUE #491: Add load/save methods to Context

ineiti:
Closes #491 
Once https://github.com/dedis/onet/pull/59 is merged.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -479,6 +479,9 @@ func (s *Service) tryLoad() error {
 	if err != nil {
 		return err
 	}
+	if msg == nil {
+		return nil
+	}
 	var ok bool
 	s.SkipBlockMap, ok = msg.(*SkipBlockMap)
 	if !ok {
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -12,11 +12,6 @@ import (
 
 	"time"
 
-	"io/ioutil"
-	"os"
-
-	"fmt"
-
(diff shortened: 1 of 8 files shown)


skipchain/skipchain.go
```diff
@@ -472,40 +466,33 @@ func (s *Service) lenSkipBlocks() int {
...
+		return err
 	}
-	if len(b) > 0 {
-		_, msg, err := network.Unmarshal(b)
-		if err != nil {
-			return fmt.Errorf("Couldn't unmarshal: %s", err)
-		}
-		log.Lvl3("Successfully loaded")
-		s.SkipBlockMap = msg.(*SkipBlockMap)
+	if msg == nil {
```
nikkolasg: If `err == nil`, can `msg == nil` ? Sounds strange.

ineiti: I already thought about https://github.com/dedis/onet/issues/66 - should an empty/non-existant file return an error or just nil data?

nikkolasg: I think it should return an error such as https://golang.org/pkg/os/#Open and https://golang.org/pkg/io/ioutil/#ReadFile

PR #775, opened 2017-01-12
Conode runscript 721

### LINKED ISSUE #721: cothorityd-runscript
create a script for cothorityd so that it gets restarted even if the daemon dies. Perhaps add an option to update the packet.

ineiti:
Closes #721


run_conode.sh
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env bash
+set -e
+
+VERSION=1.0-pre1
+MAILCMD=mail
+MAILADDR=linus.gasser@epfl.ch
```
nikkolasg: Do you have any reference as how to use run_conode ? in the README ? 
If so, do you specifiy that it sends a log to this address automatically if mail is installed ?

run_conode.sh
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env bash
+set -e
+
+VERSION=1.0-pre1
```
nikkolasg: Well, we should have a consistent versionning system, putting at two places seems like trouble for the future...

nikkolasg: I guess it's possible to make onet.Server prints out the version or stg ? Or conode binary ?

ineiti: hmm - can't run onet.Server here... But I can do some grep-magick...

run_conode.sh
```diff
@@ -0,0 +1,221 @@
...
+	exec $RUN_CONODE "$ALL_ARGS"
+}
+
+migrate(){
+	PATH_VERSION=$PATH_CO/conode/version
+	if [ -d $PATH_CO/cothorityd ]; then
+		echo "Moving cothorityd-directory to conode"
+		mv $PATH_CO/cothorityd $PATH_CO/conode
+		echo 0.9 > $PATH_VERSION
+	elif [ -d $PATH_CO/cothority ]; then
```
nikkolasg: is it not the same thing ?

run_conode.sh
ineiti: Only when called with `-mail`. I'll add that to the output.

run_conode.sh
ineiti: no, there has been some `cothorityd` and some `cothority`...

nikkolasg:
I don't quite get the difference between `run_conode.sh <public> <private>` and just `cd conode; go build && ./conode`. 
In general, can you add a very short sentence to say what is run_conode.sh useful for and how to use it ?

ineiti:
`run_conode.sh` gives you:
- migration of data in {.config,Library}/{cothority,conode}
- loop around `conode` if it dies
- automatic updates (-update)
- sending of log-files (-mail)

conode/README.md
```diff
@@ -72,6 +72,22 @@ To start your conode with the default (private) configuration file, located at `
...
+* migration of data if we change it
+* loop around conode if it quits
+* automatic updates (-update)
+* sending of log-files (-mail)
+
+You can run it with all extras like so:
+
+`./run_conode.sh -mail -update`
+
+And it will send an email to DEDIS with the last 200 lines of log-output every time conode restarts.
```
nikkolasg: Without the `And`.

PR #789, opened 2017-01-17
Run conode 788

### LINKED ISSUE #788: run_conode.sh error
`run_conode.sh` has some errors:
- `go install`-lines are wrong
- it should check for correct installation of `mail`-command if it's asked for
- it should check for presence of `Description` in `private.toml` and copy the `Description` of `public.toml` over, if it is there.

ineiti:
Closes #788


nikkolasg:
Should we already merge this now or wait 'till you fix the update stuff you talked about ? I don't really have an opinion as long as both are in an issue somewhere (which is the case for #788 !) :)

ineiti:
I just opened #790 , so I think you can merge this one and test it tomorrow on your server. Also send a message to @Liamsi on slack that he can test this out...

PR #804, opened 2017-01-26
33c3 pop

ineiti:
Merging back pop-party to master


conode/conode.go
```diff
@@ -22,6 +22,7 @@ import (
 	_ "github.com/dedis/cothority/cosi/service"
 	_ "github.com/dedis/cothority/guard/service"
 	_ "github.com/dedis/cothority/identity"
+	_ "github.com/dedis/cothority/pop/service"
```
nikkolasg: Mhhhh... I'm not so sure about that. What about having a "demo" folder or something that clearly marks that at least this pop service has been hacked in 3 days  for a demo and is in no way more than a demo ? 
I'm not so sure about including it in the default list of services to offer when running a conode...

pop/.gitignore
```diff
@@ -0,0 +1,4 @@
+*pdf
+key*odg
```
nikkolasg: I see there is still a paperkey.odg in the files of the PR. Is it intended ?

pop/33c3/.gitignore
```diff
@@ -0,0 +1,2 @@
+server.crt
+server.key
```
nikkolasg: can we add the "main", or "33c3" binary ?

pop/.gitignore
ineiti: yes, that's the template used to create random paperkeys.

conode/conode.go
ineiti: we could even do a `cothority_hacks` or `cothority_samples` or such repo where we put that code inside. Then we can say that every change in ONet needs to be reflected in `cothority_template`, then `cothority` and finally `cothority_hacks`.

pop/33c3/.gitignore
ineiti: yes, we can ;)

conode/conode.go
ineiti: After discussion we will merge `pop` back into master, but the website for `33c3` which was quite hacky (but running!) will go into the `archives`.

pop/app.go
```diff
@@ -0,0 +1,353 @@
...
+var client *service.Client
+
+// Config represents either a manager or an attendee configuration.
+type Config struct {
+	Private abstract.Scalar
+	Public  abstract.Point
+	Index   int
+	Address network.Address
+	Final   *service.FinalStatement
+}
```
nikkolasg: Can you give a very short comment about what is each field ? for example, I've no idea what Index means nor Address.

pop/app.go
```diff
@@ -0,0 +1,353 @@
...
+	newConfig()
+	host, port, err := net.SplitHostPort(c.Args().First())
+	if err != nil {
+		return err
+	}
+	addrs, err := net.LookupHost(host)
+	if err != nil {
+		return err
+	}
+	addr := network.NewAddress(network.PlainTCP, fmt.Sprintf("%s:%s", addrs[0], port))
```
nikkolasg: I think you can change that to `network.NewTCPAddress(...)`

pop/app.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	mainConfig.Address = addr
+	log.Info("Successfully linked with", addr)
+	writeConfig()
+	return nil
+}
+
+// sets up a configuration
+func orgConfig(c *cli.Context) error {
+	log.Info("Org: Config", mainConfig.Address.String())
```
nikkolasg: Should be sufficient when removing the `.String()`

pop/app.go
```diff
@@ -0,0 +1,353 @@
...
+// Config represents either a manager or an attendee configuration.
+type Config struct {
+	Private abstract.Scalar
+	Public  abstract.Point
+	Index   int
+	Address network.Address
+	Final   *service.FinalStatement
+}
+
+var mainConfig *Config
```
nikkolasg: maybe I reviewed it before but I still don't like this global variable.. !

pop/app.go
```diff
@@ -0,0 +1,353 @@
...
+		log.Fatal("Please give a public key")
+	}
+	str := c.Args().First()
+	if !strings.HasPrefix(str, "[") {
+		str = "[" + str + "]"
+	}
+	str = strings.Replace(str, "\"", "", -1)
+	str = strings.Replace(str, "[", "", -1)
+	str = strings.Replace(str, "]", "", -1)
+	str = strings.Replace(str, "\\", "", -1)
```
nikkolasg: what are all of these custom rules ... ?

pop/app.go
```diff
@@ -0,0 +1,120 @@
...
+		},
+	}
+	app.Before = func(c *cli.Context) error {
+		log.SetDebugVisible(c.Int("debug"))
+		return nil
+	}
+	app.Run(os.Args)
+}
+
+// links this pop to a cothority
```
ineiti: Should have more comments.

ineiti:
Will make better separation.

PR #806, opened 2017-01-27
changed path in all scripts

ineiti:
I hope these are the last changes for `v1`...


nikkolasg:
Sounds fine. Is there a `go get gopkg.in/dedis/onet.v1` in your scripts before ?

ineiti:
In the test-scripts? No, as all the cothority is depending on `onet.v1`, it supposes that the repo is available.

nikkolasg:
Ok.

PR #811, opened 2017-01-27
make symlink if authorized_keys exist

### LINKED ISSUE #810: `cisc follow` - don't overwrite .ssh/authorized_keys

ineiti:
Closes #810

### DIFF
--- a/cisc/cisc.go
+++ b/cisc/cisc.go
@@ -241,7 +241,8 @@ func kvDel(c *cli.Context) error {
  * ssh:device:server / ssh_public_key
  * where 'ssh' is a fixed string, 'device' is the device where the private
  * key is stored and 'server' the server that should add the public key to
- * its authorized_keys.
+ * its authorized_keys. For safety reasons, this function saves to
+ * authorized_keys.cisc. But the sysadmin can make a link.
  */
 func sshAdd(c *cli.Context) error {
 	cfg := loadConfigOrFail(c)
--- a/cisc/cisc.go
+++ b/cisc/cisc.go
@@ -238,11 +238,22 @@ func kvDel(c *cli.Context) error {
 /*
  * Commands related to the ssh-handling. All ssh-keys are stored in the
  * identity-sc as
- * ssh:device:server / ssh_public_key
+ *
+ *   ssh:device:server = ssh_public_key
+ *
  * where 'ssh' is a fixed string, 'device' is the device where the private
  * key is stored and 'server' the server that should add the public key to
- * its authorized_keys. For safety reasons, this function saves to
- * authorized_keys.cisc. But the sysadmin can make a link.
+ * its authorized_keys.
+ *
+ * For safety reasons, this function saves to authorized_keys.cisc instead
+ * of overwriting authorized_keys. If authorized_keys doesn't exist,
+ * a symbolic link to authorized_keys.cisc is created.
+ *
+ * If you want to use your own authorized_keys but also allow keys in
+ * authorized_keys.cisc to log in to your system, you can add the following
+ * line to /etc/ssh/sshd_config
+ *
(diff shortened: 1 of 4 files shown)


cisc/lib.go
```diff
@@ -94,18 +96,16 @@ func (cfg *ciscConfig) proposeSendVoteUpdate(p *identity.Config) {
...
 func (cfg *ciscConfig) writeAuthorizedKeys(c *cli.Context) {
 	var keys []string
 	dir, _ := sshDirConfig(c)
-	authFile := dir + "/authorized_keys"
-	// Make backup
-	b, err := ioutil.ReadFile(authFile)
-	if err == nil {
-		err = ioutil.WriteFile(authFile+".back", b, 0600)
-		log.ErrFatal(err)
+	authFileOrig := filepath.Join(dir, "authorized_keys")
```
nikkolasg: nice `filepath`. We use `path` in most of our code, we should migrate to using this one since it's apparently supporting windows..

ineiti: I was not aware of the difference - but looking at the code I see that `path` merely joins the elements using `/`, while `filepath` tries to be a bit more intelligent.

cisc/lib.go
```diff
@@ -94,18 +96,16 @@ func (cfg *ciscConfig) proposeSendVoteUpdate(p *identity.Config) {
...
+	authFile := authFileOrig + ".cisc"
+	if _, err := os.Stat(authFileOrig); os.IsNotExist(err) {
+		log.Info("Making link from authorized_keys to authorized_keys.cisc")
+		os.Symlink(authFile, authFileOrig)
```
nikkolasg: Here you do a symlink automatically when the `authorized_keys.cisc` does not exists, is that right ?
But in that case I don't understand the comment above 
>  For safety reasons, this function saves to authorized_keys.cisc. But the sysadmin can make a link.

It seems like onet does the link automatically, and that we still do overwrite `authorized_keys` vanilla, since we write to the symlink ? 
There's something I don't get ...

ineiti: Wrong comment, sorry.

1. it always writes to `authorized_keys.cisc`
2. if `authorized_keys` does not exist, it makes a symbolic link pointing to `authorized_keys.cisc`

nikkolasg: +1 for 1.
Why 2. ? If we're anyway using a different file, why using the regular file at all ?

ineiti: Convenience - if you're using _only_ cisc, it will make a symlink for you, and it works out-of-the box. If you already have an `authorized_keys`, then it will not touch it.

Best would be if we could combine the two files somewhat - like copying `authorized_keys` to `authorized_keys.manual` and then merge that with the keys from CISC in the `authorized_keys`-file. But I'm afraid it'll be **too intelligent**. Ideal would be if `sshd` would allow for multiple `authorized_keys` - for that one would have to add

```
AuthorizedKeysFile ~/.ssh/authorized_keys ~/.ssh/authorized_keys.cisc
```

to `/etc/ssh/sshd_config`. I'll add that to the `README.md`

cisc/cisc.go
```diff
@@ -241,7 +241,8 @@ func kvDel(c *cli.Context) error {
  * ssh:device:server / ssh_public_key
  * where 'ssh' is a fixed string, 'device' is the device where the private
  * key is stored and 'server' the server that should add the public key to
- * its authorized_keys.
+ * its authorized_keys. For safety reasons, this function saves to
+ * authorized_keys.cisc. But the sysadmin can make a link.
```
nikkolasg: Nitpicking:
> authorized_keys.cisc and a link from authorized_keys is made if the latter does not exists.
(or stg like that :D)

ineiti:
Better?

PR #816, opened 2017-02-01
Pop service - 1/2

ineiti:
This is a first part of the Proof-of-personhood code - the service.

It is set up as a suite of good and nice commits. So you can have a look at each commit and understand how the thing grows. And yes, each commit passes all tests ;)

### DIFF
--- a/pop/service/api.go
+++ b/pop/service/api.go
@@ -1,0 +1,207 @@
+package service
+
+import (
+	"bytes"
+
+	"github.com/BurntSushi/toml"
+	"github.com/satori/go.uuid"
+	"gopkg.in/dedis/crypto.v0/abstract"
+	"gopkg.in/dedis/crypto.v0/base64"
+	"gopkg.in/dedis/crypto.v0/eddsa"
+	"gopkg.in/dedis/onet.v1"
+	"gopkg.in/dedis/onet.v1/crypto"
+	"gopkg.in/dedis/onet.v1/log"
+	"gopkg.in/dedis/onet.v1/network"
+)
+
+const (
+	// ErrorWrongPIN indicates a wrong PIN
+	ErrorWrongPIN = iota + 4100
+	// ErrorInternal indicates something internally went wrong - see the
+	// error message
+	ErrorInternal
+	// ErrorOtherFinals indicates that one or more of the other conodes
+	// are still missing the finalization-step
+	ErrorOtherFinals
+)
+
+func init() {
+	network.RegisterMessage(&FinalStatement{})
+	network.RegisterMessage(&PopDesc{})
+}
+
+// FinalStatement is the final configuration holding all data necessary
+// for a verifier.
+type FinalStatement struct {
+	// Desc is the description of the pop-party.
+	Desc *PopDesc
+	// Attendees holds a slice of all public keys of the attendees.
+	Attendees []abstract.Point
+	// Signature is created by all conodes responsible for that pop-party
+	Signature []byte
+}
+
+// The toml-structure for (un)marshaling with toml
+type finalStatementToml struct {
+	Desc      *popDescToml
+	Attendees []string
+	Signature string
+}
+
+// NewFinalStatementFromToml creates a final statement from a toml slice-of-bytes.
+func NewFinalStatementFromToml(b []byte) (*FinalStatement, error) {
+	fsToml := &finalStatementToml{}
+	_, err := toml.Decode(string(b), fsToml)
+	if err != nil {
+		return nil, err
+	}
+	sis := []*network.ServerIdentity{}
+	for _, s := range fsToml.Desc.Roster {
+		uid, err := uuid.FromString(s[2])
+		if err != nil {
+			return nil, err
+		}
+		pub, err := crypto.String64ToPub(network.Suite, s[3])
+		if err != nil {
+			return nil, err
+		}
+		sis = append(sis, &network.ServerIdentity{
+			Address:     network.Address(s[0]),
+			Description: s[1],
+			ID:          network.ServerIdentityID(uid),
+			Public:      pub,
+		})
+	}
+	rostr := onet.NewRoster(sis)
+	desc := &PopDesc{
+		Name:     fsToml.Desc.Name,
+		DateTime: fsToml.Desc.DateTime,
+		Location: fsToml.Desc.Location,
+		Roster:   rostr,
+	}
+	atts := []abstract.Point{}
+	for _, p := range fsToml.Attendees {
+		pub, err := crypto.String64ToPub(network.Suite, p)
+		if err != nil {
+			return nil, err
+		}
+		atts = append(atts, pub)
+	}
+	sig := make([]byte, 64)
+	sig, err = base64.StdEncoding.DecodeString(fsToml.Signature)
+	// TODO: sign and verify signature
+	if err != nil {
+		return nil, err
+	}
+	return &FinalStatement{
+		Desc:      desc,
+		Attendees: atts,
+		Signature: sig,
+	}, nil
+}
+
+// ToToml returns a toml-slice of byte and an eventual error.
+func (fs *FinalStatement) ToToml() ([]byte, error) {
+	rostr := [][]string{}
+	for _, si := range fs.Desc.Roster.List {
+		str, err := crypto.PubToString64(nil, si.Public)
+		if err != nil {
+			return nil, err
+		}
+		sistr := []string{si.Address.String(), si.Description,
+			uuid.UUID(si.ID).String(), str}
+		rostr = append(rostr, sistr)
+	}
+	descToml := &popDescToml{
+		Name:     fs.Desc.Name,
+		DateTime: fs.Desc.DateTime,
+		Location: fs.Desc.Location,
+		Roster:   rostr,
+	}
+	atts := []string{}
+	for _, p := range fs.Attendees {
+		str, err := crypto.PubToString64(nil, p)
+		if err != nil {
+			return nil, err
+		}
+		atts = append(atts, str)
+	}
+	fsToml := &finalStatementToml{
+		Desc:      descToml,
+		Attendees: atts,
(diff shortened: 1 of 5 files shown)


pop/service/api.go
```diff
@@ -0,0 +1,215 @@
...
+	ErrorOtherFinals
+)
+
+func init() {
+	network.RegisterMessage(&FinalStatement{})
+	network.RegisterMessage(&PopDesc{})
+}
+
+// FinalStatement is the final configuration holding all data necessary
+// for a verifier.
```
nikkolasg: what is a verifier ?

pop/service/api.go
```diff
@@ -0,0 +1,215 @@
...
+
+// The toml-structure for (un)marshaling with toml
+type finalStatementToml struct {
+	Desc      *popDescToml
+	Attendees []string
+	Signature string
+}
+
+// NewFinalStatementFromString creates a final statement from a string.
+func NewFinalStatementFromString(s string) *FinalStatement {
```
nikkolasg: why string ? In toml there is `toml.Unmarshal` which takes a slice of bytes. I'm just wondering if there is a reason or not. The caller of the library, will always have to do the casting `string(myBytesRead)` because he'll read `[]bytes` most of the time with `ioutil.ReadFile(statementPath)`, so maybe change that to `[]byte` ?

pop/service/api.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	return &FinalStatement{
+		Desc:      desc,
+		Attendees: atts,
+		Signature: sig,
+	}
+}
+
+// PointToB64 converts an abstract.Point to a base64-point.
+func PointToB64(p abstract.Point) string {
```
nikkolasg: All the similar functions below look nice but what's different than with `onet/crypto/key.go` ? Seems the difference is `onet` uses `io.Reader` and `io.Pointer`. That can easily be added there if needed I think..

pop/service/api.go
nikkolasg: I know returning error is more verbose, but it's how golang does it. The fact that this function does **not** return any errors worries me because it assumes the string is a valid statement. This will probably make something fail later in unexpected situations. I'd gladly see this function returns an error...
Places where things can go wrong:
 + `toml.Decode` at 52
 + `uuid.FromString()` at 59
 + `append(atts, B64ToPoint(p))` at 80: point can be nil and then panic after !
 + `base64.StdEncoding.DecodeString(fsToml.Signature)` line 83

pop/service/api.go
```diff
@@ -0,0 +1,215 @@
...
+		DateTime: fsToml.Desc.DateTime,
+		Location: fsToml.Desc.Location,
+		Roster:   rostr,
+	}
+	atts := []abstract.Point{}
+	for _, p := range fsToml.Attendees {
+		atts = append(atts, B64ToPoint(p))
+	}
+	sig := make([]byte, 64)
+	sig, err = base64.StdEncoding.DecodeString(fsToml.Signature)
```
nikkolasg: should this signature not be verified after being decoded ? is it done somewhere else ? if so, why is it not done here ?

pop/service/api.go
```diff
@@ -0,0 +1,215 @@
...
+	err = scalar.UnmarshalBinary(buf)
+	if err != nil {
+		log.Error(err)
+		return nil
+	}
+	return scalar
+}
+
+// ToToml returns a toml-string.
+func (fs *FinalStatement) ToToml() string {
```
nikkolasg: Same for here, returning an error seems way safer.. or if reluctant, at least put a comment saying that the return value can be empty and that means an invalid statement or stg.

pop/service/api.go
nikkolasg: Again all theses functions do not return any error... I'm not a fan of assuming things are OK when it comes externally :D  One non-return-error solution is also to clearly specify in comment that the returned point can be nil if there is an error // the returned string can be empty if something went wrong.

pop/service/api.go
```diff
@@ -0,0 +1,215 @@
...
+	return string(buf.Bytes())
+}
+
+// PopDesc holds the name, date and a roster of all involved conodes.
+type PopDesc struct {
+	// Name and purpose of the party.
+	Name string
+	// DateTime of the party. It is in the following format:
+	//   YYYY-MM-DD HH:mm
+	DateTime string
```
nikkolasg: Should we provide time in UTC or stg ?

pop/service/api.go
```diff
@@ -0,0 +1,215 @@
...
+type popDescToml struct {
+	Name     string
+	DateTime string
+	Location string
+	Roster   [][]string
+}
+
+// Hash of this structure - calculated by hand instead of using network.Marshal.
+func (p *PopDesc) Hash() []byte {
+	if p == nil {
```
nikkolasg: why is this line that looks like a hack is needed ?

pop/service/api.go
ineiti: I hesitated to add a PR against 

https://github.com/dedis/onet/blob/master/crypto/key.go

is it worth it? Then we'd have all combinations of

`(Read|Write)(Point|Scalar)(Hex|64)`

pop/service/api.go
ineiti: signature creation is not done yet :( I'll add a comment.

pop/service/api.go
ineiti: `toml` uses strings and not bytes - but as it's used with file-`Open`, I'll convert it to `[]byte` anyway.

ineiti: error is cool. No need to justify ;)

pop/service/api.go
nikkolasg: I find it clearly worth it and cleaner... It should be quite fast to do and merge since you already coded the most part. If you don't have time, I can try to make a quick one.

pop/service/api.go
nikkolasg: `toml` has support for `Reader` (though it just reads all so ... it's cheating :D ), `string` and `[]byte` :) Cool thx.

pop/service/api.go
ineiti: added comment.

pop/service/api.go
ineiti: yes, should not happen (tm).

pop/service/service.go
```diff
@@ -0,0 +1,258 @@
...
+				return nil, onet.NewClientErrorCode(ErrorInternal, err.Error())
+			}
+			rep := <-s.ccChannel
+			if rep == nil {
+				return nil, onet.NewClientErrorCode(ErrorOtherFinals,
+					"Not all other conodes finalized yet")
+			}
+		}
+	}
+	s.data.Final.Signature = []byte("signed")
```
nikkolasg: `crypto/sign/{VerifySchnorr,Schnorr}` is really easy now, you think it's worth putting there ? Or did you filed an issue for that ?

pop/service/struct.go
```diff
@@ -0,0 +1,74 @@
...
+type CheckConfig struct {
+	PopHash   []byte
+	Attendees []abstract.Point
+}
+
+// CheckConfigReply sends back an integer for the Pop:
+// - 0 - no popconfig yet
+// - 1 - popconfig, but other hash
+// - 2 - popconfig with the same hash but no attendees in common
+// - 3 - popconfig with same hash and at least one attendee in common
```
nikkolasg: Do you use these differents variable somewhere? I mean do you have a 
```go
switch reply.PopStatus {
  case 1:
  case 2:
  ...
}
```
If so, it'd be nice to put constant values for those then. like `PopConfigAbsent` and `PopConfigDiffentHash` `PopConfigNoCommonAttendee` or something like that to avoid comparing hardcoded values.

pop/service/service.go
```diff
@@ -0,0 +1,258 @@
...
+	checkConfigReplyID = network.RegisterMessage(CheckConfigReply{})
+}
+
+// Service represents data needed for one pop-party.
+type Service struct {
+	// We need to embed the ServiceProcessor, so that incoming messages
+	// are correctly handled.
+	*onet.ServiceProcessor
+	path string
+	data *saveData
```
nikkolasg: Ok I'm seeing that you did an issue about that if I understand correctly ? 
https://github.com/dedis/cothority/issues/818 Is this what it is about ?

ineiti: Yes, this is thought as a first integration of Pop in the cothority. First estimate of integration-time: 1 day. Actual integration time (up to now): 2 weeks ;)

pop/service/service.go
```diff
@@ -0,0 +1,258 @@
...
+	// Public key of linked pop
+	Public abstract.Point
+	// The final statement
+	Final *FinalStatement
+}
+
+// PinRequest prints out a pin if none is given, else it verifies it has the
+// correct pin, and if so, it stores the public key as reference.
+func (s *Service) PinRequest(req *PinRequest) (network.Message, onet.ClientError) {
+	if req.Pin == "" {
```
ineiti: As we can get the server to provide it's signature to a party, we want to make sure that only legit users can do so. So we use the PIN to link the user to the server. The user has to proof he owns the server and has access to it.
This way the services authenticating the users _can_ trust the servers, instead of having to trust only a list of party-ids.

nikkolasg: > As we can get the server to provide it's signature to a party, we want to make sure that only legit users can do so

Does that imply that pop party members HAVE to be a member of a cothority ?? I did not understood that, but that sounds like a big hindrance... Anyway, let's continue off-github... :D

pop/service/service.go
ineiti: same issue as above: the signature needs to be the collective signature, which is not done yet...

pop/service/struct.go
ineiti: it's only checked to be OK in `CheckConfigReply`.

pop/service/service.go
ineiti: Did it anyway. But it's unsecure with our current implementation of the cosi-protocol.

Issue #818, opened 2017-02-03
More options for pop-party


ineiti:
In chronological order, I propose:
1. #905 - getting to know the repos
2. #907 - start coding in cothority/onet
3. #906 - more coding in cothority/onet
4. #910 - eventually do something in the frontend?
5. #911 - create a nodejs module
6. #908 - first work on the theory, then implement it
7. #909 - desert....

nikkolasg:
You should do a Epic issue for this one, looks like a big one !

ineiti:
Yes, perhaps with one of the summer-interns.

PR #819, opened 2017-02-03
Pop app - 2/2

### LINKED ISSUE #817: Port Proof-of-Personhood
The code from the 33c3-pop-party-demo should be ported back to the cothority, so it can be used later.

ineiti:
This adds the application to the pop-party. Some things are still missing, see #818 - but they will come later.
Closes #817

### DIFF
--- a/pop/app.go
+++ b/pop/app.go
@@ -1,7 +1,10 @@
 package main
 
 import (
+	"encoding/base64"
+	"errors"
 	"os"
+	"path"
 
 	"github.com/dedis/cothority/cosi/check"
 	_ "github.com/dedis/cothority/cosi/protocol"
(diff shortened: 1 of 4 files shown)


pop/app.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+	appCli.Flags = []cli.Flag{
+		cli.IntFlag{
+			Name:  "debug,d",
+			Value: 0,
+			Usage: "debug-level: 1 for terse, 5 for maximal",
+		},
+		cli.StringFlag{
+			Name:  "config,c",
+			Value: "~/.config/cothority/pop",
```
nikkolasg: Not for now, but what do you think about having some kind of utility in onet to return the "good" configuration folder ? `~/.config` is very linux, but I remember you saying that Mac is more USers/Library or stg ?

ineiti: yes, this and `solveUserHomeDir` which translates `~` to the correct home-directory.

PR #847, opened 2017-03-31
cosi doesn't hang anymore 839

### LINKED ISSUE #839: cosi-service roster-error
Now: when cosi-service receives a signature request with a roster where he is not the root, it will fail.

Wanted: also sign with a roster if we're not the first node.

ToDo: 
- write test with failing behaviour if we're not the first node
- use `GenerateNaryTreeWithRoot` instead of `GenerateBinaryTree` in `cosi/service/cosi.go:52`

christophebertrand:
closes #839

### DIFF
--- a/cosi/service/cosi_test.go
+++ b/cosi/service/cosi_test.go
@@ -22,43 +22,23 @@ func TestServiceCosi(t *testing.T) {
 	hosts, el, _ := local.GenTree(5, false)
 	defer local.CloseAll()
 
-	// Send a request to the service
-	client := NewTestClient(local)
-	msg := []byte("hello cosi service")
-	log.Lvl1("Sending request to service...")
-	res, err := client.SignatureRequest(el, msg)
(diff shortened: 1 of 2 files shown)


cosi/service/cosi_test.go
```diff
@@ -34,6 +34,33 @@ func TestServiceCosi(t *testing.T) {
...
+	defer local.CloseAll()
+
+	// Send a request to the service to the second host
+	client := NewTestClient(local)
+	msg := []byte("hello cosi service")
+	serviceReq := &SignatureRequest{
+		Roster:  el,
+		Message: msg,
+	}
+	dst := el.List[1]
```
ineiti: Can you do a `for _, dst := range el.List` and then test for all hosts? I think this would be the best test we can do for that part.

ineiti:
Don't worry about `jvss/simulation` failing - it will go away.

PR #849, opened 2017-03-31
remove jvss/simulation from testing

### LINKED ISSUE #848: jvss/simulation fails
For some tests now, `jvss/simulation` fails. As this is about to be rewritten (is it?), I propose we remove it from the tests.

ineiti:
Closes #848


nikkolasg:
What's happening with `run_conode.sh` ? Why not just adding the exception to the argument is not sufficient ?

ineiti:
Oups - thanks for the catch.

PR #853, opened 2017-04-28
made ID use consistent - 2/2

### LINKED ISSUE #324: Even more refinements dealing with IDs (uuids)
We could even do better then #320:
- [ ] consistent comparing of uuids: sometimes we use `==`, `!=` and sometimes we use 
  `uuid.Equal(uuid1, uuid2)` (or wrapper around it: `Equals`); make comparing IDs consistent
- [ ] make the testing for an empty ID more readable (see https://github.com/dedis/cothority/pull/320#discussion_r58048234), not only in production code but also also in test code

Optional:
- [ ] where needed,  we implement a `String` and an `Equals` function for the specific uuid type. The implementations are the same for all different types (as all are currently uuids). For sure, there is a more clever way to avoid repeating the code (even if its only 3 lines per function)

christophebertrand:
closes #324

### DIFF
--- a/skipchain/msgs.go
+++ b/skipchain/msgs.go
@@ -41,12 +41,12 @@ func (vId VerifierID) String() string {
 	return uuid.UUID(vId).String()
 }
 
-// Equal returns true if and only if the given VerifierID equals the current one.
-func (vId VerifierID) Equal(vId2 VerifierID) bool {
-	return uuid.Equal(uuid.UUID(vId), uuid.UUID(vId2))
+// Equal returns true if and only if the given VerifierID equals vId.
+func (vId VerifierID) Equal(vID2 VerifierID) bool {
+	return uuid.Equal(uuid.UUID(vId), uuid.UUID(vID2))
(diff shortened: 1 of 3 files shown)


skipchain/msgs.go
```diff
@@ -36,6 +36,21 @@ func init() {
...
+func (vId VerifierID) String() string {
+	return uuid.UUID(vId).String()
+}
+
+// Equal returns true if and only if the given VerifierID equals the current one.
+func (vId VerifierID) Equal(vId2 VerifierID) bool {
+	return uuid.Equal(uuid.UUID(vId), uuid.UUID(vId2))
+}
+
+//IsNil returns true iff the VerifierID is Nil
```
ineiti: `s://Is:// Is:`

ineiti:
Waiting for merge of onet#180, then I'll re-launch Travis to look if it's OK.

ineiti:
It's passing Travis now.

PR #854, opened 2017-05-10
Skipchain 835 2

ineiti:
This is the newest, shiniest skipchain. It features:

- better separation of the `api` and the service
- simplified handling with outsourcing of `chainiac`-features


ineiti:
Will be split in smaller parts

PR #856, opened 2017-05-22
Step-by-step copy of the randhound

ineiti:
This is a clean copy to master for the randhound-service.


ineiti:
The api should take some tests.

randhound/protocol/utils.go
```diff
@@ -21,9 +21,11 @@ func (rh *RandHound) newSession(nodes int, groups int, purpose string, timestamp
 	var err error
 
 	if timestamp.IsZero() {
-		timestamp = time.Now()
+		timestamp = time.Now().UTC()
```
ineiti: I think doing `time.Unix()` https://golang.org/pkg/time/#Time.Unix would be better here. Then you don't have to worry about UTC or others.

randhound/protocol/utils.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	// Process purpose string
+	if _, err := miscBuf.WriteString(purpose); err != nil {
+		return nil, err
+	}
+
+	// Process time stamp
+	timestamp = timestamp.UTC()
+	t, err := timestamp.MarshalBinary()
```
ineiti: I vote for using `time.Unix()` instead of doing a `MarshalBinary()` and `UTC()` everywhere. If you use `time.Unix()`, I think you can get rid of all `.UTC()`-calls all over the place.

ineiti:
Rebased against master and implemented `time.Unix()`

ineiti:
Merge conflicts just mean that you need to replace the files - no need to merge them.
Merge strategy: this branch overwrites master files.

PR #865, opened 2017-05-22
SkipChain Manager 2/2

### LINKED ISSUE #859: add scmgr

ineiti:
Closes #859

### DIFF
--- a/scmgr/app.go
+++ b/scmgr/app.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/dedis/cothority/skipchain"
+	"gopkg.in/dedis/onet.v1"
 	"gopkg.in/dedis/onet.v1/log"
 	"gopkg.in/dedis/onet.v1/network"
 	"gopkg.in/urfave/cli.v1"
@@ -237,7 +238,7 @@ func addWeb(c *cli.Context) error {
 		log.Fatal("Please give skipchain-id and html-file to save")
 	}
 	for i, s := range c.Args() {
-		log.Print(i, s)
+		log.Info(i, s)
 	}
 	cfg := getConfigOrFail(c)
 	sb := cfg.Sbm.GetFuzzy(c.Args().First())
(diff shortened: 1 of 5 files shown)


scmgr/app.go
```diff
@@ -0,0 +1,562 @@
...
+
+// Returns the number of calls.
+func add(c *cli.Context) error {
+	log.Info("Adding a block with a new group")
+	if c.NArg() < 2 {
+		return errors.New("Please give group-file and id to add")
+	}
+	group := readGroup(c, 1)
+	cfg := getConfigOrFail(c)
+	sb := cfg.Sbm.GetFuzzy(c.Args().First())
```
nikkolasg: the joy of naming ;)

scmgr/app.go
```diff
@@ -0,0 +1,562 @@
...
+	if sb == nil {
+		return errors.New("didn't find latest block in local store")
+	}
+	client := skipchain.NewClient()
+	guc, cerr := client.GetUpdateChain(sb.Roster, sb.Hash)
+	if cerr != nil {
+		return errors.New("while updating chain: " + cerr.Error())
+	}
+	if len(guc.Update) == 1 {
+		log.Info("No new block available")
```
nikkolasg: cant you return here since there's nothing new ?

scmgr/app.go
```diff
@@ -0,0 +1,562 @@
...
+		return errors.New("couldn't read config: " + err.Error())
+	}
+
+	// Get the list of genesis block
+	genesis := sbl{}
+	for _, sb := range cfg.Sbm.SkipBlocks {
+		if sb.Index == 0 {
+			genesis = append(genesis, sb)
+		}
+	}
```
nikkolasg: Seems like this block of code is present multiple times. Might be good to factor it in its own utility method maybe ? As you deem fit.

scmgr/app.go
```diff
@@ -0,0 +1,562 @@
...
+						if _, exists := sisAll[si.ID]; !exists {
+							log.Print("Adding", si)
+							sisNew[si.ID] = si
+							sisAll[si.ID] = si
+						}
+					}
+				}
+			}
+		}
+	}
```
nikkolasg: All of the above brings quite some conditional complexity -> 6 inner conditions. It'd be best if it were reduced to 2 or 3, but if it does not make sense (i.e. it only adds new functions only being used here), then it's fine.

scmgr/app.go
ineiti: yes, but it might be useful to have the genesis-id, and hash of the last block anyway.

PR #866, opened 2017-05-22
Skipchain 1/2

### LINKED ISSUE #858: re-structure skipchains

ineiti:
Closes #858

### DIFF
--- a/skipchain/api.go
+++ b/skipchain/api.go
@@ -7,10 +7,6 @@ import (
 	"gopkg.in/dedis/onet.v1/network"
 )
 
-// TODO:
-// - change 'Propose*' to 'Add*'
-// - find better names for Roster- and Data-SkipChain
-
 const (
 	// ErrorBlockNotFound indicates that for any number of operations the
 	// corresponding block has not been found.
--- a/skipchain/api.go
+++ b/skipchain/api.go
@@ -1,15 +1,15 @@
 package skipchain
 
 import (
-	"bytes"
-
+	"gopkg.in/dedis/crypto.v0/abstract"
 	"gopkg.in/dedis/onet.v1"
 	"gopkg.in/dedis/onet.v1/log"
 	"gopkg.in/dedis/onet.v1/network"
 )
 
-// TODO - correctly convert the BFT-signature to CoSi-Signature by removing
-//	the exception-field - has to wait for new cosi-library in crypto
+// TODO:
+// - change 'Propose*' to 'Add*'
+// - find better names for Roster- and Data-SkipChain
 
 const (
 	// ErrorBlockNotFound indicates that for any number of operations the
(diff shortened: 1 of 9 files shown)


nikkolasg:
Regarding: https://github.com/dedis/cothority/blob/3b34227870c3b393e33e802aee75a3a1c4472216/skipchain/skipchain.go#L76-L79
Why a random ID for the backlink of a genesis block ? We can't differentiate a random invalid  backlink from a "nil" backlink (i.e. belong to genesis block) from a good backling in this case. Is it possible to just set `nil` as done for the genesisID field in the last line ?

skipchain/api.go
```diff
@@ -40,146 +40,132 @@ func NewClient() *Client {
...
-// reference.
-func (c *Client) CreateRootControl(elRoot, elControl *onet.Roster, baseHeight, maxHRoot, maxHControl int, ver VerifierID) (root, control *SkipBlock, cerr onet.ClientError) {
-	log.Lvl2("Creating root roster", elRoot)
-	root, cerr = c.CreateRoster(elRoot, baseHeight, maxHRoot, ver, nil)
-	if cerr != nil {
-		return
+// StoreSkipBlock asks the cothority to store the new skipblock, and eventually
+// attach it to the 'latest' skipblock.
+//  - latest is the skipblock where the new skipblock is appended. If el and d
+//   are nil, a new skipchain will be created with 'latest' as genesis-block.
```
nikkolasg: i find it confusing that
1) `StoreSkipBlock` accepts a nil `latest` and creates a genesis block in that case
2) `CreateGenesis` is specialized in doing only that, with all the important parameters to decide at creation-time.

Which one to choose ? Seems 2) is better for creating a skipchain, so maybe 1) should not allow creating a skipchain, and return an error in that case ?

skipchain/api_test.go
```diff
@@ -16,8 +18,31 @@ func init() {
 	network.RegisterMessage(&testData{})
 }
 
-func TestClient_ProposeSkipBlock(t *testing.T) {
+func TestClient_CreateGenesis(t *testing.T) {
+	l := onet.NewTCPTest()
```
nikkolasg: Damn we still require a TCP test because of the skipchain client ... erf.

skipchain/skipchain.go
```diff
@@ -37,44 +37,79 @@ const skipblocksID = "skipblocks"
...
+// added.
+func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, onet.ClientError) {
+	prop := psbd.NewBlock
 	var prev *SkipBlock
+	var changed []*SkipBlock
 
-	if !psbd.LatestID.IsNull() {
+	if psbd.LatestID.IsNull() {
+		// A new chain is created, suppose all arguments in SkipBlock
+		// are correctly set up
```
nikkolasg: What are these arguments and most importantly, what is an incorrectly set up argument ? what does it mean ? I would prefer relying the less possible on any client "correct" setup as possible. Can the server do the verification here ? Is there any verification possible ?

skipchain/skipchain.go
```diff
@@ -37,44 +37,79 @@ const skipblocksID = "skipblocks"
...
+		prop.Index = 0
+		prop.Height = prop.MaximumHeight
+		prop.ForwardLink = make([]*BlockLink, 0)
+		// genesis block has a random back-link:
+		bl := random.Bytes(32, random.Stream)
```
nikkolasg: same comment as before for the random backlink.

skipchain/skipchain.go
```diff
@@ -144,331 +173,427 @@ func (s *Service) ProposeSkipBlock(psbd *ProposeSkipBlock) (network.Message, one
...
+		delete(s.blockRequests, string(unknown))
+		return nil, errors.New("Couldn't get updated block in time: " + unknown.Short())
+	}
+	return block, nil
+}
+
+// forwardSignature receives a signature request of a newly accepted block.
+// It only needs the 2nd-newest block and the forward-link.
+func (s *Service) forwardSignature(env *network.Envelope) {
+	err := func() error {
```
nikkolasg: Can this func be removed ?

skipchain/skipchain.go
```diff
@@ -144,331 +173,427 @@ func (s *Service) ProposeSkipBlock(psbd *ProposeSkipBlock) (network.Message, one
...
+	id := s.Sbm.Store(gbr.SkipBlock)
+	s.save()
+	log.Lvl3("Sending block to channel")
+	s.blockRequests[string(id)] <- gbr.SkipBlock
+}
+
+// verifyFollowBlock makes sure that a signature-request for a forward-link
+// is valid.
+func (s *Service) bftVerifyFollowBlock(msg []byte, data []byte) bool {
+	err := func() error {
```
nikkolasg: Can it be removed ?

skipchain/skipchain.go
```diff
@@ -144,331 +173,427 @@ func (s *Service) ProposeSkipBlock(psbd *ProposeSkipBlock) (network.Message, one
...
 }
 
-// VerifyShardFunc makes sure that the cothority of the child-skipchain is
-// part of the root-cothority.
-func (s *Service) VerifyShardFunc(msg []byte, sb *SkipBlock) bool {
-	if sb.ParentBlockID.IsNull() {
-		log.Lvl3("No parent skipblock to verify against")
+// verifyNewBlock makes sure that a signature-request for a forward-link
+// is valid.
+func (s *Service) bftVerifyNewBlock(msg []byte, data []byte) bool {
```
nikkolasg: It's exactly the same as `bftVerifyBlock` ?

skipchain/api.go
ineiti: `StoreSkipBlock` is the actual message being sent to the service. `CreateGenesis` is a convenience-method that handles some of the set up of the skipblock, and then calls `StoreSkipBlock`.

skipchain/api.go
```diff
@@ -40,146 +40,132 @@ func NewClient() *Client {
...
+//  - el is the new roster for that block. If el is nil, the previous roster
+//   will be used.
+//  - d is the data for the new block. It can be nil. If it is not of type
+//   []byte, it will be marshalled using `network.Marshal`.
+func (c *Client) StoreSkipBlock(latest *SkipBlock, el *onet.Roster, d network.Message) (reply *StoreSkipBlockReply, cerr onet.ClientError) {
```
ineiti: `v2` doesn't have a `latest` anymore:

https://github.com/dedis/cothority/blob/skipchain_835_2/skipchain/msgs.go#L30

But it's very useful to have the `latest`, because it gives you the previous roster, so you know whom to contact:

https://github.com/dedis/cothority/blob/skipchain_835_2/skipchain/api.go#L140

skipchain/api.go
```diff
@@ -40,146 +40,132 @@ func NewClient() *Client {
...
+	if el == nil && d == nil {
+		newBlock = latest
+	} else {
+		newBlock = latest.Copy()
+		if el != nil {
+			newBlock.Roster = el
+		}
+		if d != nil {
+			var ok bool
+			newBlock.Data, ok = d.([]byte)
```
ineiti: if it's a `[]byte` I don't do `network.Marshal`. Else I do. So I need the `ok` here.

skipchain/skipchain.go
ineiti: Changed comment - there is a verification some lines later:

```
		err := s.verifyBlock(prop)
```

skipchain/skipchain.go
ineiti: The random-backlink is used here to create a random hash. Without it two genesis-blocks with the same parameters on the same roster would have the same hash, and thus the same ID.
We can differentiate them by the `Index`, which is `0` for a genesis-block.

skipchain/skipchain.go
ineiti: No, it's the function that adds a forward-link for a future block.

skipchain/skipchain.go
ineiti: No, when signing a forward-link, we need to have consensus that it's OK to do so.

skipchain/skipchain.go
ineiti: I'll have to have a closer look tomorrow...

skipchain/skipchain.go
nikkolasg: I was talking about the `err := func() error }` thing. Why not writing the code inside the top level method ?

skipchain/skipchain.go
ineiti: Both are needed for the moment. I opened an issue for the v2-skipchains:

https://github.com/dedis/cothority/issues/870

skipchain/skipchain.go
ineiti: oh, sorry. The advantage here is that you can have only one `log.Error`, instead of having to `log.Error(); return` for every if-clause that you write.

PR #868, opened 2017-05-22
Propagate: error-handling and longer wait

ineiti:
Small cleanup

### DIFF
--- a/messaging/propagate.go
+++ b/messaging/propagate.go
@@ -1,6 +1,7 @@
 package messaging
 
 import (
+	"errors"
 	"sync"
 
 	"time"
@@ -73,7 +74,7 @@ type propagationContext interface {
 func NewPropagationFunc(c propagationContext, name string, f PropagationStore) (PropagationFunc, error) {
 	pid, err := c.ProtocolRegister(name, func(n *onet.TreeNodeInstance) (onet.ProtocolInstance, error) {
 		p := &Propagate{
-			sd:               &PropagateSendData{[]byte{}, 1000},
+			sd:               &PropagateSendData{[]byte{}, 100000},
 			TreeNodeInstance: n,
 			received:         0,
 			subtreeCount:     n.TreeNode().SubtreeCount(),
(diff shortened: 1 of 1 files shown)


messaging/propagate.go
```diff
@@ -73,7 +74,7 @@ type propagationContext interface {
 func NewPropagationFunc(c propagationContext, name string, f PropagationStore) (PropagationFunc, error) {
 	pid, err := c.ProtocolRegister(name, func(n *onet.TreeNodeInstance) (onet.ProtocolInstance, error) {
 		p := &Propagate{
-			sd:               &PropagateSendData{[]byte{}, 1000},
+			sd:               &PropagateSendData{[]byte{}, 100000},
```
nikkolasg: Can't this be a global variable ? Looks like a good fit (for once). Otherwise one can add an arguement to take the timeout but i don't think that's necessary. What do you think ?

ineiti: The timeout can be chosen by the caller, but it's not available upon creation of the protocol :( See https://github.com/dedis/cothority/pull/868/files/3d0b2cf919271a2456e5574e87d891cb831e9713#diff-1de3dc5ac1cd86438180ef0429247212R147

So it's variable, but we might have to wait to know what the user wants.

messaging/propagate.go
```diff
@@ -17,6 +18,9 @@ func init() {
 	network.RegisterMessage(PropagateReply{})
 }
 
+// How long to wait before timing out on waiting for the time-out.
+const initialWait = 100000
```
nikkolasg: 1)
> before timing out on waiting for the time-out.

Maybe just `before timeout` ?

2) What do you think about putting that as a exported variable instead ? Just proposing here but IIRC we had so many experiments requirings so many different timeouts because of long setup, overloaded nodes, etc that it may be useful to have it exported. If you don't think so, then I'll merge it like this, I just wanted to raise the point.

ineiti: It jokingly describes the bad behaviour of the protocol: instead of correctly waiting for the chosen timeout by the user, which comes through 
```
type PropagationFunc func(el *onet.Roster, msg network.Message, msec int) (int, error)
```
This is in fact a timeout _before_ the chosen time-out is set. Probably would need to be re-designed, so that it doesn't need to have a hard-coded timeout before waiting for the chosen timeout.

So instead of patching this with an exported variable, I'd keep it like this. It's bad and should be re-designed...

PR #881, opened 2017-05-24
Cisc qrcode implementation

ineiti:
Adds the possibility to print a qrcode of the current cisc-configuration.

### DIFF
--- a/cisc/commands.go
+++ b/cisc/commands.go
@@ -49,7 +49,7 @@ func init() {
 			{
 				Name:    "qrcode",
 				Aliases: []string{"qr"},
-				Usage:   "print out the qrcode of the identity",
+				Usage:   "print out the qrcode of the identity-skipchain and a node for contact",
 				Action:  idQrcode,
 			},
 		},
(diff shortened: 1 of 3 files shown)


cisc/commands.go
```diff
@@ -46,6 +46,12 @@ func init() {
 				Usage:   "check the health of the cothority",
 				Action:  idCheck,
 			},
+			{
+				Name:    "qrcode",
+				Aliases: []string{"qr"},
+				Usage:   "print out the qrcode of the identity",
```
nikkolasg: the identity ? isn't it the configuration ? Or do I misunderstand ?
We should start to have a glossary maybe regarding the different skipchain's usage: "server identity", "configuration", etc

ineiti: good idea - I'll add a README.md to the `identity`-directory.

The QRCode is used to connect a phone to the identity-skipchain. It contains an IP:Port and the skipchain-id where the identity is stored.

nikkolasg: Ok so I would at least put it `the qrcode of the identity skipchain` then.

PR #884, opened 2017-05-24
run_conode improvement

ineiti:
adding wait-flag to run_conode.sh, so it can be used more easily with `screen`.

### DIFF
--- a/conode/run_conode.sh
+++ b/conode/run_conode.sh
@@ -84,7 +84,28 @@ runLocal(){
 		exit 1
 	fi
 	NBR=$1
-	DEBUG=${2:-1}
+	shift
+	WAIT=""
+	DEBUG=1
+	while [ "$1" ]; do
+		case $1 in
+		-update)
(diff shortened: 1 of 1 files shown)


nikkolasg:
I'm sorry but I have trouble understanding the purpose of that change. Why would you want to wait ?

ineiti:
I start having scripts that look like:

```bash
#!/bin/bash
export GOPATH=~
pkill -f conode
screen -S conode -d -m ~/cothority/conode/run_conode.sh local 64 3 -wait
```

Without the `-wait`, `screen` exits and terminates all conodes. Ideally there would be a `-deamon`-flag to `run_conode.sh`, but `-wait` was simpler.

ineiti:
Now that I use the `run_conode.sh -local`-command for setting up _real_ conodes, I should perhaps change the whole interface. Oh my...

ineiti:
This can be merged - new issue in #888

PR #885, opened 2017-05-24
Identity 666

### LINKED ISSUE #666: Fix Identity skipchain
Dangerous issue
- [x] Change `config` to `data` in identity-skipchain
- [x] Include votes of devices into skipblock and verify correct votes in counting
- [x] Add a correct verification-method with `RegisterVerification`

ineiti:
Closes #666

### DIFF
--- a/identity/api_test.go
+++ b/identity/api_test.go
@@ -253,7 +253,7 @@ func TestVerificationFunction(t *testing.T) {
 	data2.Votes["two2"] = &sig
 	id := s0.getIdentityStorage(c1.ID)
 	require.NotNil(t, id, "Didn't find identity")
-	_, cerr := s0.skipchain.ProposeData(id.SCRoot, id.SCData, data2)
+	_, cerr := s0.skipchain.StoreSkipBlock(id.SCData, nil, data2)
 	require.NotNil(t, cerr, "Skipchain accepted our fake block!")
 
 	// Unhack: verify that the correct way of doing it works, even if
@@ -261,7 +261,7 @@ func TestVerificationFunction(t *testing.T) {
 	sig, err = crypto.SignSchnorr(network.Suite, c1.Private, hash)
 	log.ErrFatal(err)
 	data2.Votes["one1"] = &sig
-	_, cerr = s0.skipchain.ProposeData(id.SCRoot, id.SCData, data2)
+	_, cerr = s0.skipchain.StoreSkipBlock(id.SCData, nil, data2)
 	log.ErrFatal(err)
 	log.ErrFatal(c1.DataUpdate())
 
--- a/identity/api_test.go
+++ b/identity/api_test.go
@@ -26,10 +26,10 @@ func TestIdentity_ConfigNewCheck(t *testing.T) {
 	c1 := NewIdentity(el, 50, "one")
 	log.ErrFatal(c1.CreateIdentity())
 
-	conf2 := c1.Config.Copy()
+	conf2 := c1.Data.Copy()
 	kp2 := config.NewKeyPair(network.Suite)
 	conf2.Device["two"] = &Device{kp2.Public}
-	conf2.Data["two"] = "public2"
+	conf2.Storage["two"] = "public2"
 	log.ErrFatal(c1.ProposeSend(conf2))
 
 	log.ErrFatal(c1.ProposeUpdate())
@@ -39,7 +39,7 @@ func TestIdentity_ConfigNewCheck(t *testing.T) {
 	o2, ok := al.Device["two"]
 	assert.True(t, ok)
 	assert.True(t, kp2.Public.Equal(o2.Point))
-	pub2, ok := al.Data["two"]
+	pub2, ok := al.Storage["two"]
 	assert.True(t, ok)
 	assert.Equal(t, "public2", pub2)
 }
@@ -78,10 +78,10 @@ func TestIdentity_ConfigUpdate(t *testing.T) {
 
 	c2 := NewTestIdentity(el, 50, "two", l)
 	c2.ID = c1.ID
-	log.ErrFatal(c2.ConfigUpdate())
+	log.ErrFatal(c2.DataUpdate())
 
-	assert.NotNil(t, c2.Config)
-	o1 := c2.Config.Device[c1.DeviceName]
+	assert.NotNil(t, c2.Data)
+	o1 := c2.Data.Device[c1.DeviceName]
 	if !o1.Point.Equal(c1.Public) {
 		t.Fatal("Owner is not c1")
 	}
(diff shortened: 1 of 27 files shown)


identity/service.go
```diff
@@ -69,64 +75,80 @@ type Storage struct {
...
-	log.Lvl3(s, "Sending config-update")
-	return &ConfigUpdateReply{
-		Config: sid.Latest,
+	reply, cerr := s.skipchain.GetUpdateChain(sid.SCRoot.Roster, sid.SCData.Hash)
+	if cerr != nil {
+		return nil, cerr
+	}
+	if len(reply.Update) > 1 {
+		log.Lvl3("Got new data")
+		// TODO: check that update-chain is legit
```
nikkolasg: What is legit ?

identity/service.go
```diff
@@ -69,64 +75,80 @@ type Storage struct {
...
+		sid.SCData = reply.Update[len(reply.Update)-1]
+		_, dataInt, err := network.Unmarshal(sid.SCData.Data)
+		if err != nil {
+			return nil, onet.NewClientErrorCode(ErrorDataMissing, err.Error())
+		}
+		sid.Latest = dataInt.(*Data)
```
nikkolasg: Can I make a special block with a non `Data` data inside that will make this line panic ?

identity/service.go
```diff
@@ -195,43 +221,108 @@ func (s *Service) ProposeVote(v *ProposeVote) (network.Message, onet.ClientError
...
+	err := func() error {
+		if sb.Index == 0 {
+			log.Lvl4("Always accepting genesis-block")
+			return nil
+		}
+		_, dataInt, err := network.Unmarshal(sb.Data)
+		if err != nil {
+			return errors.New("got unknown packet")
+		}
+		data, ok := dataInt.(*Data)
```
nikkolasg: again for the panic

identity/service.go
```diff
@@ -195,43 +221,108 @@ func (s *Service) ProposeVote(v *ProposeVote) (network.Message, onet.ClientError
...
+			if pub := dataLatest.Device[dev]; pub != nil {
+				if err := crypto.VerifySchnorr(network.Suite, pub.Point, hash, *sig); err != nil {
+					return err
+				}
+				sigCnt++
+			} else {
+				log.Lvl2("Not representative signature detected:", dev)
+			}
+		}
+		if sigCnt >= dataLatest.Threshold || sigCnt == len(dataLatest.Device) {
```
nikkolasg: For this whole section, it seems possible that if I sucessfully create one valid signature from one valid device, then I can just copy paste it enough time so that the verification pass. Is it possible ? 
If yes, we should add a way to distinct between different signers...

identity/service.go
ineiti: all forward-links are valid.

ineiti: In the `v2`-code, the chains are handled in `Bunch`es that check for every new block that at least one forward-link is valid.

identity/service.go
ineiti: A rogue `skipchain`-node might do this, yes. But not a user.

identity/service.go
ineiti: `ok` checks if it's really a `Data`-block. The `network.Unmarshal` will return an initialised block anyway, so I don't see how I could get a `nil` here. Perhaps I'm missing something?

identity/service.go
ineiti: No, because in line 299 we take the `pub` of the corresponding device and use that to verify the Schnorr-signature. So if you copy/paste, you won't have the signature corresponding to that device.

identity/service.go
nikkolasg: Not nil, but it can be any other type than `Data` if you don't check the type, or at least with an `..., ok`.

identity/service.go
nikkolasg: So it will be resolved in v2 then, ok .

identity/service.go
nikkolasg: Then it should be checked I believe. We are working on decentralization so we must allow for rogue nodes ;)

identity/service.go
```diff
@@ -195,43 +225,108 @@ func (s *Service) ProposeVote(v *ProposeVote) (network.Message, onet.ClientError
...
+			return errors.New("No backlinks stored")
+		}
+		s.identitiesMutex.Lock()
+		defer s.identitiesMutex.Unlock()
+		var latest *skipchain.SkipBlock
+		for _, id := range s.Identities {
+			if id.SCData.Hash.Equal(sb.BackLinkIDs[0]) {
+				latest = id.SCData
+			}
+		}
```
nikkolasg: If I understand correctly, this means that for each block to verify, each node iterates over all its blocks to find the backlink ? 
Seems trouble for scalability, we might want to change that for v2 at least... Or maybe I did get it wrong ;)

identity/service.go
nikkolasg: But from what I read, it seems still possible that if there is 10  times the same signature in `data.Votes` from a device which is known before, so present in `dataLatest.Devices`, then the signature count `sigCnt` will be 10. No ?
In that case, it's a problem since an attacker only has to compromise one device now...

identity/service.go
ineiti: I just verified, and indeed, there is a `// TODO - verify new config` in there ;)

identity/service.go
ineiti: Done

identity/service.go
ineiti: That's what is done here.

identity/api_test.go
```diff
@@ -194,36 +195,81 @@ func TestCrashAfterRevocation(t *testing.T) {
...
+	sig, err = crypto.SignSchnorr(network.Suite, c1.Private, hash)
+	log.ErrFatal(err)
+	data2.Votes["one1"] = &sig
+	_, cerr = s0.skipchain.StoreSkipBlock(id.SCData, nil, data2)
+	log.ErrFatal(err)
+	log.ErrFatal(c1.DataUpdate())
+
+	if len(c1.Data.Device) != 2 {
+		t.Fatal("Should have two owners now")
+	}
```
nikkolasg: If not too much, just adding a test where there is a wrong signature, like random bytes, would be nice :)

identity/service.go
ineiti: `data.Votes` is a map, so you cannot put more than one vote for the same device in it.

nikkolasg: Damn right. I should not work on Monday.

identity/service.go
ineiti: With the `Bunch`es, it's only a map-lookup.

identity/api_test.go
ineiti: It's not a random test now, but a wrong signature.

identity/service.go
ineiti: Well, 1 bad comment out of 20 is not that bad ;)

nikkolasg:
Ok I wait for travis then I merge.

PR #895, opened 2017-05-31
Shiny new skipchain

### LINKED ISSUE #894: New skipchain-methods
Simplify the skipchains further and break compatibility with v1.1 and student-code from EPFL 2017-spring semester.

ineiti:
Closes #894 

Will be PRed against `v2`, using master for now till `kyber` is correctly merged.


lbarman:
Are there stuff that we need to keep in here (regarding our discussion of this morning where we agreed for a new start) ?
Otherwise let's close this and work here https://github.com/dedis/cothority/tree/skipchain-v2.0

PR #898, opened 2017-06-28
adding interface to RegisterVerification

ineiti:
`RegisterVerification` in skipchains accepted only a `Context`, so that it can only be called from another service. But it makes also sense to register a verification-function outside of a service, so the `Context` has been replaced by an interface accepting anything that can return the service.


nikkolasg:
> But it makes also sense to register a verification-function outside of a service
Can you provide some more details about why it makes sense, what's the use case ?

ineiti:
> Can you provide some more details about why it makes sense, what's the use case ?

One use case is to define the verification-function in the `conode.go`, so you can create a simple skipchain with a given verification-function, without having to write a service for it.

nikkolasg:
Sorry to ask but I wanna make sure I understand before I merge:
Why do you want to write a service for creating a simple skipchain. Isn't the `skipchain` service already supposed to take care of that ? Is it not just enough to have a way for the `skipchain` service to call a callback function that say `yes/no` on any new block ?

ineiti:
The previous `RegisterVerification`-method took a `Context` to find the `skipchain.Service`-instance. But the `Context` is only available in the service, something Jeff pointed out during the code-challenge ;)

As you say, the verification-function is given a skipblock and needs to return true/false, so no need for a service. But without a service, you don't have access to a `Context`. The other structure giving access to the services is a `Server`, also available in `conode.go`. With my previous patch on onet, `Server.Service` exists now, and we can extend the `RegisterVerification` to accept either a `Context` or a `Server`, as both implement the `Service`-method.

As a sidenote: I just read http://blog.jessitron.com/2017/06/the-most-productive-circumstances-for.html and feel terribly guilty. So ask all the questions you want! Please!

nikkolasg:
mhhh I see. So the dependency graph is quite complex here too. To register a function, you need to pass something that can return any service, so it can cast to the right service to call a function on that service. The ideal solution would have been just to call the right function on the skipchain service directly since any caller has access to the service. This is related to the Service API, inter-service communication etc... Anyway, ok, LGTM.

PR #899, opened 2017-06-29
Skipchain goroutines

ineiti:
The skipchains to asynchronous updating of the forward-links and propagation of the new blocks. This is handy and good for performance (you don't have to wait for all forward-signatures and propagation before being able to go on), but in the tests it's a nuisance, as the go-routines are still active.

This patch adds a counter to the skipchain-service, so that we can test if it's done propagating.


nikkolasg:
In `messaging/propagate.go`:
>  PropagationFunc starts the propagation protocol and blocks until
 all children stored the new value or the timeout has been reached.
 The return value is the number of nodes that acknowledged having
 stored the new value or an error if the protocol couldn't start.

But then in this PR, you do:
```go
s.propagatingMutex.Lock()
s.propagating++
replies, err := s.propagate(roster, &PropagateSkipBlocks{blocks}, propagateTimeout)
s.propagating--
s.propagatingMutex.Unlock()
```
Since the `s.propagate` is supposed to be blocking, I don't see the point of doing all of this. Why is propagate not blocking ? i.e. is it already running in a go routine ? 

Moreover, I'm not sure about the approach of `lock, increment, doStuff, decrement, unlock`: it's essentially doing nothing since it holds the lock all the time the propagate function is running. The `IsPropagating` is supposed to always return 0 since you always decrement before releasing the mutex...

Enlighten me please ;)

ineiti:
the `Service.startPropagation` is called in `Service.forwardSignature`. The latter is a `ProcessorFunc` and as such it is called in a go-routine.

What is questionable is the locking over the whole `s.propagate`, as it makes the `s.propagation`-variable `0` or `1`. So I could simply remove the counter, and rename `IsPropagating` to `WaitPropagationDone` which would try to take the mutex and block if it is propagating. But I'm not sure it will behave in a useful way, as there might be multiple forward-signing requests waiting, and then one node might think it's done, but then get another request. So my `waitPropagationFinished`-function will not be able to restart checking all nodes in case one was waiting.

So probably this would be more correct, but it looks even worse:

```
s.propagatingMutex.Lock()
s.propagating++
s.propagatingMutex.Unlock()
replies, err := s.propagate(roster, &PropagateSkipBlocks{blocks}, propagateTimeout)
s.propagatingMutex.Lock()
s.propagating--
s.propagatingMutex.Unlock()
```

Oh, the joys of distributed multi-threaded programming.

ineiti:
Where is travis?

nikkolasg:
Ok I now understand better. So from what I gather, the only problem here is we have no guarantee about the finished execution of `service.forwardSignature`. Is there not a blocking call when calling `forwardSignature` on the client that returns only when all blocks are propagated ?

Anyway, it makes more sense with the `unlocking` in the middle now.

ineiti:
it's here:

```
			if err := s.SendRaw(back.Roster.RandomServerIdentity(),
				&ForwardSignature{i + 1, prev.Hash, prop,
					prev.GetForward(0)}); err != nil {
				// This is not a critical failure - we have at least
				// one forward-link
				log.Error("Couldn't get old block to sign")
			} else {
				changed = append(changed, back)
			}
```

So no blocking call.

nikkolasg:
Travis is not there due to the playground stuff in makefile no ?

ineiti:
the `make test_playground` is for easier testing of part of the repo - instead of doing format-checking, lint-checking and testing the whole repo, it loops 10 times over one part - here the skipchain-tests. This is one of the old travis-hacks we used some time ago - remember those spurious network-errors? Well, `test_playground` was there to help ;)

I think Travis was overloaded. Even though the travis-status did not show anything. But all commits are checked now by Travis.

Travis points now to `make test` again, and I fixed the merge-error.

PR #902, opened 2017-07-11
Skipchain search

ineiti:
Sometimes we need to be able to search by the index of the block. This is a linear implementation.

### DIFF
--- a/skipchain/api.go
+++ b/skipchain/api.go
@@ -165,3 +165,12 @@ func (c *Client) GetSingleBlock(roster *onet.Roster, id SkipBlockID) (reply *Ski
 		&GetSingleBlock{id}, reply)
 	return
 }
+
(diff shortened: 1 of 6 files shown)


skipchain/api.go
```diff
@@ -165,3 +165,12 @@ func (c *Client) GetSingleBlock(roster *onet.Roster, id SkipBlockID) (reply *Ski
 		&GetSingleBlock{id}, reply)
 	return
 }
+
+// GetSingleBlockByIndex searches for a block with the given index following the genesis-block.
```
nikkolasg: what is the index of a block ? I thought we referenced blocks by their hash.

ineiti: It's what is called the `height` in bitcoin - the number of the block.

nikkolasg: So a block can be referenced by its number (increasing id ?) in the chain AND/OR by its hash ?

PR #903, opened 2017-07-11
Android app patches

ineiti:
Small patches

### DIFF
--- a/identity/service.go
+++ b/identity/service.go
@@ -351,7 +351,6 @@ func (s *Service) propagateDataHandler(msg network.Message) {
 		case *ProposeSend:
 			p := msg.(*ProposeSend)
 			sid.Proposed = p.Propose
-			sid.Proposed.Votes = make(map[string]*crypto.SchnorrSig)
 		case *ProposeVote:
 			v := msg.(*ProposeVote)
 			d := sid.Latest.Device[v.Signer]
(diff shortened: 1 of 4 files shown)


identity/service.go
```diff
@@ -351,9 +351,7 @@ func (s *Service) propagateDataHandler(msg network.Message) {
 		case *ProposeSend:
 			p := msg.(*ProposeSend)
 			sid.Proposed = p.Propose
-			if len(sid.Proposed.Votes) == 0 {
-				sid.Proposed.Votes = map[string]*crypto.SchnorrSig{}
-			}
+			sid.Proposed.Votes = make(map[string]*crypto.SchnorrSig)
```
nikkolasg: what is the reason you removed the `if` ? Do you *always* setup the votes by default when receiving a `ProposeSend` ?

ineiti: Dang - now I'm not sure anymore. It could be advantageous for the client to send the proposition directly with his vote - one less message. I erroneously thought we need to ensure no signatures on proposition `n` validate a proposition `n+1`, but that cannot happen, as it's signatures. I'll change it back to the if, as there is no attack possible.

ineiti: In fact it can be removed altogether, as the `Proposed.Votes` is only used when a `ProposeVote` is received.

PR #914, opened 2017-07-19
adding a newblocks list to indicate blocks that are about to be signed

### LINKED ISSUE #913: Need blocking of skipblocks during BFTCosi
When two clients ask to add a new block in parallel, skipblock will fail:

Given 3 nodes (node0..node2) and a block A
t0: Client1 asks node0 to add a block B to A
t0: Client2 asks simultaneously node5 to add a block C to A
t1: node0 asks nodes1..2 to check validity of block B
t1: node2 asks nodes0..1 to check validity of block C and create collective signature
t2: nodes0..2 approve block B and create collective signature, sending it to node0
t2: nodes0..2 approve block C and create collective signature, sending it to node2
t3: node0 propagates block A, B through system
t3: node2 propagates block A, C through system
t4: node1 wants to add block D to block A, but this will fail, because this block will not be approved, due to the presence of B or C.

At t2 we created a fork in our beloved skipchain, because we collectively signed two different future blocks :( 
My first idea is to use a `signatureinprogress`-list of blocks that are about to be signed, and then look up this list in the BFT-verification function. But this list is quite brittle, as we have to take care to remove failed signature attempts from it.

ineiti:
Closes #913

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -321,7 +321,6 @@ func (s *Service) forwardSignature(env *network.Envelope) {
 			return errors.New("Couldn't get signature")
 		}
 		s.Sbm.Lock()
-		log.Lvl1("Adding forward-link to", target.Index)
 		target.AddForward(&BlockLink{fs.ForwardLink.Hash, sig.Sig})
 		s.Sbm.Unlock()
 		s.startPropagation([]*SkipBlock{target})
@@ -468,7 +467,7 @@ func (s *Service) bftVerifyNewBlock(msg []byte, data []byte) bool {
 		return true
 	}()
 	if !ok {
-		s.newBlocks.rmBlock(prevSB.Hash, newSB.Hash)
+		s.newBlocks.rmBlock(prevSB.Hash)
 	}
 	return ok
 }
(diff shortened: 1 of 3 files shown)


skipchain/skipchain.go
```diff
@@ -418,47 +423,53 @@ func (s *Service) bftVerifyFollowBlock(msg []byte, data []byte) bool {
...
+			if !ok {
+				log.Lvlf2("Found no user verification for %x", ver)
+				return false
+			}
+			if !f(msg, newSB) {
+				return false
+			}
 		}
+		return true
+	}()
```
nikkolasg: You like thoses anonymous functions :D  But here it can be made a bit simpler still in my opinion.
```go
ok := true 
 for _, ver := range newSB.VerifierIDs {
        f, ok := s.verifiers[ver]
 	if !ok || !f(msg, newSB) {
                ok = false
                break
 	}
  }
```

ineiti: it's shorter, but you miss the debug-output that the user-verification is missing. Perhaps

```
	ok := true
	for _, ver := range newSB.VerifierIDs {
		f, ok := s.verifiers[ver]
		if !ok {
			log.Lvlf2("Found no user verification for %x", ver)
		} else {
			ok &= f(msg, newSB)
		}
		if !ok{
			break
		}
	}
```

ineiti:
Closing this as it's not the correct thing to do. Talking with Bryan, this should:
- only allow block-requests from the leader
- other nodes collect transactions and send them to the leader once he wants to create a block

So the most simple thing to do is to send all requests to the leader (suppose it's the first entry of the `Roster` for the moment) and let the leader reply with an error if a request is going on.

PR #915, opened 2017-08-02
removed AfterTest calls

ineiti:
Leftovers from previous testing.


nikkolasg:
Seems like "btcsuite" has broken API or stg ? From Travis:
>../../btcsuite/btcutil/block.go:82: b.msgBlock.SerializeNoWitness undefined (type *wire.MsgBlock has no field or method SerializeNoWitness)

ineiti:
Yes, this should be fixed in #917

PR #916, opened 2017-08-02
better blocking on new requests

### LINKED ISSUE #913: Need blocking of skipblocks during BFTCosi

ineiti:
Closes #913

### DIFF
--- a/skipchain/api.go
+++ b/skipchain/api.go
@@ -23,6 +23,9 @@ const (
 	ErrorVerification
 	// ErrorOnet indicates an error from the onet framework
 	ErrorOnet
+	// ErrorBlockInProgress indicates that currently a block is being formed
+	// and propagated
+	ErrorBlockInProgress
 )
 
 // Client is a structure to communicate with the Skipchain
@@ -69,7 +72,7 @@ func (c *Client) StoreSkipBlock(latest *SkipBlock, el *onet.Roster, d network.Me
 		}
 		latestID = latest.Hash
 	}
-	host := latest.Roster.RandomServerIdentity()
+	host := latest.Roster.Get(0)
 	reply = &StoreSkipBlockReply{}
 	cerr = c.SendProtobuf(host, &StoreSkipBlock{latestID, newBlock}, reply)
 	if cerr != nil {
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -650,7 +650,6 @@ func (s *Service) newBlockEnd(sb *SkipBlock) bool {
 	if _, processing := s.newBlocks[string(sb.Hash)]; !processing {
 		return false
 	}
-	network.Suite.Scalar().Zero()
 	delete(s.newBlocks, string(sb.Hash))
 	return true
 }
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -40,13 +40,13 @@ const skipblocksID = "skipblocks"
 type Service struct {
 	*onet.ServiceProcessor
 	// Sbm is the skipblock-map that holds all known skipblocks to this service.
-	Sbm              *SkipBlockMap
-	propagate        messaging.PropagationFunc
-	verifiers        map[VerifierID]SkipBlockVerifier
-	blockRequests    map[string]chan *SkipBlock
-	lastSave         time.Time
-	propagating      int64
-	propagatingMutex sync.Mutex
+	Sbm           *SkipBlockMap
+	propagate     messaging.PropagationFunc
+	verifiers     map[VerifierID]SkipBlockVerifier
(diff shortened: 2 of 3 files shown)


skipchain/skipchain.go
```diff
@@ -623,6 +634,27 @@ func (s *Service) startPropagation(blocks []*SkipBlock) error {
...
+	return true
+}
+
+func (s *Service) newBlockEnd(sb *SkipBlock) bool {
+	s.newBlocksLock.Lock()
+	defer s.newBlocksLock.Unlock()
+	if _, processing := s.newBlocks[string(sb.Hash)]; !processing {
+		return false
+	}
+	network.Suite.Scalar().Zero()
```
nikkolasg: I guess that can be removed

skipchain/api.go
```diff
@@ -23,6 +23,9 @@ const (
 	ErrorVerification
 	// ErrorOnet indicates an error from the onet framework
 	ErrorOnet
+	// ErrorBlockInProgress indicates that currently a block is being formed
+	// and propagated
+	ErrorBlockInProgress
```
nikkolasg: Also indicates somewhere that change of protocol that only the leader can initiate requests for a new block to be signed, that's why this error exists in the first place too no ?

skipchain/skipchain.go
ineiti: Oups - abused IntelliJ to look up the krypto-library ;)

skipchain/api.go
ineiti: Not only that, but also if a block is begin formed and you want to add a block at the same index as the one being formed.

skipchain/skipchain.go
```diff
@@ -107,6 +112,15 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, on
...
 		}
+		if len(prev.ForwardLink) > 0 {
+			return nil, onet.NewClientErrorCode(ErrorBlockContent,
+				"the latest block already has a follower")
+		}
+		if !s.newBlockStart(prev) {
+			return nil, onet.NewClientErrorCode(ErrorBlockInProgress,
+				"this skipchain-id is currently processing a block")
+		}
+		defer s.newBlockEnd(prev)
```
nikkolasg: Just to make sure: where is the check that there's no fork on the skipchain (i.e. same parent block for two distinct blocks), and is that check called after this defer ?

ineiti: It's line 119 - it checks that the leader is not adding another block to the `prev`, which is the latest ackknowledged block.

skipchain/skipchain.go
```diff
@@ -68,6 +69,10 @@ type Service struct {
 // added.
 func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, onet.ClientError) {
 	prop := psbd.NewBlock
+	if !s.ServerIdentity().Equal(prop.Roster.Get(0)) {
```
nikkolasg: Mhhh. Is this not hazardous ? 
Since the `prop.Roster.Get(0)` is coming from the client, the client could basically make anyone propose a block if he can re-arrange the Roster the way he wants, and it will still pass, as it's still the same roster ID or stg ?

ineiti: as far as I can see, this is not a problem. Changing the roster means a `view-change`. So as long as the verification-functions allow this view-change, that's OK.

nikkolasg: Does the verification function checks the ordering of the indivudual keys in the roster of a block ?

ineiti: In fact every new block can have a completely different `Roster`, so the problem is far worse ;)
This will only be better in skipchain.v2 where there will be a unified way of handling the roster. So if there are no other comments, I think this is ready to merge.

PR #918, opened 2017-08-02
uncomment btcsuite because its not go-getable

### LINKED ISSUE #917: byzcoin fails to test
There has been a change in 

github.com/btcsuite/btcutil

which makes byzcoin fail to compile :(

ineiti:
Closes #917


nikkolasg:
Maybe add a link to https://github.com/btcsuite/btcutil/issues/103 so we know why it has been commented out and that one day we can be confident to put it back (or delete it further :D )

PR #920, opened 2017-08-31
Docker conode

ineiti:
Making a docker-container for conodes.


nikkolasg:
Is it possible to have a little description on how the keys are created , where, and what needs to be stored between different executions (if we want to restart it) ?

ineiti:
Is it OK if I add the following link?

https://hub.docker.com/r/dedis/conode/

nikkolasg:
Sure that sounds great "for more info, see bla bla bla". One issue though still: can you specify that it works naturally if one gives the same command that it will re-use the same keys. Or is there something to change?

Issue #906, opened 2017-07-14
Attendee - keep more than one token


ineiti:

ineiti:
Done

PR #926, opened 2017-09-06
Add multiple parties support and merge process

### LINKED ISSUE #912: replace md5 with shasum or other
The `test.sh`-file uses `md5` to compare files, but this doesn't exist on linux (windows?) - replace `md5` with `shasum`, `cmp` or `diff` and verify it exists on both.

IlyaMarkovMipt:
* Add support multiple parties per conode
* Add support merging parties
Closes #912 #908 #906 #907

### DIFF
--- a/pop/service/service.go
+++ b/pop/service/service.go
@@ -29,33 +29,54 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-
 	"math/big"
+	"sort"
+	"strings"
+	"time"
 
-	"github.com/dedis/cothority/cosi/protocol"
+	"github.com/dedis/cothority/bftcosi"
 	"github.com/dedis/cothority/messaging"
 	"gopkg.in/dedis/crypto.v0/abstract"
 	"gopkg.in/dedis/crypto.v0/random"
 	"gopkg.in/dedis/onet.v1"
+	"gopkg.in/dedis/onet.v1/crypto"
 	"gopkg.in/dedis/onet.v1/log"
 	"gopkg.in/dedis/onet.v1/network"
 )
+
+func init() {
+	onet.RegisterNewService(Name, newService)
+	network.RegisterMessage(&saveData{})
+	checkConfigID = network.RegisterMessage(checkConfig{})
+	checkConfigReplyID = network.RegisterMessage(checkConfigReply{})
+	mergeConfigID = network.RegisterMessage(mergeConfig{})
+	mergeConfigReplyID = network.RegisterMessage(mergeConfigReply{})
+}
 
 // Name is the name to refer to the Template service from another
 // package.
 const Name = "PoPServer"
 const cfgName = "pop.bin"
-const protoCoSi = "CoSiFinal"
+const bftSignFinal = "BFTFinal"
+const bftSignMerge = "PopBFTSignMerge"
+
+const propagFinal = "PoPPropagateFinal"
+
+const timeout = 60 * time.Second
+
+// SIGSIZE size of signature
+const SIGSIZE = 64
+
+// DELIMETER in locations field in PopDesc
+// const timeout = 60 * time.Second
+const DELIMETER = "; "
 
 var checkConfigID network.MessageTypeID
 var checkConfigReplyID network.MessageTypeID
-
-func init() {
-	onet.RegisterNewService(Name, newService)
-	network.RegisterMessage(&saveData{})
-	checkConfigID = network.RegisterMessage(CheckConfig{})
-	checkConfigReplyID = network.RegisterMessage(CheckConfigReply{})
-}
+var mergeConfigID network.MessageTypeID
+var mergeConfigReplyID network.MessageTypeID
+var mergeCheckID network.MessageTypeID
+var mergeCheckReplyID network.MessageTypeID
 
 // Service represents data needed for one pop-party.
 type Service struct {
@@ -64,10 +85,14 @@ type Service struct {
 	*onet.ServiceProcessor
 	path string
 	data *saveData
-	// channel to return the configreply
-	ccChannel chan *CheckConfigReply
 	// propagate final message
-	Propagate messaging.PropagationFunc
+	PropagateFinalize messaging.PropagationFunc
+	// propagate merge info
+	PropagateMerging messaging.PropagationFunc
+	// Sync tools
+	// key of map is ID of party
+	// synchronizing inside one party
+	syncs map[string]*sync
 }
 
 type saveData struct {
@@ -75,8 +100,26 @@ type saveData struct {
 	Pin string
 	// Public key of linked pop
 	Public abstract.Point
-	// The final statement
-	Final *FinalStatement
+	// The final statements
+	// key of map is ID of party
+	Finals map[string]*FinalStatement
+	// The info used in merge process
+	// key is ID of party
+	merges map[string]*merge
+}
+
+type merge struct {
+	// Map of final statements of parties that are going to be merged together
+	statementsMap map[string]*FinalStatement
+	// Flag tells that message distribution has already started
+	distrib bool
+}
+
+type sync struct {
+	// channel to return the configreply
+	ccChannel chan *checkConfigReply
+	// channel to return the mergereply
+	mcChannel chan *mergeConfigReply
 }
 
 // PinRequest prints out a pin if none is given, else it verifies it has the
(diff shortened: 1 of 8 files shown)


pop/service/service_no_race_test.go
```diff
@@ -0,0 +1,81 @@
...
+		log.ErrFatal(err)
+		fr.Signature = sg
+		msg, err := srvcs[2*i+1].FinalizeRequest(fr)
+		require.Nil(t, err)
+		require.NotNil(t, msg)
+		_, ok := msg.(*finalizeResponse)
+		require.True(t, ok)
+	}
+
+	log.Info("Group 1, Server:", srvcs[0].ServerIdentity())
```
ineiti: remove all `log.Info` in tests with `log.Lvl1` and `log.Lvl2`. The reason is that `go test` should not print anything to the command-line, only `go test -v`. But `log.Info` prints to the command-line regardless of the `-v`-argument.

pop/service/service_no_race_test.go
```diff
@@ -0,0 +1,81 @@
+// +build !race
```
ineiti: I think it should build without races - look at it as an exercise in concurrent programming - very important when building distributed systems ;)

pop/service/service_no_race_test.go
```diff
@@ -0,0 +1,81 @@
...
+	log.Info("Group 2, Server:", srvcs[3].ServerIdentity())
+	cc.Final = srvcs[0].data.Finals[hash[0]]
+	cc.ID = []byte(hash[1])
+	srvcs[0].SendRaw(r.List[2], cc)
+	meta := srvcs[2].data.merges[hash[1]]
+	// Here is involuntary race condition solved by waiting in cycle
+	// on timeout
+	// In this case I can't wait till the end of process because
+	// I test here only one message
+	Eventually(t, func() bool { return len(meta.statementsMap) == len(descs) },
```
ineiti: If the `Eventually` creates races, then it should go away. Either you wait for the message, or you create a channel that you can wait on.

pop/service/service.go
```diff
@@ -29,14 +29,17 @@ import (
...
 	"fmt"
-
 	"math/big"
+	"sort"
+	"strings"
+	"time"
 
-	"github.com/dedis/cothority/cosi/protocol"
-	"github.com/dedis/cothority/messaging"
+	"gopkg.in/dedis/cothority.v1/bftcosi"
```
ineiti: Everything should depend on `dedis/cothority/` and not on `cothority.v1`!

pop/service/service.go
```diff
@@ -97,124 +149,314 @@ func (s *Service) PinRequest(req *PinRequest) (network.Message, onet.ClientError
...
+	m.statementsMap[hash] = newFinal
+
+	s.save()
+	return &finalizeResponse{newFinal}, nil
 }
 
-// PropagateFinal saves the new final statement
-func (s *Service) PropagateFinal(msg network.Message) {
-	fs, ok := msg.(*FinalStatement)
+/* ------------InterConode Messages ----------- */
```
ineiti: we don't use these kinds of comments. Please remove.

pop/service/service.go
```diff
@@ -97,124 +149,314 @@ func (s *Service) PinRequest(req *PinRequest) (network.Message, onet.ClientError
...
+		log.Error("Couldn't send reply:", err)
+	}
+}
+
+// MergeConfigReply processes the response after MergeConfig message
+func (s Service) MergeConfigReply(req *network.Envelope) {
+	log.Lvlf2("MergeConfigReply: %s from %s got %v",
+		s.ServerIdentity(), req.ServerIdentity.String(), req.Msg)
+	mcrVal, ok := req.Msg.(*mergeConfigReply)
+	var mcr *mergeConfigReply
```
ineiti: Merge this line and the below line using `mcr :=`

pop/service/service.go
```diff
@@ -97,124 +149,314 @@ func (s *Service) PinRequest(req *PinRequest) (network.Message, onet.ClientError
...
+		if mcrVal.Final == nil {
+			log.Error("Empty FinalStatement in reply")
+			return nil
+		}
+		mcrVal.PopStatus = final.VerifyMergeStatement(mcrVal.Final)
+		return mcrVal
+	}()
+	if syncData, ok := s.syncs[string(mcrVal.PopHash)]; ok {
+		if len(syncData.mcChannel) == 0 {
+			syncData.mcChannel <- mcr
```
ineiti: As `mcr` is only used here, I'd put the above anonymous function inside the if, so it is not executed when it's not used.

pop/service/service.go
```diff
@@ -224,47 +466,451 @@ func (s *Service) CheckConfig(req *network.Envelope) {
 // CheckConfigReply strips the attendees missing in the reply, if the
 // PopStatus == PopStatusOK.
 func (s *Service) CheckConfigReply(req *network.Envelope) {
-	ccrVal, ok := req.Msg.(*CheckConfigReply)
-	var ccr *CheckConfigReply
-	ccr = func() *CheckConfigReply {
+	ccrVal, ok := req.Msg.(*checkConfigReply)
+	var ccr *checkConfigReply
+	ccr = func() *checkConfigReply {
```
ineiti: same as above, merge to `ccr := func()`

pop/service/service.go
```diff
@@ -224,47 +466,451 @@ func (s *Service) CheckConfig(req *network.Envelope) {
...
 		}
-		s.intersectAttendees(ccrVal.Attendees)
+		final.Attendees = intersectAttendees(final.Attendees, ccrVal.Attendees)
 		return ccrVal
 	}()
-	if len(s.ccChannel) == 0 {
-		s.ccChannel <- ccr
+	if syncData, ok := s.syncs[string(ccrVal.PopHash)]; ok {
+		if len(syncData.ccChannel) == 0 {
+			syncData.ccChannel <- ccr
```
ineiti: same as above: as `ccr` is not used outside of the if, call the function only here.

pop/service/service.go
```diff
@@ -224,47 +466,451 @@ func (s *Service) CheckConfig(req *network.Envelope) {
...
+		return false
+	}
+
+	// We need to find all local parties are supposed to merge
+	// verify that everything is correct
+	var final, finalReceived *FinalStatement
+	// local parties which will me merged during current process
+	finals := make([]*FinalStatement, 0)
+	var ok, found bool
+	for _, finalReceived = range stmtsMap {
```
ineiti: Can there be multiple parties found? If only one party can be found, you should `break` if `found == true`.

pop/service/service.go
```diff
@@ -224,47 +466,451 @@ func (s *Service) CheckConfig(req *network.Envelope) {
...
+		return false
+	}
+	return true
+}
+
+// Verification function for sighning during Merging
+func (s *Service) bftVerifyMerge(Msg []byte, Data []byte) bool {
+	stmtsMap, err := decodeMapFinal(Data)
+	if err != nil {
+		log.Error("VerifyMerge: can't decode Data: " + err.Error())
```
ineiti: We don't consider these messages as errors - it's a verification function, and up to some point it's normal that the verification fails - this is why we verify ;)

So change the `log.Error` to `log.Lvl2` or so. It's not an actual error, it's 'just' that the verification decided that it's bad data. The same goes for the rest of the method, replace `log.Error` with `log.Lvl2`.

pop/service/service.go
```diff
@@ -224,47 +466,451 @@ func (s *Service) CheckConfig(req *network.Envelope) {
...
+		s.data.Finals[string(f.Desc.Hash())] = final
+		// there is no need to support consistency of syncData and merge
+		// for old parties because their finalStatements are rewritten
+	}
+
+	s.save()
+	return true
+}
+
+/* --------------Propagation function-------------- */
```
ineiti: you can remove all those `----`-comments, we don't use it in our other packages.

pop/service/service.go
```diff
@@ -224,47 +466,451 @@ func (s *Service) CheckConfig(req *network.Envelope) {
...
+
+// PropagateFinal saves the new final statement
+func (s *Service) PropagateFinal(msg network.Message) {
+	fs, ok := msg.(*FinalStatement)
+	if !ok {
+		log.Error("Couldn't convert to a FinalStatement")
```
ineiti: Again, `log.Error` should be `log.Lvl2` here.

pop/service/service.go
```diff
@@ -224,47 +466,451 @@ func (s *Service) CheckConfig(req *network.Envelope) {
...
+	if err := fs.Verify(); err != nil {
+		log.Error(err)
+		return
+	}
+	*s.data.Finals[string(fs.Desc.Hash())] = *fs
+	s.save()
+	log.Lvlf2("%s Stored final statement %v", s.ServerIdentity(), fs)
+}
+
+/* ----------------Utilite functions--------------- */
```
ineiti: s/Utilite/Utility/

But remove the `---`-comment anyway.

pop/service/service.go
```diff
@@ -224,47 +466,451 @@ func (s *Service) CheckConfig(req *network.Envelope) {
...
+	}
+	s.save()
+	return nil
+}
+
+// Merge sends MergeConfig to all parties,
+// Receives Replies, updates info about global merge party
+// When all merge party's info is saved, merge it and starts global sighning process
+// After all, sends StoreConfig request to other conodes of own party
+func (s *Service) Merge(final *FinalStatement, m *merge) (*FinalStatement,
```
ineiti: I prefer the following structure of the file:
`init()`
constants and structure-definitions
public methods on structure
private methods on structure
package public functions
package private functions

Please re-arrange.

pop/service/service.go
IlyaMarkovMipt: Yes, it's possible that one `conode` holds several parties and some of them may be merged into one and it's crucial to refresh info for each party included in merge config.

PR #928, opened 2017-09-06
correctly locking block-request replies

ineiti:
Closes a race when getting a reply to a block-request.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -384,7 +384,9 @@ func (s *Service) getBlockReply(env *network.Envelope) {
 	id := s.Sbm.Store(gbr.SkipBlock)
 	s.save()
 	log.Lvl3("Sending block to channel")
+	s.blockRequestsMutex.Lock()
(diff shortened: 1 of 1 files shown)


nikkolasg:
I have to say I'm a bit perplex by this. Normally there should be no need to have a lock around a channel because that's already thread safe to send and receive on the channel. 
Is the map of channels the thing you want to protect by using the mutex ? 
Maybe it works but feels very weird, more like hot fix than fixing the underlying problem, BUT the good news is I don't know nothing about the codebase you're fixing so I'm probably wrong ;)

ineiti:
It's definitively a bad fix to a bad problem. In the new skipchain.v2 I want to have an actual protocol to do the skipblock-propagation and updates, so I won't need these hacks anymore of sending single messages between service-instances.

And, yes, the lock is for accessing the `map`, not for the channel.

nikkolasg:
>  I want to have an actual protocol to do the skipblock-propagation and updates, so I won't need these hacks anymore of sending single messages between service-instances.

 I see nothing wrong with using message between Services !
Anyway ok !

PR #931, opened 2017-09-08
testing skipchains - some more

ineiti:
Waiting for propagations to be finished in skipchains and only testing with 10 parallel blocks.

### DIFF
--- a/skipchain/skipchain_test.go
+++ b/skipchain/skipchain_test.go
@@ -442,6 +442,7 @@ func TestService_StoreSkipBlockSpeed(t *testing.T) {
 }
 
 func TestService_ParallelStore(t *testing.T) {
+	log.Print("Starting Parallel")
 	nbrRoutines := 30
 	local := onet.NewLocalTest()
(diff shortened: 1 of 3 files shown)


skipchain/skipchain_test.go
```diff
@@ -616,3 +579,44 @@ func makeHELS(local *onet.LocalTest, nbr int) ([]*onet.Server, *onet.Roster, *Se
...
+				propagating = true
+			}
+		}
+		if propagating {
+			time.Sleep(time.Millisecond * 100)
+		}
+	}
+}
+
+func checkBacklinks(services []*Service, sb *SkipBlock) {
```
nikkolasg: what's the reason for this here ? I mean, this check is already supposed to happen somewhere else right ?

ineiti: It used to be in two tests, no only one of them is remaining. But I wanted to clean up a bit and put all private functions to the bottom of the file. I'll inline it, as it's only used once now.

PR #933, opened 2017-09-12
Cisc new auth

IlyaMarkovMipt:
Add authentication in `identity`.

There are two types of authentication:
 * with `pop-token`
 * with public keys


ineiti:
Could this use a `git rebase master`? It looks like it has things inside that were already merged, no?

ineiti:
And please provide a description - commit messages are not enough.

PR #945, opened 2017-09-26
Pop readme

ineiti:
Added a readme to the pop-directory.
Also updated the help-messages for the commands.

Depends on onet#228

### DIFF
--- a/pop/README.md
+++ b/pop/README.md
@@ -1,0 +1,119 @@
+# Proof of Personhood
+
+pop command is used by one of the two groups of users:
+
+- organizers (org)
+- attendees (attendee)
+
+Each group has its own set of commands to start, finalize and use the keys.
+
+## Organizers
+
+### Link
+
+First every organizer has to link to his conode:
+
+```bash
+pop org link 127.0.0.1:2002
+# Read pin from command line
+pop org link 127.0.0.1:2002 PIN
+```
+
+### Store the configuration
+
+Now each organizer registers the description that they got together:
+`description.toml` - example:
+
+```toml
+Name = "Proof-of-Personhood Party"
+DateTime = "2017-08-08 15:00 UTC"
+Location = "Earth, City"
+[[servers]]
+  Address = "tcp://127.0.0.1:2002"
+  Public = "lWHQZiMkxchMd7hfpn18HLqJUqiGRpIzSTyFYzWD9Zc="
+  Description = "Cot-1"
+[[servers]]
+  Address = "tcp://127.0.0.1:2004"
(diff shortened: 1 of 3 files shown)


nikkolasg:
I would say whether to merge or not depends on which point of view do we need to read the readme. Looking at it from an external point of view, the README needs much more information .
From developer-who-already-knows-pop point a view, it seems fine !

Which one do we want ?

ineiti:
For the external point of view we'll have to do something like the ongoing progress for skipchain.v2... This is mostly for the students so they have something to follow.

pop/README.md
```diff
@@ -0,0 +1,119 @@
+# Proof of Personhood
+
+pop command is used by one of the two groups of users:
```
nikkolasg: is used by both you mean ? Otherwise which group is using `pop` if there is only one ?

pop/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+- organizers (org)
+- attendees (attendee)
+
+Each group has its own set of commands to start, finalize and use the keys.
+
```
nikkolasg: Regarding the rest of the section I would probably add a little more description such as: `Each organizers must run their own conode [link-to-conode] in order to setup a party`

pop/README.md
```diff
@@ -0,0 +1,119 @@
...
+```bash
+pop org link 127.0.0.1:2002
+# Read pin from command line
+pop org link 127.0.0.1:2002 PIN
+```
+
+### Store the configuration
+
+Now each organizer registers the description that they got together:
+`description.toml` - example:
```
nikkolasg: Each organizers should assemble their identity generated when first running the conode. The description.toml contains the list of identities of all organizers and some additional information about the party.

(stg like that ;) )

pop/README.md
```diff
@@ -0,0 +1,119 @@
...
+  Address = "tcp://127.0.0.1:2004"
+  Public = "hbCeQID3RhQ0tVI8G/XkmyJv+Xr8qosVpCyJUppZcPw="
+  Description = "Cot-2"
+```
+
+```toml
+pop org config description.toml
+```
+
+This will print out the hash of the description, which should be the same
```
nikkolasg: Should it be verified ? What happens if not ? (stg to write down ?)

pop/README.md
ineiti: Each + singular: Each organizer ... his identity...

ineiti:
There is a saying in French: "j'ai donné le petit doigt et il m'a pris la main"... I just wanted to write a couple of lines to have a minimal documentation ;)

PR #947, opened 2017-09-26
Limit skipchains 935

### LINKED ISSUE #935: Limit skipchains
Instead of accepting every new request to create a skipchain, limit it to authenticated requests from hosts that can sign a message.

Extending the existing CISC:
1. when `cisc admin link` automatically setting up a private/public key pair and linking that
2. when `cisc id create` per default use that key pair

Cleaning up the CISC commands

Comment CISC

ineiti:
Closes #935


ineiti:
Superseded by #961

PR #954, opened 2017-11-07
Skipchain v2 effort

lbarman:
in progress


jeffallen:
Current plan is to do this with object store API.

PR #956, opened 2017-11-10
dont marshall bytes

### LINKED ISSUE #955: Create skipchains
Now: When creating a skipchain, the `data` field is always marshalled, even if it is of type `[]byte`. This is contrary to how the `data` field is handled in adding a block.

Wanted: `CreateSkipchain` should not marshal the data if it's of type `[]byte`

ineiti:
Closes #955

### DIFF
--- a/skipchain/api_test.go
+++ b/skipchain/api_test.go
@@ -23,10 +23,7 @@ func TestClient_CreateGenesis(t *testing.T) {
 	_, roster, _ := l.GenTree(3, true)
 	defer l.CloseAll()
 	c := newTestClient(l)
-	_, cerr := c.CreateGenesis(roster, 1, 1, VerificationNone,
-		[]byte{1, 2, 3}, nil)
-	require.NotNil(t, cerr)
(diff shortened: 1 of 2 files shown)


skipchain/api_test.go
```diff
@@ -23,10 +23,7 @@ func TestClient_CreateGenesis(t *testing.T) {
 	_, roster, _ := l.GenTree(3, true)
 	defer l.CloseAll()
 	c := newTestClient(l)
-	_, cerr := c.CreateGenesis(roster, 1, 1, VerificationNone,
-		[]byte{1, 2, 3}, nil)
-	require.NotNil(t, cerr)
-	_, cerr = c.CreateGenesis(roster, 1, 0, VerificationNone,
+	_, cerr := c.CreateGenesis(roster, 1, 0, VerificationNone,
```
jeffallen: I was expecting to see one test of each kind. Instead, it seems like you removed the previous test for []byte{1,2,3}. Why?

PR #964, opened 2017-12-08
Use boltdb for storing skipchains

kc1212:
SkipBlockMap still exists because it's used in scmgr. Change to use the database is planned in a future PR.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -39,8 +39,7 @@ const skipblocksID = "skipblocks"
 // Service handles adding new SkipBlocks
 type Service struct {
 	*onet.ServiceProcessor
-	// Sbm is the skipblock-map that holds all known skipblocks to this service.
-	Sbm                *SkipBlockMap
+	db                 *SkipBlockDB
 	propagate          messaging.PropagationFunc
 	verifiers          map[VerifierID]SkipBlockVerifier
 	blockRequestsMutex sync.Mutex
@@ -99,7 +98,7 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, on
 		defer s.newBlockEnd(prop)
 
 		if !prop.ParentBlockID.IsNull() {
-			parent := s.Sbm.GetByID(prop.ParentBlockID)
+			parent := s.db.GetByID(prop.ParentBlockID)
 			if parent == nil {
 				return nil, onet.NewClientErrorCode(ErrorParameterWrong,
 					"Didn't find parent")
@@ -111,7 +110,7 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, on
 
 	} else {
 		// We're appending a block to an existing chain
-		prev = s.Sbm.GetByID(psbd.LatestID)
+		prev = s.db.GetByID(psbd.LatestID)
 		if prev == nil {
 			return nil, onet.NewClientErrorCode(ErrorBlockNotFound,
 				"Didn't find latest block")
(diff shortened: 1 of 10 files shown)


ineiti:
LGTM

skipchain/skipchain.go
```diff
@@ -270,12 +261,10 @@ func (s *Service) GetSingleBlockByIndex(id *GetSingleBlockByIndex) (*SkipBlock,
 // GetAllSkipchains returns a list of all known skipchains
 func (s *Service) GetAllSkipchains(id *GetAllSkipchains) (*GetAllSkipchainsReply, onet.ClientError) {
 	// Write all known skipblocks to a map, thus removing double blocks.
-	chains := map[string]*SkipBlock{}
-	s.Sbm.Lock()
-	for _, sb := range s.Sbm.SkipBlocks {
-		chains[string(sb.SkipChainID())] = sb
+	chains, err := s.db.dump()
```
jeffallen: I am not a super-fan of such a simple, overused word as "dump" for a method name. Maybe "all()"? Especially if later we need a filtering thing called select(func f(...) bool).

skipchain/struct.go
```diff
@@ -363,21 +397,32 @@ func (bl *BlockLink) VerifySignature(publics []kyber.Point) error {
 
 // SkipBlockMap holds the map to the skipblocks. This is used for verification,
 // so that all links can be followed.
+// TODO remove when scmgr is updated
```
jeffallen: Why can't this be done now?

skipchain/struct.go
```diff
@@ -405,18 +450,64 @@ func (sbm *SkipBlockMap) Store(sb *SkipBlock) SkipBlockID {
 	return sb.Hash
 }
 
+// Store stores the given SkipBlock in the service-list
+func (db *SkipBlockDB) Store(sb *SkipBlock) SkipBlockID {
+	sbOld, err := db.get(sb.Hash)
```
jeffallen: There's a kind of "use after unlock" race here, not in Go, but with respect to the transactional validity of the database. It seems to me that all the actions of Store (get, if exists, then copy forward links, otherwise store as is) need to be in one boltdb txn.

skipchain/struct.go
```diff
@@ -535,3 +627,64 @@ func (sbm *SkipBlockMap) GetFuzzy(id string) *SkipBlock {
...
+		}
+
+		sb = sbMsg.(*SkipBlock).Copy()
+		return nil
+	})
+
+	return sb, err
+}
+
+// dump returns all the data in the database as a map
```
jeffallen: This feels scary if we are talking about a multi-gigabyte database. Should there be a safety limit on it, telling the caller, "errors.New("result would be too big")"

Not very nice for the caller, but better than blowing up RAM, crashing, etc.

kc1212: All the dump related methods are going to be removed after @ineiti finishes his skipchain PR. What do you think @ineti, should we add some limit on this now or remove it after we merge?

skipchain/struct.go
kc1212: Inside `db.get`, I make a copy of whatever is returned by the transaction. So the operations in `Store` is done on a copy of the data. That should be OK right?

skipchain/struct.go
kc1212: Discussed this with @ineiti. Since he's changing the config structure in scmgr, we wanted it to make it easier to merge when his PR is done. That is, instead of merging both skipchain and scmgr at once, we plan to merge them separately.

skipchain/skipchain.go
kc1212: `all` like a high order function, e.g. `all(func (k Key, v Value) bool) (map[Key]Value, error)`?

kc1212: Actually, what I described is usually called filter. All returns a bool if all items satisfies the predicate.

skipchain/struct.go
ineiti: I'd remove it after merge. The `GetAllSkipchains` method to the service needs to be removed and replaced by a `GetSkipchainUpdate` where you can give a maximum range of skipchains to be returned.

skipchain/struct.go
jeffallen: If there are two requests to add blocks that arrive on two TCP streams concurrently, then they will execute in two goroutines. If they are both writes of new skipblocks, I think they will be racy with respect to each other.

kc1212:
Fixed (I hope) the consistency issue in the latest commits that @jeffallen pointed out, thanks! There are two ways of doing it, the first is put operations that require a consistent view of the database in a single transaction, the other is to use a mutex. I tried to use the mutex solution and I think it gets quite confusing, because then we'll have two locking mechanisms - transaction and mutex. Hence, I went for the transaction approach.

PR #965, opened 2017-12-08
Adapt to changes in API for random in Suite

jeffallen:


bftcosi/bftcosi.go
```diff
@@ -290,7 +290,7 @@ func (bft *ProtocolBFTCoSi) handleCommitment(msgs []commitChan) error {
 			if len(bft.tempPrepareCommit) < len(bft.Children()) {
 				continue
 			}
-			commitment = bft.prepare.Commit(nil, bft.tempPrepareCommit)
+			commitment = bft.prepare.Commit(bft.Suite().RandomStream(), bft.tempPrepareCommit)
```
ineiti: Question: why is this not just `bft.prepare.Commit(bft.Suite(), ...)`?

Same for `Commit` and `CreateCommitment` below.

cosi/protocol/cosi.go
```diff
@@ -221,7 +221,7 @@ func (c *CoSi) handleCommitment(in *Commitment) error {
 	}
 
 	// go to Commit()
-	out := c.cosi.Commit(nil, c.tempCommitment)
+	out := c.cosi.Commit(c.Suite().RandomStream(), c.tempCommitment)
```
ineiti: Question: Why not `c.Suite()` instead of `c.Suite().RandomStream()`?

identity/api.go
```diff
@@ -196,7 +195,7 @@ func (i *Identity) popAuth(au *Authenticate, atts []kyber.Point) (*CreateIdentit
 			break
 		}
 	}
-	sigtag := anon.Sign(i.Client.Suite().(anon.Suite), random.Stream, au.Nonce,
+	sigtag := anon.Sign(i.Client.Suite().(anon.Suite), i.Client.Suite().RandomStream(), au.Nonce,
```
ineiti: Why is the randomstream still here? I thought we moved that to suite...

same on line 108.

identity/service_test.go
```diff
@@ -33,11 +33,12 @@ func TestService_CreateIdentity2(t *testing.T) {
...
 	ci.Data = il
 	ci.Roster = el
-	ci.Nonce = random.Bytes(nonceSize, random.Stream)
+	ci.Nonce = make([]byte, nonceSize)
+	random.Bytes(ci.Nonce, random.New())
 	service.auth.nonces[string(ci.Nonce)] = struct{}{}
 	ctx := []byte(ServiceName + service.ServerIdentity().String())
 
-	ci.Sig = anon.Sign(tSuite.(anon.Suite), random.Stream, ci.Nonce,
+	ci.Sig = anon.Sign(tSuite, tSuite.RandomStream(), ci.Nonce,
```
ineiti: why separate argument here for randomstream?

pop/app.go
```diff
@@ -439,7 +439,7 @@ func attSign(c *cli.Context) error {
 	msg := []byte(c.Args().First())
 	ctx := []byte(c.Args().Get(1))
 	Set := anon.Set(party.Final.Attendees)
-	sigtag := anon.Sign(cothority.Suite.(anon.Suite), random.Stream, msg,
+	sigtag := anon.Sign(cothority.Suite.(anon.Suite), random.New(), msg,
```
ineiti: Repeat: why separate random-argument?

cosi/protocol/cosi.go
jeffallen: When we switch to kyber/sign/cosi, it will be, because there the signature of Commit is:
`func Commit(suite Suite) (kyber.Scalar, kyber.Point)`

identity/api.go
jeffallen: Fixed.

PR #966, opened 2017-12-11
skipchain_auth_kyber_3_949

### LINKED ISSUE #935: Limit skipchains

### LINKED ISSUE #949: Skipchain-authentication
Instead of everybody having the right to create new skipchains, this right should be limited.

Depending on `skipchain.AuthSkipchain`, different access-rights are defined:
  - 0 : don't verify the roster of the skipblocks, accept everything
  - 1 : accept only skipchains that are being followed, and if at least one of the conodes is in the roster of a followed skipchain
  - 2 : only accept new blocks if all conodes are in one of the followed rosters. This is only for new blocks where no other skipchain is stored.

### LINKED ISSUE #981: scmgr usability
```
$ ./scmgr sc a
Adding a block with a new group
Please give group-file and id to add 
```
but then:

```
$ ./scmgr sc a ../conode/co3/public.toml 2610
Adding a block with a new group
open 2610: no such file or directory Couldn't open group definition file
```

It should (1) ask for them in the right order, and/or (2) not use positional arguments.

ineiti:
This is an implementation of #949 with the current skipchain. I'd like to take this as a base for the documentation and the new storage-system.

It also adds a `link` command to restrict who is allowed to communicate with the service:
`scmgr admin link` automatically setting up a private/public key pair and linking that to the skipchain-service

Closes #949
Closes #981

### DIFF
--- a/scmgr/app.go
+++ b/scmgr/app.go
@@ -161,7 +161,7 @@ func linkQuery(c *cli.Context) error {
 	return errors.New("not yet implemented")
 }
 
-func followAddId(c *cli.Context) error {
+func followAddID(c *cli.Context) error {
 	cfg := getConfigOrFail(c)
 	if c.NArg() != 2 {
 		return errors.New("please give: skipchain_id ip:port")
--- a/scmgr/app.go
+++ b/scmgr/app.go
@@ -439,7 +439,7 @@ func dnsList(c *cli.Context) error {
 	return nil
 }
 
-// lsIndex writes one index-file for every known skipchain and an index.html
+// lsIndex writes one index-file for every known skipchain and an index.js
 // for all skiplchains.
 func dnsIndex(c *cli.Context) error {
 	output := c.Args().First()
@@ -447,7 +447,7 @@ func dnsIndex(c *cli.Context) error {
 		return errors.New("Missing output path")
 	}
 
-	cleanHTMLFiles(output)
+	cleanJSFiles(output)
 
 	cfg, err := loadConfig(c)
 	if err != nil {
@@ -473,8 +473,8 @@ func dnsIndex(c *cli.Context) error {
 
 		// Write the genesis block file
 		content, _ := json.Marshal(block)
-		log.Infof("Writing %s.html", block.GenesisID)
-		err := ioutil.WriteFile(filepath.Join(output, block.GenesisID+".html"), content, 0644)
+		log.Infof("Writing %s.js", block.GenesisID)
+		err := ioutil.WriteFile(filepath.Join(output, block.GenesisID+".js"), content, 0644)
 
 		if err != nil {
 			log.Info("Cannot write block-specific file")
@@ -486,9 +486,9 @@ func dnsIndex(c *cli.Context) error {
 		log.Info("Cannot convert to json")
 	}
 
-	// Write the json into the index.html
-	log.Infof("Storing an index of all blocks to index.html")
-	err = ioutil.WriteFile(filepath.Join(output, "index.html"), content, 0644)
+	// Write the json into the index.js
+	log.Infof("Storing an index of all blocks to index.js")
+	err = ioutil.WriteFile(filepath.Join(output, "index.js"), content, 0644)
 	if err != nil {
 		log.Info("Cannot write in the file")
 	}
@@ -553,15 +553,15 @@ func dnsUpdate(c *cli.Context) error {
 	return cfg.save(c)
 }
 
-// Remove every file matching *.html in the given directory
-func cleanHTMLFiles(dir string) error {
+// Remove every file matching *.js in the given directory
+func cleanJSFiles(dir string) error {
 	files, err := ioutil.ReadDir(dir)
 	if err != nil {
 		return err
 	}
 
 	for _, f := range files {
-		if strings.HasSuffix(f.Name(), ".html") {
+		if strings.HasSuffix(f.Name(), ".js") {
 			err := os.Remove(filepath.Join(dir, f.Name()))
 			if err != nil {
 				return err
@@ -572,14 +572,14 @@ func cleanHTMLFiles(dir string) error {
 	return nil
 }
 
-// JSON skipblock element to be written in the index.html file
+// JSON skipblock element to be written in the index.js file
 type jsonBlock struct {
 	GenesisID string
 	Servers   []string
 	Data      []byte
 }
 
-// JSON list of skipblocks element to be written in the index.html file
+// JSON list of skipblocks element to be written in the index.js file
 type jsonBlockList struct {
 	Blocks []jsonBlock
 }
--- a/scmgr/app.go
+++ b/scmgr/app.go
@@ -20,6 +20,7 @@ import (
 	"github.com/BurntSushi/toml"
 	bolt "github.com/coreos/bbolt"
 	"github.com/dedis/cothority"
+	"github.com/dedis/cothority/identity"
 	"github.com/dedis/cothority/skipchain"
 	"github.com/dedis/kyber"
 	"github.com/dedis/kyber/util/encoding"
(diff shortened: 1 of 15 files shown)


scmgr/commands.go
```diff
@@ -0,0 +1,168 @@
...
+	groupsDef := "[group-definition]"
+	return []cli.Command{
+		{
+			Name:  "admin",
+			Usage: "administer the skipchain-service",
+			Subcommands: []cli.Command{
+				{
+					Name:    "link",
+					Usage:   "link with a skipchain-service",
+					Aliases: []string{"l"},
```
jeffallen: "admin list" and "admin link" both use abbreviation "l"

scmgr/app.go
```diff
@@ -149,11 +67,125 @@ func main() {
...
+		var err error
+		other, err = findLinkFromAddress(cfg, publicOrIP)
+		if err != nil {
+			return errors.New("couldn't parse address: " + err.Error())
+		}
+	}
+	if other == nil {
+		return errors.New("didn't find host with ip or public key: " + publicOrIP)
+	}
+	return skipchain.NewClient().SettingAuthentication(other.Conode, other.Private, lvl)
```
jeffallen: I tried to set level 2 and I got "websocket: close 4103: this authentication type is not yet supported"

scmgr/app.go
```diff
@@ -149,11 +67,125 @@ func main() {
...
+	if err != nil {
+		return err
+	}
+	cfg := getConfigOrFail(c)
+	if len(cfg.Link) == 0 {
+		return errors.New("no link yet")
+	}
+	publicOrIP := c.Args().First()
+	// Check if it's a public key and if we can find it
+	other := cfg.Link[publicOrIP]
```
jeffallen: My private.toml file has Public = "big-thing-in-hex" and when I did "scmrg admin auth big-thing-in-hex 1" it worked. But my public.toml file is another obvious place where I could find my public key, and it is a base64-encoded version of the hex decoded big-thing-in-hex. I should be able to give the public key in either format.

scmgr/commands.go
```diff
@@ -0,0 +1,168 @@
...
+
+func getCommands() []cli.Command {
+	groupsDef := "[group-definition]"
+	return []cli.Command{
+		{
+			Name:  "admin",
+			Usage: "administer the skipchain-service",
+			Subcommands: []cli.Command{
+				{
+					Name:    "link",
```
jeffallen: I think there needs to be a "unlink" command, because knowing that you can delete the scmgr config file is pretty arcane knowledge.

skipchain/api.go
```diff
@@ -184,3 +229,77 @@ func (c *Client) GetSingleBlockByIndex(roster *onet.Roster, genesis SkipBlockID,
...
+// used to to allow/disallow new blocks. Only if SettingAuthentication(true) has
+// been called is this active.
+func (c *Client) DelFollow(si *network.ServerIdentity, clientPriv kyber.Scalar, sbid SkipBlockID) onet.ClientError {
+	return nil
+}
+
+// ListFollow returns the list of latest skipblock of all skipchains that are followed
+// for authentication purposes.
+func (c *Client) ListFollow(si *network.ServerIdentity) (*ListFollowReply, onet.ClientError) {
+	return nil, nil
```
jeffallen: Not implemented yet?

skipchain/api.go
```diff
@@ -184,3 +229,77 @@ func (c *Client) GetSingleBlockByIndex(roster *onet.Roster, genesis SkipBlockID,
...
+	}
+	req.Signature = sig
+	return c.SendProtobuf(si, req, nil)
+}
+
+// DelFollow asks the conode to remove a skipchain-id from the list of skipchains that are
+// used to to allow/disallow new blocks. Only if SettingAuthentication(true) has
+// been called is this active.
+func (c *Client) DelFollow(si *network.ServerIdentity, clientPriv kyber.Scalar, sbid SkipBlockID) onet.ClientError {
+	return nil
```
jeffallen: Not implemented yet?

skipchain/msgs.go
```diff
@@ -16,23 +20,41 @@ func init() {
...
+		&ListFollowReply{},
 		// - Internal calls
 		// Propagation
 		&PropagateSkipBlocks{},
 		// Request forward-signature
 		&ForwardSignature{},
 		// Request updated block
 		&GetBlock{},
-		// Reply with updated block
+		// ERReply with updated block
```
jeffallen: huh?

skipchain/skipchain.go
```diff
@@ -664,16 +888,104 @@ func (s *Service) newBlockEnd(sb *SkipBlock) bool {
 	return true
 }
 
+// authenticate searches if this node or any follower-node can verify the
+// schnorr-signature.
+func (s *Service) authenticate(msg []byte, sig []byte) bool {
+	if err := schnorr.Verify(Suite, s.ServerIdentity().Public, msg, sig); err == nil {
```
jeffallen: At this point, msg and sig are untrusted input from the attacker. convince me that (1) the amount of mem/cpu they can consume here is bounded (I'm already 99% convinced), and (2) even though schnorr.Verify is not constant time, there is no timing channel that leaks private data.

skipchain/skipchain.go
```diff
@@ -664,16 +888,104 @@ func (s *Service) newBlockEnd(sb *SkipBlock) bool {
...
+	s.storageMutex.Lock()
+	defer s.storageMutex.Unlock()
+	// Also accept blocks that are stored in the FollowIDs
+	for _, id := range s.Storage.FollowIDs {
+		if id.Equal(sb.SkipChainID()) {
+			return true
+		}
+	}
+	switch AuthSkipchain {
+	case 1:
```
jeffallen: It seems like case 2 is missing here? (Not implemented yet?)

scmgr/app.go
ineiti: Perhaps we need to get rid of base64 representation of anything. It's a bit shorter, but it's confusing me because in some places we use hex, and in others base64...

ineiti: So @nikkolasg would like to have hex-only everywhere?

skipchain/skipchain.go
ineiti: If you DOS the system, you can consume a lot of CPU here.
The data you leak is how many clients are signed up to the service.

scmgr/app.go
ineiti: Implemented differently now.

skipchain/skipchain.go
ineiti: Implemented differently now.

kc1212:
This will go in master first or boltdb?

ineiti:
I think boltdb goes in first, then this.

ineiti:
And one more - even though I will definitively have to redo the commit-messages...

jeffallen:
I managed to make a crash like this:

```
bash-3.2$ ./scmgr admin l -private ../conode/co1/private.toml 
./scmgr admin l -private ../conode/co1/private.toml 
bash-3.2$ ./scmgr admin f -id 01 -any 127.0.0.1:7002
./scmgr admin f -id 01 -any 127.0.0.1:7002
bash-3.2$ ./scmgr admin f -id 01 -any 127.0.0.1:7004
./scmgr admin f -id 01 -any 127.0.0.1:7004
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x43d9066]

goroutine 1 [running]:
main.adminFollow(0xc4200acc60, 0x100, 0xc4200acc60)
	/Users/jallen/go/src/github.com/dedis/cothority/scmgr/app.go:147 +0x3a6
```

jeffallen:
There is a stray log.Info("si, sisNew:"... in app.go, I think it was there from before.

jeffallen:
Another crash:

```
$ ./scmgr admin d 01 127.0.0.1:7003
./scmgr admin d 01 127.0.0.1:7003
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x43d9d32]

goroutine 1 [running]:
main.adminDelfollow(0xc4200c29a0, 0x0, 0xc4200c29a0)
	/Users/jallen/go/src/github.com/dedis/cothority/scmgr/app.go:204 +0x272
```

In this case port 7003 was not listening.

PR #973, opened 2017-12-14
Adapt to change in onet for no base64 in toml files

jeffallen:
Automatically convert public.toml files using the migrate
feature of run_conode.sh.

### DIFF
--- a/conode/conode.go
+++ b/conode/conode.go
@@ -13,7 +13,12 @@
 package main
 
 import (
+	"bufio"
+	"encoding/base64"
+	"encoding/hex"
+	"fmt"
 	"os"
+	"strings"
 
 	"github.com/dedis/cothority"
 	"github.com/dedis/onet/app"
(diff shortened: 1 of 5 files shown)


conode/conode.go
```diff
@@ -128,3 +133,29 @@ func checkConfig(c *cli.Context) error {
 	}
 	return check.Config(tomlFileName, c.Bool("detail"))
 }
+
+// Convert toml files from base64-encoded kyes to hex-encoded keys
+func convert64(c *cli.Context) error {
+	scanner := bufio.NewScanner(os.Stdin)
```
ineiti: Please toml-read the file as a `app.GroupToml` and work on that...

And there should be a `private.toml` somewhere - shouldn't that be also converted?

conode/run_conode.sh
```diff
@@ -118,16 +122,22 @@ runLocal(){
...
+			mv $co/public.toml $co/public.toml.bak
+			$CONODE_BIN convert64 < $co/public.toml.bak > $co/public.toml
+		    fi
 		fi
 
 		if [ ! -d $co ]; then
 			echo -e "127.0.0.1:$((7000 + 2 * $n))\nConode_$n\n$co" | $CONODE_BIN setup
 		fi
-		$CONODE_BIN -c $co/private.toml -d $DEBUG &
+		$CONODE_BIN server -c $co/private.toml -d $DEBUG &
```
ineiti: shouldn't the `private.toml` also be migrated?

conode/run_conode.sh
```diff
@@ -255,6 +267,14 @@ migrate(){
 			fi
 			echo $VERSION > $PATH_VERSION
 			;;
+		1.2-1)
```
ineiti: I really think this should not be `v1`, but rather `v2-alpha` or so. We're breaking existing code here!

conode/conode.go
jeffallen: Private was always in hex. Thus part of the point of this fix to reduce the confusion.

PR #974, opened 2017-12-14
Adapt to changes in kyber: no more util/hash

jeffallen:
Replace the util/hash package with clearer, simpler code.

### DIFF
--- a/cosi/check/check.go
+++ b/cosi/check/check.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/dedis/cothority/cosi/crypto"
 	"github.com/dedis/cothority/cosi/service"
-	"github.com/dedis/kyber/util/hash"
 )
 
 // RequestTimeOut is how long we're willing to wait for a signature.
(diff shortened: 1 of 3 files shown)


cosi/check/check.go
```diff
@@ -138,10 +137,12 @@ func checkList(list *onet.Roster, descs []string, detail bool) error {
...
 // the roster el to create the collective signature.
 // In case the signature fails, an error is returned.
-func signStatement(client *service.Client, read io.Reader, el *onet.Roster) (*service.SignatureResponse,
-	error) {
+func signStatement(client *service.Client, read io.Reader, el *onet.Roster) (*service.SignatureResponse, error) {
 	suite := client.Suite().(kyber.HashFactory)
-	msg, _ := hash.Stream(suite.Hash(), read)
+
+	h := suite.Hash()
+	io.Copy(h, read)
```
ineiti: why not use `h.Write` here? Also in `signStatement`?

jeffallen: io.Copy is to go from a io.Reader to an io.Writer. h.Write() is to send a []byte into an io.Writer.  io.Copy allocates a 32k buffer and loops on Read/Write to/from that buffer. If I used h.Write I would need to do the same. See io.Copy's internals here: https://golang.org/src/io/io.go#L361

PR #975, opened 2017-12-15
GetFuzzy for boltdb

kc1212:


skipchain/struct.go
```diff
@@ -645,6 +645,45 @@ func (sbm *SkipBlockMap) GetFuzzy(id string) *SkipBlock {
...
+		log.Error("Failed to decode " + id)
+		return nil
+	}
+	if len(match) == 0 {
+		log.Error("id is empty")
+		return nil
+	}
+
+	var sb *SkipBlock
+	db.View(func(tx *bolt.Tx) error {
```
ineiti: You have to make three passes through the keys... The idea is that a user will most probably use the prefix, so you should check for _all_ prefixes first. The next probably use-case is that the user gives the suffix, so you have to check for _all_ suffixes second. The most unprobably use-case is that the user takes a string in the middle, so you have to check for _all_ places in the end.

Come to think of it, for the `anywhere` case, the user might catch the hex-number in a nibble, so this will not be caught here. That's a good comment to make, that the `anywhere` check only checks for whole bytes, not for nibbles.

skipchain/struct_test.go
```diff
@@ -183,6 +184,36 @@ func sign(msg SkipBlockID, servers []*onet.Server, l *onet.LocalTest) (*bftcosi.
 	return &bftcosi.BFTSignature{Sig: sig, Msg: msg, Exceptions: nil}, nil
 }
 
+func TestSkipBlock_GetFuzzy(t *testing.T) {
+	db, fname := setupSkipBlockDB(t)
```
ineiti: I think some more elaborate tests should be done:

```
list = {{1,2,3,6,4}, {2,3,4,7,5}}
search({1}) == 0
search({2}) == 1
search({3}) == 0
search({4}) == 0
search({5}) == 1
search({6}) == 0
search({7}) == 1
```
I think that should cover most of the test-cases.

skipchain/struct.go
kc1212: Regarding the anywhere case, I'm wondering if that's really needed. It's difficult for me to imagine a case where the user would use that feature. Further, if the user makes a typo and an unintended block is matched, it may cause confusion further down the road. Should we only match the prefix and then the suffix?

jeffallen: Why do we support anything other than prefix match? If this is just about usability, then just prefix match is enough to make it easy to type sc ids by hand.

ineiti: sometimes I like matching the end of a string - as it's not complicated and already in the code, I think it's a nice feature.

PR #977, opened 2017-12-18
App flag order

ineiti:
Putting the `-c` and `-d` in front of the command.

Also fixes skipchain-error by following onet.

### DIFF
--- a/conode/run_conode.sh
+++ b/conode/run_conode.sh
@@ -37,7 +37,7 @@ main(){
 		exit 1
 	fi
 	gopath="$(go env GOPATH)"
-	
+
 	if ! echo $PATH | grep -q $gopath/bin; then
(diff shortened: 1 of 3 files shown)


conode/run_conode.sh
```diff
@@ -135,7 +135,7 @@ runLocal(){
 		if [ ! -d $co ]; then
 			echo -e "127.0.0.1:$((7000 + 2 * $n))\nConode_$n\n$co" | $CONODE_BIN setup
 		fi
-		$CONODE_BIN -d $DEBUG server -c $co/private.toml &
+		$CONODE_BIN -d $DEBUG -c $co/private.toml server &
```
jeffallen: There is another one to fix on line 222.

PR #987, opened 2017-12-22
Remove redundant append

kc1212:
Since the block is appended after the propagation anyway


kc1212:
Closing, this change may change skipchain behavour

PR #988, opened 2017-12-22
Skipchain and bftcosi should work if < 1/3 nodes fail

kc1212:
This PR enables bftcosi and skipchain to continue to function when less than a third of the nodes are down. For bftcosi, we changed the tree structure to be a star structure for better fault tolerance properties. Instead of waiting for all messages, the root would wait for as many commit and response messages as possible until some timeout, and then compute the aggregate. Similarly for skipchain, we changed many of its internal protocols to keep on making progress when enough, but not necessarily all, messages arrive. We also added a feature for a node to recover after going down by synchronising skipblocks with other nodes.

The biggest downside of this PR is the timeouts. They must be tuned for their specific application. Currently, they are set to what works best locally. As a consequence, some tests do not work on travis for reasons described in https://github.com/dedis/cothority/issues/1000. We hope to address these issues in future work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ install:
   - go get github.com/dedis/goveralls
 
 script:
-  - make test
+  - make test_playground
 
 notifications:
   email: false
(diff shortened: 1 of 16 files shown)


ineiti:
Please provide a description in the PR.

PR #990, opened 2017-12-29
Removed unused aggCommit

kc1212:
Doesn't appear to be used anywhere


kc1212:
Perhaps change the name to `SendToChildrenInParallel_IgnoreTErrs`, but maybe too long?

PR #993, opened 2018-01-09
Raise the FD limit in order to make tests pass on MacOS

jeffallen:


cosi/simulation/cosi_test.go
```diff
@@ -13,6 +14,23 @@ import (
 var tSuite = suites.MustFind("Ed25519")
 
 func TestMain(m *testing.M) {
+	// Raising the FD limit like this might belong in a more central place,
```
jeffallen: I suspect this would fail to build with Win32. Checking now via cross-compilation.

PR #997, opened 2018-01-17
Make it possible to test node failure in "go test"

jeffallen:


scmgr/test.sh
```diff
@@ -193,6 +193,7 @@ testFailure() {
 	setupGenesis
 	testOK runSc skipchain block add --roster public.toml $ID
 
+	# -n: newest, so #4 is the one that is dead now
```
ineiti: Thanks for that comment - last time I had to lookup `man pkill` ;)

jeffallen:
These commits went into #1003

PR #1004, opened 2018-01-23
Make vartime suites work

jeffallen:
When compiled with "-tags vartime", the suite for cothority is P256.

When the integration tests are run with env var TAGS set to "-tags vartime", the suite for cothority is P256.

When the same env variable is set, run_conode.sh compiles conode with vartime, and the default suite is P256. If you set up a new node, it's suite will be P256. 

The suite is named in the toml files, but if it is not named it defaults to Ed25519, so old files can be used unchanged.

There are still lots of parts of the system that do not allow you to configure the suite at runtime (they use cothority.Suite instead). But at least with these changes, we've been able to identify and fix incorrect assumptions about key lengths.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 install:
-  - go get githib.com/dedis/onet
+  - go get github.com/dedis/onet
   - ( cd $GOPATH/src/github.com/dedis/onet ; git checkout p256 )
   - go get -t ./...
   - go get github.com/dedis/Coding || true
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: go
 install:
   - go get -t ./...
   - go get github.com/dedis/Coding || true
+  - ( cd $GOPATH/src/github.com/dedis/onet ; git checkout p256 )
 
 script:
   - make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: go
 
 install:
+  - go get githib.com/dedis/onet
+  - ( cd $GOPATH/src/github.com/dedis/onet ; git checkout p256 )
   - go get -t ./...
   - go get github.com/dedis/Coding || true
-  - ( cd $GOPATH/src/github.com/dedis/onet ; git checkout p256 )
 
 script:
   - make test
--- a/bftcosi/bftcosi.go
+++ b/bftcosi/bftcosi.go
@@ -21,7 +21,7 @@ import (
 	"github.com/dedis/onet/log"
 )
 
-const timeout = 1
+const defaultTimeout = 1 * time.Second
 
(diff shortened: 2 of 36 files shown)


.travis.yml
```diff
@@ -1,10 +1,10 @@
 language: go
 
 install:
+  - go get github.com/dedis/onet
+  - ( cd $GOPATH/src/github.com/dedis/onet ; git checkout p256 )
```
ineiti: That can't be right in master...

bftcosi/bftcosi.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/dedis/onet/log"
 )
 
-const timeout = 1
+const defaultTimeout = 1 * time.Second
```
ineiti: This should be in another PR, no?

bftcosi/packets.go
```diff
@@ -70,6 +71,10 @@ func (bs *BFTSignature) Verify(s network.Suite, publics []kyber.Point) error {
 	origCommit := s.Point()
 	pointLen := s.PointLen()
 	sigLen := pointLen + s.ScalarLen()
+
+	if len(bs.Sig) < sigLen {
```
ineiti: Shouldn't this be `!=`?

identity/api_test.go
```diff
@@ -75,7 +83,8 @@ func TestIdentity_StoreKeys(t *testing.T) {
 	signature := make(chan []byte)
 	c := node.(*cosi.CoSi)
 	c.RegisterSignatureHook(func(sig []byte) {
-		signature <- sig[:64]
+		log.LLvl3("sig", len(sig))
```
ineiti: `s/LLvl/Lvl/`

messaging/propagate.go
```diff
@@ -17,7 +17,7 @@ func init() {
 }
 
 // How long to wait before timing out on waiting for the time-out.
-const initialWait = 100000
+const initialWait = 100000 * time.Millisecond
```
ineiti: can we have the timeouts in another PR, please?

skipchain/msgs.go
```diff
@@ -58,8 +58,8 @@ func init() {
 		&ProtoExtendSignature{},
 		&ProtoExtendRoster{},
 		&ProtoExtendRosterReply{},
-		&ProtoGetUpdate{},
-		&ProtoBlockReply{},
+		&ProtoGetBlocks{},
```
ineiti: Also another PR for the `GetBlocks`

jeffallen:
This PR is totally screwed up, replacing with several that are correct.

bftcosi/packets.go
jeffallen: No, there's a variable length mask appended onto the end (cosi.go, line 204).

PR #1006, opened 2018-01-25
adding error testing to dispatch

ineiti:
The propagation-protocol doesn't return errors from sending messages.

### DIFF
--- a/messaging/propagate.go
+++ b/messaging/propagate.go
@@ -176,7 +176,10 @@ func (p *Propagate) Dispatch() error {
 					for _, e := range err {
 						errs = append(errs, e.Error())
 					}
-					return errors.New(strings.Join(errs, "\n"))
+					if len(err) > p.allowedFailures {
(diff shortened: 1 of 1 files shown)


messaging/propagate.go
```diff
@@ -161,19 +163,29 @@ func (p *Propagate) Dispatch() error {
...
+				if err := p.SendToParent(&PropagateReply{}); err != nil {
+					return err
+				}
 			}
 			if p.IsLeaf() {
 				process = false
 			} else {
 				log.Lvl3(p.ServerIdentity(), "Sending to children")
-				p.SendToChildrenInParallel(&msg.PropagateSendData)
+				if err := p.SendToChildrenInParallel(&msg.PropagateSendData); err != nil {
```
kc1212: It's ok have some failures here as long as it's not more than `allowedFailures`

ineiti: It's not in R4M yet ;) But thanks for the heads-up.

PR #1009, opened 2018-01-25
Fixes so that suite can be changed at runtime

jeffallen:
cothority.Suite defaults to P256 when vartime is enabled.

Fixes to make it work with P256 (i.e. places where the
length of a signature was hard coded).

Fixes to integration tests to follow these changes.
(But pop/test.sh and cisc/test.sh were already broken and
need to be fixed separately.)

Make run_conode.sh simpler, so that it can run from other
repositories as well unchanged.

### DIFF
--- a/identity/api_test.go
+++ b/identity/api_test.go
@@ -46,14 +46,17 @@ func TestIdentity_PinRequest(t *testing.T) {
 	require.Equal(t, pub, srvc.auth.adminKeys[0])
 }
 
-func TestIdentity_StoreKeys(t *testing.T) {
+func suiteSkip(t *testing.T) {
+	// Some of these tests require Ed25519, so skip if we are currently
+	// running with another suite.
 	if tSuite != suites.MustFind("Ed25519") {
-		// StoreKeys eventually calls eddsa.Verify, which will not work
-		// with non?Ed25519 suites.
-		t.Log("Wrong suite, this service will not run.")
+		t.Skip("current suite is not compatible with this test, skipping it")
 		return
 	}
+}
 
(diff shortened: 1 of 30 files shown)


identity/api_test.go
```diff
@@ -46,6 +47,13 @@ func TestIdentity_PinRequest(t *testing.T) {
 }
 
 func TestIdentity_StoreKeys(t *testing.T) {
+	if tSuite != suites.MustFind("Ed25519") {
+		// StoreKeys eventually calls eddsa.Verify, which will not work
+		// with non?Ed25519 suites.
```
kc1212: small typo non?Ed25519 -> non-Ed25519

pop/service/service_test.go
```diff
@@ -24,22 +24,38 @@ func TestMain(m *testing.M) {
 	log.MainTest(m)
 }
 
+func svc(t *testing.T, local *onet.LocalTest, n int) *Service {
+	servers := local.GenServers(n)
+	services := local.GetServices(servers, serviceID)
+	if services[0] == nil {
+		t.Log("The service is not available to test due to incompatible suite.")
+		return nil
```
kc1212: are we able to do something like `t.Skip` when this happens, I think it'll be cleaner than having `service == nil` in every test.

ineiti: That's even a
```
require.NotNil(t, services[0], "incompatible suite")
```
no?

jeffallen: It is not a fatal error: it is just that the current definition of cothority.Suite does not allow this service to be instantiated, so testing it is not possible. t.Skip is the right thing.

PR #1012, opened 2018-01-29
cleanup conode.go

ineiti:
making method in cli-definition and converting dedis-cothority.toml to hex

### DIFF
--- a/dedis-cothority.toml
+++ b/dedis-cothority.toml
@@ -1,40 +1,15 @@
 [[servers]]
-  Address = "tcp://78.46.227.60:7770"
+  Address = "tcp://conode.gasser.blue:7770"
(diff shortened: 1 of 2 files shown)


dedis-cothority.toml
```diff
@@ -1,39 +1,40 @@
 [[servers]]
   Address = "tcp://78.46.227.60:7770"
-  Public = "5eI+WFOaCdMhHY+g+zR11IZV4MBtg+k8jm59FqqHwQY="
+  Public = "e5e23e58539a09d3211d8fa0fb3475d48655e0c06d83e93c8e6e7d16aa87c106"
```
jeffallen: But this whole file is End Of Life, right? Don't we expect to manage this via CISC now?

ineiti: yes, soon... but I think we still should have a list of more reliable nodes. So I should perhaps clean this up.

PR #1013, opened 2018-01-29
Implement leader catchup

jeffallen:
If a leader does not know of the latest block that a client
mentions, but it does know of the chain, it can ask others in the
chain's roster to catch it up.

Test it by nuking blocks from the db in order to simulate what would
happen if you started a leader with a database backup that was out of
date.

Also: Test follower catchup, which was already working correctly.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -47,7 +47,6 @@ type Service struct {
 	verifiers          map[VerifierID]SkipBlockVerifier
 	blockRequestsMutex sync.Mutex
 	blockRequests      map[string]chan *SkipBlock
-	lastSave           time.Time
 	newBlocksMutex     sync.Mutex
 	newBlocks          map[string]bool
 	storageMutex       sync.Mutex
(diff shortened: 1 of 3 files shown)


skipchain/skipchain.go
```diff
@@ -145,8 +144,29 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, on
...
 		if prev == nil {
-			return nil, onet.NewClientErrorCode(ErrorBlockNotFound,
-				"Didn't find latest block")
+			// Did not find the block they claim is the latest.
+
+			// If we don't know this chain, we give up (so that
+			// they cannot make us run useless chainSyncs and attack
+			// other conodes).
+			var gen *SkipBlock
+			if gen = s.db.GetByID(psbd.NewBlock.SkipChainID()); gen == nil {
```
ineiti: Question: isn't
```
gen := s.db.GetByID(psbd.NewBlock.SkipChainID())
if gen == nil{
```
more go-idiomatic?

PR #1015, opened 2018-01-29
Skipchain improvement

ineiti:
new method to add public keys - this will be used by the identity-service to allow for secure creation of new blocks.

Plus a small fix for not searching nil-blocks.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -572,7 +572,7 @@ func (s *Service) AddFollow(add *AddFollow) (*EmptyReply, error) {
 			roster := onet.NewRoster([]*network.ServerIdentity{si})
 			last, err := s.getLastBlock(roster, add.SkipchainID)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


skipchain/skipchain.go
```diff
@@ -698,6 +698,18 @@ func (s *Service) NewProtocol(ti *onet.TreeNodeInstance, conf *onet.GenericConfi
 	return
 }
 
+// AddClientKey can be used by other services to add a key so
```
jeffallen: Where is the authentication on this? Why can't anyone who can talk on the websocket call this and add themselves?

ineiti: Because this is not exposed on the websocket and can only be called internally from one service to another.

One thing I'd like to do is some kind of Authentication-service, that can handle the authentication for all services. Now we have one for the identity, one for the skipchain and one for pop. And they don't all support {pin,public,pop} authentication. Which is crazy :(

PR #1017, opened 2018-01-31
Remove ClientError

jeffallen:
Also: export some named errors that callers can check if they need to.

### DIFF
--- a/pop/app.go
+++ b/pop/app.go
@@ -123,7 +123,7 @@ func orgLink(c *cli.Context) error {
 	addr := network.NewTCPAddress(fmt.Sprintf("%s:%s", addrs[0], port))
 	pin := c.Args().Get(1)
 	if err := client.PinRequest(addr, pin, cfg.OrgPublic); err != nil {
-		if err.ErrorCode() == service.ErrorWrongPIN && pin == "" {
+		if err == service.ErrorWrongPIN && pin == "" {
 			log.Info("Please read PIN in server-log")
 			return nil
 		}
--- a/pop/app.go
+++ b/pop/app.go
@@ -269,10 +269,10 @@ func orgFinal(c *cli.Context) error {
 		log.Lvl2("Final statement already here:\n", "\n"+string(finst))
 		return nil
 	}
-	fs, cerr := client.Finalize(cfg.Address, party.Final.Desc,
+	fs, err := client.Finalize(cfg.Address, party.Final.Desc,
 		party.Final.Attendees, cfg.OrgPrivate)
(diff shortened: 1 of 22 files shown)


pop/app.go
```diff
@@ -123,7 +123,7 @@ func orgLink(c *cli.Context) error {
 	addr := network.NewTCPAddress(fmt.Sprintf("%s:%s", addrs[0], port))
 	pin := c.Args().Get(1)
 	if err := client.PinRequest(addr, pin, cfg.OrgPublic); err != nil {
-		if err.ErrorCode() == service.ErrorWrongPIN && pin == "" {
+		if err == service.ErrorWrongPIN && pin == "" {
```
ineiti: That works without using the `.Error()`?

randhound/randhound.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/dedis/onet"
 	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
+	"gopkg.in/dedis/crypto.v0/abstract"
```
ineiti: that's probably wrong here

randhound/proof.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/dedis/kyber/poly"
 	"github.com/dedis/kyber/util/random"
 	"github.com/dedis/onet/crypto"
+	"gopkg.in/dedis/crypto.v0/abstract"
```
ineiti: that's probably wrong here

randhound/randhound.go
jeffallen: This is dead code that is waiting to be revived. I thought you were working on that?

pop/app.go
jeffallen: No, I mean I have taken care of that correction.

pop/service/api.go
```diff
@@ -57,17 +37,17 @@ func NewClient() *Client {
...
 }
 
 // StoreConfig sends the configuration to the conode for later usage.
-func (c *Client) StoreConfig(dst network.Address, p *PopDesc, priv kyber.Scalar) onet.ClientError {
+func (c *Client) StoreConfig(dst network.Address, p *PopDesc, priv kyber.Scalar) error {
 	si := &network.ServerIdentity{Address: dst}
 	sg, e := schnorr.Sign(cothority.Suite, priv, p.Hash())
 	if e != nil {
-		return onet.NewClientError(e)
+		return e
```
kc1212: nitpick: can we change it to `err` to be consistent?

PR #1018, opened 2018-01-31
kyberjs: add weierstrass form of elliptic curves

gnarula:
Can be used to implement NIST Curves. Provided parameters for P256


jeffallen:
Super cool. Is "npm test" working as the README mentions? If so, do you want to add it to .travis?

external/js/kyber/lib/group/nist/params.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+module.exports = {
+  p256: {
+    name: "P256",
+    bitSize: 256,
+    p: "ffffffffffffffffffffffff00000000000000000000000001000000ffffffff",
```
jeffallen: It seems surprising to me that something so fundamental as the p256 parameters would be stored in strings. How do we know they are parsed one single time into the internal format of the BN library?

external/js/kyber/lib/group/nist/point.js
```diff
@@ -0,0 +1,357 @@
...
+ */
+Point.prototype.unmarshalBinary = function(bytes) {
+  if (bytes.constructor !== Uint8Array) {
+    throw new TypeError();
+  }
+  const byteLen = this.ref.curve._coordLen();
+  if (bytes.length != 1 + 2 * byteLen) {
+    throw new Error();
+  }
+  if (bytes[0] != 4) {
```
jeffallen: A comment here explaining why 4 would be helpful. (And a mention in the docs that only only format 4 encoding is accepted. It seems like this error is missing an error message?

external/js/kyber/lib/group/nist/point.js
```diff
@@ -0,0 +1,357 @@
...
+    throw new Error();
+  }
+  let x = new BN(bytes.slice(1, 1 + byteLen), 16);
+  let y = new BN(bytes.slice(1 + byteLen), 16);
+  let zero = new BN(0);
```
jeffallen: Is there a way in javascript to put a constant like this in a place where it is initialized one single time?

external/js/kyber/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "elliptic",
+    "curve25519",
+    "nist"
+  ],
+  "author": "Gaurav Narula <gnarula94@gmail.com>",
+  "homepage": "https://github.com/dedis/cothority",
+  "dependencies": {
+    "bn.js": "^4.11.8",
+    "elliptic": "^6.4.0",
+    "jsdoc-to-markdown": "^4.0.1"
```
jeffallen: That seems like it should be a dev dependency?

external/js/kyber/lib/util/index.js
```diff
@@ -0,0 +1,35 @@
+"use strict";
+
+/**
+ * Class PRNG defines a PRNG using a Linear Congruent Generator
+ * Javascript doesn't have a seedable PRNG in stdlib and this
+ * implementation is to be used for deterministic outputs in tests
```
jeffallen: This is a dangerous thing to include in a crypto library. I understand it is only for test, so then it needs to NOT be in the main library.

external/js/kyber/README.md
```diff
@@ -0,0 +1,48 @@
+KyberJS
```
jeffallen: You need two disclaimers in here in big bold writing: 1. This is developmental, and not ready for protecting production data. 2. This is not a constant time implementation, and likely has timing side channels that can be attacked.

external/js/kyber/test/group/nist/index.js
```diff
@@ -0,0 +1,755 @@
+const kyber = require("../../../index.js");
```
jeffallen: We should see about creating a set of test vectors using a reference impl of p256 (maybe boringSSL would be a nice choice), and then ensuring that both Kyber-go and kyber-js calculate all of them correctly.

gnarula:
Done :)

jeffallen:
The build failure was due to flakiness in cothority. I have restarted the build.

We've GOT to get that fixed. How embarrassing. :)

external/js/kyber/lib/group/nist/params.js
gnarula: ah, right now every Curve instance will end up creating a new BN object for every curve param. Better to initialize them only once in `params.js`

external/js/kyber/lib/util/index.js
gnarula: right, moving this into `test/`

external/js/kyber/lib/group/nist/point.js
```diff
@@ -0,0 +1,359 @@
...
+ * Returns a Point with data embedded in the y coordinate
+ *
+ * @param {Uint8Array} data - data to embed with length <= embedLen
+ * @param {function} [callback] - to generate a random byte array of given length
+ *
+ * @throws {TypeError} if data is not Uint8Array
+ * @throws {Error} if data.length > embedLen
+ * @return {module:group/nist~Point}
+ */
+Point.prototype.embed = function(data, callback) {
```
nikkolasg: Is there no "stream" like things in Javascript when you just call stg like `stream.Next(32)` to get the next 32 bytes of output ? 
I guess it's not very Javascript-like, so at least is there somewhere where you provide a `random callback` using https://developer.mozilla.org/en-US/docs/Web/API/RandomSource/getRandomValues ? If not that'd be great to have, so developers don't have to always code the same callback.

external/js/kyber/lib/group/nist/point.js
```diff
@@ -0,0 +1,359 @@
...
+  let dl = this.embedLen();
+  if (data.length > dl) {
+    throw new Error("data.length > dl");
+  }
+
+  if (dl > data.length) {
+    dl = data.length;
+  }
+
+  callback = callback || crypto.randomBytes;
```
nikkolasg: Oh sorry that's there :)

gnarula:
@nikkolasg added an example index.html and another one in README :)

PR #1019, opened 2018-02-01
pass correct error message

ineiti:
The `var` defined the error message for a wrong pin, but the app needed an error message for an empty pin. Changed the error message and referenced the correct package.


jeffallen:
Sorry I scrambled that up.

PR #1025, opened 2018-02-02
Adding schnorr signatures to JS

nikkolasg:
This adds another module `kyber.sign.schnorr.{sign,verify}` to sign and verify schnorr signatures. So far it only uses what's available p256. When the ed25519 code from @gnarula will be ready, I'll try to add a compatibility test with eddsa verification.

This also adds coverage reports locally and on travis. Just run `npm run cover` to see a good resume of the current coverage.

### DIFF
--- a/external/js/kyber/lib/sign/schnorr/schnorr.js
+++ b/external/js/kyber/lib/sign/schnorr/schnorr.js
@@ -1,0 +1,128 @@
+const group = require("./../../group");
+// XXX to be changed to an interface style
+const nist = group.nist;      
+const curve = nist.Curve;
+const scalarCons = nist.Scalar;
+const pointCons = nist.Point;
+
+const hash = require("hash.js");
+/**
+ *
+ * Sign computes a Schnorr signature over the given message.
+ * 
+ * @param privateKey private key scalar to sign with
+ * @param message message over which the signature is computed
+ * @return signature as a Uint8Array
+ * */
+function Sign(suite, privateKey, message) {
+    if (suite.constructor !== curve) {
+        throw "first argument must be a suite";
+    }
+    if (privateKey.constructor !== scalarCons) {
+        throw "second argument must be a scalar";
+    }
+    if (message.constructor !== Uint8Array) {
+        throw "third argument must be Uint8Array";
+    }
+
+    // generate r & R
+    const r = suite.scalar().pick();
+    const R = suite.point().mul(r,null);
+    const buffR = R.marshalBinary();
+
+    // generate public key
+    const pub = suite.point().mul(privateKey,null);
+    
+    // generate challenge
+    const challenge = hashSchnorr(suite,buffR, pub.marshalBinary(), 
+                         message) 
+
+    // generate signature
+    const s = suite.scalar().mul(privateKey,challenge)
+    s.add(s,r)
+
+    // concatenate R || s
+    const buffS = s.marshalBinary();
+    const buffSig = new Uint8Array(buffR.length + buffS.length);
+    buffSig.set(buffR);
+    buffSig.set(buffS,buffR.length);
+    return buffSig;
+}
+
+/**
+ * 
+ * Verify verify if the signature of the message is valid under the given public
+ * key.
+ *
+ * @params suite suite to use
+ * @params publicKey public key under which to verify the signature
+ * @params message message that is signed
+ * @params signature signature made over the given message
+ * @return boolean true if signature is valid or false otherwise.
+ * */
+function Verify(suite, publicKey, message, signature) {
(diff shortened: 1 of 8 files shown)


external/js/kyber/lib/sign/schnorr/schnorr.js
```diff
@@ -0,0 +1,128 @@
+const group = require("./../../group");
+// XXX to be changed to an interface style
```
nikkolasg: @gnarula is also doing the work of putting an generic interface, like Kyber. Once ready i'll move this code towards using it.

external/js/kyber/lib/sign/schnorr/schnorr.js
```diff
@@ -0,0 +1,128 @@
...
+                         message) 
+
+    // generate signature
+    const s = suite.scalar().mul(privateKey,challenge)
+    s.add(s,r)
+
+    // concatenate R || s
+    const buffS = s.marshalBinary();
+    const buffSig = new Uint8Array(buffR.length + buffS.length);
+    buffSig.set(buffR);
```
ineiti: I'm a JS-noob: how can you set a `const` to another value? Should `const buffSig` with `let buffSig`?

nikkolasg: I don't understand. I'm also not a JS expert, but it's the only way I know of filling a buffer. I can use let probably but why should I, if it's meant to be a constant anyway ? (i.e. the variable is meant to always be this `Uint8Array` but the content of the array may change)

ineiti: Looking at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const it seems that this should be replaced with `let buffSig`.

nikkolasg:
Removed the whitespaces I've seen !

external/js/kyber/lib/sign/schnorr/schnorr.js
nikkolasg: Why ? 
> The const declaration creates a read-only reference to a value. It does not mean the value it holds is immutable, just that the variable identifier cannot be reassigned. For instance, in the case where the content is an object, this means the object's contents (e.g., its parameters) can be altered.
So it's exactly this case. What's wrong ?

ineiti: OK - I didn't see this on the link from above. Where do you read this?

ineiti:
I propose to use @gnarula 's suggestion: https://github.com/w0rp/ale - what do you think?

external/js/kyber/lib/sign/schnorr/schnorr.js
nikkolasg: Second paragraph of description !

nikkolasg:
I've installed it and used it. Works well :) (and will likely be useful for any other languages BTW)

external/js/kyber/lib/sign/schnorr/schnorr.js
```diff
@@ -0,0 +1,127 @@
...
+    throw "third argument must be Uint8Array";
+  }
+
+  // generate r & R
+  const r = suite.scalar().pick();
+  const R = suite.point().mul(r, null);
+  const buffR = R.marshalBinary();
+
+  // generate public key
+  const pub = suite.point().mul(privateKey, null);
```
jeffallen: This is surprising, I would have expected that the private and public keys arrive together in a keypair object, so that the public does not need to be derived on each call. Then I went to the Go implementation and was also surprised to see the same thing there.

So: I'm not asking you to change this, but why is it this way?

external/js/kyber/lib/sign/schnorr/schnorr.js
```diff
@@ -0,0 +1,127 @@
...
+  const buffS = s.marshalBinary();
+  const buffSig = new Uint8Array(buffR.length + buffS.length);
+  buffSig.set(buffR);
+  buffSig.set(buffS, buffR.length);
+  return buffSig;
+}
+
+/**
+ *
+ * Verify verify if the signature of the message is valid under the given public
```
jeffallen: s/Verify verify/Verify verifies/

external/js/kyber/lib/sign/schnorr/schnorr.js
nikkolasg: I don't have a definite answer but I would say that it's much safer and simpler this way, even for a little performance penalty. If you give a public key as argument, you should make sure it's the corresponding public key from the secret key otherwise your signature will be wrong. To do that check, you anyway need to do the scalar base multiplication.

PR #1026, opened 2018-02-02
correcting wrong assumptions of toml-files

ineiti:
Fixes a bug in the shell-parsing of toml-files

### DIFF
--- a/pop/app.go
+++ b/pop/app.go
@@ -141,21 +141,21 @@ func orgConfig(c *cli.Context) error {
 	pdFile := c.Args().First()
 	buf, err := ioutil.ReadFile(pdFile)
 	if err != nil {
-		return fmt.Errorf("error: %s - while reading: %s", err, pdFile)
+		return fmt.Errorf("error: %s - while reading pop-description: %s", err, pdFile)
 	}
 	err = decodePopDesc(string(buf), desc)
 	if err != nil {
-		return fmt.Errorf("error: %s - while decoding: %s", err, pdFile)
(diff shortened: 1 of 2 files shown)


pop/app.go
```diff
@@ -653,6 +653,7 @@ type shortDescGroupToml struct {
 // decode config of several groups into array of rosters
 func decodeGroups(buf string) ([]*service.ShortDesc, error) {
 	decodedGroups := make(map[string][]shortDescGroupToml)
+	log.Print(buf)
```
jeffallen: Is this left in on purpose?

pop/test.sh
```diff
@@ -411,10 +411,10 @@ EOF
 Location = "Earth, City$n"
 EOF
 		echo "[[parties.servers]]" >> pop_merge.toml
-		sed -n "$((5*$n-4)),$((5*$n))p" public.toml >> pop_merge.toml
+		sed -n "$((5*$n-3)),$((5*$n))p" public.toml >> pop_merge.toml
```
jeffallen: These gave me incredible difficulty the last time I had to debug them. How about the next time someone has to touch them, they find a better solution than these fiddly undocumented integres? I promise if it's me I'll do it!

ineiti: I think the integers are OK - I would care more about having a way to correctly cut up the `public.toml` into different servers and then merge them. In fact, we _do_ have that...

ineiti: OK, added it to the code.

PR #1030, opened 2018-02-05
JS:  adding protobuf definitions and network interface

nikkolasg:
That PR creates a `cothority` javascript package that sends and receives protobuf packets automatically. A reasonable list of protobuf packet definitions have been taken from [the student repo](https://github.com/dedis/student_17_mobile/tree/master/cpmac/app/shared/lib/dedjs/protobuf/src/models).

This module is not *finished* yet, but it already offers a basic interface to speak to the cothority and is tested. I would prefer go step by step and already try to review/change/commit on this one than rather going along "too fast".


external/js/cothority/dist/bundle.min.js
```diff
@@ -0,0 +1 @@
...ray(16):i?new i(16):new Array(16)).constructor=b,b.prototype.make=function(t){var r;if(1===t){var i=(new Date).getTime();i!==g?m=0:m++,g=i;var n,o=l(i);!function(e,t){for(var r=0,i=0;i<8;i++)r+=e[i]*t,e[i]=Math.floor(r%256),r=Math.floor(r/256)}(o,1e4),c(o,a(1,178,29,210,19,129,64,0)),m>0&&c(o,l(m)),n=p(o,8),this[3]=255&n,n=p(o,8),this[2]=255&n,n=p(o,8),this[1]=255&n,n=p(o,8),this[0]=255&n,n=p(o,8),this[5]=255&n,n=p(o,8),this[4]=255&n,n=p(o,8),this[7]=255&n,n=p(o,8),this[6]=15&n;var s=e(2,255);this[8]=s[0],this[9]=s[1];var u=e(6,255);for(u[0]|=1,u[0]|=2,r=0;r<6;r++)this[10+r]=u[r]}else if(4===t){var f=e(16,255);for(r=0;r<16;r++)this[r]=f[r]}else{if(3!==t&&5!==t)throw new Error("UUID: make: invalid version");var h="",d="object"==typeof arguments[1]&&arguments[1]instanceof b?arguments[1]:(new b).parse(arguments[1]);for(r=0;r<16;r++)h+=String.fromCharCode(d[r]);h+=arguments[2];var w=3===t?v(h):y(h);for(r=0;r<16;r++)this[r]=w.charCodeAt(r)}return this[6]&=15,this[6]|=t<<4,this[8]&=63,this[8]|=128,this},b.prototype.format=function(e){var r,i;return"z85"===e?r=function(e,t){if(t%4!=0)throw new Error("z85_encode: invalid input length (multiple of 4 expected)");for(var r="",i=0,o=0;i<t;)if(
```
jeffallen: Why are minified files checked in? To me, these are like binaries, they don't belong in Git.

external/js/cothority/lib/misc/misc.js
```diff
@@ -0,0 +1,60 @@
...
+exports.reverseHex = reverseHex;
+
+/**
+ * Convert a byte buffer to a hexadecimal string.
+ * @param {Uint8Array} buffer
+ *
+ * @throws {TypeError} when buffer is not Uint8Array
+ * @returns {string} hexadecimal representation
+ */
+function uint8ArrayToHex(buffer) {
```
jeffallen: I don't see where these are used. Why are they in this library? If they are used for testing, can they be in the testing code? Also: error handling for invalid hex is not done.

external/js/cothority/lib/protobuf/build/build_proto.js
```diff
@@ -0,0 +1,19 @@
+const protobuf = require('protobufjs');
```
jeffallen: Are these auto-generated? Can you add some docs on how to generate them?

external/js/cothority/test/net/net.js
```diff
@@ -0,0 +1,45 @@
+const mocha = require("mocha");
```
jeffallen: is mocha unused? Or is it the magic that makes the describe keyword? (Sorry, I am a JS idiot.)

external/js/cothority/dist/bundle.min.js
nikkolasg: I see your point. Mine was to be able to already offer an online working version of the script directly (just `<script src="https://github.com/dedis/cothority/external/js/kyber"></script>`). But that may not be the best option indeed. We need to think of a strategy to be able to deliver online scripts as these. 
I'll remove it.

external/js/cothority/lib/protobuf/build/build_proto.js
nikkolasg: So the `protobufjs` is an external protobuf library. From that and the list of protobuf definitions we generate a JSON representation, that gets embedded in the general library. 
There is an instruction on how to build the JSON representation (if you wanna add a new protobuf definition) in the README.

external/js/cothority/lib/misc/misc.js
nikkolasg: Yep I've added them a little bit sooner than expected. I'm working on the next PR and I'm using them only now. Will remove them in this PR.

external/js/cothority/test/net/net.js
nikkolasg: Indeed I think it's not needed to import it. The Mocha binary called by `npm run test` so I guess it's already "imported" by default.

PR #1032, opened 2018-02-06
JS: introducing skipchain client

nikkolasg:
This PR introduces different JS module to be able to talk to a `Skipchain` service
 - `cothority.ServerIdentity` / `Roster`: simply the JS translation from the Go structures.
 - `net.RosterSocket`: tries different addresses in a random order from a roster until one replies. Useful for load balancing and reliability of the JS code.
 - `skipchain.Client`: uses a `RosterSocket` to contact the `Skipchaiǹ` service. The only functionality so far is to retrieve the latest updates, it "follows" the chain. The test to verify the signature are implemented but not yet tested, it will be once the Ag**** project comes to close to conclusion (by the end of February I expect).
The test for skipchains actually runs  a Go binary that runs some conode, create a skipchain and puts some data in it, so interoperability with the Go code can now be tested by Travis :)


external/js/cothority/lib/identity.js
```diff
@@ -0,0 +1,146 @@
...
+     * */
+    get id() {
+        throw new Error("not implemented yet");
+    }
+
+    /**
+    * Parse cothority roster toml string into a Roster object.
+    * @example
+    * // Toml needs to adhere to the following format
+    * // where public has to be a base64 encodable string.
```
jeffallen: Watch out! We do not use base64 in cothority anymore.

nikkolasg: Yeah I switched to hex but forgot to change the comments, thanks ;)

external/js/cothority/lib/misc/misc.js
```diff
@@ -36,6 +36,45 @@ function hexToUint8Array(hex) {
     });
 }
 
+/**
```
jeffallen: I really really really hate having a lib/misc module here. misc and/or util packages are dumping grounds for all the things we haven't thought carefully about.

I'm asking myself two things: 1. can this compare code be in the test directory? 2. surely there is a well-known module in the NPM universe for uint8ArrayToHex and hexToUint8Array (for god's sake, they even had a module called "left_pad"!)

I know I'm being a Go prima-donna here. If you and @gnarula tell me "this is just how it in JS" I'll believe you and shut up.

external/js/cothority/lib/identity.js
```diff
@@ -0,0 +1,227 @@
...
+ * */
+class ServerIdentity {
+    /*
+     * Returns a new ServerIdentity from the public key and address.
+     * @param {Uint8Array} public key in bytes format
+     * @param {string} address tcp address of the cothority node
+     * @param {string} description of the conode. Can be null.
+     * @return a ServerIdentity
+     * */
+    constructor(publicKey, address,description) {
```
jeffallen: s/,/, /

I miss gofmt.

external/js/cothority/lib/identity.js
```diff
@@ -0,0 +1,227 @@
...
+        if (publicKey.constructor !== Uint8Array)
+            throw TypeError;
+        this.pub = publicKey
+        this.addr = address;
+        this._description = description;
+        // id of the identity
+        const url = 'https://dedis.epfl.ch/id/' + misc.uint8ArrayToHex(this.pub);
+        this._id = new UUID(5, 'ns:URL', url).export();
+        // tcp + websocket address
+        let parts = address.replace("tcp://", "").split(":");
```
jeffallen: address will shortly start with tls:// and this will break.

external/js/cothority/lib/identity.js
```diff
@@ -0,0 +1,227 @@
...
+        return this.tcpAddr;
+    }
+
+    /**
+     * point returns the point representation of the public key
+     *
+     * @param {kyber.Group} group the expected group from which the public key belongs
+     * @returns {kyber.Point} the kyber.Point representing the public key
+     */
+    point(group) {
```
jeffallen: It seems strange to postpone identifying the group until here. Why doesn't the constructor take a point as it's first argument?

external/js/cothority/lib/identity.js
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    /**
+     * aggregateKey returns the aggregate public key for this server.
+     * It is the sum of all public keys of the identities of this Roster.
+     *
+     * @param {kyber.Group} group The group to use to compute the aggregate key.
+     * @returns {kyber.Point} The aggregate key
+     */
+    aggregateKey(group) {
```
jeffallen: As a continuation of the previous comment: if the point was stored in the identity, you'd already know the group and not need to pass it in here.

external/js/cothority/lib/identity.js
```diff
@@ -0,0 +1,227 @@
...
+    *
+    * @throws {TypeError} when toml is not a string
+    * @return {Roster} roster
+    */
+    static fromTOML(toml) {
+        if (typeof toml !== 'string')
+            throw new TypeError;
+
+        const roster = topl.parse(toml);
+        const identities = roster.servers.map((server) =>
```
jeffallen: There is now a Suite line in the TOML files (if not present then suite = "Ed25519"). You can use that to look up the group by name use it to unmarshal the hex into a point.

external/js/cothority/lib/misc/misc.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const BASE64 = require("base-64");
```
jeffallen: No longer used?

external/js/cothority/lib/misc/misc.js
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Compares two Uint8Array buffers. If constant is true, it does the comparison in constant time.
+ *
+ * @param {Uint8Array} arr1 first buffer
+ * @param {Uint8Array} arr2 second buffer
+ * @param {Boolean} constant if true, constant time comparison is done.
+ * @returns {Boolean} True if both buffers are equals, false otherwise.
+ */
+function uint8ArrayCompare(arr1, arr2,constant) {
```
jeffallen: s/,/, /

external/js/cothority/lib/net/net.js
```diff
@@ -3,27 +3,28 @@
 const topl = require('topl');
 const UUID = require('pure-uuid');
 const protobuf = require('protobufjs')
+const co = require('co');
+const shuffle = require("crypto-shuffle");
```
jeffallen: single quotes (does this matter to JS, or is it just for OCD people like me?)

external/js/cothority/lib/protobuf/build/models/skip/skipblock.proto
```diff
@@ -17,8 +17,8 @@ message SkipBlock {
     required bytes data = 9;
     required Roster roster = 10;
     optional bytes hash = 11;
-    optional BlockLink forward = 12;
-    optional BlockLink children = 13;
+    repeated BlockLink forward = 12;
```
jeffallen: Where is the code that generates these? Why did it generate them wrong before?

external/js/cothority/lib/skipchains/skipchains.js
```diff
@@ -0,0 +1,120 @@
...
+        this.group = group;
+    }
+
+    /**
+     * Returns the latest known skipblockID
+     *
+     * @returns {string} hexadecimal encoded skipblockID
+     */
+    get latestID() {
+        return misc.uint8ArrayToHex(this.lastID);
```
jeffallen: If there was a separate javascript object for SkipblockID, then you could implement toString on it, and this would be a bit more readable/automatic right?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString

external/js/cothority/lib/skipchains/skipchains.js
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * updateChain asks for the latest block of the skipchain with all intermediate blocks.
+     * It automatically verifies the transition from the last known skipblock ID to the
+     * latest one returned. It also automatically save the latest good known
+     * roster from the latest block.
+     * @return {Promise} A promise which resolves with the latest skipblock if
+     * all checks pass.
+     */
+    latestBlock() {
+        const requestName = "GetUpdateChain";
```
jeffallen: Sorry, yesterday I changed the behavior of GetUpdateChain in #1034. For now, this is good enough, but to handle evolving rosters, the client needs to check that latest is in fact a terminal, and if not, it needs to keep calling GetUpdateChain on the latest.Roster, to keep following the chain as it moves to other servers.

external/js/cothority/lib/skipchains/skipchains.js
```diff
@@ -0,0 +1,120 @@
...
+     * @returns {string} hexadecimal encoded skipblockID
+     */
+    get latestID() {
+        return misc.uint8ArrayToHex(this.lastID);
+    }
+
+    /**
+     * updateChain asks for the latest block of the skipchain with all intermediate blocks.
+     * It automatically verifies the transition from the last known skipblock ID to the
+     * latest one returned. It also automatically save the latest good known
```
jeffallen: save? Where? I don't see this yet...

external/js/cothority/lib/skipchains/skipchains.js
```diff
@@ -0,0 +1,120 @@
...
+
+            // move to the next block
+            currBlock = nextBlock;
+        }
+    }
+
+    /**
+     * verify if the link is a valid signature over the given message for the given roster
+     *
+     * @param {Roster} roster THe roster who created the signature
```
jeffallen: s/TH/Th/

external/js/cothority/lib/skipchains/skipchains.js
```diff
@@ -0,0 +1,120 @@
...
+            // the skipchain, i.e. we skip the biggest number of blocks
+            const lastLink = forwardLinks[forwardLinks.length-1];
+            // XXX to change later to source_hash, dest_hash, dst_roster_id
+            const message = nextBlock.hash;
+            const roster = identity.Roster.fromProtobuf(currBlock.roster);
+            if (!this.verifyForwardLink(roster,message,lastLink))
+                return [null,new Error("Forward link incorrect!")];
+
+            // move to the next block
+            currBlock = nextBlock;
```
jeffallen: currBlock and nextBlock are confusing here, it would be easier with simply block[i-1] and block[i].

external/js/cothority/lib/misc/misc.js
nikkolasg: > I really really really hate having a lib/misc module here. misc and/or util packages are dumping grounds for all the things we haven't thought carefully about.

While it may not be the best thing to do, it is still tested and kept under control. If you want I can rename the package "bitutils" to make it clearer ? I'm not sure how it's done in JS but relying on an external library for something as trivial as that does not seem like better to me. The less dependency the better.

external/js/cothority/lib/identity.js
nikkolasg: On the top level then ? There is not now at least.  
https://github.com/dedis/cothority/blob/master/dedis-cothority.toml
I'll try to implement the suite search functionality in kyber-js then.

external/js/cothority/lib/identity.js
nikkolasg: I now split via the `://` separation "word".

external/js/cothority/lib/identity.js
nikkolasg: This is done at the top level currently even if it does not exists, and it defaults to edwards25519

external/js/cothority/lib/net/net.js
nikkolasg: Since I installed the eslint-er it seems everything is put as double quotes. And from what I know it does not matter.

nikkolasg:
> Where is the code that generates these? Why did it generate them wrong before?

My guess is it was generated by hand. We would need to make a go parser script that parse all of the structs of the codebase, and put that in protobuf, and then, by hand, manually exclude the one that don't make sense. Or use a whitelist of struct name. Stg like that. It does not exists at the moment ;)

So I just corrected the definition by hand for the moment to be able to receive a list of forward links instead of just one.

nikkolasg:
> Sorry, yesterday I changed the behavior of GetUpdateChain in #1034. For now, this is good enough, but to handle evolving rosters, the client needs to check that latest is in fact a terminal, and if not, it needs to keep calling GetUpdateChain on the latest.Roster, to keep following the chain as it moves to other servers.
What do you mean by a "terminal" roster ? 
WHat I'm doing now is checking if I got the latest data from "the" roster, and "the" roster is set continuously to each skipblock I process.

external/js/cothority/lib/skipchains/skipchains.js
nikkolasg: Sorry. Save in a variable ? :D

external/js/cothority/lib/skipchains/skipchains.js
nikkolasg: Ok. I find it much easier in my mind to process `"the current block"/"the next block"` rather than `"block[i]" / "block[i-1]"` because it does not tell which block are you verifying at the moment. I'm verifying the current block.

ineiti:
node-tests fail - moving back to WIP.

jeffallen:
In https://github.com/dedis/cothority/pull/1038 I made it so that if a WS client asks a conode for updates, it will only give updates from it's own database. If the last block it knows about has a forward link onward to some other conodes, because this conode is no longer in the roster for this chain, then the client will need to make a new WS connection to continue to follow the chain with the new roster.

(I haven't looked at your most recent code yet to see if it's already working this way.)

external/js/cothority/lib/skipchain/skipchain.js
```diff
@@ -0,0 +1,217 @@
...
+          data = yield client.socket.send(requestStr, responseStr, request);
+        } catch (err) {
+          return Promise.reject(err);
+        }
+        // verifies it's all correct
+        var lastBlock, err;
+        [lastBlock, err] = client.verifyUpdateChainReply(data);
+        if (!err) {
+          // tries again with random conodes
+          nbErr++;
```
jeffallen: It seems like there is a missing continue here.

external/js/cothority/lib/skipchain/skipchain.js
```diff
@@ -0,0 +1,217 @@
+//"use strict";
```
jeffallen: 💥 👎

external/js/cothority/lib/skipchain/skipchain.js
```diff
@@ -0,0 +1,217 @@
...
+      if (forwardLinks.length == 0)
+        return [null, new Error("No forward links included in the skipblocks")];
+
+      // only take the highest link since we move "as fast as possible" on
+      // the skipchain, i.e. we skip the biggest number of blocks
+      const lastLink = forwardLinks[forwardLinks.length - 1];
+      // XXX to change later to source_hash, dest_hash, dst_roster_id
+      const message = nextBlock.hash;
+      const roster = identity.Roster.fromProtobuf(currBlock.roster);
+      //var err = this.verifyForwardLink(roster, message, lastLink);
```
jeffallen: Is this left commented out on purpose?

external/js/cothority/lib/skipchain/skipchain.js
```diff
@@ -0,0 +1,217 @@
...
+        [lastBlock, err] = client.verifyUpdateChainReply(data);
+        if (!err) {
+          // tries again with random conodes
+          nbErr++;
+        }
+        // update the roster
+        client.lastRoster = identity.Roster.fromProtobuf(lastBlock.roster);
+        client.lastID = lastBlock.hash;
+        // if there is nothing new stop !
+        if (!lastBlock.forward || lastBlock.forward.length == 0) {
```
nikkolasg: @jeffallen See here. I think that's what I do. I take the latest roster available. See if there is any forward link. If there are any left I continue to contact the cothority with the latest roster to get updates.

external/js/cothority/lib/skipchain/skipchain.js
nikkolasg: Yes because I'd like to advance on the Agora stuff as quickly as possible so the code that verifies is there but untested. It will be once I have more time.

PR #1033, opened 2018-02-06
kyberjs: add edwards25519

gnarula:
Adds support for ed25519 curve. Tests check for multiplication with vectors from http://ed25519.cr.yp.to/python/sign.input


jeffallen:
I know this isn't ready for review yet, but I was wondering why the docs do not show the return value. There's a => like there's supposed to be something on the right side of the arrow, but nothing is there.  https://github.com/dedis/cothority/blob/9909363df93e48fa29d52e5215049cfc022e70a6/external/js/kyber/doc/doc.md#Sign

nikkolasg:
Great job. Could you add a big issue (with important tag or stg) specifying that this current implementation of ed25519 is not constant time, and should be fixed as soon as possible ? One example to inspire ourself fo do that is the NaCl JS port: https://github.com/dchest/tweetnacl-js/blob/master/nacl.js#L680 
Since we're quite busy this month, I won't push for having a constant time implementation now but it's something that really needs to be done.
Thanks again !

Looking good to merge for me.

gnarula:
@jeffallen that was because the comments were not according to the JSDoc syntax. Fixed it in the latest commit.

@nikkolasg sure, will open an issue when this gets merged

ineiti:
Just went very fast over it - did you create some random test-vectors from the go-kyber library to test against?

gnarula:
Yes, used the same byte arrays in go and tested the output against the go implementation. Where the public interface didn't allow getting a particular value (like say the X coordinate)  I made sure it is the same numerical value which you get from ed25519.py on ed25519 site.

I didn't run the test vectors from ed25519 site  on go though

PR #1038, opened 2018-02-08
Finish removing SendRaw from skipchain

### LINKED ISSUE #1034: Finish removing of `SendRaw` in skipchain
After @jeffallen work on skipchains to update using a protocol instead of `SendRaw`, one of the methods still needs to be taken care of. `GetUpdateChain` still uses `SendRaw` to fetch the latest block.

Suggested work:
- remove `callGetBlock`, `deprecatedProcessorGetBlockReply` and `deprecatedProcessorGetBlock` altogether and replace calls to `callGetBlock` with calls to `syncChain`.

jeffallen:
Change GetUpdateChain to operate on local blocks only. Delegate
responsibility to the client to go contact other servers according
to the latest roster (implemented in api.go, so that existing
clients do not see a change in behaviour).

Also: Added a test to check a worry I had about if new servers
which were invited into a chain would work correctly.

Fixes #1034.

### DIFF
--- a/skipchain/api.go
+++ b/skipchain/api.go
@@ -2,6 +2,7 @@ package skipchain
 
 import (
 	"errors"
+	"fmt"
 	"math/rand"
 
 	"github.com/dedis/cothority"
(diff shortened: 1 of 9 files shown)


skipchain/api.go
```diff
@@ -177,10 +177,33 @@ func (c *Client) CreateRootControl(elRoot, elControl *onet.Roster,
...
-	r := roster.RandomServerIdentity()
-	err = c.SendProtobuf(r,
-		&GetUpdateChain{latest}, reply)
-	return
+	for {
+		r2 := &GetUpdateChainReply{}
+		r := roster.RandomServerIdentity()
+		err = c.SendProtobuf(r, &GetUpdateChain{LatestID: latest}, r2)
+		if err != nil {
+			return
```
ineiti: I think you should not return here but try another server from the list. Either iterate through the roster, or randomly iterate through it. But some of the servers might be down.

skipchain/api_test.go
```diff
@@ -99,6 +99,73 @@ func TestClient_CreateRootInter(t *testing.T) {
...
+				h1 := sb1.Height
+				h2 := sb2.Height
+				log.Lvl3("sbc1.Height=", sb1.Height)
+				log.Lvl3("sbc2.Height=", sb2.Height)
+
+				height := h1
+				if h2 < height {
+					height = h2
+				}
+				if !bytes.Equal(sb1.ForwardLink[height-1].Hash(),
```
ineiti: `Hash` is a `SkipBlockID` which has an `Equal`. So this could be:
```
require.True(t, sb1.ForwardLink[height-1].Hash().Equal(sb2.Hash),
"Forward-pointer of update", up, "is different from hash in", up+1 )
```
just in case a student happens to read our code to know how to use things ;)

skipchain/api.go
```diff
@@ -177,10 +177,33 @@ func (c *Client) CreateRootControl(elRoot, elControl *onet.Roster,
...
+		}
+		// Why would this happen?
+		if len(r2.Update) == 0 {
```
kc1212: If this shouldn't happen, then we need to at least log something or return an error?

nikkolasg: There may be malicious servers that can return this. So it can happen, and by Murphy's law, will happen. So the best thing here is to retry another server again as @ineiti suggested a few lines above.

nikkolasg:
I'm allowing to comment since I'm coding the same functionality JS side. It's not the purpose of this PR but it should be the case that this function verifies the given received blocks if they are correct or not. For hte moment it's just "trust this server to get the right blocks" and no verification is done at the moment.  
I'm trying to do this verification at the moment, you can have a look [here](https://github.com/dedis/cothority/pull/1032/files#diff-81d9d13de34e000fb4ce16889ba53546). I haven't yet done the "contact the newest roster as long as you can" thing again.

skipchain/api.go
```diff
@@ -177,10 +177,33 @@ func (c *Client) CreateRootControl(elRoot, elControl *onet.Roster,
...
+			return
+		}
+		// The reply always has the original block they
+		// searched for. The first time through, keep it.
+		if reply.Update == nil {
+			reply.Update = append(reply.Update, r2.Update...)
```
nikkolasg: BTW is it possible to avoid a copy here just by using `reply.Update = r2.Update` ? (since `r2.Update` is anyway `nil`.

ineiti:
rebase and resolve conflicts

skipchain/skipchain.go
```diff
@@ -1285,11 +1218,6 @@ func newSkipchainService(c *onet.Context) (onet.Service, error) {
 		s.GetSingleBlock, s.GetSingleBlockByIndex, s.GetAllSkipchains,
 		s.CreateLinkPrivate, s.Unlink, s.AddFollow, s.ListFollow,
 		s.DelFollow, s.Listlink))
-	s.ServiceProcessor.RegisterStatusReporter("Skipblock", s.db)
```
ineiti: Why is there no status reporter anymore?

jeffallen: Merge damage. Fixing.

Issue #1036, opened 2018-02-07
Remove example protocols


ineiti:
The example protocol has been superseded by cothority_template. Unfortunately it's still used in one place: #940

ineiti:
Closed by #1043

PR #1042, opened 2018-02-08
Clean up status service, client

### LINKED ISSUE #940: status/service/status_test.go depends on example
With the new stable version, the file `status/service/status_test` uses the example-channel.

BRANCH: v1.2
FILE: status/service/status_test
TODO: use the `Count`-protocol for testing if all children are available.

jeffallen:
Clean up the type of the status reply, remove a call to an example
that will disappear with #1036.

Correctly display status replies which have multiple parts of the
server reporting stats.

Fixes #940.

### DIFF
--- a/skipchain/struct.go
+++ b/skipchain/struct.go
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strconv"
 	"time"
 
 	bolt "github.com/coreos/bbolt"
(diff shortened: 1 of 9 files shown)


skipchain/struct.go
```diff
@@ -522,6 +523,21 @@ func NewSkipBlockDB(db *bolt.DB, bn string) *SkipBlockDB {
 	}
 }
 
+// GetStatus is a function that returns the status report of the db.
+func (db *SkipBlockDB) GetStatus() onet.Status {
+	out := make(map[string]string)
+	db.DB.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(db.bucketName))
+		s := b.Stats()
+		out["Blocks"] = strconv.Itoa(s.KeyN)
```
kc1212: Why call it "Blocks"? Isn't this the number of keys?

ineiti: for skipchains it's the number of blocks.

PR #1049, opened 2018-02-09
Omnicon

kc1212:
Implementation of Omnicon, described in https://eprint.iacr.org/2017/406.pdf. This PR aims to replace the current cosi and bftcosi implementations with a single cosi implementation. The new cosi implementation uses 3-level with a leader, sub-leader and children. If a sub-leader fails, the tree is re-generated with a new sub-leader. The new bftcosi should perform two rounds of the same cosi to represent the prepare and commit stages of PBFT. Further, duplicate implementation of the cosi crypto primitives are removed, only kyber/sign/cosi is used in this PR.

Original impelementation by @Racater):
https://github.com/dedis/student_17_bftcosi/


bftcosi/bftcosi.go
```diff
@@ -1,798 +1,239 @@
...
-func (bft *ProtocolBFTCoSi) RegisterOnSignatureDone(fn func(*BFTSignature)) {
-	bft.onSignatureDone = fn
-}
-
-// Shutdown closes all channels in case we're done
-func (bft *ProtocolBFTCoSi) Shutdown() error {
-	defer func() {
-		// In case the channels were already closed
-		recover()
+	go func() {
```
jeffallen: How do you know this goroutine will terminate?

bftcosi/bftcosi.go
```diff
@@ -1,798 +1,239 @@
...
-	if bft.IsLeaf() {
-		return bft.startResponse(RoundPrepare)
+	pi, err := bft.CreateProtocol(name, bft.Tree())
+	if err != nil {
+		return nil, err
 	}
-	return bft.sendToChildren(&ch)
-}
+	cosiProto := pi.(*protocol.CoSiRootNode)
+	cosiProto.CreateProtocol = bft.CreateProtocol
```
jeffallen: Please error-check this type conversion or explain why not.

bftcosi/bftcosi.go
```diff
@@ -1,798 +1,239 @@
...
+	// NOTE We set it to n / 10 to have every sub-leader manage 10 nodes.
+	// This setting is bad if there are thousands of nodes as the root
+	// would need to manage hundres of sub-leaders.
```
jeffallen: hundreds

jeffallen: Not asking you to do it now, but if we need to make this adjustable later, will it be possible?

bftcosi/bftcosi.go
```diff
@@ -1,798 +1,239 @@
...
-	tempPrepareResponse []kyber.Scalar
-	// temporary buffer of the public keys for nodes that responded
-	tempPrepareResponsePublics []kyber.Point
-	// temporary buffer of "commit" responses
-	tempCommitResponse []kyber.Scalar
+	// prepCosiProtoName is the cosi protocol name for the prepare phase
+	prepCosiProtoName string
+	// commitCosiProtoName is the cosi protocol name for the commit phase
+	commitCosiProtoName string
+	// prepSigChan is the channel for reading the prepare phase signature
```
jeffallen: what is the []byte that goes through this channel? I understand it as a signal only, in which case it should be a struct{}.

bftcosi/bftcosi.go
```diff
@@ -1,798 +1,239 @@
...
 	}
 
-	// verify if the signature is correct
-	data := sha512.Sum512(ch.Signature.Msg)
-	bftPrepareSig := &BFTSignature{
-		Sig:        ch.Signature.Sig,
-		Msg:        data[:],
-		Exceptions: ch.Signature.Exceptions,
+	// prepare phase (part 2)
+	prepSig := <-bft.prepSigChan
```
jeffallen: I see a lot of blocking channel reads here. Are timeouts planned for a future iteration?

If a timeout was implemented by close(bft.prepSigChan), then prepSig would end up nil here, which is not checked, but I guess/hope cost.Verify would Do The Right Thing.

cosi/README.md
```diff
@@ -78,6 +74,23 @@ cosi check -g $COTHORITY
 
 This will first contact each server individually and then check a few random collective signing group constellations. If there are connectivity problems, due to firewalls or bad connections, for example, you will see a "Timeout on signing" or similar error message.
 
+## Description
+The purpose of this work is to implement a robust and scalable consensus algorithm using CoSi protocol and handling some exceptions. **The CoSi tree is a three level tree** to make a compromise between the two-level tree, making the root-node vulnerable to DoS, and a more than three level tree, slowing the algorithm because of the RTT between the root node and the leaves.
+
+The tree is composed of a leader (root-node), and some groups of equal size, each having a sub-leader (second level nodes) and members (leaves). The groups composition are defined by the leader.
```
jeffallen: s/groups/group/

cosi/README.md
```diff
@@ -78,6 +74,23 @@ cosi check -g $COTHORITY
...
+
+We want to **handle non-responding nodes**, no matter where they are in the tree. If a leaf is failing, then it is ignored in the CoSi commitment. If a sub-leader is non-responding, then the leader (root node) recreates the group designing another sub-leader from the group members. And finally, if the leader is failing, the protocol restarts using another leader.
```
jeffallen: s/designing/selecting/

jeffallen: Is this "if leader is failing" part implemented already?

cosi/protocol/gen_tree.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/dedis/onet"
+	"github.com/dedis/onet/network"
+)
+
+// GenTrees will create a given number of subtrees of the same number of nodes.
+// Each generated subtree will have the same root.
+// Each generated tree have a root with one child (the subleader)
+// and all other nodes in the tree will be the subleader children.
+// NOTE: register being not implementable with the current API could hurt the scalability tests
+func GenTrees(roster *onet.Roster, nNodes, nSubtrees int) ([]*onet.Tree, error) {
```
jeffallen: It seems like this code could be in onet/treenode.go. Not saying it should be, just checking your opinion... is it general purpose, or cosi-only?

cosi/protocol/gen_tree.go
```diff
@@ -0,0 +1,114 @@
...
+		var err error
+		trees[i], err = GenSubtree(treeRoster, 1)
+		if err != nil {
+			return nil, err
+		}
+
+		start = end
+	}
+
+	//l.Trees[tree.ID] = tree
```
jeffallen: remove please

cosi/protocol/helper_functions.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/dedis/kyber/sign/cosi"
+	"github.com/dedis/onet"
+	"github.com/dedis/onet/log"
+	"github.com/dedis/onet/network"
+)
+
+// generateCommitmentAndAggregate generates a personal secret and commitment
+// and returns respectively the secret, an aggregated commitment and an aggregated mask
+func generateCommitmentAndAggregate(s cosi.Suite, t *onet.TreeNodeInstance, publics []kyber.Point,
+	structCommitments []StructCommitment) (kyber.Scalar, kyber.Point, *cosi.Mask, error) {
```
jeffallen: Go types usually don't have "struct" in them.

cosi/protocol/helper_functions.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, err
+	}
+
+	log.Lvl3(t.ServerIdentity().Address, "is done aggregating responses with total of",
+		len(responses), "responses")
+	return aggResponse, nil
+}
+
+// GetSubleaderIDs returns a slice of subleaders for tree
+func GetSubleaderIDs(tree *onet.Tree, nNodes, nSubtrees int) ([]network.ServerIdentityID, error) {
```
jeffallen: why are these exported? (next function too)

cosi/protocol/protocol.go
```diff
@@ -0,0 +1,346 @@
+package protocol
```
jeffallen: I stopped reviewing here, and will pick it back up tomorrow morning.

bftcosi/bftcosi.go
kc1212: It terminates when the prepProto (an instance of cosi), which it eventually does after a timeout.

bftcosi/bftcosi.go
kc1212: Yes, we can add a `NSubtrees` property in bftcosi so that it is configurable. The value is propagated down to cosi and sub-cosi, so nothing else needs to be changed. Actually, there is a better way to compute `NSubtrees`, I'm updating the code.

bftcosi/bftcosi.go
kc1212: It's the signature from the prepare phase of bftcosi. It is used for verification in Dispatch.

bftcosi/bftcosi.go
kc1212: It's possible, but tricky. The bft.Timeout is not a global timeout for the whole protocol, but passed down to cosi (I will make this clear in the comments). The way that cosi uses the timeout is that if a subleader does not respond after the timeout while waiting for commits, then it'll re-arrange the tree and try a new sub-leader from the original set of children. This process is repeated until it has tried all children. It is used again while waiting for responses. So it's difficult to put a precise bound on it. I am adding a rather long timeout, which is bft.Timeout * NSubtrees, which is better than nothing.

cosi/README.md
kc1212: No, will update the readme.

cosi/protocol/gen_tree.go
kc1212: This is somewhat cosi-only, as it only generates 3-level trees. We may, however, simplify the code by making use of existing onet tree generation functions. I will add a comment.

cosi/protocol/helper_functions.go
kc1212: This is somewhat of a convention in cothority... you have to wrap the types in a struct with onet.TreeNode to make it work with protocol channels, see https://github.com/dedis/cothority_template/blob/master/protocol/struct.go#L22

I remember Linus expressing a need to fix this but we never go around doing a major onet refactor.

cosi/protocol/helper_functions.go
kc1212: These are used in the simulations package, but I've changed some other functions to be private.

cosi/protocol/protocol.go
kc1212: Thanks. My latest commits and comments hopefully address your review so far.

cosi/protocol/helper_functions.go
jeffallen: Oh sorry, I knew this, duh.

cosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,279 @@
...
+		err = p.SendToParent(&Response{response})
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+//HandleStop is called when a Stop message is send to this node.
```
jeffallen: I don't understand this Stopping mechanism. Can you document it better? (Maybe it already is someplace else, if so sorry.)

cosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,279 @@
...
+	return nil
+}
+
+// Start is done only by root and starts the subprotocol
+func (p *CoSiSubProtocolNode) Start() error {
+	log.Lvl3(p.ServerIdentity().Address, "Starting subCoSi")
+	if p.Msg == nil {
+		return errors.New("subprotocol does not have a proposal msg")
+	}
+	// p.Data can be nil
```
jeffallen: yeah, that's obvious from the code, but when and why is it nil, that's what the reader is wondering.

cosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,279 @@
...
+	if p.Publics == nil || len(p.Publics) < 1 {
+		return errors.New("subprotocol has invalid public keys")
+	}
+	if p.verificationFn == nil {
+		return errors.New("subprotocol has an empty verification fn")
+	}
+	if p.Timeout < 10 {
```
jeffallen: It seems like you want 10 * time.Second here?

cosi/service/cosi.go
```diff
@@ -45,64 +46,84 @@ type SignatureResponse struct {
...
+
+	// wait for reply
+	var sig []byte
+	select {
+	case sig = <-p.FinalSignature:
+	case <-time.After(p.Timeout + time.Second):
+		return nil, errors.New("protocol timed out")
+	}
+
+	// the hash is the message cosi actually signs, ideally cosi protocol
```
jeffallen: Would not be a good time to fix that? If cosi knows it, cosi should give it?

cosi/simulation/server_list
```diff
@@ -0,0 +1,4 @@
+iccluster084.iccluster.epfl.ch
```
jeffallen: What's this? Why is it checked in?

pop/service/api_test.go
```diff
@@ -54,6 +54,7 @@ func TestFinalStatement_Verify(t *testing.T) {
 	require.NotNil(t, fs.Verify())
 	h, err := fs.Hash()
 	log.ErrFatal(err)
+	// TODO how to sign??
```
jeffallen: So, how?

pop/service/service.go
```diff
@@ -98,7 +100,11 @@ type Service struct {
 	// Sync tools
 	// key of map is ID of party
 	// synchronizing inside one party
-	syncs map[string]*sync
+	syncs map[string]*syncChans
+	// verifyFinalBuffer is a temporary buffer for bftVerifyFinal results
+	verifyFinalBuffer sync.Map
```
jeffallen: This needs way more documentation. Why do we need these sync.Maps? Who are the producers, consumers? How do we know that they are cleaned up?

pop/service/service.go
```diff
@@ -726,23 +756,25 @@ func (s *Service) signAndPropagate(final *FinalStatement, protoName string,
...
 	if err != nil {
 		return err
 	}
 
 	root.Data = data
-	done := make(chan bool)
-	root.RegisterOnDone(func() {
-		done <- true
-	})
+	root.Timeout = time.Second * 5
```
jeffallen: 5 before time.Second, please.

skipchain/skipchain.go
```diff
@@ -1252,6 +1249,14 @@ func (s *Service) tryLoad() error {
 	return nil
 }
 
+// sliceToArr does what the name suggests, we need it to turn a slice into
```
jeffallen: 👍 for code non-reuse. I hate little util packages with this kind of crap in them. :)

(PS: This is not even sarcastic. Leave that right how it is!)

skipchain/struct.go
```diff
@@ -511,11 +511,16 @@ func (fl *ForwardLink) Copy() *ForwardLink {
 // Verify checks the signature against a list of public keys. This list must
 // be in the same order as the Roster that signed the message.
 // It returns nil if the signature is correct, or an error if not.
-func (fl *ForwardLink) Verify(suite network.Suite, pubs []kyber.Point) error {
+func (fl *ForwardLink) Verify(suite cosi.Suite, pubs []kyber.Point) error {
 	if bytes.Compare(fl.Signature.Msg, fl.Hash()) != 0 {
 		return errors.New("wrong hash of forward link")
 	}
-	return fl.Signature.Verify(suite, pubs)
+	// this calculation must match the one in omnicon/bftcosi
```
jeffallen: Then maybe they should be one and the same code... (yeah, ignore my last comment, sometimes I do not like copy/paste)

cosi/protocol/protocol.go
```diff
@@ -0,0 +1,354 @@
...
+
+	// send challenge to every subprotocol
+	for _, coSiProtocol := range runningSubProtocols {
+		subProtocol := coSiProtocol
+		subProtocol.ChannelChallenge <- structChallenge
+	}
+
+	// get response from all subprotocols
+	responses := make([]StructResponse, 0)
+	errChan := make(chan error, 1)
```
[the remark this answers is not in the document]
kc1212: You are right, we should not swallow errors which may make debugging more difficult later on. I did it this way is because the logic is simpler when I need to read the channel.

cosi/protocol/struct.go
```diff
@@ -0,0 +1,108 @@
...
+func (m *cosiSuite) Hash() hash.Hash {
+	return sha512.New()
+}
+
+func (m *cosiSuite) RandomStream() cipher.Stream {
+	return m.r
+}
+
+// EdDSACompatibleCosiSuite is a custom suite made to be compatible with eddsa because
+// cothority.Suite uses sha256 but EdDSA uses sha512.
```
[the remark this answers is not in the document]
kc1212: Raised a bug for this https://github.com/dedis/cothority/issues/1083

cosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,279 @@
...
+	// internodes channels
+	ChannelAnnouncement chan StructAnnouncement
+	ChannelCommitment   chan StructCommitment
+	ChannelChallenge    chan StructChallenge
+	ChannelResponse     chan StructResponse
+}
+
+// NewDefaultSubProtocol is the default sub-protocol function used for registration
+// with an always-true verification.
+func NewDefaultSubProtocol(n *onet.TreeNodeInstance) (onet.ProtocolInstance, error) {
```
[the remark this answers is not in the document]
kc1212: These are used in external packages, e.g. cosi/service, or any other service that want it. For the other services, they need to do the registration for both the main protocol and sub protocol.

cosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,279 @@
...
+			verifyChan <- p.verificationFn(p.Msg, p.Data)
+		}()
+	}
+
+	if errs := p.SendToChildrenInParallel(&announcement.Announcement); len(errs) > 0 {
+		log.Lvl3(p.ServerIdentity().Address, "failed to send announcement to all children")
+	}
+
+	// ----- Commitment -----
+	commitments := make([]StructCommitment, 0)
```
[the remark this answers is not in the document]
kc1212: We are not sure how many to expect. If the subprotocol is running on the root, it should only receive one commitment from the subleader. Further, some child may fail but we want to still do something with the successful children.

cosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,279 @@
...
+			}
+			commitments = append(commitments, commitment)
+		case <-time.After(p.Timeout):
+			// the timeout here should be shorter than the main protocol timeout
+			// because main protocol waits on the channel below
+			p.subleaderNotResponding <- true
+			return nil
+		}
+	} else {
+		t := time.After(p.Timeout / 2)
```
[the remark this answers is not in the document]
kc1212: The timeout should be shorter than the timeout for receiving commits (i.e. p.Timeout), a few lines above, hence it is reduced, /2 is somewhat arbitrary. I will add a comment to clarify.

cosi/protocol/protocol.go
```diff
@@ -0,0 +1,354 @@
...
+		subProtocolName:  subProtocolName,
+		suite:            suite,
+	}
+
+	return c, nil
+}
+
+// Shutdown stops the protocol
+func (p *CoSiRootNode) Shutdown() error {
+	if !p.hasStopped {
```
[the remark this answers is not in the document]
kc1212: It seems `hasStopped` doesn't do anything in protocol and `Shutdown` doesn't do any cleanup at the moment. I think it's useful to close some of the channels, which I will do, in a non-racy way.

cosi/protocol/sub_protocol.go
kc1212: Thanks, updating the comment.

cosi/protocol/sub_protocol.go
kc1212: I wanted 10 nanoseconds, basically to make sure the timeout is realistic. I'll change it to 10 * time.Nanosecond to clarify.

cosi/service/cosi.go
kc1212: The comment is perhaps a bit misleading, I will update it. Cosi knows, but we also need to verify that cosi signed the right message. There are two cases:
1, cosi returns message and signature, so we need to check the message matches the original message and the signature is valid.
2, cosi returns only the signature, then we only check that the signature matches with the original message. If cosi signed a different message, then the final verification will fail.
Hence, in both cases, we arrive at the same result. Whether cosi returns the message is not a big concern in my opinion.

pop/service/api_test.go
kc1212: An artifact from debugging...

cosi/protocol/doc.go
```diff
@@ -0,0 +1,34 @@
...
+"Keeping Authorities "Honest or Bust" with Decentralized Witness Cosigning"
+
+The protocol has four messages:
+	- Announcement which is sent from the root down the tree and announce the proposal
+	- Commitment which is sent back up to the root, containing an aggregated commitment from all nodes
+	- Challenge which is sent from the root down the tree and contains the aggregated challenge
+	- Response which is sent back up to the root, containing the final aggregated signature, then used by the root to sign the proposal
+
+The protocol uses five files:
+- struct.go defines the messages sent around and the protocol constants
```
ineiti: for godoc to look nice, this has to be indented. We use two-space indents, please adjust also the lines above ("the protocol has four messages:...")

cosi/protocol/doc.go
```diff
@@ -0,0 +1,34 @@
...
+Under-the-hood, there are two protocols. A main protocol which only runs on the
+root node and a sub-protocol that runs on all nodes (including the root). The
+tree structure of the sub-protocol is illustrated below.
+     root
+       |
+       |
+   sub-leader
+    /       \
+   /         \
+leaf_1 ... leaf_n
```
ineiti: Please make sure you have two spaces before all these lines, else it will look very bad in godoc.

cosi/protocol/gen_tree_test.go
```diff
@@ -0,0 +1,381 @@
...
+	"github.com/dedis/onet"
+	"github.com/dedis/onet/log"
+	"github.com/dedis/onet/network"
+)
+
+func TestMain(m *testing.M) {
+	log.MainTest(m)
+}
+
+//tests the root of the trees
```
ineiti: Nit-picking here and later:
`s/\/\/(\w)/\/\/ \1/`

PR #1050, opened 2018-02-09
Authentication service: first ideas

ineiti:
Ideas how to create one and only one authentication service in the cothority, so we don't have three implementations like now: one for cisc, one for skipchains and one for pop.

For review, start looking at `authentication/service{,_test}.go`


jeffallen:
We are not currently working on this, we can always come back to it.

PR #1056, opened 2018-02-13
Evoting

qantik:
Port of dedis/student_17_evoting into cothority.


jeffallen:
I am removing the experimental tag in kyber. Will merge this when that is done.

qantik:
I don't understand why there is a race condition in ```github.com/dedis/kyber/xof/blake```. It's not
obvious to me where kyber is invoked concurrently.

jeffallen:
I will find a solution for this.

kc1212:
The commit c4a9535 introduces `sync.Map`. I'm not sure which version of go we support, but if it includes 1.8 or earlier versions, then it won't compile. @ineiti @jeffallen do we have any commitment on this matter?

jeffallen:
I've been working under the assumption that Go 1.8 is far enough back to test.

kc1212:
If 1.8 is supported, then `sync.Map` won't compile, as it's a 1.9 feature. You may want to use a regular map with a mutex around it @qantik

qantik:
The race conditions vanished when using ```cothority.Suite```. The global suite is gone and ```sync.Map``` has been replaced by ordinary mutexes. The evoting tests now all pass on travis. For some reason there is one test failing in the ```pop``` package.

ineiti:
Once #1061 is merged you can rebase against master and try again.

evoting/api_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/dedis/cothority"
+	"github.com/dedis/onet"
+)
+
+func TestPing(t *testing.T) {
+	local := onet.NewTCPTest(cothority.Suite)
+	defer local.CloseAll()
+
+	// 	_, roster, _ := local.GenTree(3, true)
```
ineiti: Is that test used?

evoting/protocol/dkg.go
```diff
@@ -0,0 +1,226 @@
+package protocol
```
ineiti: Is this a copy of the onchain-secrets/protocol/dkg.go?

evoting/service/service.go
```diff
@@ -0,0 +1,443 @@
...
+		if err := req.Election.Store(req.Election); err != nil {
+			return nil, err
+		}
+
+		if err = master.Store(&lib.Link{ID: genesis.Hash}); err != nil {
+			return nil, err
+		}
+
+		return &evoting.OpenReply{ID: genesis.Hash, Key: secret.X}, nil
+	case <-time.After(2 * time.Second):
```
ineiti: That should be a timeout.

evoting/service/service.go
```diff
@@ -0,0 +1,443 @@
...
+
+	service.RegisterHandlers(service.Ping, service.Link, service.Open, service.Login,
+		service.Cast, service.GetBox, service.GetMixes, service.Shuffle,
+		service.GetPartials, service.Decrypt, service.Reconstruct,
+	)
+
+	service.state.schedule(3 * time.Minute)
+	service.node = onet.NewRoster([]*network.ServerIdentity{service.ServerIdentity()})
+
+	log.Lvl3("Pin:", service.pin)
```
ineiti: This (and the `"0", //nonce(6)`) should be made better. probably add the `nonce(6)` and print it using `log.Info`.

evoting/service/state.go
```diff
@@ -0,0 +1,85 @@
+package service
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+)
+
+const limit = 5
```
ineiti: Is this limit the time a user is connected? Then it should be named as such. Also, we started using direct `time.Duration` instead of `int`.

evoting/service/state_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNonce(t *testing.T) {
+	n1, n2, n3 := nonce(10), nonce(10), nonce(10)
+	assert.Equal(t, 10, len(n1), len(n2), len(n3))
```
ineiti: I think you mean
```
assert.Equal(t, 10, len(n1))
assert.Equal(t, 10, len(n2))
assert.Equal(t, 10, len(n3))
```

evoting/service/state_test.go
```diff
@@ -0,0 +1,52 @@
...
+	assert.NotEqual(t, n1, n2, n3)
```
ineiti: Same here - the test should be on three lines (n1, n2 - n2, n3 - n3, n1)

qantik: Fixed.

evoting/service/state_test.go
qantik: Fixed.

evoting/service/state.go
qantik: Fixed.

evoting/service/service.go
qantik: Fixed.

evoting/protocol/dkg.go
qantik: Yes. The three files `dkg.go`, `dkg_struct.go` and `dkg_test.go` are from the onchain-secrets project. Also the `DKGSimulate` function in `lib/utils.go` is taken from `onchain-secrets/protocol/onchain_test.go`.

evoting/api_test.go
qantik: Fixed.

evoting/service/service.go
qantik: Fixed.

PR #1058, opened 2018-02-14
Skipchain signature

ineiti:
This changes the forward-signatures to something more reasonable like:

```
type ForwardLink struct {
	// From - where this forward link comes from
	From SkipBlockID
	// To - where this forward link points to
	To SkipBlockID
	// NewRoster is only set to non-nil if the From block has a
	// different roster from the To-block.
	NewRoster *onet.Roster
	// Signature is calculated on the
	// sha256(From.Hash()|To.Hash()|NewRoster.ID)
	// In the case that NewRoster is nil, the signature is
	// calculated on the sha256(From.Hash()|To.Hash())
	Signature bftcosi.BFTSignature
}
```

This prevents re-use of a ForwardSignature in a wrong context. It also allows later to get updates
using `ForwardSignature`s only, as it will be enough jumping from one ForwardSignature to the next, given that the `NewRoster` will indicate the Roster that will be responsible for the new Block.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -9,6 +9,7 @@
 package skipchain
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"strconv"
@@ -89,34 +90,33 @@ var ErrorProcessing = errors.New("this skipchain-id is currently processing a bl
 // skipchain after verification that it fits and no other block already has been
 // added.
 func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, error) {
+	// Initial checks on the proposed block.
 	prop := psbd.NewBlock
 	if !s.ServerIdentity().Equal(prop.Roster.Get(0)) {
 		return nil, errors.New(
 			"only leader is allowed to add blocks")
-
 	}
 	if len(s.Storage.Clients) > 0 {
 		if psbd.Signature == nil {
 			return nil, errors.New(
 				"cannot create new skipblock without authentication")
-
 		}
 		if !s.authenticate(psbd.NewBlock.CalculateHash(), *psbd.Signature) {
 			return nil, errors.New(
 				"wrong signature for this skipchain")
-
 		}
 	}
 	var prev *SkipBlock
 	var changed []*SkipBlock
 
 	if psbd.LatestID.IsNull() {
-		log.Lvl3("Creating new skipchain with roster", psbd.NewBlock.Roster.List)
 		// A new chain is created
+		log.Lvl3("Creating new skipchain with roster", psbd.NewBlock.Roster.List)
 		prop.Index = 0
 		prop.Height = prop.MaximumHeight
-		prop.ForwardLink = make([]*BlockLink, 0)
-		// genesis block has a random back-link:
+		prop.ForwardLink = make([]*ForwardLink, 0)
+		// genesis block has a random back-link, so that two
+		// identical genesis blocks have a different ID.
 		var bl [32]byte
 		random.Bytes(bl[:], random.New())
 		prop.BackLinkIDs = []SkipBlockID{SkipBlockID(bl[:])}
@@ -125,11 +125,9 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, er
 		err := s.verifyBlock(prop)
 		if err != nil {
 			return nil, err
-
 		}
 		if !s.newBlockStart(prop) {
 			return nil, ErrorProcessing
-
 		}
 		defer s.newBlockEnd(prop)
 
@@ -138,7 +136,6 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, er
 			if parent == nil {
 				return nil, errors.New(
 					"Didn't find parent")
-
 			}
 			parent.ChildSL = append(parent.ChildSL, prop.Hash)
 			changed = append(changed, parent)
@@ -146,8 +143,9 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, er
 		changed = append(changed, prop)
 
 	} else {
-		log.Lvlf3("Adding block with roster %+v to %x", psbd.NewBlock.Roster.List, psbd.LatestID)
+
 		// We're appending a block to an existing chain
+		log.Lvlf3("Adding block with roster %+v to %x", psbd.NewBlock.Roster.List, psbd.LatestID)
 		prev = s.db.GetByID(psbd.LatestID)
 		if prev == nil {
 			// Did not find the block they claim is the latest. So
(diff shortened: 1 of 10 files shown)


skipchain/struct.go
```diff
@@ -373,7 +374,7 @@ func (sb *SkipBlock) Copy() *SkipBlock {
 	b := &SkipBlock{
 		SkipBlockFix: sb.SkipBlockFix.Copy(),
 		Hash:         make([]byte, len(sb.Hash)),
-		ForwardLink:  make([]*BlockLink, len(sb.ForwardLink)),
+		ForwardLink:  make([]*ForwardLink, len(sb.ForwardLink)),
```
jeffallen: With this change, you have invalidated all the existing .DB files. We will need to re-establish the production network. However, we also need to change the rosters in the production conodes to refer to each other via TLS. So we need to do both at the same time.

We need a better strategy for evolving on-disk formats. Is now the time to do it? Or maybe with the first change after v2?

nikkolasg:
Why `NewRoster` entirely ? Is it not possible to just put the ID ? A `ForwardLink` is only signed by the *current* roster (the one that is responsible for the block wher the forward link is attached), not the "*forward*" roster IIRC.

skipchain/struct_test.go
```diff
@@ -118,11 +118,11 @@ func TestSkipBlock_Hash2(t *testing.T) {
...
 func TestBlockLink_Copy(t *testing.T) {
 	// Test if copy is deep or only shallow
-	b1 := &BlockLink{}
-	b1.Sig = []byte{1}
+	b1 := &ForwardLink{}
+	b1.Signature.Sig = []byte{1}
 	b2 := b1.Copy()
-	b2.Sig = []byte{2}
-	if bytes.Equal(b1.Sig, b2.Sig) {
+	b2.Signature.Sig = []byte{2}
```
jeffallen: This test doesn't make sense, it's testing something obvious (set to two different slices, then check that they are different). It would make more sense to mutate b1.Signature after the Copy() and confirm that b2 does NOT see the change made to b1. i.e. b1.Signature.Sig[0] = 0.

skipchain/struct.go
nikkolasg: > We need a better strategy for evolving on-disk formats. Is now the time to do it? 

it's definitely a good time to think about it

ineiti:
`Why NewRoster and not RosterID`?

The goal is to be able to follow the forward-links only to get to the latest block. If I get the `RosterID`, I cannot know the aggregate public key that is allowed to sign. If I have the aggregate public key, I cannot know if an exception is valid or not. So I need the complete roster (if it changed).

The best would be to have a `diff` of the rosters, but this was too complicated to have by the end of February ;)

nikkolasg:
> If I get the RosterID, I cannot know the aggregate public key that is allowed to sign. If I have the aggregate public key, I cannot know if an exception is valid or not. So I need the complete roster (if it changed).

But wait I don't get it. The signature in the new struct you show is made by the "current roster" no ? Why do we need the new roster to verify this signature ??
And what do it mean `sha256(From.Hash()|To.Hash()|NewRoster)` ? WHat's `sha256(roster)` ?

skipchain/struct.go
ineiti: +1 - I opened #1059 for this discussion. For the moment we're in 'API-breaking mode', so the things will break regularly anyway. But for the changes in the upcoming `v2` we definitively must have a good way of defining that.

ineiti:
If you have an old block and want to get to the latest, you need to use more than one `ForwardLink`. Now if the roster changes during the jumps, you need to be able to verify the new `ForwardLink`. If you don't do that, you have to download both the `ForwardLink`s plus the `SkipBlock`. With the addition of the `NewRoster`, only downloading the `ForwardLink` is enough.

Another one of these Bryan-ideas ;)

nikkolasg:
So you want to be able to only download forward link now ?
...
I would much more tempted to separate a skipblock into a `SkipblockHeader` and `SkipblockData` as usually done elsewhere. A `SkipblockHeader` would have a fixed size whereas a `ForwardLink` now has variable length size. If we have a roster of 50 nodes, then **each** forward link has `(32 (public key) + 16 (address) + XXX (description))` bytes more. This information is just redundant, it just imposes a higher load for a gain we don't even measure (maybe it is a gain, maybe it is not if we arrive to make a good split between the data that a block contains and the other informations).

ineiti:
You'll have the same size if you just download the skipblock-headers, even bigger. And if the roster doesn't change in between blocks, in the case of skipblock-headers, you'll download it anyway. So the solution I propose here is already an optimized one of what you're proposing.

skipchain/struct.go
```diff
@@ -442,34 +443,83 @@ func (sb *SkipBlock) updateHash() SkipBlockID {
...
-		bftcosi.BFTSignature{
+	return &ForwardLink{
+		Signature: bftcosi.BFTSignature{
 			Sig:        sigCopy,
 			Msg:        msgCopy,
 			Exceptions: exsCopy,
 		},
+		From:      fl.From,
+		To:        fl.To,
+		NewRoster: fl.NewRoster,
```
kc1212: NewRoster is not exactly a copy, but maybe it's ok in this situation.

skipchain/skipchain.go
```diff
@@ -863,105 +827,91 @@ func (s *Service) bftVerifyNewBlock(msg []byte, data []byte) (out bool) {
...
-		roster.List, src.Index, dst.Index, fwl)
-	if len(fwl) > 0 {
-		return errors.New("forward-link got signed during our signing")
-	}
-	src.ForwardLink = []*BlockLink{fwd}
-	if err = src.VerifyForwardSignatures(); err != nil {
-		return errors.New("Wrong BFT-signature: " + err.Error())
+// verifyFollowBlock makes sure that a signature-request for a forward-link
+// is valid.
+func (s *Service) bftVerifyFollowBlock(msg []byte, data []byte) bool {
```
kc1212: I didn't see where the `From` part of the forward link is verified in this function. Is that somewhere else or there is no need to check it explicitly (as it's already a part of the hash and it's signed).

nikkolasg:
> You'll have the same size if you just download the skipblock-headers, even bigger.

Yes the roster will be the same size of course. But now:
 - the conodes have to store twice the same data without any gain *from a conode's perspective*. It does not provide more availability or whatsoever. 

 - the client that wishes to update to latest block downloads only once the roster  for each intermediate blocks (using both methods), I agree, but for the last block it downloads it twice with this method. How much does the client gain with this method compared to what a **fixed** bigger size of a header, I'm not sure. This would have been nice to check.

We already had trouble when having redundant data in the past with cothority, my main concern is to avoid duplication of data (without clear benefit) because that means lots of trouble when there is an inconsistencies (vs simplicity).

skipchain/skipchain.go
ineiti: It's done towards the end where a `ForwardLink` is created and then his Hash is compared to the final hash:

```
		fl := NewForwardLink(target, fs.Newest)
		if bytes.Compare(fl.Hash(), msg) != 0 {
```

So if the forward-link would be wrong, the hash would be changed.

ineiti:
Yes, duplication is evil, this is true. Here I think it is OK because it helps us to get smaller proofs that a current skipblock is valid.

But the best solution would be to add a `RosterDiff` in the `ForwardLink` - this is in fact what Bryan suggested. But making a `RosterDiff` looked too long to go into `v2`... So I came up with this solution.

The most important part of this PR is to have the `From` - the `Roster` was more of a 'bonus' because it came up in a discussion and it seemed to be easy enough to add it. A next step would be to have a `GetUpdateForwardLinks` or so that only returns the forward-links and the latest block.

PR #1061, opened 2018-02-15
Adjust to cisc to TLS and fix Status

ineiti:
Make cothority and cisc work correctly with TLS.
Restore previous behavior of the Status protocol, so that the web UI works again.


scmgr/test.sh
```diff
@@ -1,26 +1,26 @@
 #!/usr/bin/env bash
 
-DBG_TEST=1
+DBG_TEST=3
```
jeffallen: Don't think you wanted these changes.

skipchain/skipchain.go
```diff
@@ -585,7 +585,7 @@ func (s *Service) AddFollow(add *AddFollow) (*EmptyReply, error) {
 		}
 		log.Lvlf2("%s FollowSearch %s %x", s.ServerIdentity(), add.Conode, add.SkipchainID)
 	case FollowLookup:
-		si := network.NewServerIdentity(cothority.Suite.Point().Null(), network.NewAddress(network.PlainTCP, add.Conode))
+		si := network.NewServerIdentity(cothority.Suite.Point().Null(), network.NewAddress(network.TLS, add.Conode))
```
jeffallen: This can't work in heterogeneous networks. I think add.Conode needs to be a ServerIdentifier.

jeffallen:
There are other changes needed to make Status work again.

cisc/cisc.go
```diff
@@ -380,19 +392,18 @@ func scRoster(c *cli.Context) error {
...
+		return err
+	}
 	if id == nil {
 		scList(c)
 		return errors.New("Please indicate skipchain-id")
 	}
-	if err := id.DataUpdate(); err != nil {
-		return err
-	}
-	if err := id.ProposeUpdate(); err != nil {
```
kc1212: I haven't looked at CISC code in detail, but why are these not needed anymore?

scmgr/test.sh
```diff
@@ -262,7 +262,7 @@ testConfig(){
 	CFG=$OLDCFG
 
 	# $CFG/data cannot be empty
-	testFail find "$CFG/data" -maxdepth 0 -type d -empty
+	testFail [ -d "$CFG/data" ]
```
kc1212: Here, the directory can be empty where as before the directory cannot. Is this intended?

cisc/cisc.go
jeffallen: Sorry, no idea. This was part of Linus' hacking around on Friday.

scmgr/test.sh
jeffallen: Sorry, no clue. I will send Linus email to ask him about this apparent comment/code incompatibility when he gets back.

kc1212: Actually, I think I wrote the `find "$CFG/data" -maxdepth 0 -type d -empty` part. I did not want it to be empty because `scmgr` should write stuff into that folder. I'm not sure why Linus changed it.

kc1212: If there's no failure in the more strict version of the test, then we should change it back.

jeffallen: OK, I've sent him email, he'll look at it when he's back.

cisc/cisc.go
kc1212: After a more in-depth look, I think this is OK. `DataUpdate` and `ProposeUpdate` is called in `cfg.findSC`.

PR #1062, opened 2018-02-16
cothorityjs: add evoting protobuf definitions

gnarula:
Added a protobuf file to communicate with the evoting service


ineiti:
Please add a short description to the PR.

PR #1066, opened 2018-02-19
Cisc: Adding key/value from a CSV file

nikkolasg:
This PR adds a small functionality to cisc to be able to give a CSV input file to put on a "cisc chain".
This is needed for Agora as we're gonna have a bunch of results that we'll put into a plaintext CSV file and we want to easily push that to the skipchain.


cisc/cisc.go
```diff
@@ -547,13 +548,52 @@ func kvAdd(c *cli.Context) error {
...
+	}
+	id, err := cfg.findSC(c.Args().Get(1))
+	if err != nil {
+		scList(c)
+		return errors.New("Please give skipchain-id")
+	}
+	file := c.Args().First()
+	fd, err := os.Open(file)
+	log.ErrFatal(err)
+	reader := csv.NewReader(fd)
```
jeffallen: This would be more efficient if you fed a bufio.Reader to it (https://godoc.org/bufio#NewReader). I don't know if the expected scale justifies this.

cisc/cisc.go
```diff
@@ -547,13 +548,52 @@ func kvAdd(c *cli.Context) error {
 	value := c.Args().Get(1)
 	prop := id.GetProposed()
 	prop.Storage[key] = value
+	return addKv(c, cfg, id, prop)
+}
+
+// kvAddCsv reads the input CSV file, lookup its key column and inputs all the
+// // csv file at once to the skipchain. The key is given from the "column" args
```
jeffallen: double double slash ::))

cisc/cisc.go
```diff
@@ -547,13 +548,52 @@ func kvAdd(c *cli.Context) error {
...
+	column := c.Int("column")
+	log.Info("Column names to insert in cisc: " + strings.Join(records[0], ":"))
+	log.Info("Column name to use as a key: " + records[0][column])
+	prop := id.GetProposed()
+	for _, row := range records[1:] {
+		key := row[column]
+		value := strings.Join(row, ",")
+		prop.Storage[key] = value
+	}
+	return addKv(c, cfg, id, prop)
```
jeffallen: Is there a limit to the number that can be added at once? This feels like it is inviting a complaint from someone someday, who puts 50000 keys in and wonders why it doesn't work. (You can tell me, "we'll fix it then", and it would be fine with me.)

nikkolasg: Linus told me he tested it with more than 1Mb but not more. I wrote a sample csv file that I expect to handle is 16k (250 rows with 11 column with mostly integers). 
So I think we need @ineiti to confirm how cisc works because my understanding was the following: I should minimizes the number of new blocks created so it's less bandwidth consuming on the client's side (we're talking about a country on African continent). So I wanted to go with "one go". But maybe proposing sequentially new data does not increase the number of final blocks on the skipchain ? 

I think later on, it would be easy to add an option or stg to make it more "stream"-like. 
But this question is completely legitimate and should also be asked for the regular command line input command. So if you're fine with it, I propose we merge this one and I open an issue to not forget about this. Tell me what you think.

cisc/cisc.go
nikkolasg: See my comment below. it's related in my opinion. The way we read the file and the way we propose new data should be considered.

PR #1072, opened 2018-02-20
js: build before publishing

gnarula:
Forgot to build before publishing the previous version to npm. Added a hook that ensures this doesn't happen again in future.


gnarula:
Not R4M yet (depends on #1071). I can't move the card because I don't have write permissions ZenHub yet.

PR #1077, opened 2018-02-21
[cisc] replacing csv input by generic file input

nikkolasg:
This pr fixes the functionality of a previous PR I made because I just realized it's better to make it a bit more generic to fit all uses case. And because I haven't thought that through enough before, I realised the csv functionality is not exactly fitting my use case.
So now it's simpler: input is a file, and it's mapped to the filename or a given key. So the user has to decide what to do with the value instead of cisc deciding how to treat the value.
The test is adapted and works.

### DIFF
--- a/cisc/cisc.go
+++ b/cisc/cisc.go
@@ -560,12 +560,12 @@ func kvAdd(c *cli.Context) error {
 	return addKv(c, cfg, id, prop)
 }
 
-// kvAddCsv reads the input file, and stores it in the data. The key is the
+// kvAddFile reads the input file, and stores it in the data. The key is the
 // name of the file by default or overridden by the key flag. Do not use with
 // big files as it reads all at once.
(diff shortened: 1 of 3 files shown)


cisc/cisc.go
```diff
@@ -551,10 +560,10 @@ func kvAdd(c *cli.Context) error {
 	return addKv(c, cfg, id, prop)
 }
 
-// kvAddCsv reads the input CSV file, lookup its key column and inputs all the
-// csv file at once to the skipchain. The key is given from the "column" args
-// and the value is the whole full row, unmodified.
-func kvAddCsv(c *cli.Context) error {
+// kvAddCsv reads the input file, and stores it in the data. The key is the
```
jeffallen: This comment is out of date. (Sorry for the lame comment.)

PR #1078, opened 2018-02-21
build protobuf files before publishing

gnarula:
Build protobuf files from `external/js/protobuf` as a part of prepublish hook.

### DIFF
--- a/external/js/cothority/package-lock.json
+++ b/external/js/cothority/package-lock.json
@@ -1,13 +1,13 @@
 {
   "name": "@dedis/cothority",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


jeffallen:
Can you comment on how/why the dependencies keep getting updated? Is that an automatic thing, or on purpose? I'm all for updating, but I'm also trying to understand how we could get some stability if we wanted it.

gnarula:
In this case I did update them before sending the PR. By default the updates ensure backwards compatibility. In `package.json` the dependencies usually follow [Semantic Versioning](https://semver.org/)

Consider
```json
....
dependencies: {
  "package-foo": "^1.2.3"
}
...
```
On running `npm update` package foo will be updated to the latest `1.x.x` release which is bound to be backwards compatible if the package distributor is following semver. Now that I think of it, we aren't adhering to semver in the js packages we've published so far. That's something that should probably be ensured in future.

PR #1085, opened 2018-02-27
adding sciper lookup to evoting

ineiti:
This adds a lookup call to the evoting module.

### DIFF
--- a/evoting/service/service.go
+++ b/evoting/service/service.go
@@ -2,6 +2,11 @@ package service
 
 import (
 	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/dedis/kyber"
@@ -172,6 +177,47 @@ func (s *Service) Login(req *evoting.Login) (*evoting.LoginReply, error) {
 	admin := master.IsAdmin(req.User)
 	token := s.state.register(req.User, admin)
 	return &evoting.LoginReply{Token: token, Admin: admin, Elections: elections}, nil
+}
+
+// LookupSciper calls https://people.epfl.ch/cgi-bin/people/vCard?id=sciper
+// to convert scipers to names
+func (s *Service) LookupSciper(req *evoting.LookupSciper) (*evoting.LookupSciperReply, error) {
+	url := fmt.Sprintf("https://people.epfl.ch/cgi-bin/people/vCard?id=%s", req.Sciper)
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	reply := &evoting.LookupSciperReply{}
+	search := regexp.MustCompile("[:;]")
+	for _, line := range strings.Split(string(body), "\n") {
+		fr := search.Split(line, 2)
+		if len(fr) != 2 {
+			continue
+		}
+		value := strings.Replace(fr[1], "CHARSET=UTF-8:", "", 1)
+		switch fr[0] {
+		case "FN":
+			reply.FullName = value
+		case "EMAIL":
+			reply.Email = value
+		case "TITLE":
+			reply.Title = value
+		case "URL":
+			reply.URL = value
+		}
+	}
+	if reply.FullName == "" {
+		return nil, errors.New("didn't find sciper")
+	}
+
+	log.Lvl2("Got vcard: %v", reply)
+	return reply, nil
 }
 
 // Cast message handler. Cast a ballot in a given election.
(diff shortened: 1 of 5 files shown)


evoting/service/service.go
```diff
@@ -174,6 +179,47 @@ func (s *Service) Login(req *evoting.Login) (*evoting.LoginReply, error) {
...
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
```
jeffallen: body.Close()

evoting/service/service.go
```diff
@@ -174,6 +179,47 @@ func (s *Service) Login(req *evoting.Login) (*evoting.LoginReply, error) {
...
+// LookupSciper calls https://people.epfl.ch/cgi-bin/people/vCard?id=sciper
+// to convert scipers to names
+func (s *Service) LookupSciper(req *evoting.LookupSciper) (*evoting.LookupSciperReply, error) {
+	url := fmt.Sprintf("https://people.epfl.ch/cgi-bin/people/vCard?id=%s", req.Sciper)
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
jeffallen: Please wrap resp.Body in an io.LimitReader with a limit of 100k to prevent using too much RAM in case people.epfl.ch loses it's mind and spams us.

evoting/service/service.go
```diff
@@ -174,6 +179,47 @@ func (s *Service) Login(req *evoting.Login) (*evoting.LoginReply, error) {
...
+	}
+
+	reply := &evoting.LookupSciperReply{}
+	search := regexp.MustCompile("[:;]")
+	for _, line := range strings.Split(string(body), "\n") {
+		fr := search.Split(line, 2)
+		if len(fr) != 2 {
+			continue
+		}
+		value := strings.Replace(fr[1], "CHARSET=UTF-8:", "", 1)
```
jeffallen: If fr[1] does not start with CHARSET=UTF-8, return "error unknown character encoding" to help the next poor sucker debug what went wrong.

evoting/service/service.go
gnarula: we can perhaps throw an error early if Content-Type isn't `text/v-card`

evoting/service/service.go
```diff
@@ -174,6 +179,47 @@ func (s *Service) Login(req *evoting.Login) (*evoting.LoginReply, error) {
...
+			reply.FullName = value
+		case "EMAIL":
+			reply.Email = value
+		case "TITLE":
+			reply.Title = value
+		case "URL":
+			reply.URL = value
+		}
+	}
+	if reply.FullName == "" {
```
jeffallen: Seems possible that FN would be empty for a known user, but probably less possible that EMAIL would ever be empty.

evoting/service/service.go
```diff
@@ -174,6 +179,47 @@ func (s *Service) Login(req *evoting.Login) (*evoting.LoginReply, error) {
...
+		return nil, errors.New("didn't find sciper")
+	}
+
+	log.Lvl2("Got vcard: %v", reply)
```
jeffallen: The default log level for product servers is now 2; I think logging precise data other than "this connection from this remote addr" should be at level 3.

evoting/service/service.go
```diff
@@ -174,6 +179,47 @@ func (s *Service) Login(req *evoting.Login) (*evoting.LoginReply, error) {
 	return &evoting.LoginReply{Token: token, Admin: admin, Elections: elections}, nil
 }
 
+// LookupSciper calls https://people.epfl.ch/cgi-bin/people/vCard?id=sciper
+// to convert scipers to names
+func (s *Service) LookupSciper(req *evoting.LookupSciper) (*evoting.LookupSciperReply, error) {
+	url := fmt.Sprintf("https://people.epfl.ch/cgi-bin/people/vCard?id=%s", req.Sciper)
```
jeffallen: If sciper contains characters other than [0-9], then error. If len(sciper) > 20 error. We want to be polite net citizens and not pass on crazy input to them.

evoting/service/service.go
ineiti: even after `ioutil.ReadAll()`?

evoting/service/service.go
ineiti: Not all fields are of that charset type:
```
BEGIN:VCARD
VERSION:2.1
FN;CHARSET=UTF-8:Linus Gasser
N;CHARSET=UTF-8:Gasser;Linus;;;
ADR;TYPE=WORK;CHARSET=UTF-8:EPFL IC IINFCOM DEDIS ; BC 208 (Bâtiment BC) ; Station 14 ; CH-1015 Lausanne
EMAIL;TYPE=INTERNET:linus.gasser@epfl.ch
URL:https://people.epfl.ch/linus.gasser
TITLE;CHARSET=UTF-8:
TEL;TYPE=WORK:+41216936770 
ORG:EPFL
END:VCARD
```

PR #1093, opened 2018-02-28
Add test to evoting

ineiti:
This adds an end-to-end test using all the service methods of the evoting system, to make sure that it really works. The other tests use internal functions and not only the external ones.

### DIFF
--- a/evoting/protocol/decrypt_test.go
+++ b/evoting/protocol/decrypt_test.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/dedis/onet"
+	"github.com/dedis/onet/log"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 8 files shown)


evoting/protocol/decrypt_test.go
```diff
@@ -72,11 +73,10 @@ func runDecrypt(t *testing.T, n int) {
...
 	case <-decrypt.Finished:
-		// partials, _ := election.Partials()
-		// for _, partial := range partials {
-		// 	fmt.Println(partial)
-		// 	assert.False(t, partial.Flag)
-		// }
+		partials, _ := election.Partials()
+		for _, partial := range partials {
+			log.Lvl1(partial)
+		}
```
kc1212: no assertion here?

evoting/protocol/shuffle_test.go
```diff
@@ -66,12 +67,12 @@ func runShuffle(t *testing.T, n int) {
...
-		// mixes, _ := election.Mixes()
+		box, _ := election.Box()
+		mixes, _ := election.Mixes()
 
-		// x, y := lib.Split(box.Ballots)
-		// v, w := lib.Split(mixes[0].Ballots)
-		// fmt.Println(lib.Verify(mixes[0].Proof, election.Key, x, y, v, w))
+		x, y := lib.Split(box.Ballots)
+		v, w := lib.Split(mixes[0].Ballots)
+		log.Lvl1(lib.Verify(mixes[0].Proof, election.Key, x, y, v, w))
```
kc1212: should we check the verification result?

evoting/protocol/decrypt_test.go
ineiti: I only uncommented this from Andrea's code - the assertion there was failing. So I thought that I might at least replace the `fmt.Pringln` with a `log.Lvl1`...

PR #1097, opened 2018-02-28
evoting: Election - replace Data with Candidates and add MaxChoices

gnarula:
Replaces `Election.Data` with `Election.Candidates`. Changed the field type to `[]uint32` since scipers will be stored in it. Also introduced `Election.MaxChoices` to store maximum votes allowed in a ballot. Added a `Start` field to denote election start unix timestamp.

### DIFF
--- a/evoting/lib/election.go
+++ b/evoting/lib/election.go
@@ -35,7 +35,7 @@ type Election struct {
 	Stage  uint32                // Stage indicates the phase of the election.
 
 	Candidates  []uint32 // Candidates is the list of candidate scipers.
-	MaxChoices  uint32   // MaxChoices is the max votes in allowed in a ballot.
+	MaxChoices  int      // MaxChoices is the max votes in allowed in a ballot.
 	Description string   // Description in string format.
 	Start       int64    // Start denotes the election start unix timestamp
 	End         int64    // End (termination) datetime as unix timestamp.
--- a/evoting/lib/election.go
+++ b/evoting/lib/election.go
@@ -37,6 +37,7 @@ type Election struct {
 	Candidates  []uint32 // Candidates is the list of candidate scipers.
 	MaxChoices  uint32   // MaxChoices is the max votes in allowed in a ballot.
 	Description string   // Description in string format.
+	Start       int64    // Start denotes the election start unix timestamp
 	End         int64    // End (termination) datetime as unix timestamp.
 }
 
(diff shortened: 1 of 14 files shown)


evoting/lib/election.go
```diff
@@ -34,9 +34,10 @@ type Election struct {
 	Key    kyber.Point           // Key is the DKG public key.
 	Stage  uint32                // Stage indicates the phase of the election.
 
-	Data        []byte // Data can hold auxiliary information.
-	Description string // Description in string format.
-	End         string // End (termination) date.
+	Candidates  []uint32 // Candidates is the list of candidate scipers.
+	MaxChoices  uint32   // MaxChoices is the max votes in allowed in a ballot.
+	Description string   // Description in string format.
+	End         string   // End (termination) date.
```
jeffallen: In Go, normally this would always be a time.Time. The only exception is at the edge of a system, where you are interacting with another system (i.e. using protobuf to send this via to JS?). So, does this one qualify for the exception, or are you going to make this time.Time, like Rob Pike intended? :)

ineiti: I would do both as `int64` and comment that they are in unix time format...

gnarula: 👍 changed in the new commit.

evoting/lib/election.go
```diff
@@ -34,9 +34,11 @@ type Election struct {
 	Key    kyber.Point           // Key is the DKG public key.
 	Stage  uint32                // Stage indicates the phase of the election.
 
-	Data        []byte // Data can hold auxiliary information.
-	Description string // Description in string format.
-	End         string // End (termination) date.
+	Candidates  []uint32 // Candidates is the list of candidate scipers.
+	MaxChoices  uint32   // MaxChoices is the max votes in allowed in a ballot.
```
ineiti: it's nice to have this as `uint32`, but for convenience I would put this as `int`. Else you have to always cast it when you compare lengths.

Except if @jeffallen says that this has to be `uint32` ;) But then I guess we'll have a ton of other places where it should be `uint32` and not `int`...

jeffallen: Again, the Go way is, "int unless something about the data type actually requires that you tell the exact number of bits". When crossing systems, from Go to JS or from Go to the kernel, this is sometimes when you should use explicit numbers of bits.

Is that helpful?

gnarula: +1 to saving some casts while counting in future. I'll set it as `int32` in the protobuf, and since values are in the range [0-9] we won't lose any info in Go either.

Issue #1100, opened 2018-02-28
TestClient_ParallelWrite might be too heavy for a normal laptop


kc1212:
The `TestClient_ParallelWrite` test introduced in https://github.com/dedis/cothority/commit/0d5ed122f185ca1ce0284c5812777f447adf376a doesn't work locally on my machine 

The error I get is
```
Didn't finish in time
goroutine profile: total 22405
22269 @ 0x102f39c 0x102f48e 0x1006534 0x10061db 0x137676f 0x105dc51
#	0x137676e	github.com/dedis/onet.(*TreeNodeInstance).waitDispatch+0x47e		/Users/cong/go/src/github.com/dedis/onet/treenode.go:406
#	0x137676e	github.com/dedis/onet.(*TreeNodeInstance).dispatchMsgReader+0x47e	/Users/cong/go/src/github.com/dedis/onet/treenode.go:441
...
and many more stack traces
...
```

If I comment this test out then all skipchain tests pass.

PR #1101, opened 2018-02-28
Make TestClient_ParallelWrite shorter

kc1212:
@jeffallen please have a look. Turning the number of blocks down to 50 allows the test to pass (in about 66 seconds on my machine). If you think there's a deeper issue in the code please feel free to close this and we can discuss. We may also need to think about what happens if there are too many store block requests, should the server stop accepting blocks if there are too many in the queue?

Fixes https://github.com/dedis/cothority/issues/1100

### DIFF
--- a/skipchain/api_test.go
+++ b/skipchain/api_test.go
@@ -427,7 +427,7 @@ type testData struct {
 
 func TestClient_ParallelWrite(t *testing.T) {
 	numClients := 20
-	numWrites := 100
+	numWrites := 50
 	if testing.Short() {
 		numClients = 2
 	}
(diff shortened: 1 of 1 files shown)


ineiti:
We should perhaps do a `big bad global lock` in the `skipchain`-package, to make sure there is only one request pending. Or, if you want to make this clean, create a queue of incoming requests, and then handle them one by one.

But again, this is definitively something we should tackle with some asynchronous interface, where a client can request the inclusion of a key/value pair, and the leader periodically fetches all requests to include them in a block.

I promised Darik we have this by the end of March - but it's getting closer fast ;)

jeffallen:
Fine with me. It already has a less demanding config for t.Short, where are you seeing that 50 works but 100 not? And how does it not work?

kc1212:
It's described in https://github.com/dedis/cothority/issues/1100, basically timeouts

PR #1102, opened 2018-03-01
Genesis id 1088

### LINKED ISSUE #1088: skipchain - hide `GensisID`
Because `skipblock.GenesisID` is nil for the genesis block we should make it private and only use `SkipchainID()` wherever the GenesisID is needed. This would avoid quite some pitfalls for users.

ineiti:
Closes #1088 

Instead of making `GenesisID` private, as proposed, it is replaced by `SkipChainID()` wherever possible and a comment on the `GenesisID` field is added.

### DIFF
--- a/scmgr/app.go
+++ b/scmgr/app.go
@@ -359,7 +359,7 @@ func scAdd(c *cli.Context) error {
 	}
 	cfg.Db.Store(ssbr.Latest)
 	log.ErrFatal(cfg.save(c))
-	log.Infof("Added new block %x to chain %x", ssbr.Latest.Hash, ssbr.Latest.GenesisID)
+	log.Infof("Added new block %x to chain %x", ssbr.Latest.Hash, ssbr.Latest.SkipChainID())
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


skipchain/struct.go
```diff
@@ -259,7 +259,9 @@ type SkipBlockFix struct {
 	// SkipBlockParent points to the SkipBlock of the responsible Roster -
 	// is nil if this is the Root-roster
 	ParentBlockID SkipBlockID
-	// GenesisID is the ID of the genesis-block.
+	// GenesisID is the ID of the genesis-block. For the genesis-block, this
+	// is null. The SkipBlockID() methods returns the correct ID both for
```
jeffallen: s/methods/method/

scmgr/app.go
```diff
@@ -601,9 +601,9 @@ func cleanJSFiles(dir string) error {
 
 // JSON skipblock element to be written in the index.js file
 type jsonBlock struct {
-	GenesisID string
-	Servers   []string
-	Data      []byte
+	SkipchainID string
```
jeffallen: This will require a matching fix on status.dedis.ch web app.

PR #1103, opened 2018-03-01
adding timing to see where the tests are slow

ineiti:
Don't merge - just see which tests are _slow_


ineiti:
Removed slow tests for the moment.

PR #1109, opened 2018-03-02
Correctly add genesis block

kc1212:
Previously, in https://github.com/dedis/cothority/pull/1102, `psbd.NewBlock.GenesisID.IsNull()` was changed to `psbd.NewBlock.SkipChainID().IsNull()`. This change affected the logic of `StoreSkipBlock` because `psbd.NewBlock.SkipChainID().IsNull()` is always false and the `TestService_ParallelStoreBlock` test started failing.

I'm curious why this bug didn't break more tests.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -169,7 +169,7 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, er
 	var prev *SkipBlock
 	var changed []*SkipBlock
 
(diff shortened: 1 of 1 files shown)


ineiti:
I think that most of the time a new genesis block is proposed with an empty `Hash`, so this would pass, as `SkipChainID` would return that empty `Hash` and it would return true for `IsNull`. But if you copy an existing block and send it here, it wouldn't pass.

Anyway, using `Index` here is definitively better!

PR #1115, opened 2018-03-06
Fix documentation to say that LatestID may be updated

### LINKED ISSUE #1110: StoreSkipBlock potentially overwrites prev when it is supplied by the Client
From reading the code, `prev` is overwritten on this line https://github.com/dedis/cothority/blob/4844826d4b0fc2b8402bcaa36a980e6f76b45ac9/skipchain/skipchain.go#L263 even when it is supplied by the client https://github.com/dedis/cothority/blob/4844826d4b0fc2b8402bcaa36a980e6f76b45ac9/skipchain/skipchain.go#L215

Unless I missed something, I think this is not the right behaviour. 

If the client knows the latest block and wants to add the new block after it, we should respect its wish.

kc1212:
I give my reasons for why fixing the documentation may be a better idea in this issue https://github.com/dedis/cothority/issues/1110

Closes #1110

### DIFF
--- a/skipchain/api.go
+++ b/skipchain/api.go
@@ -26,7 +26,7 @@ func NewClient() *Client {
 }
 
 // StoreSkipBlockSignature asks the cothority to store the new skipblock, and
-// eventually attach it to after the target skipblock.
+// eventually attach it after the target skipblock.
 //  - target is a skipblock, and the new skipblock is going to be added after
 //    it, not necessarily immediately after it.  The caller should use the
 //    genesis skipblock as the target. But for backward compatibility, any
--- a/skipchain/api.go
+++ b/skipchain/api.go
@@ -72,7 +72,7 @@ func (c *Client) StoreSkipBlockSignature(latest *SkipBlock, ro *onet.Roster, d n
 		}
 		sig = &signature
 	}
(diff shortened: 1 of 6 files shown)


kc1212:
New version available: changed `LatestID` to `TargetSkipChainID`. The client is encouraged to supply the `TargetSkipChainID`, but for backward compatibility reasons, it can also be used as a skipblock ID.

skipchain/api.go
```diff
@@ -25,24 +25,27 @@ func NewClient() *Client {
 	return &Client{Client: onet.NewClient(cothority.Suite, "Skipchain")}
 }
 
-// StoreSkipBlockSignature asks the cothority to store the new skipblock, and eventually
-// attach it to the 'latest' skipblock.
-//  - latest is the skipblock where the new skipblock is appended. If ro and d
-//   are nil, a new skipchain will be created with 'latest' as genesis-block.
+// StoreSkipBlockSignature asks the cothority to store the new skipblock, and
+// eventually attach it to after the target skipblock.
```
jeffallen: s/it to after/it after/         I think that's easier to read, no?

jeffallen:
I agree with this, but I'm leaving it to Linus to merge it since I'm only halfway here these days.

PR #1117, opened 2018-03-07
exclude node modules from node target and support WS in nativescript

gnarula:
The build for `node` target included all the dependencies in minified form (think of this as JS equivalent of a fat JAR). This commit excludes the dependencies in the node build (since npm is capable of resolving them).

This is required for [nativescript-nodeify](https://github.com/EddyVerbruggen/nativescript-nodeify) to parse the build and replace `ws` with a nativescript compatible dependency.

TL;DR: makes it possible to use the package in nativescript land :)


gnarula:
Don't merge yet please. found a small issue on browser

ineiti:
If you uses zenhub, you can simply put it in `WIP` and we will try not to merge it ;)

PR #1121, opened 2018-03-11
Fix shuffle verification bug

qantik:
This fixes a long-standing bug in the Neff shuffle verification routing inside ```evoting/protocol```.
As it turned out the code making sure that only the most recent ballot of a user is fetched from
the skipchain invalidated the correct ballot order.
Thus the Neff shuffle code in ```kyber/shuffle``` works fine.

### DIFF
--- a/evoting/service/service.go
+++ b/evoting/service/service.go
@@ -192,14 +192,14 @@ func (s *Service) Login(req *evoting.Login) (*evoting.LoginReply, error) {
 }
 
 // LookupSciper calls https://people.epfl.ch/cgi-bin/people/vCard?id=sciper
-// to convert scipers to names
+// to convert Sciper numbers to names.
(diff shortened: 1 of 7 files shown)


evoting/service/service.go
```diff
@@ -192,22 +192,22 @@ func (s *Service) Login(req *evoting.Login) (*evoting.LoginReply, error) {
 }
 
 // LookupSciper calls https://people.epfl.ch/cgi-bin/people/vCard?id=sciper
-// to convert scipers to names
+// to convert Sciper numbers to names.
 func (s *Service) LookupSciper(req *evoting.LookupSciper) (*evoting.LookupSciperReply, error) {
 	if len(req.Sciper) != 6 {
-		return nil, errors.New("sciper should be 6 digits only")
+		return nil, errors.New("Sciper should be 6 digits only")
```
ineiti: Errors are supposed to be lower-case, so that you can join them with other errors.

PR #1122, opened 2018-03-12
Add Subtitle, MoreInfo and Footer fields

gnarula:
Fields suggested after discussion with Lindo about the layout.


kc1212:
@gnarula do you know why the evoting build is failing?

gnarula:
@kc1212 Fixed with the latest commit. Forgot to change a test before.

PR #1126, opened 2018-03-14
Bftcosi optimisation byzgen - master

kc1212:
The algorithm does not wait for all messages to arrive, but proceed when
2/3 of the messages are at the root.


kc1212:
I feel this PR shouldn't go in master. It changes a lot of the functionality to a protocol is kept simply for historical purposes. Maybe it should only go into the byzgen_poc branch. What do you think @ineiti ?

ineiti:
I also think it should be closed - thanks anyway.

PR #1127, opened 2018-03-15
Bftcosi optimisation byzgen - byzgen_poc

kc1212:
This PR implements an optimised version of bftcosi where the root node does not wait for all replies from its children, it makes progress when 2/3rd of the replies arrive. This so that in an event of a failure, we do not need to wait for a long timeout.

### DIFF
--- a/bftcosi/bftcosi.go
+++ b/bftcosi/bftcosi.go
@@ -267,6 +267,15 @@ func (bft *ProtocolBFTCoSi) Dispatch() error {
 // signature will be nil.
 func (bft *ProtocolBFTCoSi) Signature() *BFTSignature {
 
+	bft.successfulMut.Lock()
+	ok := bft.successful
+	bft.successfulMut.Unlock()
+	if !ok {
+		return &BFTSignature{
+			Msg: bft.Msg,
+		}
+	}
+
 	// create exceptions for the nodes that were late to respond in the
 	// commit phase
 	var i int
@@ -278,17 +287,6 @@ func (bft *ProtocolBFTCoSi) Signature() *BFTSignature {
 				Commitment: bft.Suite().Point().Null(),
 			}
 			i++
-		}
-	}
-
-	bft.successfulMut.Lock()
-	ok := bft.successful
-	bft.successfulMut.Unlock()
-	if !ok {
-		return &BFTSignature{
-			Sig:        nil,
-			Msg:        bft.Msg,
-			Exceptions: exceptions,
 		}
 	}
 
@@ -411,7 +409,16 @@ func (bft *ProtocolBFTCoSi) handleChallengePrepare(msg challengePrepareChan) err
 		bft.prepare.Challenge(ch.Challenge)
 	}
 	go func() {
-		bft.verifyChan <- bft.VerificationFunction(bft.Msg, bft.Data)
+		select {
+		case bft.verifyChan <- bft.VerificationFunction(bft.Msg, bft.Data):
+		case <-time.After(bft.Timeout):
+			log.Error(bft.Name(), "verification didn't complete after", bft.Timeout)
+			// we might not have a reader on bft.verifyChan if nodes exit early
+			select {
+			case bft.verifyChan <- false:
+			default:
+			}
+		}
 	}()
 	if bft.IsLeaf() {
 		return bft.handleResponsePrepare(bft.responseChan)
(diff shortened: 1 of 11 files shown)


ineiti:
Please provide a description for the PR.

bftcosi/bftcosi.go
```diff
@@ -110,34 +107,43 @@ type collectStructs struct {
...
 	// temporary buffer of "commit" responses
 	tempCommitResponse []kyber.Scalar
+
+	committedInPreparePhase bool
+	committedInCommitPhase  bool
 }
 
 // NewBFTCoSiProtocol returns a new bftcosi struct
-func NewBFTCoSiProtocol(n *onet.TreeNodeInstance, verify VerificationFunction) (*ProtocolBFTCoSi, error) {
+func NewBFTCoSiProtocol(n *onet.TreeNodeInstance, verify VerificationFunction, to time.Duration) (*ProtocolBFTCoSi, error) {
```
ineiti: Here you are changing the signature of the method, which is not allowed during the current development-cycle (no API-changes until September 2018 ;)

bftcosi/bftcosi.go
```diff
@@ -210,59 +217,84 @@ func (bft *ProtocolBFTCoSi) Dispatch() error {
...
+			}
+			bft.successfulMut.Lock()
+			bft.successful = true
+			bft.successfulMut.Unlock()
+			// we return here even when prepare is not done yet,
+			// because we may not have received the challenge for
+			// the prepare phase
+			return nil
+		case <-time.After(bft.Timeout):
+			return fmt.Errorf("timeout waiting for challenge - " +
```
ineiti: `errors.New` if there is no string-formatting to do.

bftcosi/bftcosi.go
```diff
@@ -210,59 +217,84 @@ func (bft *ProtocolBFTCoSi) Dispatch() error {
...
-		for _, ex := range bft.tempExceptions {
-			if ex.Commitment.Equal(bft.Suite().Point().Null()) {
-				bftSig.Exceptions = append(bftSig.Exceptions, ex)
+	// create exceptions for the nodes that were late to respond in the
+	// commit phase
+	var i int
+	exceptions := make([]Exception, bft.AllowedExceptions)
+	for _, tn := range bft.Children() {
+		if _, ok := bft.respondedNodesCommit[tn.ServerIdentity.Public]; !ok {
+			exceptions[i] = Exception{
```
ineiti: With regard to the error we saw - are we guaranteed here that the following holds true? `i < bft.AllowedExceptions`?

bftcosi/packets.go
```diff
@@ -187,11 +189,9 @@ type responseChan struct {
 }
 
 // Exception represents the exception mechanism used in BFTCosi to indicate a
-// signer did not want to sign.
-// The index is the index of the public key of the cosigner that do not want to
-// sign.
-// The commit is needed in order to be able to
-// correctly verify the signature
+// signer did not want to sign.  The index is the index of the public key of
```
ineiti: can you do a `:%s/   / /g`?

kc1212:
@ineiti sorry about that, description is up

pop/service/service.go
```diff
@@ -708,7 +708,13 @@ func (s *Service) PropagateFinal(msg network.Message) {
 //signs FinalStatement with BFTCosi and Propagates signature to other nodes
 func (s *Service) signAndPropagate(final *FinalStatement, protoName string,
 	data []byte) error {
-	tree := final.Desc.Roster.GenerateNaryTreeWithRoot(2, s.ServerIdentity())
+
```
ineiti: no empty lines at beginning of method.

bftcosi/bftcosi.go
kc1212: I'm merging to the byzgen_poc branch. Is this also not allowed?

bftcosi/bftcosi.go
kc1212: I can't provide a 100% guarantee, but it should. The only way for it to be false would be a programmer error, which should make the code panic and then we need to fix the bug.

bftcosi/packets.go
kc1212: This is the result of my auto comment formatting, to keep it fully justified. Vim's `gq` does this by default.

PR #1130, opened 2018-03-16
evoting: check end date on election creation

gnarula:
Ensures elections aren't created with an end time before the current time.

### DIFF
--- a/evoting/service/open_test.go
+++ b/evoting/service/open_test.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"testing"
+	"time"
 
 	"github.com/dedis/onet"
 	"github.com/dedis/onet/network"
@@ -65,7 +66,7 @@ func TestOpen_CloseConnection(t *testing.T) {
 	assert.NotNil(t, err)
 }
 
-func TestOpen_Full(t *testing.T) {
+func TestOpen_EndLessThanNow(t *testing.T) {
 	local := onet.NewLocalTest(cothority.Suite)
 	defer local.CloseAll()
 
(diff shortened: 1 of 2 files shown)


evoting/service/open_test.go
```diff
@@ -77,6 +78,22 @@ func TestOpen_Full(t *testing.T) {
 	master.GenChain(nil)
 
 	election := &lib.Election{}
+	_, err := s.Open(&evoting.Open{Token: token, ID: master.ID, Election: election})
+	assert.NotNil(t, err)
```
kc1212: Since `Open` might return other errors, could you check that `err.Error()` contains the "election cannot end before current time" string please?

gnarula: 👍 added assertion in new commit

evoting/service/service.go
```diff
@@ -111,6 +111,11 @@ func (s *Service) Open(req *evoting.Open) (*evoting.OpenReply, error) {
 		return nil, err
 	}
 
+	// sanity check - do not allow elections to be created retrospectively
+	if req.Election.End < time.Now().Unix() {
```
ineiti: We need to take care here and do an audit of the code, starting right here: if only the 'entry-point-service' checks for valid data, then we have a security problem. So all of these things should be verified in a verification function that is applied to the evoting-skipchain.

Whenever a new master- or voting-skipchain is created, the corresponding verification-functions need to be appended to the creation of the genesis-block:

https://github.com/dedis/cothority/blob/master/evoting/lib/chains.go#L12

`skipchain.VerificationStandard` should be replaced with `election.ElectionVerification` or so.

PR #1133, opened 2018-03-16
Improve godoc; add synopsis

kc1212:
Documentation improvement on all packages


evoting/app/app.go
```diff
@@ -1,3 +1,4 @@
+// This is command line interface for communicating with the evoting service.
```
ineiti: s/is command/is a command/

ftcosi/README.md
```diff
@@ -40,6 +40,43 @@ etc.) are not yet handled.
 The purpose of the project is to test scalability and robustness of this
 service on a testbed and to have a well-documented reusable code for it.
 
+## Implementation
+The protocol has four messages: 
+* Announcement which is sent from the root down the tree and announce the
```
ineiti: s/\*/-/

ftcosi/README.md
```diff
@@ -40,6 +40,43 @@ etc.) are not yet handled.
...
+
+```
+       root
+         |
+         |
+     sub-leader
+      /       \
+     /         \
+  leaf_1 ... leaf_n
+```
```
ineiti: We need the same for `CoSi` and link to there from here. Or write it here and the copy it to the cosi-protocol.

kc1212: Added a section on implementation in cosi.

PR #1134, opened 2018-03-19
Evoting Transactions

### LINKED ISSUE #1092: evoting - save private keys
The conodes should store their shares of the private key and load it again when restarted.

### LINKED ISSUE #1132: evoting verification functions
Currently the election service checks the validity of the input only at the entry-point to the service, but not in each node when a new block is created.

Whenever a new master- or voting-skipchain is created, the corresponding verification-functions need to be appended to the creation of the genesis-block:

https://github.com/dedis/cothority/blob/master/evoting/lib/chains.go#L12

`skipchain.VerificationStandard` should be replaced with `election.ElectionVerification` or so.

I think that @qantik is already looking into that.

qantik:
This pull request adds transactions to the evoting package. Now, all conodes of an election have to agree on a new block before it is appended. As a consequence the old login mechanism via tokens is history and in general the codebase is much simpler. Furthermore, some API messages
have been adjusted with the addition of a new ```GetElections``` message.

For the moment the signature check on a transaction is void because I'm not sure what parts of it have
effectively to be signed.

Closes #1132 
Closes #1092


evoting/lib/chains.go
```diff
@@ -1,22 +1,31 @@
...
+)
 
-// New creates a new skipchain for a given roster and stores data in the genesis block.
-func New(roster *onet.Roster, data interface{}) (*skipchain.SkipBlock, error) {
-	return client.CreateGenesis(roster, 1, 1, skipchain.VerificationStandard, data, nil)
+// NewSkipchain creates a new skipchain for a given roster and verification function.
+func NewSkipchain(roster *onet.Roster, verifier []skipchain.VerifierID, data interface{}) (
+	*skipchain.SkipBlock, error) {
+	client := skipchain.NewClient()
+	return client.CreateGenesis(roster, 1, 1, verifier, data, nil)
```
ineiti: We should start using higher values for these numbers. Else we cannot take advantage of the _skip_ part in skipchains ;) Something like `8, 4`, which will create a link every 8th block and up to 4096 blocks in the past.

Have a look at this definition: https://github.com/dedis/cothority/blob/master/skipchain/api.go#L100

evoting/lib/chains.go
```diff
@@ -1,22 +1,31 @@
...
+func Store(id skipchain.SkipBlockID, roster *onet.Roster, data ...interface{}) error {
+	client := skipchain.NewClient()
+	reply, err := client.GetUpdateChain(roster, id)
 	if err != nil {
-		return nil, err
+		return err
+	}
+
+	for _, d := range data {
+		_, err = client.StoreSkipBlock(reply.Update[len(reply.Update)-1], nil, d)
```
ineiti: According to https://github.com/dedis/cothority/blob/master/skipchain/api.go#L85, this should now be the skipchain-id.

So you won't even need the call to `GetUpdateChain` anymore.

evoting/lib/election.go
```diff
@@ -53,94 +58,56 @@ type footer struct {
...
 
-	n, numMixes, numPartials := len(election.Roster.List), 0, 0
-	for _, block := range chain {
-		_, blob, _ := network.Unmarshal(block.Data, cothority.Suite)
-		if _, ok := blob.(*Mix); ok {
-			numMixes++
-		} else if _, ok := blob.(*Partial); ok {
-			numPartials++
+	n, mixes, partials := len(election.Roster.List), 0, 0
+	for _, block := range reply.Update {
```
ineiti: This will break with a non-1 `BaseHeight` as proposed above, because `GetUpdateChain` will return the _shortest_ path from the genesis to the last block.

Furthermore, this might take quite some time if we have 1000s of votes on the skipchain. My proposition is to store that information in the service and save it every time it is updated (what we do in most other services). Even though it means that the service needs to hold state then.

For a future version, we'll be able to use the skipchains in a more database-like way and have the whole state available in a provable way at every block. But that's for later.

Also think about what to do with the other `GetUpdateChain` calls you have in the code.

evoting/lib/transaction.go
```diff
@@ -0,0 +1,117 @@
...
+	Ballot   *Ballot
+	Mix      *Mix
+	Partial  *Partial
+
+	User      uint32
+	Signature []byte
+}
+
+// UnmarshalTransaction decodes a data blob to a transaction structure.
+func UnmarshalTransaction(data []byte) *Transaction {
```
ineiti: We should ban the use of `network.Marshal` - it doesn't play nice with non-onet code :( Can you replace that with regular `protobuf.Encode` and `protobuf.Decode` calls?

evoting/lib/transaction.go
```diff
@@ -0,0 +1,117 @@
...
+	_, blob, _ := network.Unmarshal(data, cothority.Suite)
+	transaction, _ := blob.(*Transaction)
+	return transaction
+}
+
+// NewTransaction constructs a new transaction for the given arguments.
+func NewTransaction(data interface{}, user uint32, signature []byte) *Transaction {
```
ineiti: Question: what is the `signature` here?

evoting/service/service.go
```diff
@@ -459,94 +410,91 @@ func (s *Service) NewProtocol(node *onet.TreeNodeInstance, conf *onet.GenericCon
...
-		}
-
-		if admin && !election.IsCreator(stamp.user) {
-			return nil, errNotCreator
-		} else if !admin && !election.IsUser(stamp.user) {
-			return nil, errNotPart
-		}
-
-		return election, nil
+	if transaction.Verify(skipblock.GenesisID, skipblock.Roster) != nil {
```
ineiti: `return transaction.Verify(...) == nil`

evoting/lib/transaction.go
qantik: Fixed.

evoting/lib/transaction.go
qantik: That's the front-end signature.

evoting/lib/chains.go
qantik: Fixed.

evoting/lib/election.go
qantik: Keeping track of the election stage turned out to be more complicated than I initially thought. Especially now with the transaction verification function that is executed on all nodes. There are
cases when it is not clear. E.g. during protocols the leaf nodes never know if the protocol has finished correctly.
I decided to drop the field altogether. The front-end can also use the getter messages like GetMixes and GetPartials to determine the election stage.
The shared secret however is now stored on disk. This fixes #1092.

evoting/lib/chains.go
qantik: I'm not sure how I can store a new block by just using the skipchain ID?
https://github.com/dedis/cothority/blob/master/skipchain/api.go#L85 still takes the latest block
as an argument.

PR #1140, opened 2018-03-21
making testing of bft easier

ineiti:
Correctly fail BFTCoSi if the server is stopped.
Also cleans up correctly by not calling `Shutdown` itself.

### DIFF
--- a/bftcosi/bftcosi.go
+++ b/bftcosi/bftcosi.go
@@ -221,9 +221,12 @@ func (bft *ProtocolBFTCoSi) Dispatch() error {
 	var gotChallengePrepare bool
 	for {
 		select {
-		case c := <-bft.challengePrepareChan:
+		case c, ok := <-bft.challengePrepareChan:
+			if !ok {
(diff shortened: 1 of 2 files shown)


kc1212:
Travis is complaining about data races - moving back to WIP.

alsosweet:
When running bftcosi_test.go used bf=2,test fail.Does BFTCoSi support binary tree? I run go test just change bf=2

kc1212:
@alsosweet For fault-tolerance reasons, it won't work with binary trees, only broadcast trees (star graphs).

kc1212:
@alsosweet I should add that bftcosi is deprecated, the new version is called byzcoinx which you can find in the mater branch https://github.com/dedis/cothority/tree/master/byzcoinx

alsosweet:
@kc1212 ok, but it test fine in cothority master

kc1212:
@alsosweet most likely it'll be fine if there are no failures. but if a node fails and it's not a leaf node, then binary trees won't work.

PR #1149, opened 2018-03-27
add evoting Dockerfile

gnarula:
Created a new Dockerfile which adds the evoting service to conode.go
before the build.

The resulting image can be pushed to Dockerhub. `dedis/conode:evoting` perhaps?

### DIFF
--- a/conode/experimental.go
+++ b/conode/experimental.go
@@ -1,0 +1,8 @@
+package main
+
+/*
+This imports the experimental services that will not be used in the
+stable branch.
+*/
+
+import _ "github.com/dedis/cothority/evoting/service"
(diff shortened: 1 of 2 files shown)


ineiti:
we had an `experimental.go` in `conode/`-directory that included all services that are experimental and not stable, but it disappeared. I propose we create that file again and put all imports for services that are not in `v2`.

ineiti:
Sorry, I was not clear enough. Here is what I proposed:
- create a file `conode/experimental.go` that includes all services that are not included in `conode/conode.go`
- add that file to `stable/remove_files`, so that it will be removed when building `v2`
done ;)

gnarula:
third time lucky maybe?

PR #1153, opened 2018-03-29
create stable of latest master

ineiti:
Merging latest changes of master into `v2`.


ineiti:
LGTM - who merges now? We're in a deadlock, because we both worked on it and cannot merge it anymore ;)

jeffallen:
First step is to merge #1155 . Then do a new "make create_stable" from that.

PR #1156, opened 2018-04-03
evoting: reintroduce Election Stage

gnarula:
The frontend requires a way to filter out elections based on their
current stage to enable/disable buttons and render the elections
at an appropriate position. This commit caches the current phase
of the election in the Stage field.

Another alternative would've been to expose methods on the service that allow checking the phase for every election but that would be quite expensive.

### DIFF
--- a/evoting/lib/election.go
+++ b/evoting/lib/election.go
@@ -10,6 +10,18 @@ import (
 	"github.com/dedis/cothority/skipchain"
 )
 
+// ElectionState is the type for storing the stage of Election.
+type ElectionState uint32
+
+const (
+	// Running depicts that an election is open for ballot casting
+	Running ElectionState = iota + 1
+	// Shuffled depicts that the mixes have been created
+	Shuffled
+	// Decrypted depicts that the partials have been decrypted
+	Decrypted
+)
+
 func init() {
 	network.RegisterMessages(Election{}, Ballot{}, Box{}, Mix{}, Partial{})
 }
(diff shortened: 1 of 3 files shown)


evoting/lib/election.go
```diff
@@ -10,6 +10,15 @@ import (
 	"github.com/dedis/cothority/skipchain"
 )
 
+const (
```
jeffallen: These constants should be typed, so that they can only be used where they are expected to be. See https://golang.org/doc/effective_go.html#constants.

Thus:
```
type ElectionState uint32
const (
  Running ElectionState = iota + 1
)
```

Sometimes at the edge of the system (i.e. where protobufs are sent over websockets) this can't work and you want to use integers instead, but I don't think this is that case.

evoting/lib/election.go
```diff
@@ -27,6 +36,7 @@ type Election struct {
 	Roster    *onet.Roster          // Roster is the set of responsible nodes.
 	Key       kyber.Point           // Key is the DKG public key.
 	MasterKey kyber.Point           // MasterKey is the front-end public key.
+	Stage     uint32                // Stage indicates the phase of election and is used for filtering in frontend
```
jeffallen: s/uint32/ElectionState/

evoting/service/service.go
```diff
@@ -245,7 +245,8 @@ func (s *Service) GetElections(req *evoting.GetElections) (*evoting.GetElections
 		if err != nil {
 			return nil, err
 		}
-		if election.IsUser(req.User) || election.IsCreator(req.User) {
+		if (election.IsUser(req.User) || election.IsCreator(req.User)) &&
```
jeffallen: This if would be easier to read if the two sides of the && were two nested if's:
```
if election.IsUser(req.User) || election.IsCreator(req.User) {
  if req.Stage == 0 || req.Stage == election.Stage {
    // append...
  }
}
```
The two checks are checking different things, and as such each deserves it's own line, and maybe one or both deserve comments to explain subtleties.

gnarula:
addressed the review comments

PR #1158, opened 2018-04-03
Add a new batch setup mode to be used in container start up

jeffallen:


jeffallen:
I need this to be checked in before I can test the changes I'm going to propose to the Dockerfile to use it.

kc1212:
I don't fully understand the goal of this PR. Doesn't app.InteractiveConfig do the key generation, file-saving and so on? If you need to make it non-interactive, I think it's possible to give the ip:port to ./conode setup in stdin.

jeffallen:
Yes, but then it does all the "get public IP" magic, etc. I want the minimal way to make a valid private.toml, and we will use kubernetes service discovery to talk to the servers afterwards.

conode/conode.go
```diff
@@ -129,6 +149,48 @@ func setup(c *cli.Context) error {
...
+
+		conf := &app.CothorityConfig{
+			Suite:       cothority.Suite.String(),
+			Public:      pub,
+			Private:     priv,
+			Address:     serverBinding,
+			Description: c.String("description"),
+		}
+
+		dir := cfgpath.GetConfigPath("conode")
```
kc1212: Maybe better to use `DefaultName`?

conode/conode.go
```diff
@@ -129,6 +149,48 @@ func setup(c *cli.Context) error {
...
+		out := path.Join(dir, "private.toml")
+		err := conf.Save(out)
```
kc1212: Are you sure that `app.InteractiveConfig` won't overwrite this file? I had a quick look at `app.InteractiveConfig` and didn't see anything that checks for existing configs.

conode/conode.go
```diff
@@ -129,6 +149,48 @@ func setup(c *cli.Context) error {
 	if c.String("debug") != "" {
 		log.Fatal("[-] Debug option cannot be used for the 'setup' command")
 	}
+
+	if c.Bool("batch") {
```
kc1212: I'm not sure about the naming, it's not doing a batch task. Maybe `--non-interactive` (if `app.InteractiveConfig` is actually non-interactive).

PR #1159, opened 2018-04-03
evoting: verify signature on every user transaction

gnarula:
This ensures the signature is verified for all user transactions.

### DIFF
--- a/evoting/lib/transaction.go
+++ b/evoting/lib/transaction.go
@@ -2,8 +2,10 @@ package lib
 
 import (
 	"errors"
+	"strconv"
 	"time"
 
+	"github.com/dedis/kyber/sign/schnorr"
 	"github.com/dedis/onet"
 	"github.com/dedis/onet/network"
 	"github.com/dedis/protobuf"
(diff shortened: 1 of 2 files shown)


evoting/lib/transaction.go
```diff
@@ -75,8 +77,29 @@ func NewTransaction(data interface{}, user uint32, signature []byte) *Transactio
...
+		election, _ = GetElection(roster, genesis)
+	}
+	// Master or Link transaction
+	if election == nil {
+		return nil
+	}
+	message := election.Master
+	for _, c := range strconv.Itoa(int(t.User)) {
+		d, _ := strconv.Atoi(string(c))
+		message = append(message, byte(d))
```
kc1212: Is it not possible to do `append(message, c)`?

jeffallen:
As we discussed, this signature is probably open to a replay attack, so the next iteration will be to add a timestamp, and sign it. On the receiving side, if the signature is valid, then the timestamp can be trusted, and then it needs to be checked for freshness.

evoting/lib/transaction.go
jeffallen: Would have to change the Javascript side too. This is not really ideal, but it works, and I think it is important to keep moving forward on this.

PR #1166, opened 2018-04-09
cothorityjs: add LeaderSocket

gnarula:
LeaderSocket reads a roster and allows sending messages to the
leader node. As of now the leader node is the first node in the
roster.

### DIFF
--- a/external/js/cothority/lib/net/index.js
+++ b/external/js/cothority/lib/net/index.js
@@ -1,10 +1,10 @@
 "use strict";
 
 const net = require("./net");
-const Socket = net.Socket;
-const RosterSocket = net.RosterSocket;
+const { Socket, RosterSocket, LeaderSocket } = net;
 
 module.exports = {
-    Socket,
-    RosterSocket,
+  Socket,
+  RosterSocket,
+  LeaderSocket
 };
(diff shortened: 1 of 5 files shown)


gnarula:
@nikkolasg This is something @ineiti and I discussed earlier this morning. From what I understand, it's only the leader that has the capability to add a new block to the skipchain. This would allow saving some internal requests from conode N to the leader and offer some performance improvements. All the nodes still end up verifying the transaction before it's appended to the chain.

ineiti:
New blocks always have to be proposed by the leader. If another node wants to propose new blocks, this requires a view-change. The new _omniledger_ (insert great music here) implementation will allow nodes to collect transactions that will be collected by the leader and inserted into a new block.

nikkolasg:
I'm not sure centralizing all client requests on one server actually improves performance and reliability. Usually you would expect any nodes can receive transactions and there's some kind of gossiping network that shares every transactions seen. Also, there's lot of problems concerning the fact that only the leader has access to the transactions, he can just choose whatever he wants and no other nodes will be able to detect if he is purposedly dropping some transactions or not, etc...

Anyway, I don't want to interpose, just wanted to raise my concerns about this approach, so it's always in the back of your mind when thinking about the final solution... :)

ineiti:
You're completely right with your comments, this is what we're working on in the new omniledger implementation. But this will unfortunately not be ready for the epfl e-voting schedule.

PR #1172, opened 2018-04-11
adding new api call to fetch proposed config

ineiti:
This adds a new api call to the pop service so that the mobile frontend doesn't have to use pastebin anymore for passing the configurations.

Every time a new configuration is stored using `StoreConfig`, the node sends the configuration to all included nodes as a proposed configuration. The owners of the other nodes can now fetch this proposed configuration, inspect it, and if they agree, they can propose it themselves to their nodes.


pop/app.go
```diff
@@ -203,10 +203,56 @@ func orgConfig(c *cli.Context) error {
...
 
+// read all newly proposed configs
+func orgProposed(c *cli.Context) error {
+	if c.NArg() < 1 {
+		return errors.New("Please give IP:Port of the conode you want to read proposed configs from")
+	}
+	client := service.NewClient()
+	proposed, err := client.GetProposals(network.Address("tls://" + c.Args().First()))
+	if err != nil {
+		log.Print()
```
kc1212: remove?

pop/service/api.go
```diff
@@ -213,6 +225,20 @@ func encodeMapFinal(stmts map[string]*FinalStatement) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// ToToml returns a string that can be decoded using toml.
+func (desc *PopDesc) ToToml() (string, error) {
```
kc1212: where is this function used? further, this ToToml isn't consistent with FinalStatement.ToToml, which returns ([]byte, error).

pop/service/service.go
```diff
@@ -91,9 +91,9 @@ type Service struct {
 	path string
 	data *saveData
 	// propagate final message
-	PropagateFinalize messaging.PropagationFunc
-	// propagate merge info
-	PropagateMerging messaging.PropagationFunc
+	propagateFinalize messaging.PropagationFunc
+	// propagate possible new descriptions
+	propagateDescription messaging.PropagationFunc
```
kc1212: I imagine this change would change the API

ineiti: Got me. But this should have been private anyway, nobody else should be calling the `propagateMerging` function that is not initialised anywhere!

pop/service/api.go
ineiti: Removed it - it's a mess:
- reading from the file it's a `PopDescGroupToml`
- getting from the service it's a `PopDesc`
- `shortDescToml` is used for the final statement

So probably `shortDescToml` should be replaced by `PopDescGroupToml`. That's an exercise for later...

PR #1173, opened 2018-04-11
Test for db name update

kc1212:
These are the tests concerning https://github.com/dedis/onet/pull/401
we keep them in cothority because the behaviour is primarily observed in conode

### DIFF
--- a/conode/test.sh
+++ b/conode/test.sh
@@ -9,6 +9,7 @@ main(){
     setupConode
     test Build
     test Conode
+    test Database
     stopTest
 }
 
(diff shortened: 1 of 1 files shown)


conode/test.sh
```diff
@@ -21,6 +22,23 @@ testConode(){
     testFail runCo 1 check -g public.toml
 }
 
+testDatabase(){
+    runCoBG 1
+    sleep .1
```
ineiti: Is this needed? I think that `runCoBG` checks whether the conode started correctly. Same below.

conode/test.sh
```diff
@@ -21,6 +22,23 @@ testConode(){
...
+    hexPK=$(cat co1/public.toml  | grep 'Public = '  | cut -d = -f 2 - | tr -d ' "')
+    shaPK=$(ls -t "$CONODE_SERVICE_PATH" | head -n 1 | sed 's/.db$//g')
+    testOK mv "$CONODE_SERVICE_PATH/$shaPK.db" "$CONODE_SERVICE_PATH/$hexPK.db"
+
+    # run conode again
+    pkill conode
+    rm -f "$COLOG"1.log.dead
+    runCoBG 1
+    sleep .1
+    testOK ls "$CONODE_SERVICE_PATH/$shaPK.db"
```
ineiti: `testOK [ -f "$CONODE_SERVICE_PATH/$shaPK.db" ]`

PR #1174, opened 2018-04-11
evoting: use skipchain service for creating new skipchains and appending blocks

gnarula:
Changes the evoting service to use the skipchain service to create new skipchains and append new blocks.

Blocks appended during the execution of a protocol (Shuffle, Decrypt) still make use of the websocket API since they need to communicate with the leader.

Also updated setStage method to use the skipchain service to retrieve the latest block. This should be considerably faster than iterating over all blocks one at a time.

### DIFF
--- a/evoting/lib/election.go
+++ b/evoting/lib/election.go
@@ -1,6 +1,7 @@
 package lib
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/dedis/kyber"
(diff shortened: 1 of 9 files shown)


evoting/lib/election.go
```diff
@@ -75,30 +77,28 @@ func GetElection(roster *onet.Roster, id skipchain.SkipBlockID) (*Election, erro
...
 	if err != nil {
-		return err
+		return errors.New("error getting latest skipblock")
 	}
+	transaction := UnmarshalTransaction(latest.Data)
 
-	if len(mixes) == 0 {
-		e.Stage = Running
-	} else if len(partials) == len(e.Roster.List) {
+	// TODO: check for length and check if chain is corrupt
```
ineiti: We can suppose that the skipchain service is not corrupt. Before the skipchain service stores a new block in its database, it checks that it is a valid block.

evoting/service/service.go
```diff
@@ -587,13 +590,18 @@ func (s *Service) load() error {
 	return nil
 }
 
+func (s *Service) db() *skipchain.SkipBlockDB {
```
jeffallen: This seems to not be used?

gnarula: Aha, looks like I copied this in haste from ocs and forgot to remove it

PR #1179, opened 2018-04-13
adding better roster handling

ineiti:
This adds the following four commands to `cisc skipchain roster`:
- show - prints the current roster
- add - adds a node (either ip:port or public.toml) to the roster
- remove - removes a node from the roster
- set - sets the roster

Merge this first into #1182, then merge the result into #1181 and finally merge #1181 into master.

### DIFF
--- a/cisc/cisc.go
+++ b/cisc/cisc.go
@@ -17,9 +17,11 @@ import (
 	"path"
 	"strings"
 
+	"github.com/BurntSushi/toml"
 	"github.com/dedis/cothority"
 	"github.com/dedis/cothority/identity"
 	"github.com/dedis/cothority/pop/service"
+	status "github.com/dedis/cothority/status/service"
 	"github.com/dedis/kyber"
 	"github.com/dedis/kyber/sign/schnorr"
 	"github.com/dedis/kyber/util/encoding"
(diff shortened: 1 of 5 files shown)


cisc/commands.go
```diff
@@ -107,11 +107,39 @@ func getCommands() cli.Commands {
...
+						cli.Command{
+							Name:      "show",
+							Aliases:   []string{"s"},
+							Usage:     "shows the current roster of the skipchain",
+							ArgsUsage: "[skipchain-id]",
+							Action:    scRosterShow,
+						},
+						cli.Command{
+							Name:      "set",
+							Aliases:   []string{"a"},
```
kc1212: conflicts with "add" below

cisc/cisc.go
```diff
@@ -387,6 +410,79 @@ func scRoster(c *cli.Context) error {
...
+	if id.Proposed == nil {
+		log.Info("New roster has been accepted")
+	}
+	return nil
+}
+
+func getServerIdentity(fileIP string) *network.ServerIdentity {
+	if _, err := os.Stat(fileIP); err == nil {
+		// Suppose it's a roster file
+		g, err := getGroupString(fileIP)
```
kc1212: There is a possibility that the file exists but has a name in the format of "ip:port". There may be some confusion because the user might expect that the "ip:port" to be used directly rather than what is in the file. I'd prefer to use explicit flags for these kind of options, e.g. `--toml` and `--addr`.

ineiti: Muggle alarm! Magicians like magic...

PR #1181, opened 2018-04-13
don't return network.Message

ineiti:
Returning `network.Message` in services is evil - it should return the correct message.

This can only be merged into master once #1182 is merged into here.

### DIFF
--- a/cisc/test.sh
+++ b/cisc/test.sh
@@ -51,7 +51,28 @@ main(){
   test Follow
   test SymLink
   test Revoke
+  test RosterEdit
   stopTest
+}
+
+testRosterEdit(){
+  runCoBG `seq 3`
+  runAddToken 3
+  runDbgCl 0 1 skipchain create -name client1 co1/public.toml
+  runGrepSed ID "s/.* //" runDbgCl 2 1 data list
+  ID=$SED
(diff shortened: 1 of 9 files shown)


ineiti:
True - got me a second time. What does @jeffallen say?

My take on this is that as long as nobody external is using it, it's OK to change. This is a real corner case, as it might break compilation of code depending on it, but it doesn't change the behaviour of the method. And it's easy to fix.

Anyway, let's listen to the wisdom of the elder.

jeffallen:
Paying the costs of strict backwards compatibility is only necessary when there is an audience who needs/wants it. It seems to me like playing fast and loose with these for the moment, especially when the previous design we are breaking was wrong, is ok.

cisc/test.sh
```diff
@@ -29,31 +29,52 @@ main(){
...
+  # test KeyAdd
+  # test KeyFile
+  # test KeyAdd2
+  # test KeyAddWeb
+  # test KeyDel
+  # test SSHAdd
+  # test SSHDel
+  # test Follow
+  # test SymLink
+  # test Revoke
```
kc1212: missed this - these need to be enabled.

PR #1184, opened 2018-04-16
return partial final statement upon request

ineiti:
For the CPMAC 2.0, we need to be able to fetch a partial final statement, even before the other nodes collectively signed it. For non-API breaking changes, introduced an optional field in the request.


kc1212:
What is CPMAC 2.0? this one? https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/CpMac.1.html

PR #1187, opened 2018-04-17
Make it possible to update the Master record for a vote

jeffallen:
In order to evolve the list of admins, roster, etc the
evoting service needs to allow updating the master record,
and should always use the latest for new elections. Adapt
the admin app to make these edits.


evoting/lib/transaction.go
```diff
@@ -99,8 +99,40 @@ func (t *Transaction) Digest(s *skipchain.Service, genesis skipchain.SkipBlockID
 // Verify checks that the corresponding transaction is valid before storing it.
 func (t *Transaction) Verify(genesis skipchain.SkipBlockID, s *skipchain.Service) error {
 	digest := t.Digest(s, genesis)
-	db := s.GetDB()
 	if t.Master != nil {
+		// Find the current master in order to compare against it.
+		m, err := GetMaster(s, genesis)
```
kc1212: The `s *skipchain.Service` is the skipchain that contains the master blocks, right? But the transaction is, to my understanding, something on the election skipchain. Don't you want to get, not the latest master, but the master that is just before the Election/transaction block?

evoting/app/app.go
```diff
@@ -21,39 +23,80 @@ var (
...
+		if *argSig == "" {
+			log.Fatal("-sig required when updating")
+		}
+		sig, err := hex.DecodeString(*argSig)
+		if err != nil {
+			log.Fatal("sig decode", err)
+		}
+		var sbid skipchain.SkipBlockID = id
+		request.ID = &sbid
+		var u uint32 = uint32(*argUser)
```
kc1212: travis is complaining about this

evoting/lib/transaction.go
jeffallen: (We figured out this misunderstanding in person.)

evoting/lib/master.go
```diff
@@ -34,18 +34,29 @@ type Link struct {
 
 // GetMaster retrieves the master object from its skipchain.
 func GetMaster(s *skipchain.Service, id skipchain.SkipBlockID) (*Master, error) {
+	// Search backwards from the end of the chain, unmarshalling each block until
+	// we find the first Master transaction. (There are Link transactions mixed in
+	// with the Masters.)
+
 	block, err := s.GetSingleBlockByIndex(
-		&skipchain.GetSingleBlockByIndex{Genesis: id, Index: 1},
+		&skipchain.GetSingleBlockByIndex{Genesis: id, Index: -1},
```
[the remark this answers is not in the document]
ineiti: I don't understand why you need that. We have `GetUpdateChain` that returns the shortest path to the latest block. What you're introducing here doesn't allow for securely getting the latest block. Or did I miss something?

PR #1190, opened 2018-04-18
evoting: allow checking if user has already voted

gnarula:
This allows the frontend to query if a user has already cast a ballot for an election. Since the check gets expensive as the skipchain grows, we try to cache the last cast ballot per election after every vote in the browser's localStorage, hence the change in `CastReply`. We iterate the skipchain to search for the last cast ballot only when the user logs in from a new browser session.


evoting/service/service.go
```diff
@@ -298,7 +299,7 @@ func (s *Service) GetElections(req *evoting.GetElections) (*evoting.GetElections
 
 // GetBox message handler to retrieve the casted ballot in an election.
 func (s *Service) GetBox(req *evoting.GetBox) (*evoting.GetBoxReply, error) {
-	election, err := lib.GetElection(s.skipchain, req.ID)
+	election, err := lib.GetElection(s.skipchain, req.ID, false, 0)
```
jeffallen: A different way to have done this would have been to introduce a new API called lib.GetElectionCheckVoted. That way, GetElection calls would not end up with this useless 0 argument in them, which is a bit arbitrary/ugly/confusing.

I'm not recommending you change this now, just pointing out that it's not always bad to expand the public API: if the new function prevents callers of the old function from getting uglified, it's a nice tradeoff.

PR #1197, opened 2018-04-21
Javalib

### LINKED ISSUE #892: Java library
There is an implementation in the private ocs-repo of a java-implementation to interact with kyber.v1, but the `Point` and `Scalar` is not generic enough to support different groups, and has some of the arguments changed with regard to the kyber-library.

Clean up java-library with regard to `Point` and `Scalar`.

aounleonardo:
Closes #892 

Change name of classes Point and Scalar to PointEd25519 and ScalarEd25519 and make them implement interfaces Point and Scalar.
Made the necessary changes elsewhere in the code, respecting the "Program to an interface" philosophy. 

~~All tests are passing except the ones in: `ch.epfl.dedis.byzgen`, probably because of containerization.~~

Please merge this into https://github.com/dedis/cothority/pull/1228 first, and then merge it into master.

### DIFF
--- a/external/java/src/main/java/ch/epfl/dedis/lib/crypto/Ed25519Point.java
+++ b/external/java/src/main/java/ch/epfl/dedis/lib/crypto/Ed25519Point.java
@@ -1,0 +1,128 @@
+package ch.epfl.dedis.lib.crypto;
+
+import ch.epfl.dedis.lib.exception.CothorityCryptoException;
+import com.google.protobuf.ByteString;
+import net.i2p.crypto.eddsa.EdDSAPublicKey;
+import net.i2p.crypto.eddsa.math.GroupElement;
+import net.i2p.crypto.eddsa.spec.EdDSAPublicKeySpec;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.xml.bind.DatatypeConverter;
+import java.security.PublicKey;
+import java.util.Arrays;
+
+public class Ed25519Point implements Point {
+    private final static Logger logger = LoggerFactory.getLogger(Ed25519Point.class);
+
+    public GroupElement element;
+
+    public Ed25519Point(ByteString pub) {
+        this(pub.toByteArray());
+    }
+
+    public Ed25519Point(PublicKey pub) {
+        this(Arrays.copyOfRange(pub.getEncoded(), 12, 44));
+    }
+
+    public Ed25519Point(String str) {
+        this(DatatypeConverter.parseHexBinary(str));
+    }
+
+    public Ed25519Point(byte[] b) {
+        element = new GroupElement(Ed25519.curve, b);
+    }
+
+    public Ed25519Point(Point p) {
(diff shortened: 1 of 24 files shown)


external/java/src/main/java/ch/epfl/dedis/lib/crypto/Ed25519Point.java
```diff
@@ -0,0 +1,128 @@
...
+        int len = bytes[0];
+        if (len > Ed25519.pubLen || len < 0) {
+            logger.info(DatatypeConverter.printHexBinary(bytes));
+            throw new CothorityCryptoException("doesn't seem to be a valid point");
+        }
+        return Arrays.copyOfRange(bytes, 1, len + 1);
+    }
+
+
+    public static Point pubStore(byte[] data) throws CothorityCryptoException {
```
ineiti: Should be called `Embed` to mirror kyber's names.

external/java/src/main/java/ch/epfl/dedis/lib/crypto/Ed25519Point.java
```diff
@@ -0,0 +1,128 @@
...
+    public EdDSAPublicKey toEdDSAPub() {
+        EdDSAPublicKeySpec spec = new EdDSAPublicKeySpec(element, Ed25519.ed25519);
+        return new EdDSAPublicKey(spec);
+    }
+
+    public Point negate() {
+        return new Ed25519Point(element.toP3().negate());
+    }
+
+    public byte[] pubLoad() throws CothorityCryptoException {
```
ineiti: Should be called `Data` to mirror kyber's names.

external/java/src/main/java/ch/epfl/dedis/lib/crypto/Ed25519Scalar.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public Scalar negate() {
+        return convert(Ed25519.prime_order).sub(this.reduce()).reduce();
+    }
+
+    public boolean isZero() {
+        return !convert(reduce()).fieldElement.isNonZero();
+    }
+
+    public Point scalarMult(Point p) {
```
ineiti: Should be defined on `Point` to mirror kyber's definitions.

external/java/src/main/java/ch/epfl/dedis/lib/crypto/Point.java
```diff
@@ -2,118 +2,20 @@
...
-                }
-                return e;
-            } catch (IllegalArgumentException e) {
-                // Will fail in about 87.5%, so try again.
-            }
-        }
-        throw new CothorityCryptoException("did not find matching point!?!");
-    }
+    public Point negate();
+    public byte[] pubLoad() throws CothorityCryptoException;
```
ineiti: Rename and add both `Embed` and `Data`.

ineiti:
Added #1257 so we can move on. Please add things that we still need to do there.

PR #1199, opened 2018-04-23
evoting: use skipchain service in protocol

gnarula:
Optimisations for Mix and Partial operations used within shuffle and decrypt protocol. They now use the skipchain service to get the blocks. The iterations are also performed from the end to search for the Mix or Partial transactions since they're always at the end of the chain in case of a finished election.

### DIFF
--- a/evoting/lib/chains.go
+++ b/evoting/lib/chains.go
@@ -36,12 +36,10 @@ func StoreUsingWebsocket(id skipchain.SkipBlockID, roster *onet.Roster, transact
 	if err != nil {
 		return err
 	}
-
 	enc, err := protobuf.Encode(transaction)
 	if err != nil {
 		return err
 	}
-
 	_, err = client.StoreSkipBlock(reply.Update[len(reply.Update)-1], nil, enc)
 	if err != nil {
 		return err
--- a/evoting/lib/election.go
+++ b/evoting/lib/election.go
@@ -136,10 +136,12 @@ func (e *Election) setStage(s *skipchain.Service) error {
 }
 
 // Box accumulates all the ballots while only keeping the last ballot for each user.
-func (e *Election) Box() (*Box, error) {
-	client := skipchain.NewClient()
-
-	block, err := client.GetSingleBlockByIndex(e.Roster, e.ID, 0)
+func (e *Election) Box(s *skipchain.Service) (*Box, error) {
+	block, err := s.GetSingleBlockByIndex(
+		&skipchain.GetSingleBlockByIndex{
+			Genesis: e.ID,
+			Index:   0,
+		})
 	if err != nil {
 		return nil, err
 	}
@@ -155,7 +157,10 @@ func (e *Election) Box() (*Box, error) {
 		if len(block.ForwardLink) <= 0 {
 			break
 		}
-		block, _ = client.GetSingleBlock(e.Roster, block.ForwardLink[0].To)
+		block, _ = s.GetSingleBlock(
+			&skipchain.GetSingleBlock{
+				ID: block.ForwardLink[0].To,
+			})
 	}
 
 	// Reverse ballot list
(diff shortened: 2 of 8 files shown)


evoting/lib/election.go
```diff
@@ -136,10 +136,14 @@ func (e *Election) setStage(s *skipchain.Service) error {
 }
 
 // Box accumulates all the ballots while only keeping the last ballot for each user.
-func (e *Election) Box() (*Box, error) {
-	client := skipchain.NewClient()
+func (e *Election) Box(s *skipchain.Service) (*Box, error) {
+	// client := skipchain.NewClient()
```
ineiti: remove unused code

evoting/lib/election.go
```diff
@@ -181,54 +188,77 @@ func (e *Election) Box() (*Box, error) {
 }
 
 // Mixes returns all mixes created by the roster conodes.
-func (e *Election) Mixes() ([]*Mix, error) {
-	client := skipchain.NewClient()
+func (e *Election) Mixes(s *skipchain.Service) ([]*Mix, error) {
 
-	block, err := client.GetSingleBlockByIndex(e.Roster, e.ID, 0)
+	// Traversing forward one by one might be expensive in a large
+	// election. Better would be to search for the Mix transactions
```
ineiti: s/would be/is/

evoting/lib/election.go
```diff
@@ -181,54 +188,77 @@ func (e *Election) Box() (*Box, error) {
...
 }
 
 // Partials returns the partial decryption for each roster conode.
-func (e *Election) Partials() ([]*Partial, error) {
-	client := skipchain.NewClient()
+func (e *Election) Partials(s *skipchain.Service) ([]*Partial, error) {
 
-	block, err := client.GetSingleBlockByIndex(e.Roster, e.ID, 0)
+	// Traversing forward one by one might be expensive in a large
+	// election. Better would be to search for the Partial transactions
```
ineiti: s/Better would be/It is better/

evoting/lib/chains.go
```diff
@@ -32,17 +32,15 @@ func NewSkipchain(s *skipchain.Service, roster *onet.Roster, verifier []skipchai
 // using websockets. Used for storing a block while executing a protocol.
 func StoreUsingWebsocket(id skipchain.SkipBlockID, roster *onet.Roster, transaction *Transaction) error {
 	client := skipchain.NewClient()
-	reply, err := client.GetUpdateChain(roster, id)
+	latest, err := client.GetSingleBlockByIndex(roster, id, -1)
```
ineiti: I think we should not use the `-1` and remove it. It opens up too big a hole for the server to give wrong blocks. `GetUpdateChain` is much more secure, as it checks the proof of the blocks.

gnarula:
Addressed the comments in the new commit

evoting/lib/election.go
```diff
@@ -181,54 +186,77 @@ func (e *Election) Box() (*Box, error) {
...
+
+		if transaction.Mix != nil {
+			// append to the mixes array
 			mixes = append(mixes, transaction.Mix)
 		}
 
-		if len(block.ForwardLink) <= 0 {
-			break
+		// keep iterating back
+		block, err = s.GetSingleBlock(&skipchain.GetSingleBlock{ID: block.BackLinkIDs[0]})
```
ineiti: Why don't you use `s.GetDB().GetByID` here?

evoting/lib/election.go
```diff
@@ -181,54 +186,77 @@ func (e *Election) Box() (*Box, error) {
...
-		if len(block.ForwardLink) <= 0 {
-			break
+		if transaction.Partial != nil {
+			partials = append(partials, transaction.Partial)
 		}
-		var err error
-		block, err = client.GetSingleBlock(e.Roster, block.ForwardLink[0].To)
+
+		// keep iterating back
+		block, err = s.GetSingleBlock(&skipchain.GetSingleBlock{ID: block.BackLinkIDs[0]})
```
ineiti: Why not `s.GetDB().GetByID`?

ineiti:
Two more comments, but then it's OK.

evoting/lib/election.go
gnarula: Fixed

evoting/lib/election.go
gnarula: Fixed

PR #1200, opened 2018-04-23
Conode: Show dev version and remove vgo

ineiti:
This PR removes the `vgo` building from the Makefile, as the usage of `.netrc` is not nice for users who simply want to create a docker image.

It also adds the `-dev-YYMMDD` tag to the master-branch, so that the nodes can be easily recognized on status.dedis.ch.

### DIFF
--- a/conode/Makefile
+++ b/conode/Makefile
@@ -39,6 +39,12 @@ docker_clean:
 
 TAG=$(shell git describe --tags)
 OUTPUT_DIR=conode-$(TAG)
+# dev is the master branch, which will get automatic date tagging for
+# distribution files.
+ifeq ($(TAG),dev)
+	TAG := $(TAG)-$(shell date +%y%m%d)
+endif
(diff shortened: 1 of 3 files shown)


conode/Makefile
```diff
@@ -1,24 +1,38 @@
...
+flags=-ldflags="$(ldflags)"
 
 all: docker
 
 docker: Dockerfile
-	[ -f $(HOME)/.netrc ] && cp $(HOME)/.netrc .
-	docker build -t $(IMAGE_NAME):$(VERSION) .
+	BUILDFLAG=$(flags)
+	docker build -t $(IMAGE_NAME):$(VERSION) --build-arg BUILDFLAG="$(ldflags)" .
 	rm -f .netrc
```
jeffallen: The rm .netrc can be deleted now. Also remove .netrc from .gitignore.

PR #1210, opened 2018-04-25
Add JSON output to status tool

jeffallen:
Which makes it more convenient to use from inside of a monitoring script.

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -4,7 +4,11 @@
 package main
 
 import (
+	"bufio"
+	"bytes"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


status/status.go
```diff
@@ -96,3 +130,16 @@ func printConn(e *status.Response) {
...
+
+func printJSON(all []se) {
+	b1 := new(bytes.Buffer)
+	e := json.NewEncoder(b1)
+	e.Encode(all)
+
+	b2 := new(bytes.Buffer)
+	json.Indent(b2, b1.Bytes(), "", "  ")
+
+	out := bufio.NewWriter(os.Stdout)
```
ineiti: Why don't you use `fmt.Print` here?

jeffallen: os.Stdout is non-buffered by default. If you know/think you are going to be writing a lot to it, then it is better to wrap it in a bufio first.

PR #1214, opened 2018-04-26
evoting: handle failures while finalizing

gnarula:
This PR contains modifications to the shuffle protocol to recover from failure during shuffling. The shuffle protocol is a chain protocol where each node in the protocol tries to perform a shuffle on the last `Mix` transaction (or `Box` in case of the leader). The node then propagates the message further down the chain. In case of a network error, the node tries to contact the next node in chain. If a node fails for any reason during the shuffle, the protocol would time out for the leader. From the administrator's perspective, restarting a protocol would first look at existing `Mix` transactions in the skipchain and the protocol will be re-initiated with the left over nodes.

Edit:

Updated the PR to handle decrypt resilience as well. The Decrypt protocol now sends a broadcast to all nodes and exits as soon as it gets 2n/3+1 replies. In case the protocol is resumed from a previously failed attempt, the leader first tries to figure out the list of nodes that have not participated in decryption before and then reinitiates the protocol.

### DIFF
--- a/evoting/lib/container.go
+++ b/evoting/lib/container.go
@@ -43,7 +43,9 @@ type Mix struct {
 	Ballots []*Ballot // Ballots are permuted and re-encrypted.
 	Proof   []byte    // Proof of the shuffle.
 
-	Node string // Node signifies the creator of the mix.
+	Node      string      // Node signifies the creator of the mix.
+	PublicKey kyber.Point // Public key of the node generating the mix
+	Signature []byte      // Signature of the public key
 }
 
 // Partial contains the partially decrypted ballots.
--- a/evoting/lib/container.go
+++ b/evoting/lib/container.go
@@ -52,8 +52,9 @@ type Mix struct {
 type Partial struct {
 	Points []kyber.Point // Points are the partially decrypted plaintexts.
 
-	Flag bool   // Flag signals if the mixes could not be verified.
-	Node string // Node signifies the creator of this partial decryption.
+	Node      string      // Node signifies the creator of this partial decryption.
+	PublicKey kyber.Point // Public key of the node generating a partial decryption
+	Signature []byte      // Signature of the public key
 }
 
 // genPartials generates partial decryptions for a given list of shared secrets.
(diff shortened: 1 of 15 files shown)


evoting/service/service.go
```diff
@@ -640,6 +672,8 @@ func new(context *onet.Context) (onet.Service, error) {
 		},
 		skipchain: context.Service(skipchain.ServiceName).(*skipchain.Service),
 	}
+	log.Lvl1("Setting skipchain propagation timeout to 5s")
+	service.skipchain.SetPropTimeout(5 * time.Second)
```
gnarula: I'm not sure if this requires a call to `skipchain.SetBFTTimeout` as well

jeffallen: Let's talk about this in person and see what we can figure out.

evoting/service/service_test.go
```diff
@@ -308,12 +313,296 @@ func TestEvolveRoster(t *testing.T) {
...
 
 	// There was a test here before to try to replace the leader.
 	// It didn't work. For the time being, that is not supported.
+
+	// The decrypt protocol tries to stop early as soon as 2n/3 + 1 nodes store a partial.
+	// However, since the leader sends a broadcast to all the n nodes initially we
+	// want the servers to be up until the goroutines terminate or the test framework complains
+	// about zombie goroutines. The call to time.Sleep ensures we dont end up with
+	// zombie goroutines
+	time.Sleep(5 * time.Second)
```
ineiti: I do understand that sometimes `time.Sleep` is the only way to go forward. But most of the time I managed to do something like:

```
for i := 0; i < 10; i++{
  if checkIfAllMessagesAreDone(){
    break
  }
  time.After(200*time.Millisecond):
}
require.NotEqual(t, 10, i, "some messages are still waiting")
```

Can you look whether that is possible? Also for the other occurrences of `time.Sleep` in previous tests. If it's too complicated, leave the `time.Sleep` in.

evoting/lib/container.go
```diff
@@ -43,15 +43,19 @@ type Mix struct {
 	Ballots []*Ballot // Ballots are permuted and re-encrypted.
 	Proof   []byte    // Proof of the shuffle.
 
-	Node string // Node signifies the creator of the mix.
+	Node      string      // Node signifies the creator of the mix.
+	PublicKey kyber.Point // Public key of the node generating the mix
```
ineiti: I prefer having this as the `ServerIdentity.ID`, so you can look up the public key in the roster of the nodes. Else you might get a rogue signature with a random private/public key. Also for the `Partial` structure.

ineiti:
Closing this in favour of #1224

PR #1215, opened 2018-04-26
Make it possible to make a Docker image from local files

jeffallen:
Before you could only make images from source fetched via "go get". Now you can use
"make docker_dev" to build from local source.

### DIFF
--- a/conode/Dockerfile-dev
+++ b/conode/Dockerfile-dev
@@ -1,0 +1,12 @@
+FROM debian:stretch-slim
+WORKDIR /root/
+COPY exe/conode.Linux.x86_64 ./conode
+COPY setup-then-start.sh .
+RUN mkdir /conode_data
+RUN mkdir -p .local/share .config
+RUN ln -s /conode_data .local/share/conode
+RUN ln -s /conode_data .config/conode
+
+EXPOSE 6879 6880
+
+CMD "./setup-then-start.sh"
--- a/conode/Makefile
+++ b/conode/Makefile
@@ -21,9 +21,14 @@ flags=-ldflags="$(ldflags)"
 
 all: docker
 
+# Use this target to build from only published sources.
 docker: Dockerfile
 	BUILDFLAG=$(flags)
 	docker build -t $(IMAGE_NAME):$(VERSION) --build-arg BUILDFLAG="$(ldflags)" .
+
+# Use this target to build from local source instead of from publish sources.
+docker_dev: Dockerfile-dev exe/conode.Linux.x86_64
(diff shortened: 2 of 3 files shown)


conode/Makefile
```diff
@@ -21,10 +21,15 @@ flags=-ldflags="$(ldflags)"
 
 all: docker
 
+# Use this target to build from only published sources.
 docker: Dockerfile
 	BUILDFLAG=$(flags)
 	docker build -t $(IMAGE_NAME):$(VERSION) --build-arg BUILDFLAG="$(ldflags)" .
 
+# Use this target to build from local source instead of from publish sources.
+docker_dev: Dockerfile-dev exe/conode.Linux.x86_64
```
ineiti: I'm not sure this is what you want:

https://stackoverflow.com/questions/9159960/order-of-processing-components-in-makefile

So according to this, it would first create the Dockerfile-dev, then the binary. Which sounds counter-intuitive...

conode/Makefile
```diff
@@ -52,10 +57,14 @@ docker_clean:
 
 # The suffix on conode exe is the result from: echo `uname -s`.`uname -m`
 # so that we can find the right one in the wrapper script.
-bindist:
+# This is in it's own rule because the Docker build needs it also.
+exe/conode.Linux.x86_64:
+	GOOS=linux GOARCH=amd64 go build $(flags) -o $@
```
ineiti: But now you miss the `OUTPUT_DIR`, now? Wouldn't it be simpler to have `docker_dev` depend on `bindist` only? Even if you create 4 binaries too many?

ineiti: OK, looking through the code I see the light: you use `exe/` as a directory for the conode-binary, and `conode-...` as directory for the bindist.

I think I would still keep the x86-binary in bindist and then pass `OUTPUT_DIR` to Dockerfile-dev. But if you prefer it this way, it's also possible.

jeffallen: I prefer it this way. Docker doesn't need any args, it knows where to find the thing Make just made for it.

conode/Dockerfile-dev
```diff
@@ -0,0 +1,12 @@
+FROM debian:stretch-slim
+WORKDIR /root/
+COPY exe/conode.Linux.x86_64 ./conode
+COPY setup-then-start.sh .
+RUN mkdir /conode_data
+RUN mkdir -p .local/share .config
+RUN ln -s /conode_data .local/share/conode
+RUN ln -s /conode_data .config/conode
```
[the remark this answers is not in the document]
jeffallen: It is untouched form before.

PR #1218, opened 2018-04-27
Prevent conode from starting when built in 32-bit mode.

jeffallen:


ineiti:
We could even put this in `Server.Start` or so, as it will always fail sooner or later.

PR #1221, opened 2018-04-30
adding payload to skipblock

### LINKED ISSUE #1220: skipchain: Add payload data
Now: all the data is in `SkipBlockFix`

Wanted: separate like in other blockchains the header and the rest of the data

Backward-compatible way: add an optional protobuf field to the `SkipBlock` struct that can hold data that is stored on the blockchain but ignored for the hash

ineiti:
Follow Bitcoin and Ethereum by having a payload that is not directly hashed into the header (`SkipBlockFix`) part. The application is responsible for hashing the payload into the header.

Closes #1220


skipchain/struct.go
```diff
@@ -343,6 +343,13 @@ type SkipBlock struct {
 	// SkipLists that depend on us, given as the first SkipBlock - can
 	// be a Data or a Roster SkipBlock
 	ChildSL []SkipBlockID
+
+	// Payload is additional data that needs to be hashed by the application
+	// itself into SkipBlockFix.Data. A normal usecase is to set
+	// SkipBlockFix.Data to the sha256 of this payload. Then the proofs
+	// using the skipblocks can return simply the SkipBlockFix, as long as they
+	// don't need the payload.
+	PayLoad []byte `protobuf:"opt"`
```
jeffallen: Payload is an english word, no need to capitalize in the middle of it, please.

PR #1224, opened 2018-05-01
Evoting shuffle resilience

ineiti:
This replaces #1214


evoting/protocol/decrypt.go
```diff
@@ -130,6 +129,7 @@ func (d *Decrypt) HandlePrompt(prompt MessagePromptDecrypt) error {
 	}
 	partial.Signature = sig
 	transaction := lib.NewTransaction(partial, d.User, d.Signature)
+	time.Sleep(time.Millisecond * time.Duration(rand.Int31n(1000)))
```
jeffallen: Is this from debugging?

PR #1226, opened 2018-05-01
evoting: check mix and partial length for determining stage

gnarula:
Changes `setStage` to check against the length of partials and mixes so that it works well with the changes introduced in #1224

### DIFF
--- a/evoting/lib/election.go
+++ b/evoting/lib/election.go
@@ -118,23 +118,28 @@ func (e *Election) setVoted(s *skipchain.Service, user uint32) error {
 }
 
 func (e *Election) setStage(s *skipchain.Service) error {
-	db := s.GetDB()
-	latest, err := db.GetLatest(db.GetByID(e.ID))
+	// threshold is the minimum number of blocks we need
+	// to complete a shuffle or a decryption. Following
+	// byzantine consensus it's set to floor(2*n/3) + 1
+	threshold := 2*len(e.Roster.List)/3 + 1
+	partials, err := e.Partials(s)
 	if err != nil {
-		return errors.New("error getting latest skipblock")
+		return err
 	}
-	transaction := UnmarshalTransaction(latest.Data)
-	if transaction == nil {
-		return errors.New("invalid transaction")
+	if len(partials) >= threshold {
+		e.Stage = Decrypted
+		return nil
 	}
 
-	if transaction.Partial != nil {
-		e.Stage = Decrypted
-	} else if transaction.Mix != nil {
-		e.Stage = Shuffled
-	} else {
-		e.Stage = Running
+	mixes, err := e.Mixes(s)
+	if err != nil {
+		return err
 	}
+	if len(mixes) >= threshold {
+		e.Stage = Shuffled
+		return nil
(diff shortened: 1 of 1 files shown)


evoting/lib/election.go
```diff
@@ -118,23 +118,25 @@ func (e *Election) setVoted(s *skipchain.Service, user uint32) error {
...
+	target := 2 * len(e.Roster.List) / 3
+	partials, err := e.Partials(s)
 	if err != nil {
-		return errors.New("error getting latest skipblock")
+		return err
 	}
-	transaction := UnmarshalTransaction(latest.Data)
-	if transaction == nil {
-		return errors.New("invalid transaction")
+	if len(partials) > target {
```
ineiti: Why is that not `>=`?

evoting/lib/election.go
```diff
@@ -118,23 +118,25 @@ func (e *Election) setVoted(s *skipchain.Service, user uint32) error {
...
 	}
 
-	if transaction.Partial != nil {
-		e.Stage = Decrypted
-	} else if transaction.Mix != nil {
+	mixes, err := e.Mixes(s)
+	if err != nil {
+		return err
+	}
+	if len(mixes) > target {
```
ineiti: Same question as above.

evoting/lib/election.go
gnarula: because the minimum required count is floor(2n/3) + 1? Considering 7 conodes, `2*7/3 == 4` and we need at least 5

evoting/lib/election.go
```diff
@@ -118,23 +118,25 @@ func (e *Election) setVoted(s *skipchain.Service, user uint32) error {
 }
 
 func (e *Election) setStage(s *skipchain.Service) error {
-	db := s.GetDB()
-	latest, err := db.GetLatest(db.GetByID(e.ID))
+	target := 2 * len(e.Roster.List) / 3
```
kc1212: shouldn't this be configured during initialisation?

gnarula: sorry, didn't quite get you there. Did you mean during election initialization as a field in the `Election` struct?

kc1212: yeah, something like that. but this is just a suggestion. if you need to get these changes in soon, then please go ahead.

evoting/lib/election.go
ineiti: OK - so you need either a comment here to explain what you wrote, or initialize `target` with `2*n/3+1`. Also, we usually call it `threshold`.
Sorry for nitpicking, but I'd like this code to be understandable also later on!

evoting/lib/election.go
ineiti: Yes, sounds reasonable. At least open an issue for that (and hope we don't forget it ;)

evoting/lib/election.go
gnarula: Fixed in the new commit 👍

evoting/lib/election.go
gnarula: opened #1227

PR #1228, opened 2018-05-02
Add java tests for OCS

### LINKED ISSUE #1205: java test for ocs
The java test for OCS is not active anymore. Different problems:

- `cothority/conode/experimental.go` should include `cothority/ocs/service` 
- `cothority/conode/Dockerfile` should create `dedis/conode:dev-YYMMDD` and link it to `dedis/conode:dev` in case we're on the master-branch and have the `dev` tag set.
- `cothority/ocs/docker/Dockerfile` should be replaced by `cothority/ocs/docker/testimg/*`
- new `cothority/ocs/docker/Dockerfile` should depend on `dedis/conode` that is defined in `cothority/conode/Dockerfile`
- `mvn test` is not called in travis.yml

aounleonardo:
- We run java tests for OCS
- change the dockerfile and makefile in cothority/conode to correctly build and tag docker containers 
- please first merge https://github.com/dedis/cothority/pull/1197 into this PR 

Closes #1205


ineiti:
Just tried out this version and it gives a nice tag : http://status.dedis.ch ;)

.travis.yml
```diff
@@ -38,3 +39,12 @@ matrix:
 
       script:
         - make test
+
+    - language: java
+
+      script:
+        - (cd conode/; make docker)
```
ineiti: why do we need `docker` and `docker_dev`?

conode/Makefile
```diff
@@ -5,10 +5,13 @@ TAG=$(shell git describe --tags)
 OUTPUT_DIR=conode-$(TAG)
 # dev is the master branch, which will get automatic date tagging for
 # distribution files.
-ifeq ($(TAG),dev)
-	TAG := $(TAG)-$(shell date +%y%m%d)
+STRIPPED_TAG = $(shell echo $(TAG) | cut -d '-' -f 1)
+VERSION := $(STRIPPED_TAG)
+ifeq ($(findstring v2,$(STRIPPED_TAG)),v2)
```
ineiti: this thing in fact did not work out, as we'll need to add a `v2` tag, but that's interfering with the `v2` branch :( Any idea how to know if we're sitting somewhere on the `v2` branch?

.travis.yml
kc1212: we don't need `make docker`, but I added it there just to make sure it builds. maybe it belongs to somewhere else instead of the java test. should I moved it to the go tests?

conode/experimental.go
```diff
@@ -5,4 +5,4 @@ This imports the experimental services that will not be used in the
 stable branch.
 */
 
-import _ "github.com/dedis/cothority/evoting/service"
```
ineiti: we need to put that back in - perhaps rename the ocs-dkg protocol for the moment?

.travis.yml
ineiti: go tests look better to me.

conode/Makefile
kc1212: I think this part might be a result of a merge. I can change it back to what it was previously. Would that be better? But doing that, I will have to hard-code the docker tag for `docker_dev`.

conode/experimental.go
kc1212: ok, I'll change the protocol name and add it back in

kc1212:
@ineiti addressed comments.

PR #1230, opened 2018-05-02
minimal changes to get the timeouts correct

ineiti:
OK - the previous PR got too intrusive and never worked. Time to start fresh...

This puts the timeouts into a much nicer way, supposing that we're mostly network bound, the timeouts are spread out evenly between the leader, the subleaders and the nodes.

Also includes putting `SendToChildren` in go-routines, so that we don't wait for children accepting the connection but not replying.

Finally some logging changes for nicer and more useful outputs.

Depends on dedis/onet#417

### DIFF
--- a/ftcosi/protocol/protocol.go
+++ b/ftcosi/protocol/protocol.go
@@ -193,8 +193,6 @@ func (p *FtCosi) Dispatch() error {
 				responsesMut.Lock()
 				responses = append(responses, response)
 				responsesMut.Unlock()
-			case <-time.After(p.Timeout):
-				errChan <- fmt.Errorf("%v", i)
 			}
 		}(i, cosiSubProtocol)
 	}
@@ -240,6 +238,7 @@ func (p *FtCosi) collectCommitments(trees []*onet.Tree,
 		wg.Add(1)
 		go func(i int, subProtocol *SubFtCosi) {
 			defer wg.Done()
+			timeout := time.After(p.Timeout / 2)
 			for {
 				select {
 				case <-subProtocol.subleaderNotResponding:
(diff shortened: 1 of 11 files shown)


skipchain/protocol.go
```diff
@@ -88,17 +87,25 @@ func NewProtocolGetBlocks(n *onet.TreeNodeInstance) (onet.ProtocolInstance, erro
 // Start sends the extend roster request to all of the children.
 func (p *ExtendRoster) Start() error {
 	log.Lvl3("Starting Protocol ExtendRoster")
-	errs := p.SendToChildrenInParallel(p.ExtendRoster)
-	if len(errs) > p.allowedFailures {
-		return fmt.Errorf("Send to children failed: %v", errs)
-	}
+	go func() {
```
jeffallen: When I see this, I think of Dave Cheny, who says, "always know how your goroutine will exit before you start it". If p.SendToChildrenInParallel blocks for some reason, this goroutine will never exit. tcp.go uses read deadlines, but not write deadlines. It seems likely that an attempt to send to a peer who is accepting TCP but not talking (or has wedged and stops talking) will result in a wedged goroutine.

But that problem is not related to what you are trying to solve here, so feel free to just file an issue and kick it down the road a bit.

ineiti: Opened dedis/onet#416

ftcosi/protocol/protocol.go
```diff
@@ -347,7 +349,9 @@ func (p *FtCosi) startSubProtocol(tree *onet.Tree) (*SubFtCosi, error) {
 	cosiSubProtocol.Publics = p.publics
 	cosiSubProtocol.Msg = p.Msg
 	cosiSubProtocol.Data = p.Data
-	cosiSubProtocol.Timeout = p.Timeout / 2
+	// We allow for one subleader failure during the commit phase, and thus
```
Racater: Why the choice to allow only one subleader failure?
Moreover, if the timeout for the commit phase is t/2 and for one subleader failure t/3, the second subleader will only have t/6 to complete before commit timeout is reached. Am I correct? @ineiti

PR #1232, opened 2018-05-02
Add a cache to reduce the load on people.epfl.ch

jeffallen:
Also increases the UI performance.


ineiti:
Can you tell me why you didn't take a map here?

PR #1236, opened 2018-05-08
Fixed Jenkins

### LINKED ISSUE #1202: Make Jenkins work
Jenkins is _mostly_ working, but some of the tests, specifically the `cothority/cosi/test.sh` fail due to bad setup. Find those tests and fix them...

For `cosi/test.sh`, 1st step is to replace calls to `setupServers` with calls to `runCoBG`.

ValentinMoullet:
Changed the way of setting up the servers in `cosi`, using `runCoBG` from `conode`

With the changes done in dedis/onet#422, Jenkins seems to now be fixed.

Fixes #1202

### DIFF
--- a/conode/test.sh
+++ b/conode/test.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-DBG_TEST=2
+DBG_TEST=1
 
 . "$(go env GOPATH)/src/github.com/dedis/onet/app/libtest.sh"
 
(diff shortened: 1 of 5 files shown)


ineiti:
The error you still see in Jenkins is due to the fact that in https://github.com/dedis/cothority/blob/master/scmgr/test.sh#L218 the `pkill -n conode` does not always kill the latest conode, because they are started in the background at https://github.com/dedis/onet/blob/master/app/libtest.sh#L261 and I guess that some of the earlier conodes take longer to start than the latest conode. So it should not be `pkill -n`, but `pkill -f 'conode.*co4'` or something similar.

PR #1237, opened 2018-05-08
Update the cisc data list display

Claudio5:
When using cisc data list -p. It now shows in addition to the proposed data who and how many people voted for the new proposed data.


ineiti:
Looks good ;) One thing that you should do (supposing the dedis/cothority remote is called `upstream`):
```
git fetch upstream
git rebase upstream/origin
git push --force
```
Then there will not be all these commits, but only one commit...

Claudio5:
Rebase done :)

PR #1246, opened 2018-05-15
syncChain must also update forward links to new blocks it finds

### LINKED ISSUE #1244: syncChain does not refresh forward links
Imainge that a node has blocks with indexes: 0 -> <- 1. It is asked to make a forward link to block 3. It calls syncChain to find block 2, and integrates it into it's DB. It now has:

 0 -> <- 1  <- 2

This issue is to fix that so that once 2 is fetched, the forward link on 1 will be also updated.

jeffallen:
Fixes #1244

### DIFF
--- a/skipchain/skipchain_test.go
+++ b/skipchain/skipchain_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"strconv"
 	"sync"
 	"testing"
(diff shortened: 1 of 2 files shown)


skipchain/skipchain_test.go
```diff
@@ -1130,3 +1131,98 @@ func TestRosterAddCausesSync(t *testing.T) {
...
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// See if 4's view of block 0 has the forward link it should have to
+	// block 1.
+
+	s4 := local.Services[servers[4].ServerIdentity.ID][skipchainSID].(*Service)
+	//leader.dumpChain("leader", genId, os.Stdout)
```
ineiti: To avoid commented commands, can you replace `fmt.Fprintf(out` in `dumpChain` with `log.Lvlf3`?

jeffallen:
This is back in WIP because on second thought, we need to be careful to follow back links of any new block we get, not just the ones directly from syncChain's first call to getBlocks.

jeffallen:
@ineiti has a better fix for this in the works.

PR #1248, opened 2018-05-16
Add support for Safari 10 on CothorityJS

yvgny:
Cothorithy JS suffers from a bug on Safari 10 where the variable scoping is not correctly handled, thus preventing the module from being loaded (see here for more informations : https://github.com/mishoo/UglifyJS2/issues/1753). These flags allow Uglify JS to handle Safari 10 by correctly mangling the file according to this bug.

### DIFF
--- a/external/js/cothority/webpack.config.js
+++ b/external/js/cothority/webpack.config.js
@@ -26,7 +26,15 @@ const nodeConfig = {
     ]
   },
   externals: [nodeExternals()],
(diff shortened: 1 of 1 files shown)


ineiti:
Will this also solve the `kyber-js` problem we have on nativescript/iOS?

yvgny:
Yes in fact the problem came from one of the @dedis scoped package and I thought it was KyberJS but it was CothorityJS ;-)

PR #1253, opened 2018-05-18
Cothority-js should not bury error strings

jeffallen:
When the retry 3 times code was added, it had the side effect of
hiding the true reason for the error. Expose the last known
error message again.


external/js/cothority/lib/net/net.js
```diff
@@ -177,6 +177,7 @@ class LeaderSocket {
     // maximum 3 times and returns on the first successful attempt
     const that = this;
     const fn = co.wrap(function*() {
+      var lastErr
```
gnarula: Please prefer `let` over `var` Ref: https://www.typescriptlang.org/docs/handbook/variable-declarations.html

PR #1259, opened 2018-05-24
Defer higher-level forwardlinks

### LINKED ISSUE #1096: Skipchain forwardlinks
The forward-links in the skipchain-service are created in two steps:
1. create forward-link from current latest skipblock to new proposed one
  - runs all verification functions
  - is crucial to life of skipchain
2. asks all cothorities (rosters) from backlinked blocks to sign a new forward-link
  - only verifies forward-link from 1.
  - if it fails it only means that there is a slow-down when wanting to catch up

In the current implementation, in `StoreSkipBlock`, both 1. and 2. are done on the same node, which is not possible if there was a roster-change in between.

Needed action: add a call from `StoreSkipBlock` to the leader of the backlinked block to create the forward-link.

ineiti:
Closes #1096 

Instead of creating higher-level forwardlinks and waiting on the creation of them, this PR defers this creation, which also allows to create these forwardlinks on a roster where the current node is not present.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -80,8 +80,8 @@ func (cl *chainLocker) lock(chain SkipBlockID) {
 		cl.chains = make(map[string]*sync.Mutex)
 	}
 
+	cl.locks++
 	if l, ok := cl.chains[string(chain)]; ok {
-		cl.locks++
 		cl.Unlock()
 		l.Lock()
 		return
@@ -207,7 +207,7 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, er
 	} else {
 
 		// We're appending a block to an existing chain.
-		log.Lvlf2("Adding block with roster %+v to %x",
+		log.Lvlf2("Preparing to add block with roster %+v to %x",
 			psbd.NewBlock.Roster.List, psbd.TargetSkipChainID)
 
 		// Check if we really want to take this block with regard to our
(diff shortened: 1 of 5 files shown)


skipchain/skipchain.go
```diff
@@ -343,16 +343,12 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, er
...
-					TargetHeight: i + 1,
-					Previous:     back.Hash,
-					Newest:       prop,
-				}); err != nil {
-				// This is not a critical failure - we have at least
-				// one forward-link
-				log.Lvl1("Couldn't get old block to sign: " + err.Error())
-			}
+			// Requesting creation of secondary forward link.
+			s.SendRaw(back.Roster.List[0], &ForwardSignature{
```
jeffallen: So this is a "fire and forget" operation, right? No feedback if it fails? Do we have a test that shows what happens if the leader is down and misses adding a forward signature?

skipchain/skipchain_test.go
```diff
@@ -1104,29 +1109,25 @@ func TestRosterAddCausesSync(t *testing.T) {
 	log.Lvl1("Add last server into roster")
 	newBlock := &SkipBlock{
 		SkipBlockFix: &SkipBlockFix{
-			MaximumHeight: 2,
```
jeffallen: Removing maxheight/baseheight here is just cleanup, right? Because they are set for "real" when creating the genesis block above?

When I was creating this test, I found that it was really fragile with respect to the config (max X base). We should wrap this in a loop that tries several different max/base configs.

skipchain/struct.go
```diff
@@ -887,7 +887,11 @@ func (db *SkipBlockDB) getFromTx(tx *bolt.Tx, sbID SkipBlockID) (*SkipBlock, err
 		return nil, nil
 	}
 
-	_, sbMsg, err := network.Unmarshal(val, cothority.Suite)
+	// For some reason boltdb changes the val before Unmarshal finishes. When
```
jeffallen: This is really important, and I just spent some time on it yesterday in OmniLedger as well. I wonder if we should put a wrapper on tx.Bucket.Get so that by default we always get a new copy. Gross, wasteful, but safe.

skipchain/skipchain.go
ineiti: It's "fire and forget", because the secondary forward-links are only needed to traverse faster.
Now that I think of it, I need to add something to correctly add forward-links if previous ones failed :( R4M -> WIP...

skipchain/skipchain_test.go
ineiti: There are some tests that create blocks with different base/max-heights, but not with failing nodes. I will have to do that because of the comment I made above with forward-links where lower-level forward-links are missing.

skipchain/struct.go
ineiti: A wrapper sounds good to me. It is a memory copy, and we do much worse (more time consuming) things with it after that, so I think copying everything from `Bucket.Get` should not slow us down a lot.

skipchain/skipchain_test.go
ineiti: Added a test for a failing leader when creating secondary forward-links. Now we can have empty forward-links - check out with `ForwardLink.IsEmpty()`. If they were `nil`, protobuf removed them from the slice :(

skipchain/struct.go
ineiti: Opened #1269

PR #1267, opened 2018-06-01
GetAllSkipchains fixes

jeffallen:
GetAllSkipchains now correctly returns only the latest block of each skipchain instead of all blocks.

It is also available to apps like scmgr that need to run it on a local database.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -536,19 +536,35 @@ func (s *Service) GetSingleBlockByIndex(id *GetSingleBlockByIndex) (*SkipBlock, 
 	return nil, errors.New("No block with this index found")
 }
 
-// GetAllSkipchains returns a list of all known skipchains
+// GetAllSkipchains returns all known last blocks of skipchains.
 func (s *Service) GetAllSkipchains(id *GetAllSkipchains) (*GetAllSkipchainsReply, error) {
-	// Write all known skipblocks to a map, thus removing double blocks.
 	chains, err := s.db.getAll()
 	if err != nil {
 		return nil, err
 	}
 
-	reply := &GetAllSkipchainsReply{
-		SkipChains: make([]*SkipBlock, 0, len(chains)),
-	}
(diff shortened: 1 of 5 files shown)


skipchain/skipchain_test.go
```diff
@@ -703,11 +703,11 @@ func TestService_AddFollow(t *testing.T) {
 	sig, err = schnorr.Sign(cothority.Suite, priv1, ssb.NewBlock.CalculateHash())
 	log.ErrFatal(err)
 	ssb.Signature = &sig
-	sbs, err := service.db.getAll()
-	log.ErrFatal(err)
-	for _, sb := range sbs {
-		services[1].db.Store(sb)
-	}
+	// sbs, err := service.db.getAll()
```
ineiti: delete or put in again?

skipchain/skipchain.go
```diff
@@ -536,19 +536,18 @@ func (s *Service) GetSingleBlockByIndex(id *GetSingleBlockByIndex) (*SkipBlock,
 	return nil, errors.New("No block with this index found")
 }
 
-// GetAllSkipchains returns a list of all known skipchains
-func (s *Service) GetAllSkipchains(id *GetAllSkipchains) (*GetAllSkipchainsReply, error) {
-	// Write all known skipblocks to a map, thus removing double blocks.
-	chains, err := s.db.getAll()
+// GetAllSkipchains returns the latest block of all known skipchains.
+func (s *Service) GetAllSkipchains(_ *GetAllSkipchains) (*GetAllSkipchainsReply, error) {
```
ineiti: Please chose one of:
1. keep the old method and add a new one (like I did with `AddForwardLink` in #1267)
2. break it completely by changing the signature, so we know where it breaks and who else is using it

I have a strong preference for 1. and changing it in September for the upcoming `onet.v3`

jeffallen:
We will fix this in a backwards compatible way. Closing for now.

PR #1272, opened 2018-06-06
Add Omniledger and Event Log to Cothority

jeffallen:
These are the commits made to the student_18_omniledger repository, followed by additional commits to touch things up and fix the build.

### DIFF
--- a/eventlog/README.md
+++ b/eventlog/README.md
@@ -14,7 +14,7 @@ auditability. Below are some of the main features that sets us apart.
 - Configurable acceptance criteria - we execute a smart-contract on all nodes,
   nodes only accept the event if the smart-contract returns a positive result.
 - Existance proof - once an event is logged, an authorised client can request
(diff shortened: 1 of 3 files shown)


kc1212:
oh dear, I guess rewriting history removes the signed commits, oh well

ineiti:
There is a patch that can sign using the public keys (sorry, insider-joke ;)

ineiti:
Anyway - this looks good - not too many outside code changes. The only surprise is that the `GetAllSkipchains` change is in here and didn't pass through master. But that's a negligible epsilon and should be ignored.

The current https://pulsar.dedis.ch timestamp is 2467. We can take the value at 2500, and if it starts with
- [0-4] - @jeffallen merges
- [5-9] - @kc1212 merges
- [a-e] - @ineiti merges
- [f] take next value

Sorry for another gamification...

PR #1282, opened 2018-06-08
Event log searching

jeffallen:
Implement a new way of managing time buckets where the start times are monotonically increasing by at least 5 seconds for every new time bucket and we remove the limit on the number of events in every time bucket. This change was made to enable the searching/filtering functionality (also in this PR) so that when we iterate over time buckets to find a list of events between a time period, we know when to stop. It wouldn't have been possible in the old code because an earlier time bucket may have a later start time.


jeffallen:
This is still work in progress, because I need to do multi-bucket tests still. But i'm looking for quick feedback in case I've totally screwed up the API.

eventlog/service.go
```diff
@@ -119,6 +119,83 @@ func (s *Service) GetEvent(req *GetEventRequest) (*GetEventResponse, error) {
 	}, nil
 }
 
+var days = 24 * time.Hour
+var years = 365 * days
```
kc1212: doesn't seem to be used

eventlog/service.go
```diff
@@ -119,6 +119,83 @@ func (s *Service) GetEvent(req *GetEventRequest) (*GetEventResponse, error) {
...
+		bend = b.Start
+		b, err = getBucketByID(v, b.Prev)
+		if err != nil {
+			// We might have found some buckets to search, so exit the loop.
+			break
+		}
+	}
+
+	reply := &SearchResponse{
+		Truncated: false,
```
kc1212: is this used?

eventlog/service.go
```diff
@@ -119,6 +119,83 @@ func (s *Service) GetEvent(req *GetEventRequest) (*GetEventResponse, error) {
...
+	}
+	if len(buckets) == 0 {
+		return reply, nil
+	}
+
+	for _, b := range buckets {
+		for _, e := range b.EventRefs {
+			ev, err := getEventByID(v, e)
+			if err != nil {
+				return nil, err
```
kc1212: If we allow failures in `getBucketByID`, we should also allow failures in `getBucketByID`, i.e., we try to return as many events as we can and let the client know that some results might be missing via the `Truncated` flag. Alternatively, we immediately fail if either of these functions returns an error, because that would indicate the time buckets are not created correctly, which is a critical error to me. I'm leaning towards the latter.

eventlog/api_test.go
```diff
@@ -143,3 +143,44 @@ func checkProof(t *testing.T, omni *omniledger.Service, key []byte, scID skipcha
...
+	leader.waitForBlock(c.ID)
+	leader.waitForBlock(c.ID)
+
+	req = &SearchRequest{ID: c.ID}
+	resp, err = c.Search(req)
+	require.Nil(t, err)
+	require.NotNil(t, resp)
+	require.Equal(t, 10, len(resp.Events))
+
+	req = &SearchRequest{ID: c.ID, From: tm0 + 3, To: tm0 + 8}
```
ineiti: Is this `From <= timestamp < To`? That might be worth a documentation in the `Search` api...

eventlog/messages.go
```diff
@@ -44,23 +43,39 @@ type InitResponse struct {
...
+// LogResponse is the reply to LogRequest.
+type LogResponse struct {
+}
+
+// SearchRequest includes all the search parameters (AND of all provided search parameters).
+type SearchRequest struct {
+	ID    skipchain.SkipBlockID
+	Topic string // Return events where Event.Topic == Topic.
+	From  int64  // Return events where When is > From.
+	To    int64  // Retursn events where When is <= To.
```
ineiti: `s/Retursn/Returns/`

I would copy/paste the same documentation to the client-API, as there is the most probably place a user will look.

ineiti:
Also has some undocumented public methods.

eventlog/service.go
jeffallen: Well, I just thought it would be useful to know. Like in case I forgot how many days in a year later. But now it's in git so I guess I can nuke it. :)

eventlog/service.go
jeffallen: It is supposed to be, thanks for the reminder.

eventlog/api_test.go
jeffallen: It is documented in the SearchRequest type.

eventlog/el/main_test.go
```diff
@@ -20,7 +20,7 @@ func TestMain(m *testing.M) {
 	log.MainTest(m)
 }
 
-const blockInterval = 100 * time.Millisecond
+const blockInterval = 1 * time.Second
```
ineiti: I started to use a lot of these:

https://github.com/dedis/cothority/pull/1292/files#diff-61a6ae16d2daf74433eb76df236b74e0R555

To verify when the new instance is created. But I think we need to do the same with the `ClientTransaction` and make a way to get the result of a `ClientTransaction` from the node.

jeffallen: Are you asking for a change here... I'm  not sure I see what you are proposing.

ineiti: I thought you were waiting for all the transactions to be included and pointed to how it should be done - polling each `interval` to see if the proof `match()`es.

eventlog/service.go
```diff
@@ -119,28 +111,112 @@ func (s *Service) GetEvent(req *GetEventRequest) (*GetEventResponse, error) {
...
+	if err == errIndexMissing {
+		// There are no events yet on this chain, so return no results.
+		return &SearchResponse{}, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	// bEnd is normally updated from the last bucket's start. For the latest
+	// bucket, bEnd is now.
+	bEnd := time.Now().UnixNano()
```
kc1212: if I understood correctly, this might not be the case if you submit an even that has a timestamp after `time.Now()` and then do a search request immediately after the event is recorded.

eventlog/service.go
```diff
@@ -119,28 +111,112 @@ func (s *Service) GetEvent(req *GetEventRequest) (*GetEventResponse, error) {
...
+		b := buckets[i]
+		for _, e := range b.EventRefs {
+			ev, err := getEventByID(v, e)
+			if err != nil {
+				log.Errorf("bucket %x points to event %x, but the event was not found: %v", bids[i], e, err)
+				return nil, err
+			}
+
+			if req.From <= ev.When && ev.When < req.To {
+				if req.Topic == "" || req.Topic == ev.Topic {
```
kc1212: future work: a useful improvement here would be to allow expressions, something like `req.Topic = "art & music"`

eventlog/service.go
jeffallen: I see what you are saying, but it seems too uncommon, let's keep this as simple as possible until it proves to be a problem.

kc1212:
moving to R4M because @jeffallen says it is ready

PR #1285, opened 2018-06-10
Make evolve and sign action names configurable

kc1212:
We make this change because the service using darcs may have its own
naming scheme for these actions. The implementation simply adds two
additional fields for these names, which must be set on initialisation.

### DIFF
--- a/eventlog/api.go
+++ b/eventlog/api.go
@@ -4,7 +4,6 @@ import (
 	"time"
 
 	"github.com/dedis/cothority/omniledger/darc"
-	"github.com/dedis/cothority/omniledger/darc/expression"
 	omniledger "github.com/dedis/cothority/omniledger/service"
 	"github.com/dedis/protobuf"
 
@@ -34,23 +33,14 @@ func NewClient(r *onet.Roster) *Client {
 	}
 }
 
-// AddWriter modifies the given darc.Rules to use expr as the authorized writer
-// to add new Event Logs. If expr is nil, the current evolution expression is used instead.
-func AddWriter(r darc.Rules, expr expression.Expr) darc.Rules {
-	if expr == nil {
-		expr = r.GetEvolutionExpr()
-	}
-	r["Spawn_eventlog"] = expr
-	return r
-}
-
 // Init initialises an event logging skipchain. A sucessful call
 // updates the ID, Signer and Darc fields of the Client. The new
 // skipchain has a Darc that requires one signature from owner.
 func (c *Client) Init(owner *darc.Signer, blockInterval time.Duration) error {
-	rules := darc.InitRules([]*darc.Identity{owner.Identity()}, []*darc.Identity{})
-	d := darc.NewDarc(AddWriter(rules, nil), []byte("eventlog owner"))
-
+	d := darc.NewDarc([]*darc.Identity{owner.Identity()}, []*darc.Identity{}, "_evolve", "_sign", []byte("eventlog owner"))
+	if err := d.AddRule("Spawn_eventlog", d.GetEvolutionExpr()); err != nil {
+		return err
+	}
 	msg := &InitRequest{
 		Owner:         *d,
 		Roster:        *c.roster,
--- a/external/proto/darc.proto
+++ b/external/proto/darc.proto
@@ -7,32 +7,38 @@ option java_outer_classname = "DarcProto";
 // These are the messages used in the API-calls
 // ***
 
-// Darc is the basic structure representing an access control. A Darc can evolve in the way that
-// a new Darc points to the previous one and is signed by the owner(s) of the previous Darc.
+// Darc is the basic structure representing an access control. A Darc can
+// evolve in the way that a new Darc points to the previous one and is signed
+// by the owner(s) of the previous Darc.
 message Darc {
   // Version should be monotonically increasing over the evolution of a Darc.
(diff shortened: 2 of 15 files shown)


omniledger/darc/identitiy.go
```diff
@@ -0,0 +1,247 @@
+package darc
```
jeffallen: `git mv identitiy.go identity.go`

omniledger/darc/struct.go
```diff
@@ -43,7 +43,16 @@ type Darc struct {
 	BaseID ID
 	// PrevID is the previous darc ID in the chain of evolution.
 	PrevID ID
-	// Rules map an action to an expression.
+	// EvolveName is the name for the evolution action, which must
+	// exist and is responsible for managing the rules for evolving a darc.
+	// This field should not be changed after initialisation.
```
jeffallen: "should not change after init" -> Then why is it public?

omniledger/darc/request.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+	}
+	d.Signatures = darcSigs
+
+	return d, nil
+}
+
+// InitRequest initialises a request, the caller must provide all the fields of
+// the request. There is no guarantee that this request is valid, please see
+// InitAndSignRequest is a valid request needs to be created.
```
jeffallen: s/is/if/

jeffallen: Maybe these two should be called "InitRequestWithSignatures" and "InitRequestAndSign" in order to explain the difference better. Also a sentence in the docs explaining why you'd use one vs the other would be helpful. (I can't propose it because I can't understand it yet.)

eventlog/api.go
```diff
@@ -34,23 +33,14 @@ func NewClient(r *onet.Roster) *Client {
...
-}
-
 // Init initialises an event logging skipchain. A sucessful call
 // updates the ID, Signer and Darc fields of the Client. The new
 // skipchain has a Darc that requires one signature from owner.
 func (c *Client) Init(owner *darc.Signer, blockInterval time.Duration) error {
-	rules := darc.InitRules([]*darc.Identity{owner.Identity()}, []*darc.Identity{})
-	d := darc.NewDarc(AddWriter(rules, nil), []byte("eventlog owner"))
-
+	d := darc.NewDarc([]*darc.Identity{owner.Identity()}, []*darc.Identity{}, "_evolve", "_sign", []byte("eventlog owner"))
```
jeffallen: It seems like NewDarc is *always* (at least in this PR) called with _evolve and _sign. So should they be defaults?

omniledger/darc/struct.go
kc1212: Because they need to be serialised by protobuf

external/proto/darc.proto
```diff
@@ -7,32 +7,38 @@ option java_outer_classname = "DarcProto";
...
+// Darc is the basic structure representing an access control. A Darc can
+// evolve in the way that a new Darc points to the previous one and is signed
+// by the owner(s) of the previous Darc.
 message Darc {
   // Version should be monotonically increasing over the evolution of a Darc.
   required uint64 version = 1;
-  // Description is a free-form field that can hold any data as required by the user.
-  // Darc itself will never depend on any of the data in here.
+  // Description is a free-form field that can hold any data as required by the
+  // user.  Darc itself will never depend on any of the data in here.
```
ineiti: Detail warning: can we have a consensus on whether to use one space or two spaces after a `.`? If both of you think two spaces are OK, so be it. But if it's half/half, it drives me crazy. I vote for one space only.

external/proto/darc.proto
```diff
@@ -7,32 +7,38 @@ option java_outer_classname = "DarcProto";
...
-  // from the previous valid Darc.
-  repeated Signature signatures = 6;
-  // VerificationDarcs are a list of darcs that the verifier needs to
-  // verify this darc. It is not needed in online verification where the
-  // verifier stores all darcs.
-  repeated Darc verificationdarcs = 7;
+  // EvolveName is the name for the evolution action, which must exist and is
+  // responsible for managing the rules for evolving a darc.  This field should
+  // not be changed after initialisation.
+  required string evolvename = 5;
```
ineiti: why do we need this and `signname` again? can't we have them hardcoded? It looks really clunky in here...

omniledger/darc/request.go
ineiti: Why not `NewRequest` and `NewRequestAndSign`?

external/proto/darc.proto
kc1212: I don't have a strong opinion on it, it's just my editor default since I don't manually reflow the text. I think @jeffallen does the same. I'm happy to change to one space. On a related note, we also have different text block widths, I have 80 characters but @jeffallen has something shorter, see the beginning of `darc.go` and the commend width for the difference. We also need to have an agreement on that.

kc1212:
Moving this back to WIP to assess whether it darc contracts can be done without this PR.

kc1212:
This change is not needed to support darc contracts.

external/proto/darc.proto
jeffallen: I type one space after period. My right margin was 70, and I've just fixed it to 80. Sorry for the noise.

PR #1286, opened 2018-06-11
Add contract for darc evolution

kc1212:
This commit adds a simple contract that checks the darc evolution
request. The functionality of Instruction.Action had to be changed to
allow this feature. That is, the darc-action for Invoke does not have
the "Invoke_" prefix anymore. This is unecessary because the re-created
darc.Request in an instruction must match the darc.Request during
verification. If the action in the instruction does match the action in
the darc, then verification will fail.

For example, when client submits an instruction, the client signs the
instruction by creating a darc.Request and then signs the request. The
request will contain instruction.Action, which would be "Invoke__evolve"
in the old implementation. This instruction is sent to the omniledger
service and it does two verification, one in the service code and the
other in the contract. (1) The service finds the corresponding darc,
converts the instruction to a darc.Request, and calls
darc.Request.Verify. The verification will fail if the darc doesn't have
"Invoke__evolve" as one of its actions (a simple way to fix it is to
copy the expression from "_evolve" to "Invoke__evolve"). (2) We need to
do the verification again in the contract because the service
verification doesn't try to decode the serialised darc. Hence, the
contract is given the same instruction, decodes it, and calls
darc.Verify. For this verification to work, the darc code will check
that the evolution is valid, e.g., the version is correct, base ID is
correct and most importantly, that the darc is signed by the owners of
the previous darc on the darc.Request representation, which *must*
use "_evolve" as its action. Verification (1) works because we can add
new rules to the darc. But verification (2) will not work unless we make
sure that the action in darc.Request is re-created in the same way, from
an instruction and from a darc (in verifyOneEvolution).

Another change is the Signer and Identity types. Before we always used
pointer types, which is unecessary because we never modify these structs
after creation. Now we use value types.

### DIFF
--- a/omniledger/service/contracts.go
+++ b/omniledger/service/contracts.go
@@ -2,7 +2,9 @@ package service
 
 import (
 	"encoding/binary"
+	"encoding/hex"
 	"errors"
+	"strings"
 	"time"
 
 	"github.com/dedis/cothority/omniledger/darc"
@@ -87,10 +89,41 @@ func LoadBlockIntervalFromColl(coll CollectionView) (time.Duration, error) {
 	return config.BlockInterval, nil
 }
 
+// LoadDarcFromColl loads a darc which should be stored in key.
+func LoadDarcFromColl(coll CollectionView, key []byte) (*darc.Darc, error) {
+	rec, err := coll.Get(key).Record()
+	if err != nil {
+		return nil, err
+	}
+	vs, err := rec.Values()
+	if err != nil {
+		return nil, err
+	}
+	if len(vs) < 2 {
+		return nil, errors.New("not enough records")
+	}
+	contractBuf, ok := vs[1].([]byte)
+	if !ok {
+		return nil, errors.New("can not cast value to byte slice")
+	}
+	if string(contractBuf) != "darc" {
+		return nil, errors.New("expected contract to be darc but got: " + string(contractBuf))
+	}
+	darcBuf, ok := vs[0].([]byte)
+	if !ok {
+		return nil, errors.New("cannot cast value to byte slice")
+	}
+	d, err := darc.NewDarcFromProto(darcBuf)
+	if err != nil {
+		return nil, err
+	}
+	return d, nil
+}
+
 // ContractConfig can only be instantiated once per skipchain, and only for
 // the genesis block.
 func (s *Service) ContractConfig(cdb CollectionView, tx Instruction, coins []Coin) (sc []StateChange, c []Coin, err error) {
-	if tx.Spawn == nil {
+	if tx.getType() != spawnType {
 		return nil, nil, errors.New("Config can only be spawned")
 	}
 	darcBuf := tx.Spawn.Args.Search("darc")
@@ -102,7 +135,7 @@ func (s *Service) ContractConfig(cdb CollectionView, tx Instruction, coins []Coi
 	if len(d.Rules) == 0 {
 		return nil, nil, errors.New("don't accept darc with empty rules")
 	}
-	if err = d.Verify(); err != nil {
+	if err = d.Verify(true); err != nil {
 		log.Error("couldn't verify darc")
 		return
 	}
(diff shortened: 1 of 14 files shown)


omniledger/service/service_test.go
```diff
@@ -389,6 +458,33 @@ func (s *ser) service() *Service {
...
+		SkipchainID: s.sb.SkipChainID(),
+		Transaction: ctx,
+	})
+	require.Nil(t, err)
+
+	// wait for the block to be processed (it should fail due to the wrong
+	// version)
+	time.Sleep(4 * s.interval)
+
+	// try to get the darc bac, we should get the genesis back instead
```
jeffallen: s/bac/back/

omniledger/service/service_test.go
```diff
@@ -389,6 +458,33 @@ func (s *ser) service() *Service {
...
+func (s *ser) testDarcEvolution(t *testing.T, d2 darc.Darc) GetProofResponse {
+	ctx := darcToCtx(t, d2, s.signer)
+	_, err := s.service().AddTransaction(&AddTxRequest{
+		Version:     CurrentVersion,
+		SkipchainID: s.sb.SkipChainID(),
+		Transaction: ctx,
+	})
+	require.Nil(t, err)
+
+	// wait for the block to be processed (it should fail due to the wrong
```
jeffallen: It should fail? This test seems to be checking for success?

omniledger/service/service_test.go
```diff
@@ -372,6 +372,75 @@ func TestService_StateChange(t *testing.T) {
...
+		Nonce:  GenNonce(),
+		Index:  0,
+		Length: 1,
+		Invoke: &invoke,
+	}
+	require.Nil(t, instr.SignBy(signer))
+	ctx := ClientTransaction{
+		Instructions: []Instruction{instr},
+	}
+	return ctx
```
jeffallen: Can we avoid using ctx to mean client transaction? ctx is well and truly a reserved word in my head for "context".

omniledger/service/service_test.go
```diff
@@ -372,6 +372,75 @@ func TestService_StateChange(t *testing.T) {
...
+	// parse the darc
+	require.True(t, resp.Proof.InclusionProof.Match())
+	_, vs, err := resp.Proof.KeyValue()
+	require.Nil(t, err)
+	d22, err := darc.NewDarcFromProto(vs[0])
+	require.Nil(t, err)
+	require.True(t, d22.Equal(d2))
+}
+
+func darcToCtx(t *testing.T, d2 darc.Darc, signer darc.Signer) ClientTransaction {
```
jeffallen: s/darcToCtx/darcToClientTx/

See next comment on ctx.

omniledger/service/service_test.go
```diff
@@ -372,6 +372,75 @@ func TestService_StateChange(t *testing.T) {
 	require.Equal(t, latest, int64(n-1))
 }
 
+func TestService_DarcEvolutionFail(t *testing.T) {
+	s := newSer(t, 1, testInterval)
+	defer s.local.CloseAll()
+	defer closeQueues(s.local)
+
+	d2 := s.darc.Copy()
```
jeffallen: This is a bit a surprising API, that you need to make a copy and then do "EvolveFrom" on the copy, referencing the source. Would it be possible/good if it worked like `d2 := s.darc.Evolve()` instead? So that Evolve() does the copy and the evolution at the same time?

(Sorry, I know this was already decided on before when darc was checked in.)

omniledger/service/contracts.go
```diff
@@ -138,6 +171,63 @@ func (s *Service) ContractConfig(cdb CollectionView, tx Instruction, coins []Coi
 // ContractDarc accepts the following instructions:
 //   - Spawn - creates a new darc
 //   - Invoke.Evolve - evolves an existing darc
-func (s *Service) ContractDarc(cdb CollectionView, tx Instruction, coins []Coin) (sc []StateChange, c []Coin, err error) {
-	return nil, nil, errors.New("Not yet implemented")
+func (s *Service) ContractDarc(coll CollectionView, tx Instruction,
+	coins []Coin) ([]StateChange, []Coin, error) {
+	if tx.Invoke == nil {
```
jeffallen: Every contract is starting with different formulations of this. Repeat after me: repetition is not good.

Maybe we could make:
```
func (i *Instruction) Require(itype InstructionType) error {
  switch itype {
    case Spawn:
      if tx.Invoke != nil || tx.Delete != nil {
        return errors.New("transaction must be a spawn")
     ...etc...
```

omniledger/service/contracts.go
```diff
@@ -138,6 +171,63 @@ func (s *Service) ContractConfig(cdb CollectionView, tx Instruction, coins []Coi
...
+			NewStateChange(Update, tx.ObjectID, ContractDarcID, darcBuf),
+		}, nil, nil
+	} else if tx.Invoke.Command == "add" {
+		return nil, nil, errors.New("not implemented")
+	}
+	return nil, nil, errors.New("invalid command: " + tx.Invoke.Command)
+}
+
+func strToObjSlice(s string) ([]byte, error) {
+	ss := strings.Split(s, ":")
```
jeffallen: maybe SplitN so that once the first colon is found it does not keep going? Cuts out a probably not real, but theoretical attack vector by giving a ginormous string with billions of colons in it.

omniledger/service/contracts.go
```diff
@@ -138,6 +171,63 @@ func (s *Service) ContractConfig(cdb CollectionView, tx Instruction, coins []Coi
...
+	if len(ss) != 2 {
+		return nil, errors.New("failed to split string into 2")
+	}
+	darcID, err := hex.DecodeString(ss[1])
```
jeffallen: Why is ss[0] not validated? Why can I pass in garbage:00ff00 and get to here? (maybe this is not important, I don't quite have the whole picture on this.)

omniledger/service/service_test.go
kc1212: wrong comment, but this function doesn't do any real assertion on the response, the tests do it.

omniledger/service/service_test.go
kc1212: I was also thinking about this issue. But doing `d2 := s.darc.Evolve(signers)` also won't work nicely because you can't change anything in the `s.darc`. You can do `d.Evolve()`, then change the darc, and then do `d.SignEvolution(signers)`. Which is also not so nice.

One solution I came up with, but haven't had time to implement yet, is to use the builder pattern. We use it a lot in java and I think it's quite elegant to configuring a complex object before finalising it.

kc1212: Created an issue - https://github.com/dedis/cothority/issues/1288

omniledger/service/contracts.go
kc1212: it's not so important because this is used in a callback for looking for darcs which are already stored in the collections, which are guaranteed to be correct and `ss[0]` should always be "darc". but it's always better to check in case this function gets used by something that has untrusted data.

kc1212:
Thanks @jeffallen addressed your comments

omniledger/service/transaction.go
```diff
@@ -353,6 +366,27 @@ func (sc StateAction) String() string {
 	}
 }
 
+type instrType int
+
+const (
+	spawnType instrType = iota
+	invokeType
+	deleteType
+	invalidInstrType
```
jeffallen: Normally I'd expect the zero value of a type to be the invalid one, so that in case the zero value is accidentally used someplace, it does not accidentally mean "spawn".

This new getType is really nice, thanks for doing it private only, very good API cleanliness. :)

PR #1287, opened 2018-06-11
Omniledger documentation

ineiti:
Integrating omniledger documentation and adding more information.


omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
...
+
+1. find the Darc instance by looking at the first 32 bytes of the `InstanceID` given in
+the instruction, here the `GenesisDarc`, and adding 32 x 0x00 bytes
+2. create a `DarcRequest` using the `InstanceID` and the `Args`
+3. verify the request corresponds to the expression of the `Invoke_Update` rule
+in the Darc instance found in 1.
+
+## Contract Arguments
+
+A contract is immutable and has the following signature in go:
```
jeffallen: Well, immutable until I type "go build" again. I don't think immutable is quite the word we want here...

s/go/Go/

omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
...
+pointing to the data the contract should work on
+- `inCoins` is mostly ignored for the moment, but can be used to pass around
+coins between different instructions
+
+Output:
+- `sc` is the slice of stateChanges the contract wants to apply to the global
+state. They will only be applied if all instructions in the ClientTransaction
+are valid, else they will be discarded
+- `outCoins` can be used to pass around coins
+- `err` if not nil, the contract indicates it failed, and all instructions in that
```
jeffallen: Wait what? Wow, I didn't understand that before. (Just checked in the code, yup!)

What's an example use case of len(tx.Instructions) > 1, where the first instruction succeeds and then a later instruction fails?

Also: It seems like if createStateChanges executes "continue clientTransactions" on the second and latter instrctions, that states will still have any states from the previous executed instructions in it. Which seems not right? If you agree, please file an issue.

omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
...
+ClientTransaction will be discarded
+
+The contract itself has access to all elements of the collection, but will mainly
+work on the data pointed to by the `tx Instruction` given as a parameter.
```
jeffallen: Explain, please:
- why the collection is read-only
- when the state changes are applied to the collection
- that client transactions will see a consistent view of the collection, but no guarantee which order they are relative to other client trans in the same block.

omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Instance Structure
+
+Every instance in omniledger is stored with the following information in the
+global state:
+
+- `InstanceID` is a globally unique identifier of that instance, composed of:
+  - `DarcID`, defining the access rights to that instance
+  - `Nonce`, which is randomly chosen. The special `Nonce` of `0` indicates
```
jeffallen: This will be updated to the new terminology.

omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
...
+the Darc responsible for all the instances starting with the same `DarcID`
+- `ContractID` points to the contract that will be called if that instance
+receives an instruction from the client
+- `Data` is interpreted by the contract and can change over time
+
+## Interaction between Instructions and Instances
+
+Every instruction sent by a client indicates the `InstanceID` it is sent to.
+Omniledger will start by verifying the authorization as described above, then
+use the `InstanceID` to look up the responsible contract for this instance and
```
jeffallen: This looking up the responsible contract is only for invoke. For Spawn, the responsible contract is specified in the insttruction.Spawn.ContractID field.

omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
...
+then send the instruction to that contract. A client can call an instance with
+one of the following three basic instructions:
+
+- `Spawn` - will ask the instance to create a new instance. The client indicates the
+requested new contract-type and the arguments. Currently only `Darc` instances can
+spawn new instances.
```
jeffallen: but user defined contracts can also spawn? because that's how we expected to continue to impl eventlog.Log()

omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
...
+
+In the omniledger service, the following contracts are pre-defined:
+
+- `GenesisReference` - points to the genesis configuration
+- `Config` - holds the configuration of omniledger
+- `Darc` - defines the access control
+
+To extend omniledger, you will have to create a new service that defines new
+contracts that will have to be registered with omniledger. An example is
+[EventLog](../../eventlog) that defines further contracts.
```
jeffallen: s/further contracts/a contract/.

omniledger/ChildTransactions.md
```diff
@@ -0,0 +1,120 @@
...
+and how the leader can fetch all new transactions from the children.
+
+## Transaction Queue and Block Generation
+
+This part of the document describes the technical details of the design and
+implementation of transaction queue and block generation for OmniLedger. The
+assumption is that the leader will not fail. We will implement view-change in
+the future, starting by eliminating stop-failure and then byzantine-failure.
+Further, we assume there exists a maximum block size of B bytes. Transaction
+Queue A transaction is similar to what is defined above, namely a key/kind/value
```
jeffallen: s/Transaction Queue// ? Otherwise I can't understand this sentence...

omniledger/ChildTransactions.md
```diff
@@ -0,0 +1,120 @@
...
+valid. The valid transactions are marked as "Verified" and the bad transactions
+are dropped and a message is printed to the audit log. Finally, the transactions
+with the "Verified" flag are sent to the leader in the `PollTxResponse` message.
+These transactions are marked as "Submitted". The `PollTxResponse` message should
+not be larger than B bytes.
+
+Upon receiving all the `PollTxResponse` message, the leader does the following:
+- remove duplicates
+- verify signature
+- sort the transactions in a random but deterministic way
```
jeffallen: "shuffle the transactions in a deterministic way"

omniledger/ChildTransactions.md
```diff
@@ -0,0 +1,120 @@
...
+of wasted messages if there are very little or no transactions. We can attempt
+to implement the simplest technique first and then try to optimise it later. For
+example, a slightly better version would be to add some delay when the blocks
+are getting smaller. But this is only a heuristic because the leader does not
+know how many transactions are in the queues of the non-root nodes.
+
+Another option would be that each node sends _one_ message to the leader if it
+has not-included transactions. This could happen when:
+- the queue has been empty and a first transaction comes in
+- a new block has been accepted, but not all transaction of the queue are in
```
jeffallen: How about piggybacking an advisory of "I have some txns you should come pick up" on the replies to request to sign a new block?

omniledger/ChildTransactions.md
```diff
@@ -0,0 +1,120 @@
...
+client should not see the success message as an indication that the transaction
+is included in the block, but that the transaction is received and may be
+accepted into a block. We do not attempt to check whether the transaction is
+valid at this point because the conode’s darc database might not be up-to-date,
+for example if it just came back online.  
+
+### Block Generation
+
+The poll method is inspired by the
+beta synchroniser, where the leader sends a message, e.g.
```
jeffallen: What is beta synchronizer?

omniledger/DataStructues.md
```diff
@@ -0,0 +1,248 @@
+Navigation: [DEDIS](https://github.com/dedis/doc/tree/master/README.md) ::
+[Cothority](https://github.com/dedis/cothority/tree/master/README.md) ::
+[Building Blocks](https://github.com/dedis/cothority/tree/master/BuildingBlocks.md) ::
```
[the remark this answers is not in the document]
ineiti: wrong path - missed `doc/`

omniledger/ChildTransactions.md
ineiti: Sounds like a possible optimisation, yes.

omniledger/ChildTransactions.md
ineiti: No idea - this is copied from @kc1212 proposition to go further...

omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
+Navigation: [DEDIS](https://github.com/dedis/doc/tree/master/README.md) ::
+[Cothority](https://github.com/dedis/cothority/tree/master/README.md) ::
+[Building Blocks](https://github.com/dedis/cothority/tree/master/BuildingBlocks.md) ::
+[OmniLedger](README.md) ::
+Contracts and Instances
+
+# Contracts and Instances
+
+A contract in omniledger is similar to a smart contract in Ethereum, except that
```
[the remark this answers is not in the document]
ineiti: It's more easy to implement. We discussed with Viktor Kuncak, and we might use his scala verification tools to create a contract that interprets jvm...

@jeffallen said that we can sell this restriction by saying that we're serious about smart contracts and don't think everybody should be able to write those. But then we'll have a hard time explaining when we actually _do_ implement them ;)

omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Authorizations
+
+Authorizations are handled using Darcs. Each Darc has a set of rules that define
+a pair of action / expression that need to be fulfilled to execute any instruction
+on the instances governed by that Darc.
+
+A Darc is always stored with the following `InstanceID`:
+
+`[32]byte{DarcBaseID}[32]byte{0}`
```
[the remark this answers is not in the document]
ineiti: Would be a possibility. But the new darc simply overwrites the old one...

omniledger/Contracts.md
```diff
@@ -0,0 +1,191 @@
...
+
+All instances spawned by that Darc (except for other Darcs) will share the same
+first half of the `InstanceID`. The second half is randomly chosen by the contract
```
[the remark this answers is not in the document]
ineiti: That feels like somebody's got bitten ;)

omniledger/Contracts.md
ineiti: Example use case is a transaction with the two instructions:
1. Get x coins from account
2. Spend the coins creating a document

So if 2. fails, we don't want to spend the coins, but just discard the whole transaction.

But: it is important that 2. works with the changed state of 1. (which empties the account), else you could take money out of your account multiple times.

omniledger/Contracts.md
ineiti: No, it still goes to the original contract - else the contract being called would have to access an unknown data structure. See:

https://github.com/dedis/cothority/pull/1292/files#diff-8b85ff3369bd70f6fe77dcbd8cdf6088R195

ineiti:
As this is hanging around, I'll add my new graph on omniledger.

PR #1291, opened 2018-06-13
Added Omniledger and Skipchain classes

ineiti:
This PR completes the java part of omniledger by adding the following classes:
- `OmniledgerRPC` to create a new omniledger, send `Instruction`s to it and ask it to create `Proof`s
- `DarcInstance` that represents a stored darc in omniledger, complete with `Evolve` method
- `SkipchainRPC` to fetch the latest block of the underlying skipchain

The first commit only updates the protobuf files, so the review can be concentrated on the second commit.


external/java/src/main/java/ch/epfl/dedis/lib/omniledger/OmniledgerRPC.java
```diff
@@ -0,0 +1,196 @@
...
+        this(d, new Configuration(r, blockInterval));
+    }
+
+    /**
+     * Instantiates an omniledger object given the byte representation. The omniledger must already have been
+     * initialized on the cothority.
+     *
+     * @param buf is the representation of the basic omniledger parameters
+     */
+    public OmniledgerRPC(byte[] buf) {
```
jeffallen: This seems to be missing an impl.

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/contracts/DarcInstance.java
```diff
@@ -0,0 +1,159 @@
...
+     * darc as "Invoke_Evolve" rule.
+     *
+     * @param newDarc the darc to replace the old darc.
+     * @param owner   must have its identity in the "Invoke_Evolve" rule
+     * @param pos     position of the instruction in the ClientTransaction
+     * @param len     total number of instructions in the ClientTransaction
+     * @return Instruction to be sent to omniledger
+     * @throws CothorityCryptoException
+     */
+    public Instruction evolveDarcInstruction(Darc newDarc, Signer owner, int pos, int len) throws CothorityCryptoException {
```
jeffallen: s/owner/signer/ I think. Because any authorized signer can ask for the evolution, not just the owner. (The default is owner only, but the owner could give others evolution rights?)

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/contracts/DarcInstance.java
```diff
@@ -0,0 +1,159 @@
...
+        TransactionId tid = evolveDarc(newDarc, owner);
+        while (true) {
+            Proof p = ol.getProof(instance.getId());
+            Instance inst = new Instance(p);
+            try {
+                darc = new Darc(inst.getData());
+                if (darc.getVersion() == newDarc.getVersion()) {
+                    break;
+                }
+                Thread.sleep(ol.getConfig().getBlockInterval().getNano() / 1_000_000);
```
jeffallen: getBlockInterval() returns a Duration, and Thread.sleep wants milis, so use toMillis: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toMillis--

external/java/src/test/java/ch/epfl/dedis/omniledger/OmniledgerRPCTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Test
+    void updateDarc() throws Exception{
+        SkipBlock previous = ol.getLatest();
+        logger.info("Previous skipblock is: {}", previous.getIndex());
+        DarcInstance dc = new DarcInstance(ol, genesisDarc);
+        logger.info("DC is: {}", dc.getId());
+        logger.info("genesisDarc is: {}", genesisDarc.getId());
+        Darc newDarc = genesisDarc.copy();
+        newDarc.setRule("spawn:darc", "all".getBytes());
```
jeffallen: Why does setRule take byte[] and not String?

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/contracts/DarcInstance.java
ineiti: My definition of `owner` is any set of identities that can resolve the expression in the `invoke:evolve` rule. I added a TODO but kept `owner`.

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/contracts/DarcInstance.java
ineiti: thanks - I did a quick google search and only found this solution... Fixed it.

external/java/src/test/java/ch/epfl/dedis/omniledger/OmniledgerRPCTest.java
ineiti: Good question - that part still comes from @kc1212 and I did not touch this part. Anyway, it will fail because the rule will not be able to be verified...

kc1212: this is because the parser combinator library in go expects `[]byte`, so the expression in go are all `[]byte`. We can definitely change it to string to in the java side because it won't try to evaluate the expression.

PR #1292, opened 2018-06-14
Darc spawn

ineiti:
This adds `Spawn` to darcs:
- if you spawn another darc, it will be directly created
- if you spawn another contract, this contract will be called and then this contract will have to verify if all is correct

Also added a `Value` contract to store any value you want... Dropbox on Omniledger, anybody?


external/java/src/main/java/ch/epfl/dedis/lib/omniledger/contracts/DarcInstance.java
```diff
@@ -135,6 +135,88 @@ public void evolveDarcAndWait(Darc newDarc, Signer owner) throws CothorityExcept
         }
     }
 
+    /**
+     * Creates an instruction for spwaning a contract.
```
jeffallen: s/spwaning/spawning/

omniledger/service/contracts.go
```diff
@@ -163,35 +168,80 @@ func (s *Service) ContractConfig(cdb CollectionView, tx Instruction, coins []Coi
...
+	default:
+		return nil, nil, errors.New("Only invoke and spawn are defined yet")
 	}
-	return nil, nil, errors.New("invalid command: " + tx.Invoke.Command)
+}
+
+// ContractValue is a simple key/value storage where you
+// can put any data inside as wished.
+func (s *Service) ContractValue(cdb CollectionView, tx Instruction, c []Coin) ([]StateChange, []Coin, error) {
+	switch {
```
jeffallen: There is no code here to validate that the instruction has exactly one of spawn, invoke, delete. Can be fixed later in a nicer way. See issue #1294 that I just filed.

omniledger/service/contracts.go
```diff
@@ -35,6 +35,10 @@ var ContractConfigID = "config"
 // ContractDarcID denotes a darc-contract
 var ContractDarcID = "darc"
 
+// ContractValueID denotes a contract that can store and update
+// key values.
```
jeffallen: Is this a toy? Is this a caprice de Linus? Is this a critical future feature? Or for testing?

Am a bit concerned we are adding a super generic thing with no structure. Who's allowed to use this?

omniledger/service/contracts.go
```diff
@@ -163,35 +168,80 @@ func (s *Service) ContractConfig(cdb CollectionView, tx Instruction, coins []Coi
...
-		newD, err := darc.NewDarcFromProto(darcBuf)
-		if err != nil {
-			return nil, nil, err
+	switch {
+	case tx.Spawn != nil:
+		if tx.Spawn.ContractID == ContractDarcID {
+			darcBuf := tx.Spawn.Args.Search("darc")
+			d, err := darc.NewDarcFromProto(darcBuf)
+			if err != nil {
+				return nil, nil, errors.New("given darc was not valid: " + err.Error())
```
jeffallen: maybe more clear: "could not be decoded"

in the context of darc, Valid() is about the signatures, right?

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/contracts/DarcInstance.java
ineiti: difficult word - I found another occurrence of `spwan`... but only in java...

omniledger/service/contracts.go
ineiti: Having only one of `Spawn`, `Invoke`, `Delete` should be checked in omniledger itself.

omniledger/service/contracts.go
ineiti: Allowed is everybody with `spawn:value` access in a darc. I use it in the java tests. And it might also be used for more generic data storage.

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/InstanceId.java
```diff
@@ -69,7 +70,18 @@ public String toString(){
         return DatatypeConverter.printHexBinary(id);
     }
 
-    public ByteString toProto(){
+    public ByteString toByteString(){
         return ByteString.copyFrom(id);
     }
+
+    public TransactionProto.ObjectID toProto(){
```
kc1212: this is a bit confusing to me, so we don't have an `ObjectID` class anymore, but the equivalent in java is called `InstanceID`?

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/Instruction.java
```diff
@@ -77,8 +78,8 @@ public Instruction(ObjectID objId, byte[] nonce, int index, int length, Delete d
     /**
      * Getter for the object ID.
```
kc1212: update comment

PR #1298, opened 2018-06-19
Fix dangling forward links on server start

jeffallen:
Also: Update timeouts in order to get big elections to finalize.

### DIFF
--- a/evoting/service/service.go
+++ b/evoting/service/service.go
@@ -2,6 +2,7 @@
 package service
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -340,6 +341,10 @@ func (s *Service) Cast(req *evoting.Cast) (*evoting.CastReply, error) {
 func (s *Service) GetElections(req *evoting.GetElections) (*evoting.GetElectionsReply, error) {
 	master, err := lib.GetMaster(s.skipchain, req.Master)
 	if err != nil {
+		log.LLvlf4("get master bad forward link: starting from %x", req.Master)
+		buf := &bytes.Buffer{}
+		lib.DebugDumpChain(s.skipchain, buf, req.Master)
+		log.LLvl4(string(buf.Bytes()))
 		return nil, err
 	}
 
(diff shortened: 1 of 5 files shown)


jeffallen:
Travis has been broken on evoting for a while, so don't look into that. Just look at the repair code.

Here's what it looks like when it is repairing the database from the production DEDIS conode:
```
1!: (                             service.new: 848) -   block a3004388a5acdcc708b31115568853acfe43f4eb85e0c5f102cf05eedc305a52: forward link to fdd40bb4dc7ad131950b6f57774d24bc97c0f1b5ab830f5d313d46ac8f25627f at level 1 invalid, fixing it.
1!: (                             service.new: 863) -   writing block a3004388a5acdcc708b31115568853acfe43f4eb85e0c5f102cf05eedc305a52
1!: (                             service.new: 848) -   block 653951e092a562c578e7e9a81e6fd09bc6f2e7837986407d2be7a9d6b208d43a: forward link to fdd40bb4dc7ad131950b6f57774d24bc97c0f1b5ab830f5d313d46ac8f25627f at level 0 invalid, fixing it.
1!: (                             service.new: 863) -   writing block 653951e092a562c578e7e9a81e6fd09bc6f2e7837986407d2be7a9d6b208d43a
1!: (                             service.new: 874) - chain 19069a0fb17ce4b020102ee5dba2d32053044fa27ca0245720c8f45187feb22d had invalid forward links
```

That is the only chain that has the problem, at least on our conode.

evoting/service/service.go
```diff
@@ -815,6 +827,54 @@ func new(context *onet.Context) (onet.Service, error) {
 	log.RegisterListener(&service.rl)
 	service.RegisterStatusReporter("evoting", service)
 
+	// Fix dangling forward pointers: for each skipchain,
+	// for each block, check the forward links looking for
+	// a dangling one. Remove them.
+	db := service.db()
+	chains, err := db.GetAllSkipchains()
```
ineiti: is that still the old that returns all skipblocks?

ineiti:
thinking about it - probably a cleaner way would've been to create a proper `skipchain_id.db` file and send that to the administrators. Anyway, that's for the next time we have to fix a db ;)

PR #1301, opened 2018-06-20
Java API for searching events

kc1212:
The go API was implemented in a previous PR, this is the Java part of
it.

Later, these API may be moved to an `EventLogInstance` class.


ineiti:
So this connects to the eventlog-service API - I think #1303 would be the correct next steps for this interface. Not sure where the `Search` should go. 

We might need something like ethereum where the contracts define how to access the internal data structure...

PR #1302, opened 2018-06-20
adding backward-compatible challenge calculation

### LINKED ISSUE #1293: Don't send challenge
https://github.com/dedis/cothority/blob/d502d742e6ef5cb7a3bdee7250f6e04046de0bd6/ftcosi/protocol/protocol.go#L177

Here the challenge is sent over the network, but only the commitment should be sent, and every node should re-compute the challenge on it's own.

ineiti:
The current cosi implementation trusts the leader to correctly calculate the challenge - but this means that the leader could propose any challenge he wants and get the nodes to sign off something they did not verify.

As we're in `v2` non-breaking mode, it will output a warning if it receives a pre-calculated challenge, but still continue. In the best case, it will calculate the challenge.

Closes #1293

### DIFF
--- a/ftcosi/protocol/struct.go
+++ b/ftcosi/protocol/struct.go
@@ -76,7 +76,12 @@ type StructCommitment struct {
 
 // Challenge is the ftcosi challenge message
 type Challenge struct {
+	// CoSiChallenge is deprecated and should not be used anymore!
 	CoSiChallenge kyber.Scalar
+	// AggregateCommit should be used by all nodes.
+	AggregateCommit kyber.Point
+	// Mask represents the nodes that participated in the signature.
+	Mask *cosi.Mask
 }
 
 // StructChallenge just contains Challenge and the data necessary to identify and
--- a/ftcosi/protocol/sub_protocol.go
+++ b/ftcosi/protocol/sub_protocol.go
@@ -64,18 +64,14 @@ func NewSubFtCosi(n *onet.TreeNodeInstance, vf VerificationFn, suite cosi.Suite)
 		c.subResponse = make(chan StructResponse)
 	}
 
-	for _, channel := range []interface{}{
-		&c.ChannelAnnouncement,
+	err := c.RegisterChannels(&c.ChannelAnnouncement,
 		&c.ChannelCommitment,
 		&c.ChannelChallenge,
-		&c.ChannelResponse,
-	} {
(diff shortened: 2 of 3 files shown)


ftcosi/protocol/struct.go
```diff
@@ -76,7 +76,12 @@ type StructCommitment struct {
 
 // Challenge is the ftcosi challenge message
 type Challenge struct {
+	// CoSiChallenge is deprecated and should not be used anymore!
```
jeffallen: This does not correspond to the code, i.e. sup_protocol.go, line 280 where you compare the locally calculated challenge with the remotely provided one. If it should not be used anymore, why are we depending on it being set in order to check it?

ftcosi/protocol/sub_protocol.go
```diff
@@ -260,6 +259,19 @@ func (p *SubFtCosi) Dispatch() error {
...
 		// generate own response and send to parent
+		if challenge.AggregateCommit == nil || challenge.Mask == nil {
+			log.Warn("Only have pre-calculated challange - this is dangerous!")
+		} else {
+			cosiChallenge, err := cosi.Challenge(p.suite, challenge.AggregateCommit,
+				challenge.Mask.AggregatePublic, p.Msg)
+			if err != nil {
+				return err
+			}
+			if !cosiChallenge.Equal(challenge.CoSiChallenge) {
```
jeffallen: I was not expecting this branch of the code to read from challenge.CoSiChallenge, since it is deprecated. I expected the output of cosi.Challenge to be used directly.

ineiti: Not sure I can do better and easier to provide both backward-compatibility and warning about bad behaviour. Any proposition how to do better here?

jeffallen: I guess what I mean is that this code is checking a value that the docs say should no longer be sent. So when it is not sent, this code will fail. If the new format (i.e. provide Agg and Mask) is provided, why not just use it 100% and ignore the old format?

kc1212: I think it must be sent because some nodes might not be on the latest code.

jeffallen: Ah, this is what I was missing, that gnerateResponse will be using the challenge nextly. OK, got it, kind of.

PR #1306, opened 2018-06-20
Do not allow verifier panic to take down the server

jeffallen:
If a verifier panics, do not crash. Instead report the error
and treat the verifier as if it failed.


ineiti:
Out of curiosity - which function was panicking?

jeffallen:
Today it was kyber/shuffle/pair.go:240, "mismatched vector lengths".

But it could have been anywhere, there are other things in evoting that could panic, including array index out of bounds, etc.

The fact that panics escape from package kyber/shuffle is proof that dedis/kyber#316 is not ready for stable.

PR #1307, opened 2018-06-20
Storing batches of blocks

ineiti:
In order to avoid storing a block with a forward-link and opening a small
window where this block can confuse the skipchain service, this PR saves
multiple blocks at the same time in a boltdb-transaction.

### DIFF
--- a/omniledger/service/proof_test.go
+++ b/omniledger/service/proof_test.go
@@ -106,8 +106,8 @@ func createSC(t *testing.T) (s sc) {
 	s.sb2.Hash = s.sb2.CalculateHash()
 	s.genesis.ForwardLink = genForwardLink(t, s.genesis, s.sb2, s.genesisPrivs)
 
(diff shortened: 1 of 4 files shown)


skipchain/skipchain_test.go
```diff
@@ -113,7 +113,7 @@ func storeSkipBlock(t *testing.T, nbrServers int, fail bool) {
 
 	// And add it again, with all nodes running
 	if fail {
-		log.Lvl3("Unpausing server ", deadServer.Address())
+		log.LLvl3("Unpausing server ", deadServer.Address())
```
jeffallen: Checked in an LL, oops.

ineiti: yes, still WIP because it doesn't work yet - I thought (again) that this was a low hanging fruit :(

ineiti:
Putting this in #1297 because one needs the other needs the one...

PR #1314, opened 2018-06-26
Update el tool and create ol tool

jeffallen:
To correctly work with our current understanding of how to use this stuff.


conode/run_conode.sh
```diff
@@ -26,6 +26,12 @@ VERSION="$VERSION_ONET-$VERSION_SUB"
 #   TAGS="-tags vartime" ./run_conode.sh
 # Note: TAGS is also used by the integration tests.
 
+# This will allow you to put the servers on a different range if you want.
+#   $ export PORTBASE=18000
+#   $ ./run_conode.sh local 3
+# Results in the servers being on 18002-18007.
+[ -z "$PORTBASE" ] && PORTBASE=7000
+
```
ineiti: We absolutely need to kill `run_conode.sh` - it really gets aweful. Of course it's OK to extend it like you do, no need to change - but `run_conode.sh` should be removed...

eventlog/api.go
```diff
@@ -51,46 +43,15 @@ func AddWriter(r darc.Rules, expr expression.Expr) darc.Rules {
...
-	// When we have a genesis block, we need to initialise one eventlog and
-	// store its ID.
-	var instID *omniledger.InstanceID
-	instID, err = c.initEventLog()
-	if err != nil {
-		return nil, err
-	}
-	c.InstanceID = *instID
-	return &c.InstanceID, nil
+	return nil, errors.New("not impl")
```
ineiti: Just remove the method...

eventlog/el/main.go
```diff
@@ -168,97 +155,71 @@ func main() {
...
-	n := c.String("name")
-	cfg, err := doCreate(n, r, 5*time.Second)
-	if err != nil {
-		return err
-	}
-	fmt.Printf("Created event log chain with ID %x.\n", cfg.ID)
-	return err
+func show(c *cli.Context) error {
+	return errors.New("not implemented")
+	// long := c.Bool("long")
```
ineiti: I would remove the commented out code. It doesn't look like something you can't write (better) in 10 minutes...

eventlog/el/main_test.go
```diff
@@ -23,6 +23,9 @@ func TestMain(m *testing.M) {
 const blockInterval = 1 * time.Second
 
 func Test(t *testing.T) {
```
ineiti: Please change the name to `TestFull` or something.

omniledger/ol/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// This is required; without it onet/log/testuitl.go:interestingGoroutines will
+// call main.main() interesting.
+func TestMain(m *testing.M) {
+	log.MainTest(m)
+}
+
+func Test(t *testing.T) {
```
ineiti: s/Test/TestFull/

omniledger/ol/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+	l := onet.NewTCPTest(cothority.Suite)
+	_, roster, _ := l.GenTree(2, true)
+
+	defer l.CloseAll()
+	defer func() {
+		// Walk the service lists, looking for Omniledgers that we can shut down.
+		for _, x := range l.Services {
+			for _, y := range x {
+				if z, ok := y.(*omniledger.Service); ok {
+					close(z.CloseQueues)
```
ineiti: I saw that problem, too - definitively something we need to find a better fix for. Keep it for the moment, but we need to be able to stop the services, at least for the tests/simulations.

omniledger/ol/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+				}
+			}
+		}
+	}()
+
+	// All this mess is to take the roster we have from onet.NewTCPTest
+	// and get it into a file that create can read.
+	g := &app.GroupToml{
```
ineiti: Sounds like a good candidate for `onet/app/config.go`

omniledger/ol/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+		g.Servers[i] = &app.ServerToml{
+			Address: si.Address,
+			Public:  si.Public.String(),
+		}
+	}
+	rf := path.Join(dir, "roster.toml")
+	err = g.Save(rf)
+	require.NoError(t, err)
+
+	t.Log("create: ")
```
ineiti: Please use `log.Lvl1` here, the go-logs are only printed at the end of the test, and as such will not be synched with other `log.Lvl*` calls.

jeffallen:
main_test.go is my answer to test.sh. I'm quite satisfied with how it is turning out. And my discovery of [cliApp.Metadata](https://godoc.org/gopkg.in/urfave/cli.v1#App.Metadata) makes it even cooler: you can have the production code stuff intermediate results in there in a typed manner, so that the test code can go in and see if things worked right without having to parse the output. (See line 72.)

What I really hated about debugging test.sh was trying to write and then later fix, the awk/sed/etc to verify if the CLI had done the right thing. With main_test.go, we keep the data typed, and inside of Go where it is ready to be worked with.

PR #1321, opened 2018-06-28
Make our java compatible with Android

jeffallen:
Remove dependency on javax.xml.bind, which is not available in Android
Remove @nonnull, which is not supported in the same way on
Java and Android. See issue #1328.


ineiti:
Do we want to have that in `cothority`? Or should it go into a separate repo first?

jeffallen:
Ok, I'll do it in another repo.

PR #1322, opened 2018-06-29
Publish cothority jar file

jeffallen:
Update `pom.xml` in preparation for publication on maven repo.

### DIFF
--- a/external/java/pom.xml
+++ b/external/java/pom.xml
@@ -6,7 +6,29 @@
 
     <groupId>ch.epfl.dedis</groupId>
     <artifactId>cothority</artifactId>
-    <version>v2.0.0-dev</version>
+    <version>2.0.1</version>
+    <name>Cothority client library for Java</name>
+    <description>A client library for communicating with cothorities (a collective authority).</description>
+    <url>https://github.com/dedis/cothority</url>
+    <licenses>
+        <license>
+            <name>GNU Affero General Public License version 3</name>
+            <url>https://opensource.org/licenses/AGPL-3.0</url>
+        </license>
+    </licenses>
+    <developers>
+        <developer>
+            <name>DEDIS Sofware Engineers</name>
+            <email>cothority@googlegroups.com</email>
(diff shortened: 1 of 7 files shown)


external/java/README.md
```diff
@@ -0,0 +1,45 @@
+# Publishing to Maven
+
+Get an account on Sonatype JIRA: https://issues.sonatype.org/secure/Signup!default.jspa
+
+Tell your username to Jeff, who will file a ticket to add you
+to have access to publish.
```
kc1212: I don't really understand this part. This instruction is to help other people to publish their own version of cothority-java? If we need our library to be on the maven repo, it's easier to just publish it directly by ourselves.

external/java/pom.xml
```diff
@@ -6,7 +6,29 @@
...
+            <name>DEDIS Sofware Engineers</name>
+            <email>cothority@googlegroups.com</email>
+            <organization>DEDIS Lab, EPFL</organization>
+            <organizationUrl>https://dedis.epfl.ch/</organizationUrl>
+        </developer>
+    </developers>
+    <scm>
+        <connection>scm:git:git://github.com/dedis/cothority.git</connection>
+        <developerConnection>scm:git:ssh://github.com:dedis/cothority.git</developerConnection>
+        <url>http://github.com/dedis/cothority/tree/master</url>
```
kc1212: Does the `<scm>` tag give us some kind of functionality? If it does, then it might not work for people who want to publish their own versions (see the other comment).

external/java/README.md
jeffallen: It is to tell you how to publish. But maybe it does not belong in the public repo.

external/java/pom.xml
jeffallen: That's all copy-pasta from Sonatype. I have no idea what it's doing.

external/java/README.md
jeffallen: OK, I have moved these notes into the "Operations Notes" file in the Google Drive.

PR #1323, opened 2018-07-02
Coin simuls

pablo-lo:
Adding a simple coin simulation which transfers coins from one account to the other. Also adding a rather hacky python script to generate plots. The script currently relies on the naming of the csv-files from the simulation to add variables to the data which are non-numerical (see issue dedis/onet#446).


ineiti:
Superseded by #1354

PR #1324, opened 2018-07-02
Cleanup omniledger

ineiti:
Cleanup I wanted to do for a long time...
Continued renaming of old structures and methods. Also added some `TODO`s that still need to be done.
Added a `contract/`-directory in omniledger for non-genesis contracts.


pablo-lo:
Codewise it looks good to me, but what's up with travis? The errors are just Java related, right?

ineiti:
Makefile is not nice yet - moving to WIP!

omniledger/service/api.go
```diff
@@ -27,6 +28,7 @@ type Client struct {
 }
 
 // NewClient instantiates a new Omniledger client.
+// TODO: this needs to be changed to avoid having an invalid Client.
```
jeffallen: I wouldn't understand how to fix this todo.

omniledger/service/api.go
```diff
@@ -168,6 +172,31 @@ func (c *Client) GetChainConfig() (*ChainConfig, error) {
 	return config, nil
 }
 
+// WaitProof will poll omniledger until a given instanceID exists.
```
jeffallen: s/omniledger/OmniLedger/

What's up with you guys, do I need to make a pre-commit hook for this? :)

omniledger/service/api.go
```diff
@@ -168,6 +172,31 @@ func (c *Client) GetChainConfig() (*ChainConfig, error) {
...
+// It will return the proof of the instance created. If value is
+// non-nil, it will wait for the value of the proof to be equal to
+// the value.
```
jeffallen: No comment on timeout behaviour. I was expecting maybe a special errTimeout val returned, but in fact as implemented timeout = valid *Proof with !Match() and err == nil.

omniledger/service/contracts.go
```diff
@@ -10,14 +10,14 @@ import (
...
 
-// zeroNonce is 32 bytes of zeroes.
-var zeroNonce = Nonce([32]byte{})
+// ZeroNonce is 32 bytes of zeroes.
+var ZeroNonce = Nonce([32]byte{})
 
-// zeroSubID is 32 bytes of zeroes as a subid for storing Darcs.
-var zeroSubID = SubID([32]byte{})
+// ZeroSubID is 32 bytes of zeroes as a subid for storing Darcs.
+var ZeroSubID = SubID([32]byte{})
```
jeffallen: I just unexported these on purpose, following the go standard of expose the least possible, and also that uninitialized structs should be useful, which they are because instance id zero means something useful.

So, are you sure you want these exported? No big deal for me, just thought it unnecessary.

omniledger/service/struct.go
```diff
@@ -42,11 +45,42 @@ func (r *roCollection) Get(key []byte) collection.Getter {
 	return r.c.Get(key)
 }
 
+func (r *roCollection) GetValues(key []byte) (value []byte, contractID string, err error) {
```
jeffallen: Why is no godoc required here? Whats the diff btw get getvalues and getvalue? Am not really a fan of big interfaces.....

omniledger/service/transaction.go
```diff
@@ -67,14 +77,38 @@ type InstanceID struct {
...
+	return bytes.Compare(iID.DarcID, other.DarcID) == 0 &&
+		bytes.Compare(iID.SubID[:], other.SubID[:]) == 0
+}
+
+// SubID is a 32-byte id.
+type SubID [32]byte
+
+// NewSubID returns a SubID given a slice of bytes.
+func NewSubID(buf []byte) SubID {
+	if len(buf) != 32 {
```
jeffallen: It seems dangerous to silently return a subid that's zero filled when they thought they were getting a correct subid. Imagine how hard to debug if I call someone, I get a []byte back from them with len()==32, and then I send it on to NewSubID. Later the routine I called starts returning 34 bytes because of reasons. Now I am getting zeroSubID but I didn't even get a runtime warning about it.

omniledger/service/transaction.go
```diff
@@ -109,6 +143,11 @@ type Argument struct {
 // Arguments is a searchable list of arguments.
 type Arguments []Argument
 
+// NewArguments is a convenience method to create Arguments.
```
jeffallen: Not a super fan of convenience methods that fit into one line. How inconvenient is it really to just say what it is in the calling code?

Smaller APIs are better, if we can manage it.

omniledger/contracts/service.go
```diff
@@ -0,0 +1,27 @@
+package contracts
+
+import (
+	"github.com/dedis/cothority/omniledger/service"
+	"github.com/dedis/onet"
+	"github.com/dedis/onet/log"
+)
+
+func init() {
+	_, err := onet.RegisterNewService("contracts", newService)
```
kc1212: it'd be nice to have a comment here to say this service is only needed to register the contracts but not an actual service

stable/script
```diff
@@ -0,0 +1,4 @@
+SRC="$1"
+DST="$2"
+
+perl -pi -e "s:VERSION = dev:VERSION = latest:" $DST/conode/Makefile
```
kc1212: could we rename this file name to something more descriptive? e.g. `dev-to-latest`

omniledger/service/api.go
ineiti: Changing the signature of the method to `NewClient(roster, ...)`. Now if you call `NewClient()`, you get a client that is not usable... Which _might_ be OK, but we get weird panics out of it.

omniledger/service/api.go
ineiti: REAL question: or is it `Omniledger`, like the java-class?

omniledger/service/api.go
ineiti: and didn't use the `value` - both fixed now.

omniledger/service/contracts.go
ineiti: We could change it by having a `NewDarcInstanceID` and then un-exporting `ZeroSubID`. What about that?

omniledger/service/struct.go
ineiti: As it's defined in the interface, it doesn't seem to be needed here. Perhaps I got the java-virus? Want me to remove it?

omniledger/service/transaction.go
ineiti: panic? Change the signature to add an `error` to the return?

stable/script
ineiti: it's the 'standard' `stable/script` name...

omniledger/service/api.go
jeffallen: In order to sync up to the paper, I made a sweep thru a while ago making all the Go code OmniLedger.

omniledger/service/struct.go
jeffallen: Either the docs will convince me why it's there, or you should remove it. :)

omniledger/service/transaction.go
jeffallen: Or change the doc and the impl to cleary say, "Will copy len(buf) or 32 bytes, whichever is smaller." And then NewSubID(nil) does something that's valid and even documented: make a zero SubID.

omniledger/service/api.go
ineiti: should we do the same for java?

omniledger/service/transaction.go
ineiti: as we defined `Nonce` and `SubID` as arrays, we can in fact make it much more simple: `SubID{}` and `Nonce{}`.
Now I removed `Zero*`, except for `ZeroDarc`, because the `Darc.ID` is a slice...

omniledger/service/contracts.go
ineiti: In fact `SubID{}` is so much fun, that `NewDarcInstanceID` is just `InstanceID{darc.ID(), SubID{}}`. So no method needed.

omniledger/service/struct.go
ineiti: kept `GetValues` instead of `GetValue` and `GetContractID`.

omniledger/service/api.go
ineiti: made `OmniLedger` in all .md and .go in omniledger-directory.

stable/script
ineiti: Added a comment.

Issue #1317, opened 2018-06-27
Collection-race


ineiti:
```
WARNING: DATA RACE
Write at 0x00c420106e18 by goroutine 98:
  github.com/dedis/cothority/omniledger/collection.(*Collection).update()
      /home/jenkins/.jenkins/workspace/cothority long test/src/github.com/dedis/cothority/omniledger/collection/singlenode.go:49 +0x152
  github.com/dedis/cothority/omniledger/collection.(*Collection).Set()
      /home/jenkins/.jenkins/workspace/cothority long test/src/github.com/dedis/cothority/omniledger/collection/manipulators.go:186 +0x602
  github.com/dedis/cothority/omniledger/service.storeInColl()
      /home/jenkins/.jenkins/workspace/cothority long test/src/github.com/dedis/cothority/omniledger/service/struct.go:112 +0x298
  github.com/dedis/cothority/omniledger/service.(*collectionDB).Store()
      /home/jenkins/.jenkins/workspace/cothority long test/src/github.com/dedis/cothority/omniledger/service/struct.go:121 +0x9b
  github.com/dedis/cothority/omniledger/service.(*Service).updateCollection()
      /home/jenkins/.jenkins/workspace/cothority long test/src/github.com/dedis/cothority/omniledger/service/service.go:404 +0xb9b
  github.com/dedis/cothority/omniledger/service.(*Service).(github.com/dedis/cothority/omniledger/service.updateCollection)-fm()
      /home/jenkins/.jenkins/workspace/cothority long test/src/github.com/dedis/cothority/omniledger/service/service.go:726 +0x55
  github.com/dedis/cothority/messaging.(*Propagate).Dispatch()
      /home/jenkins/.jenkins/workspace/cothority long test/src/github.com/dedis/cothority/messaging/propagate.go:179 +0x1a42
  github.com/dedis/onet.(*Overlay).CreateProtocol.func1()
      /home/jenkins/.jenkins/workspace/cothority long test/src/github.com/dedis/onet/overlay.go:572 +0x42

Previous read at 0x00c420106e18 by goroutine 111:
  github.com/dedis/cothority/omniledger/collection.dumpNode()
      /home/jenkins/.jenkins/workspace/cothority long test/src/github.com/dedis/cothority/omniledger/collection/proof.go:38 +0x1a3
  github.com/dedis/cothority/omniledger/collection.Getter.Proof()
[...]

PR #1329, opened 2018-07-03
Submit transactions to follower

kc1212:
We implement this functionality using a polling leader. Every follower
(including the leader) stores new transactions in a temporary buffer.
This buffer is emptied every time a polling request arrives and the
content of the buffer is sent to the leader. Followers are responsible
for checking whether they are sending the real leader. Otherwise a
malicious polling request may cause the follower to empty its buffer
prematurely.

Closes #1329


omniledger/service/service.go
```diff
@@ -28,6 +27,9 @@ const invokeEvolve darc.Action = darc.Action("invoke:evolve")
 
 // OmniledgerID can be used to refer to this service
 var OmniledgerID onet.ServiceID
+
+const collectTxProtocol = "CollectTxProtocol"
```
ineiti: that should be in `collect_tx.go`

omniledger/service/service.go
```diff
@@ -396,6 +400,18 @@ func (s *Service) updateCollection(msg network.Message) {
 		log.Error("Couldn't recreate state changes:", err.Error())
 		return
 	}
+
+	// Check whether we are storing a genesis config, if we are, also set
+	// the leaderMap.
+	if scs.isGenesisConfig() {
```
ineiti: Why is that here and not in the `Config`-contract?

omniledger/service/leader_map.go
```diff
@@ -0,0 +1,51 @@
+package service
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/dedis/onet/network"
+)
+
+type leaderMap struct {
```
ineiti: I don't understand why you need this `leaderMap` - can't you take the roster of the skipchain and look at the first entry?

omniledger/service/service.go
kc1212: I put it here because `CollectTx` is not a "full" protocol, it needs to be registered with a `getTxs` callback. This is similar to how byzcoinx does it.

omniledger/service/service.go
kc1212: probably will be removed if `leaderMap` is removed

kc1212:
@ineiti removed `leaderMap` and `isGenesisConfig`

omniledger/service/collect_tx.go
```diff
@@ -0,0 +1,119 @@
...
+		return errors.New("did not receive request")
+	}
+
+	// send the result of the callback to the root
+	resp := &CollectTxResponse{
+		Txs: p.getTxs(req.ServerIdentity, req.SkipchainID),
+	}
+	if p.IsRoot() {
+		if err := p.SendTo(p.TreeNode(), resp); err != nil {
+			return err
```
jeffallen: If either of these error cases happen, then the transactions which were taken get lost, because they are not re-put back.

omniledger/service/service.go
```diff
@@ -473,53 +476,66 @@ func (s *Service) loadLatestDarc(sid skipchain.SkipBlockID, dID darc.ID) (*darc.
...
+				if err != nil {
+					panic("DB is in bad state and cannot find skipchain anymore: " + err.Error())
+				}
+				// We assume the caller of this function is the
+				// current leader. So we generate a tree with
+				// this node as the root.
+				tree := sb.Roster.GenerateNaryTreeWithRoot(len(sb.Roster.List), s.ServerIdentity())
+				proto, err := s.CreateProtocol(collectTxProtocol, tree)
+				if err != nil {
+					panic("cannot create protocol: " + err.Error())
```
jeffallen: If someone, somehow, finds a way to make this happen, then the entire server will crash. There's got to be a better reaction than panic here., like maybe log.Error() and close(closeSignal).

omniledger/service/service.go
```diff
@@ -622,6 +638,37 @@ clientTransactions:
...
+	if err != nil {
+		return nil, err
+	}
+	if sb.Roster == nil || len(sb.Roster.List) < 1 {
+		return nil, errors.New("roster is empty")
+	}
+	return sb.Roster.List[0], nil
+}
+
+func (s *Service) getTxs(leader *network.ServerIdentity, scID skipchain.SkipBlockID) ClientTransactions {
```
jeffallen: Why does getTxs even care who the current leader is?

omniledger/service/service_test.go
```diff
@@ -65,6 +65,14 @@ func padDarc(key []byte) []byte {
...
 
 func TestService_AddTransaction(t *testing.T) {
+	testAddTransaction(t, 0)
+}
+
+func TestService_AddTransactionToFollower(t *testing.T) {
+	testAddTransaction(t, 1)
+}
+
+func testAddTransaction(t *testing.T, sendToIdx int) {
```
jeffallen: This is really nice and clear, I like.

omniledger/service/service.go
ineiti: So it can verify if it is the _real_ leader?

omniledger/service/collect_tx.go
kc1212: This is the intended behaviour when I was implementing it. We cannot guarantee every transaction that gets submitted to OL will eventually be included in the skipchain. It'll be even harder to have this guarantee when the leader starts to change (I don't know how to do it yet). But in this particular case, it should be possible to put the transactions back if there is an error.

omniledger/service/service.go
kc1212: The reason I decided to panic is that there is simply nothing we can do to recover without putting the server in a strange state. If you stop the go-routine than that would result in a non-polling leader. Then somebody has to go and fix the problem and restart the server.

skipchain/skipchain.go
```diff
@@ -821,11 +821,16 @@ func (s *Service) SetBFTTimeout(t time.Duration) {
...
 }
 
-// SetPropTimeout is used to set the propagation timeout
+// SetPropTimeout is used to set the propagation timeout.
 func (s *Service) SetPropTimeout(t time.Duration) {
 	s.propTimeout = t
 }
 
+// GetPropTimeout retrieves the current propagation timeout.
+func (s *Service) GetPropTimeout() time.Duration {
```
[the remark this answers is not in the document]
kc1212: I think `SetPropTimeout` is only meant to be used in tests, only in skipchain tests IIRC, so I didn't put a lock in `GetPropTimeout`.

omniledger/service/collect_tx.go
kc1212: Created an issue to track this problem https://github.com/dedis/cothority/issues/1337

omniledger/service/service.go
jeffallen: ok, good reasoning. Can you put it in the code, so that later when we say, "damn it, why is this panic here?" we don't forget?

kc1212:
Moving back to WIP because there seems to be a data race in collections...

kc1212:
Can't produce the race condition anymore... moving back to R4M, looks like there's already an issue tracking it https://github.com/dedis/cothority/issues/1317

omniledger/service/service.go
kc1212: sure, I updated it

PR #1330, opened 2018-07-03
Byzcoinx 1186

ineiti:
This is the optimized implementation of ByzCoinX. It returns early whenever the threshold is reached. This works very well in a permissioned setting, however, in an unpermissioned setting, it can lead to censoring of slow nodes which will not get any mining reward.

It has been quite extensively tested in all test-suites, but there might still be some corner cases where it fails.

### DIFF
--- a/byzcoinx/byzcoinx_test.go
+++ b/byzcoinx/byzcoinx_test.go
@@ -118,8 +118,7 @@ func TestBftCoSi(t *testing.T) {
 	require.Nil(t, err)
 
 	for _, n := range []int{1, 2, 4, 9, 20} {
-		nbrFault := FaultThreshold(n)
-		runProtocol(t, n, nbrFault, 0, protoName)
+		runProtocol(t, n, 0, 0, protoName)
 	}
 }
 
@@ -159,6 +158,8 @@ func TestBftCoSiFault(t *testing.T) {
 }
 
 func runProtocol(t *testing.T, nbrHosts int, nbrFault int, refuseIndex int, protoName string) {
+	log.Lvlf1("Starting with %d hosts with %d faulty ones and refusing at %d. Protocol name is %s",
+		nbrHosts, nbrFault, refuseIndex, protoName)
 	local := onet.NewLocalTest(testSuite)
 	defer local.CloseAll()
 
--- a/byzcoinx/byzcoinx_test.go
+++ b/byzcoinx/byzcoinx_test.go
@@ -118,7 +118,8 @@ func TestBftCoSi(t *testing.T) {
 	require.Nil(t, err)
 
 	for _, n := range []int{1, 2, 4, 9, 20} {
-		runProtocol(t, n, 0, 0, protoName)
+		nbrFault := FaultThreshold(n)
+		runProtocol(t, n, nbrFault, 0, protoName)
 	}
 }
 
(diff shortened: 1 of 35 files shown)


ftcosi/protocol/protocol.go
```diff
@@ -1,14 +1,17 @@
 // Package protocol is the fault tolerant cosi protocol implementation.
 //
 // For more information on the protocol, please see
-// https://github.com/dedis/cothority/blob/master/ftcosi/protocol/README.md.
+// https:// github.com/dedis/cothority/blob/master/ftcosi/protocol/README.md.
```
kc1212: don't need the space.

byzcoinx/byzcoinx_test.go
```diff
@@ -118,7 +118,8 @@ func TestBftCoSi(t *testing.T) {
 	require.Nil(t, err)
 
 	for _, n := range []int{1, 2, 4, 9, 20} {
-		runProtocol(t, n, 0, 0, protoName)
+		nbrFault := FaultThreshold(n)
+		runProtocol(t, n, nbrFault, 0, protoName)
```
kc1212: shouldn't this test check the scenario where we do not have any faults? because the faulty test is covered by `TestBftCoSiFault`

external/java/src/test/java/ch/epfl/dedis/lib/omniledger/contracts/EventLogInstanceTest.java
```diff
@@ -70,11 +71,24 @@ void testLogMore() throws Exception {
...
-            assertEquals(event, event2);
+        boolean allOK = true;
+        for (int i = 0; i < 4; i++) {
+            allOK = true;
+            Thread.sleep(2 * ol.getConfig().getBlockInterval().toMillis());
+            for (InstanceId key : keys) {
+                try {
+                    Event event2 = el.get(key);
+                    assertEquals(event, event2);
+                } catch (CothorityCryptoException e){
```
kc1212: why don't we just let the exception be thrown? the test will fail and i think that achieves the same effect.

ftcosi/protocol/protocol.go
```diff
@@ -127,72 +129,88 @@ func (p *FtCosi) Dispatch() error {
...
 	}
 
-	// signs the proposal
-	ok := <-verifyChan
-	if !ok {
-		// root should not fail the verification otherwise it would not have
-		// started the protocol
+	var secret kyber.Scalar
+	// verifies the proposal
+	verificationOk := <-verifyChan
```
kc1212: might be a good idea to add a timeout, just in case the verification function doesn't stop

ftcosi/protocol/protocol_test.go
```diff
@@ -99,27 +100,24 @@ func TestProtocol(t *testing.T) {
 	}
 }
 
-// Tests unresponsive leaves in various tree configurations
-func TestUnresponsiveLeafs(t *testing.T) {
-	nodes := []int{3, 13, 24}
-	subtrees := []int{1, 2}
+// Tests various trees configurations
+func TestProtocolQuickAnswer(t *testing.T) {
```
kc1212: how do we know these are actually the quick answers, is it possible to check how long it takes for the signature to arrive?

external/java/src/test/java/ch/epfl/dedis/lib/omniledger/contracts/EventLogInstanceTest.java
ineiti: Because we try four times to get all the proofs - so if we don't succeed in one of the first three rounds, we don't want the exception to be thrown.

ftcosi/protocol/protocol_test.go
ineiti: I count the bits in the signature and make sure it's less than the total number of nodes - that should hold up most of the time, except with very few nodes and a very fast server.

ftcosi/protocol/protocol.go
ineiti: added timeout while waiting for the verification.

Issue #1332, opened 2018-07-03
Version is shown as 2.0-$TAG-$DATE


jeffallen:
When tag is v2.0.1, which is the right format for vgo and for marking the Java and JS releases, then we get 2.x twice.

jeffallen:
Probably the real problem is that I set a v2.x.y tag, which is not allowed on the master branch or else gopkg.in does the wrong thing for gopkg.in/dedis/cothority.v2. I nuked the v2.0.1 tag, and will use the form $BRACH-$DATE (i.e. master-20180701) for publishing jar files for now.

So, I'm closing this until we see another way this is a problem.

PR #1333, opened 2018-07-04
Deploy version master-180704, switch to auto-release

jeffallen:
Deploy version master-180704, switch to auto-release since it seems to work well.

### DIFF
--- a/external/java/pom.xml
+++ b/external/java/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>ch.epfl.dedis</groupId>
     <artifactId>cothority</artifactId>
-    <version>2.0.2</version>
+    <version>master-180704</version>
     <name>Cothority client library for Java</name>
     <description>A client library for communicating with cothorities (a collective authority).</description>
     <url>https://github.com/dedis/cothority</url>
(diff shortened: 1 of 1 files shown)


kc1212:
Looking for a few clarification:
- [This](https://help.sonatype.com/repomanager2/staging-releases/configuring-your-project-for-deployment) is the guide that you followed to create a release, correct? 
- The reason for using `master-$date` is because we can't make go-releases with `v2.x.x` because versioning in go is hopeless https://github.com/dedis/cothority/issues/1332
- From what I can see here https://mvnrepository.com/artifact/ch.epfl.dedis/cothority we don't have a `master-180704`, should it be there?

jeffallen:
I followed the open source version of those instructions: https://central.sonatype.org/pages/ossrh-guide.html

The version thing is that v2.x.y tags should not be made on a branch that's not named v2. But we don't want this unstable java code checked into v2. Thus: $BRANCH-$DATE, at least until we are willing to put the Java stuff into v2.

mvnrepository is out of date: https://repo1.maven.org/maven2/ch/epfl/dedis/cothority/ has the new package. It also has the 2.0.x ones, which I have requested be deleted via a Jira ticket.

PR #1335, opened 2018-07-04
Allow timestamps a few seconds into the future

### LINKED ISSUE #1331: eventlog is too sensitive
If the client's clock is a few seconds before the server's the events are refused. Need to open up the window some.

jeffallen:
Fixes #1331.

Also: Remove forgotten LLvl2.

Also: Add missing clean to bindist, which avoids accidentally
getting a different version than expected if something is lingering
in exe/*.


ineiti:
Looks OK - although we might have to come back to this later and find a configurable way to set limits on the time-skew.

PR #1346, opened 2018-07-10
Include roster in ChainConfig

kc1212:
We add roster to ChainConfig, so that it is possible to change the
roster via the config contract. Changing the roster currently can only
be done via the client, we will add a way to rotate the roster
automatically (view-change) next.

### DIFF
--- a/omniledger/service/service.go
+++ b/omniledger/service/service.go
@@ -18,6 +18,7 @@ import (
 	"github.com/dedis/onet"
 	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
+	"github.com/dedis/protobuf"
 	"gopkg.in/satori/go.uuid.v1"
 )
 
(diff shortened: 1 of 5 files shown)


omniledger/service/service.go
```diff
@@ -581,8 +588,35 @@ func (s *Service) verifySkipBlock(newID []byte, newSB *skipchain.SkipBlock) bool
...
+	config, err := LoadConfigFromColl(&roCollection{collClone})
+	if err != nil {
+		log.Error(err)
+		return false
+	}
+	if !config.Roster.ID.Equal(newSB.Roster.ID) {
+		log.Error("rosters have unequal IDs")
+		return false
+	}
+	if len(config.Roster.List) != len(newSB.Roster.List) {
```
ineiti: Is this `if` and the following `for` not covered in `Roster.ID.Equal`?

kc1212: From looking at the code, `Roster.ID.Equal` only checks the uuid. It possible to use the same ID but a different list of server identities to "fool" the verification function if we don't check individual server identities. In this case, the ID will be wrong and an error might show up later, but I just thought it's a good idea to catch problems early.

PR #1349, opened 2018-07-12
Adding a field to AddTransaction to wait for the transaction to be included

ineiti:
Finally added a way in `AddTransaction` to tell OmniLedger to not return before the transaction is included.


omniledger/service/service.go
```diff
@@ -54,6 +54,10 @@ type Service struct {
 	// collections cannot be stored, so they will be re-created whenever the
 	// service reloads.
 	collectionDB map[string]*collectionDB
+	// holds a link for every omniledger to the latest block that is included in
+	// the collection.
+	// TODO: merge collectionDB and pollChan into olState structure.
+	state olState
```
kc1212: I'm not sure whether we need a separate field. Most of the fields in this struct are states anyway. Further, `olState` has a mutex on it. If everything is sharing the same mutex, that'd hinder performance.

omniledger/service/service.go
```diff
@@ -191,6 +195,44 @@ func (s *Service) AddTransaction(req *AddTxRequest) (*AddTxResponse, error) {
 
 	s.txBuffer.add(string(req.SkipchainID), req.Transaction)
 
+	if req.InclusionWait != nil {
```
kc1212: This is different from the `InclusionWait` documentation in `proto.go`. The documentation says if it's 0, then do nothing. But here it's check whether it's unset. Maybe use `req.InclusionWait != nil || req.InclusionWait > 0`. Alternatively, we can just change `InclusionWait` to a non-optional field, which defaults to 0 if not set. I prefer the latter a bit more because it as fewer moving parts.

omniledger/service/service.go
```diff
@@ -191,6 +195,44 @@ func (s *Service) AddTransaction(req *AddTxRequest) (*AddTxResponse, error) {
...
+				}
+				current = next
+			} else {
+				log.Lvl3("no new block")
+			}
+			*req.InclusionWait--
+		}
+		if !found {
+			return nil, errors.New("didn't find transaction in blocks")
+		}
```
kc1212: While this implementation works fine, we might be unmarshalling a lot of data, especially there are concurrent transactions. I feel it's more elegant to just maintain a map of channels that signals inclusion. Further, the client will know immediately when their block is included. Concretely, I'm imagining something like:
- on AddTransaction:
  - if InclusionWait > 0, create a channel and put it in an InclusionWait map 
  - wait on this channel
- on updateCollection:
  - we're unmarshalling `DataBody` anyway, so at that point we can check if any transaction are in the map
  - if we found one, then write to the channel
- on AddTransaction:
  - if a signal comes back, we know our transaction is included
  - if InclusionWait times out, return an error (the InclusionWait is in real-time instead of number of block intervals)

omniledger/service/service_test.go
```diff
@@ -198,14 +199,90 @@ func TestService_GetProof(t *testing.T) {
...
+	if wait > 0 {
+		w = &wait
+	}
+	ser := s.services[client]
+	_, err = ser.AddTransaction(&AddTxRequest{
+		Version:       CurrentVersion,
+		SkipchainID:   s.sb.SkipChainID(),
+		Transaction:   tx,
+		InclusionWait: w,
+	})
```
kc1212: this part can be replaced by `ser.SendTxTo` so that we don't repeat ourselves. But you'll have to change `ser.SendTxTo` to accept `InclusionWait`.

omniledger/service/service.go
ineiti: it's mostly read/write mutex for a shared resource - I don't think it'll be a big performance issue.

omniledger/service/service.go
ineiti: In fact the `dedis/protobuf` library handles pointers as optional parameters. But perhaps I should start using 'optional' in there...

ineiti: changed `proto.awk` to accept 
```
`protobuf="opt"`
```
as optional parameter.

omniledger/service/service.go
ineiti: Done.

omniledger/service/service.go
```diff
@@ -191,6 +195,23 @@ func (s *Service) AddTransaction(req *AddTxRequest) (*AddTxResponse, error) {
...
 	s.txBuffer.add(string(req.SkipchainID), req.Transaction)
 
+	if req.InclusionWait > 0 {
+		// Wait for InclusionWait new blocks and look if our transaction is in it.
+		interval, err := LoadBlockIntervalFromColl(s.GetCollectionView(req.SkipchainID))
+		if err != nil {
+			return nil, errors.New("couldn't get collectionView: " + err.Error())
+		}
+		ctHash := req.Transaction.Instructions.Hash()
+		ch := s.state.createWaitChannel(ctHash)
```
kc1212: We need to defer a `deleteWaitChannel`, otherwise the map only grows.

kc1212:
One small comment and then it's good.

PR #1350, opened 2018-07-12
OmniLedger view-change

kc1212:
The view-change logic is described in the README. The view-change
feature is disabled by default because it is not a robust implementation
yet (see the README for reasons). This is the first of many steps to
fully implement a robust view-change mechanism.

We've gotten a bit ahead of ourselves by implementing view-change before
doing any fault-tolerance tests on the non-leaders. But I started
view-change so might as well finish it. I propose the next step is to
make sure the children can fail and catch up before making view-change
more robust.

### DIFF
--- a/omniledger/contracts/value.go
+++ b/omniledger/contracts/value.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 
 	"github.com/dedis/cothority/omniledger/service"
+	"github.com/dedis/cothority/skipchain"
 )
 
 // The value contract can simply store a value in an instance and serves
@@ -19,7 +20,7 @@ var ContractValueID = "value"
 // It can spawn new value instances and will store the "value" argument in these
 // new instances.
 // Existing value instances can be "update"d and deleted.
-func ContractValue(cdb service.CollectionView, inst service.Instruction, c []service.Coin) ([]service.StateChange, []service.Coin, error) {
+func ContractValue(cdb service.CollectionView, scID skipchain.SkipBlockID, inst service.Instruction, c []service.Coin) ([]service.StateChange, []service.Coin, error) {
 	switch {
 	case inst.Spawn != nil:
 		return []service.StateChange{
--- a/omniledger/contracts/value.go
+++ b/omniledger/contracts/value.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 
 	"github.com/dedis/cothority/omniledger/service"
-	"github.com/dedis/cothority/skipchain"
 )
 
 // The value contract can simply store a value in an instance and serves
@@ -20,7 +19,7 @@ var ContractValueID = "value"
 // It can spawn new value instances and will store the "value" argument in these
 // new instances.
 // Existing value instances can be "update"d and deleted.
-func ContractValue(cdb service.CollectionView, scID skipchain.SkipBlockID, inst service.Instruction, c []service.Coin) ([]service.StateChange, []service.Coin, error) {
+func ContractValue(cdb service.CollectionView, inst service.Instruction, c []service.Coin) ([]service.StateChange, []service.Coin, error) {
 	switch {
 	case inst.Spawn != nil:
 		return []service.StateChange{
--- a/omniledger/service/contracts.go
+++ b/omniledger/service/contracts.go
@@ -3,10 +3,12 @@ package service
 import (
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/dedis/cothority"
 	"github.com/dedis/cothority/omniledger/darc"
+	"github.com/dedis/cothority/skipchain"
 	"github.com/dedis/onet"
 	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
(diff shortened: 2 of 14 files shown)


omniledger/service/heartbeat.go
```diff
@@ -0,0 +1,115 @@
+package service
+
+import (
+	"errors"
+	"sync"
+	"time"
+)
+
+type heartbeat struct {
```
jeffallen: A 3-5 line comment here explaining the general concept, and what problem it solves for us would be good here. (But no other comments needed in this file, it's clear enough other than "why does it exist?")

omniledger/service/service.go
```diff
@@ -776,20 +811,168 @@ func (s *Service) getTxs(leader *network.ServerIdentity, scID skipchain.SkipBloc
...
+	if s.heartbeats.enabled() {
+		s.heartbeats.closeAll()
+		s.heartbeatsClose <- true
 	}
 	s.pollChanWG.Wait()
 }
 
+func (s *Service) heartbeatTimeoutMonitor() {
+	go func() {
+		// empty the messages in heartbeatsClose
```
jeffallen: Why? (I have a guess it is related to running multiple times in tests, but this needs a clear explanation in the code.)

omniledger/service/service.go
```diff
@@ -776,20 +811,168 @@ func (s *Service) getTxs(leader *network.ServerIdentity, scID skipchain.SkipBloc
...
+	}
 	return s.txBuffer.take(string(scID))
 }
 
 // TestClose closes the go-routines that are polling for transactions.
 func (s *Service) TestClose() {
-	for _, c := range s.pollChan {
+	for k, c := range s.pollChan {
 		close(c)
+		delete(s.pollChan, k)
```
jeffallen: why are we not holding s.pollChanMut here?

omniledger/service/service.go
```diff
@@ -776,20 +811,168 @@ func (s *Service) getTxs(leader *network.ServerIdentity, scID skipchain.SkipBloc
...
+	}
+	if s.heartbeats.enabled() {
+		s.heartbeats.closeAll()
+		s.heartbeatsClose <- true
 	}
 	s.pollChanWG.Wait()
 }
 
+func (s *Service) heartbeatTimeoutMonitor() {
```
jeffallen: Maybe call it "monitorForLeaderFailure" or something. The heartbeat mechanism is how it gets done, but the reader of this code wants to know what is getting done.

omniledger/service/service.go
```diff
@@ -776,20 +811,168 @@ func (s *Service) getTxs(leader *network.ServerIdentity, scID skipchain.SkipBloc
...
+	}
+
+	// find the position in the proposer queue
+	latestConfig, err := s.LoadConfig(scID)
+	if err != nil {
+		return err
+	}
+	pos := func() int {
+		var ctr int
+		for _, pk := range latestConfig.Roster.Publics() {
```
jeffallen: [roster.Search](https://godoc.org/github.com/dedis/onet#Roster.Search) maybe?

omniledger/service/service.go
```diff
@@ -891,9 +1084,11 @@ func (s *Service) save() {
 // deployments.
 func newService(c *onet.Context) (onet.Service, error) {
 	s := &Service{
-		ServiceProcessor: onet.NewServiceProcessor(c),
-		contracts:        make(map[string]OmniLedgerContract),
-		txBuffer:         newTxBuffer(),
+		ServiceProcessor:  onet.NewServiceProcessor(c),
+		contracts:         make(map[string]OmniLedgerContract),
+		txBuffer:          newTxBuffer(),
+		heartbeatsTimeout: make(chan string, 1),
```
jeffallen: I expected to find heartbeats: newHeartbeats() here, because there are lots of calls to "if s.heatbeats.enabled()" that seem like they will fail if s.heartbeats is left nil. It is hard to see how this works when EnableViewChange() has not been called.

omniledger/service/service_test.go
```diff
@@ -680,6 +679,57 @@ func TestService_SetBadConfig(t *testing.T) {
...
+	// check the leader for all nodes
+	for _, service := range s.services[1:] {
+		// everyone should have the same leader after the genesis block is stored
+		leader, err := service.getLeader(s.sb.SkipChainID())
+		require.NoError(t, err)
+		require.NotNil(t, leader)
+		require.True(t, leader.Equal(s.services[1].ServerIdentity()))
+	}
+
+	// try to send a transaction to the node on index 2 (not the current leader)
```
jeffallen: not or now? So if I understand this, this is a test of "server 0 was leader, now is down, server 1 is now leader, so send tx to server 2 and see that server 1 polls server 2, finds the tx, and that we can talk to server 2 and see the tx applied".

This is an excellent end to end test, and needs better comment to point that out.

Also, it would be more interesting to ask server 3 for the inclusion proof, showing that the tx was propagated from the new leader to a follower who never heard anything about the tx.

omniledger/service/struct.go
```diff
@@ -179,9 +178,10 @@ func (c *collectionDB) Store(t *StateChange) error {
 }
 
 func (c *collectionDB) GetValueContract(key []byte) ([]byte, []byte, error) {
-	return getValueContract(&roCollection{c.coll}, key)
+	return getValueContract(c, key)
 }
 
+// TODO this function can be merged with getValuesFromRecord
```
jeffallen: Could you make an issue for this, and reference it here, and add label assistant, so we have useful jobs to give the assistants? Thx.

skipchain/skipchain.go
```diff
@@ -855,6 +894,16 @@ func (s *Service) forwardLinkLevel0(src, dst *SkipBlock) error {
 
 	// create the message we want to sign for this round
 	roster := src.Roster
+
+	if enableViewChange {
```
jeffallen: This is a really big deal, and needs a nice big comment explaining it, I think. What are the trade-offs here? What is the security risk that skipchain allows this? etc. (I think this is the right implementation, I just think a lot of reasoning is locked in your head that needs to be in comments.)

skipchain/skipchain.go
```diff
@@ -855,6 +894,16 @@ func (s *Service) forwardLinkLevel0(src, dst *SkipBlock) error {
...
+		// If the server identities in the two rosters are the same,
+		// then it might be a view-change, so we use the second roster
+		// with the new leader.
+		if isRotation(src.Roster, dst.Roster) {
```
jeffallen: I wonder if this should be a method on type onet.Roster? So that both sk and ol can use it and not duplicate code?

skipchain/struct.go
```diff
@@ -550,6 +550,22 @@ func (fl *ForwardLink) Verify(suite cosi.Suite, pubs []kyber.Point) error {
 	if bytes.Compare(fl.Signature.Msg, fl.Hash()) != 0 {
 		return errors.New("wrong hash of forward link")
 	}
+	if enableViewChange {
```
jeffallen: Yeah, definitely more commented needed. :)

omniledger/contracts/value.go
```diff
@@ -19,7 +20,7 @@ var ContractValueID = "value"
 // It can spawn new value instances and will store the "value" argument in these
 // new instances.
 // Existing value instances can be "update"d and deleted.
-func ContractValue(cdb service.CollectionView, inst service.Instruction, c []service.Coin) ([]service.StateChange, []service.Coin, error) {
+func ContractValue(cdb service.CollectionView, scID skipchain.SkipBlockID, inst service.Instruction, c []service.Coin) ([]service.StateChange, []service.Coin, error) {
```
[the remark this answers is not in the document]
kc1212: That's how I started doing it, but there are some places where we create `service.CollectionView` on the fly from a regular collection. In those cases, it's not super obvious what the `scID` should be. I will have another look at this option after I've addressed your other comments.

omniledger/service/contracts.go
```diff
@@ -113,38 +115,112 @@ func LoadDarcFromColl(coll CollectionView, key []byte) (*darc.Darc, error) {
...
+func (s *Service) invokeContractConfig(cdb CollectionView, scID skipchain.SkipBlockID, inst Instruction, coins []Coin) (sc []StateChange, c []Coin, err error) {
 	c = coins
-	if inst.Invoke.Command != "update_config" {
+	// There are two situations where we need to change the roster, first
+	// is when it is initiated by the client(s) that holds the genesis
+	// signing key, in thise case we trust the client to do the right
+	// thing. The second is during a view-change, so we need to do
+	// additional validation to make sure a malicious node doesn't freely
+	// change the roster.
+	if inst.Invoke.Command == "update_config" {
```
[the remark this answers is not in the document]
kc1212: nice catch - there was an extra check outside of the contract that makes sure the new roster has the same length as the old one. it shouldn't be there in the first place because genesis darc holders can modify the config in any way they want.

omniledger/service/service.go
kc1212: because the caller of `TestClose` holds the lock - see `tryLoad`

omniledger/service/service.go
kc1212: can't use `roster.Search` because it takes server identities instead of public keys

omniledger/service/service.go
kc1212: `s.heartbeats` is a value not a reference. so it cannot be nil. the map inside it can be nil, which is what I use to check if it has been initialised.

omniledger/service/service_test.go
kc1212: you're understanding is spot-on. I will add comments to make the test understandable and do the inclusion proof for index 3.

skipchain/skipchain.go
kc1212: made an issue for that, should be able to fix it quickly in a future PR https://github.com/dedis/cothority/issues/1362

omniledger/service/struct.go
kc1212: done - https://github.com/dedis/cothority/issues/1363

skipchain/skipchain.go
kc1212: added a comment in the `enableViewChange` declaration

kc1212:
Created https://github.com/dedis/cothority/issues/1364 which will be the next step.

PR #1351, opened 2018-07-12
avoid flooding of OmniLedger

ineiti:
This is a first proposition to make OmniLedger avoid flooding of its chain.

### DIFF
--- a/omniledger/service/service.go
+++ b/omniledger/service/service.go
@@ -373,7 +373,7 @@ func (s *Service) createNewBlock(scID skipchain.SkipBlockID, r *onet.Roster, cts
 		NewBlock:          sb,
 		TargetSkipChainID: scID,
 	}
-	log.Lvlf2("Storing skipblock with transactions %+v", ctsOK)
+	log.Lvlf3("Storing skipblock with transactions %+v", ctsOK)
 	ssbReply, err := s.skService().StoreSkipBlock(&ssb)
 	if err != nil {
 		return nil, err
@@ -430,11 +430,7 @@ func (s *Service) updateCollection(msg network.Message) {
 		return
 	}
 
-	if i, _ := sb.Roster.Search(s.ServerIdentity().ID); i == 0 {
-		log.Lvlf2("%s: Storing %d state changes %v", s.ServerIdentity(), len(scs), scs.ShortStrings())
-	} else {
-		log.Lvlf3("%s: Storing %d state changes %v", s.ServerIdentity(), len(scs), scs.ShortStrings())
-	}
+	log.Lvlf3("%s: Storing %d state changes %v", s.ServerIdentity(), len(scs), scs.ShortStrings())
 	for _, sc := range scs {
 		err = cdb.Store(&sc)
 		if err != nil {
@@ -517,6 +513,7 @@ func (s *Service) startPolling(scID skipchain.SkipBlockID, interval time.Duratio
 	closeSignal := make(chan bool)
 	go func() {
 		defer s.pollChanWG.Done()
+		var txs ClientTransactions
 		for {
 			select {
 			case <-time.After(interval):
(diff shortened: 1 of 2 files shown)


omniledger/service/service.go
```diff
@@ -571,7 +567,34 @@ func (s *Service) startPolling(scID skipchain.SkipBlockID, interval time.Duratio
 					continue
 				}
 
-				_, err = s.createNewBlock(scID, sb.Roster, txs)
+				// Pre-run transactions to look how many we can fit in the alloted time
+				// slot. Perhaps we can run this in parallel during the wait-phase?
```
kc1212: If contracts are the main things that take a long time to execute, is it not possible to put this kind of logic in `createStateChange` instead of computing all the state changes twice? perhaps we give `createStateChange` a time budge (maybe also a size budget), then it'll output the state changes as well as the transactions that haven't been processed, but we'll need to store these transactions elsewhere. we might be also unnecessarily computing these state changes too many times, for example, during `updateCollection`, we compute them again on the leader. maybe storing the state changes that we know is correct in a cache and then applying it will allow us to add more transactions.

ineiti: OK, talking about optimizations...

To list the number of executions of the `ClientTransactions` (including this PR):
1. `startPolling` to know how long the execution will take
2. `createStateChanges` called form `createNewBlock`
3. `createStateChanges` called from `updateCollection`

2. and 3. are easy to merge into one single call to `createStateChanges` by adding a cache (or whatever you want to call it)
1. is not so easy to merge, because the order in 1. is different from the order in 2. and 3., because the instructions are re-ordered.

So one thing we might be doing is removing the re-ordering (as Lefteris said the protection against censorship there is minimal) and then:
1. execute all `ClientInstructions` in `startPolling` and store the results, making sure the total duration is not more than the `interval`. Store all `StateChange`s in a cache (or whatever)
2. `createNewBlock` then only uses this cache and stores a version of the new collection in the cache
3. `updateCollection` copies the new collection in the cache to the 'official' collection

This is to optimize the CPU-usage. 

However, we also have a bandwidth restriction, because the blocks get big, at least for simple transactions like coin-transfers (for onchain-secret writes we're CPU-bound). For this, we need to:
- avoid sending the block around multiple times in `skipchain.startPropagation` - the block is already known from the byzcoinx call and only the forwardlink needs to be propagated.

And finally some network delay dependency because of too many round-trips:
- remove the propagation of `updateCollection` by creating a hook in the skipchain-service that will call omniledger if a new block has been accepted.

So when all is said, in this PR I could add at least the first part for the CPU dependency:
1. remove the re-ordering
2. `startPolling` stores the state-changes in a cache
3. `createNewBlock` stores a copy of the final state
4. `updateCollection` uses the stored final state

But then it gets a huge PR... If you think this is a good plan, I propose to keep this PR like that and then go on with the rest of the plan.

kc1212: sounds like a good plan, let's keep the PRs small too. I've already lost track of the different things I changed in the view-change PR.

omniledger/service/service.go
```diff
@@ -571,7 +567,34 @@ func (s *Service) startPolling(scID skipchain.SkipBlockID, interval time.Duratio
...
+				var txsCollect ClientTransactions
+				cdbI := s.GetCollectionView(scID)
+				now := time.Now()
+				for len(txs) > 0 {
+					if err := s.verifyClientTx(scID, txs[0]); err == nil {
```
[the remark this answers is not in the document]
ineiti: This is now in #1380

omniledger/service/service.go
```diff
@@ -661,40 +684,15 @@ func (s *Service) createStateChanges(coll *collection.Collection, cts ClientTran
 	// we could use some kind of copy-on-write technique.
 
 	cdbTemp := coll.Clone()
+	var cin []Coin
 clientTransactions:
```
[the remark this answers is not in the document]
ineiti: It seems to be used - at least in the latest code...

PR #1355, opened 2018-07-17
adding a coin contract together with the simulation

ineiti:
This is the follow-up of #1323 

This adds a coin-contract with the following possibilities:
- `Spawn` to create a new account with a 0-balance
- `Invoke:transfer` to transfer coins from one to another account
- `Invoke:mint` to create new coins (must be protected by a darc)

In addition to this, a simulation allows for testing the performance of the contract.

### DIFF
--- a/omniledger/contracts/coins.go
+++ b/omniledger/contracts/coins.go
@@ -1,0 +1,125 @@
+package contracts
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+
+	"github.com/dedis/cothority/omniledger/service"
+	"github.com/dedis/onet/log"
+)
+
+// ContractCoinID denotes a contract that can store and transfer coins.
+var ContractCoinID = "coin"
+
+var olCoin = service.NewInstanceID([]byte("olCoin"))
+
+// ContractCoin is a coin implementation that holds one instance per coin.
+// If you spawn a new ContractCoin, it will create an account with a value
+// of 0 coins.
+// The following methods are available:
+//  - mint will add the number of coins in the argument "coins" to the
+//    current coin instance. The argument must be a 64-bit uint in LittleEndian
+//  - transfer will send the coins given in the argument "coins" to the
+//    instance given in the argument "destination". The "coins"-argument must
+//    be a 64-bit uint in LittleEndian. The "destination" must be a 64-bit
+//    instanceID
+//  - fetch takes "coins" out of the account and returns it as an output
+//    parameter for the next instruction to interpret.
+//  - store puts the coins given to the instance back into the account.
+// You can only delete a contractCoin instance if the account is empty.
+func ContractCoin(cdb service.CollectionView, inst service.Instruction, c []service.Coin) (sc []service.StateChange, cOut []service.Coin, err error) {
+	cOut = c
+	switch {
+	case inst.Spawn != nil:
+		// Spawn creates a new coin account as a separate instance. The subID is
+		// taken from the hash of the instruction.
+		ca := service.InstanceID{
+			DarcID: inst.InstanceID.DarcID,
+			SubID:  service.NewSubID(inst.Hash()),
+		}
+		log.Lvlf3("Spawing coin to %x", ca.Slice())
+		return []service.StateChange{
+			service.NewStateChange(service.Create, ca, ContractCoinID, make([]byte, 8)),
+		}, c, nil
+	case inst.Invoke != nil:
+		// Invoke is one of "mint", "transfer", "fetch", or "store".
+		value, _, err := cdb.GetValues(inst.InstanceID.Slice())
+		if err != nil {
+			return nil, nil, err
+		}
+		coinsCurrent := binary.LittleEndian.Uint64(value)
+		coinsBuf := inst.Invoke.Args.Search("coins")
+		if coinsBuf == nil {
+			return nil, nil, errors.New("please give coins")
+		}
+		coinsArg := binary.LittleEndian.Uint64(coinsBuf)
+		var sc []service.StateChange
+		switch inst.Invoke.Command {
+		case "mint":
+			// mint simply adds this amount of coins to the account.
+			log.Lvl2("minting", coinsArg)
+			coinsCurrent += coinsArg
+		case "transfer":
+			// transfer sends a given amount of coins to another account.
+			if coinsArg > coinsCurrent {
+				return nil, nil, errors.New("not enough coins in instance")
+			}
+			coinsCurrent -= coinsArg
+			target := inst.Invoke.Args.Search("destination")
+			v, cid, err := cdb.GetValues(target)
+			if err == nil && cid != ContractCoinID {
+				err = errors.New("destination is not a coin contract")
+			}
+			if err != nil {
+				return nil, nil, err
+			}
+			targetCoin := binary.LittleEndian.Uint64(v)
+			var w bytes.Buffer
+			binary.Write(&w, binary.LittleEndian, targetCoin+coinsArg)
+			log.Lvlf3("transferring %d to %x", coinsArg, target)
+			sc = append(sc, service.NewStateChange(service.Update, service.NewInstanceID(target),
+				ContractCoinID, w.Bytes()))
+		case "fetch":
+			// fetch removes coins from the account and passes it on to the next
+			// instruction.
+			if coinsArg > coinsCurrent {
+				return nil, nil, errors.New("not enough coins in instance")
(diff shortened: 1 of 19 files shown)


omniledger/contracts/coins.go
```diff
@@ -0,0 +1,124 @@
...
+//    be a 64-bit uint in LittleEndian. The "destination" must be a 64-bit
+//    instanceID
+//  - fetch takes "coins" out of the account and returns it as an output
+//    parameter for the next instruction to interpret.
+//  - store puts the coins given to the instance back into the account.
+// You can only delete a contractCoin instance if the account is empty.
+func ContractCoin(cdb service.CollectionView, inst service.Instruction, c []service.Coin) (sc []service.StateChange, cOut []service.Coin, err error) {
+	cOut = c
+	switch {
+	case inst.Spawn != nil:
```
kc1212: we should use the helper functions `inst.GetType() == SpawnType` instead, this will catch errors when more than 1 of spawn/invoke/delete is set.

omniledger/contracts/coins.go
```diff
@@ -0,0 +1,124 @@
...
+		if coinsBuf == nil {
+			return nil, nil, errors.New("please give coins")
+		}
+		coinsArg := binary.LittleEndian.Uint64(coinsBuf)
+		var sc []service.StateChange
+		switch inst.Invoke.Command {
+		case "mint":
+			// mint simply adds this amount of coins to the account.
+			log.Lvl2("minting", coinsArg)
+			coinsCurrent += coinsArg
```
kc1212: this should not overflow, there needs to be a way to safely mint coins because the initial supply might very well near the limit of uint64. for example, ERC20 token contracts almost always checks for overflows.

omniledger/contracts/coins.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		coinsCurrent := binary.LittleEndian.Uint64(value)
+		coinsBuf := inst.Invoke.Args.Search("coins")
+		if coinsBuf == nil {
+			return nil, nil, errors.New("please give coins")
+		}
+		coinsArg := binary.LittleEndian.Uint64(coinsBuf)
+		var sc []service.StateChange
+		switch inst.Invoke.Command {
+		case "mint":
```
kc1212: I feel we're missing some tests regarding the mint permission. usually, only a few identities are allowed to mint, the regular accounts are only allowed to transfer, fetch and store. it'll be nice to have these tests, not only to test the functionality, but how the permissions should be configured for a real coin contract

omniledger/contracts/coins.go
```diff
@@ -0,0 +1,124 @@
...
+				return nil, nil, errors.New("not enough coins in instance")
+			}
+			coinsCurrent -= coinsArg
+			cOut = append(cOut, service.Coin{Name: olCoin, Value: coinsArg})
+		case "store":
+			// store moves all coins from the last instruction into the account.
+			cOut = []service.Coin{}
+			for _, co := range c {
+				if co.Name.Equal(olCoin) {
+					coinsCurrent += co.Value
```
kc1212: check for overflow

omniledger/service/service_test.go
```diff
@@ -254,19 +255,19 @@ func TestService_FloodLedger(t *testing.T) {
 	before := reply.Update[len(reply.Update)-1]
 
 	// Create a transaction without waiting
```
kc1212: comment should say "Create 10 transactions without waiting", or just remove it

omniledger/contracts/coins.go
```diff
@@ -0,0 +1,124 @@
...
+			sc = append(sc, service.NewStateChange(service.Update, service.NewInstanceID(target),
+				ContractCoinID, w.Bytes()))
+		case "fetch":
+			// fetch removes coins from the account and passes it on to the next
+			// instruction.
+			if coinsArg > coinsCurrent {
+				return nil, nil, errors.New("not enough coins in instance")
+			}
+			coinsCurrent -= coinsArg
+			cOut = append(cOut, service.Coin{Name: olCoin, Value: coinsArg})
```
kc1212: what happens if there is not store after fetch? does the coin get lost? is it ok for it to be lost?

omniledger/contracts/coins_test.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+	}
+	sc, co, err = ContractCoin(ct, inst, []service.Coin{})
+	require.Nil(t, err)
+	require.Equal(t, 0, len(co))
+	require.Equal(t, 2, len(sc))
+	require.Equal(t, service.NewStateChange(service.Update, coAddr2, ContractCoinID, coinOne), sc[0])
+	require.Equal(t, service.NewStateChange(service.Update, coAddr1, ContractCoinID, coinZero), sc[1])
+}
+
```
kc1212: we're missing tests for fetch and store.

omniledger/contracts/coins.go
jeffallen: Caveat emptor?

omniledger/contracts/coins_test.go
jeffallen: ...but we were not missing bugs!

omniledger/contracts/coins.go
```diff
@@ -28,97 +56,142 @@ var olCoin = service.NewInstanceID([]byte("olCoin"))
...
-			service.NewStateChange(service.Remove, inst.InstanceID, ContractCoinID, nil),
-		}, c, nil
+		return
 	}
-	return nil, nil, errors.New("didn't find any instruction")
+	err = errors.New("instruction type not allowed")
+	return
+}
+
+// iid uses sha256(in)x2 to get 64 bytes from in, and makes an InstanceID from it.
```
kc1212: `sha256(in)x2` means `sha256(sha256(in))`? I don't think the code is doing that

omniledger/contracts/coins_test.go
```diff
@@ -0,0 +1,197 @@
...
+	inst = omniledger.Instruction{
+		InstanceID: coAddr,
+		Invoke: &omniledger.Invoke{
+			Command: "fetch",
+			Args:    omniledger.Arguments{{Name: "coins", Value: coinOne}},
+		},
+	}
+
+	// Try once with no enough coins avail.
```
kc1212: "Try once without enough coins."?

omniledger/contracts/coins.go
```diff
@@ -0,0 +1,197 @@
...
+			// instruction.
+			if coinsArg > uint64(coinsCurrent) {
+				err = errors.New("not enough coins in instance")
+				return
+			}
+			coinsCurrent, err = coinsCurrent.sub(coinsArg)
+			if err != nil {
+				return
+			}
+			cOut = append(cOut, omniledger.Coin{Name: CoinName, Value: coinsArg})
```
kc1212: I don't understand why coins needs a name. From my understanding, we should enforce that store/fetch work only in the same instance (not sure how to do it).

kc1212:
@jeffallen branch still have some conflicts, otherwise i'm happy to merge.

omniledger/service/service.go
```diff
@@ -250,6 +251,7 @@ func (s *Service) SetPropagationTimeout(p time.Duration) {
 	s.storage.PropTimeout = p
 	s.storage.Unlock()
 	s.save()
+	s.skService().SetBFTTimeout(p)
```
kc1212: I'm not sure whether this should be here. The BFT timeout is different from the propagation timeout. The BFT timeout is like a time budget, where every sub-protocol in ByzCoinX will use some of the budget. The propagation timeout is always from the `message.Propagate` protocol. I think these two should be set independently.

omniledger/service/service.go
```diff
@@ -800,25 +808,19 @@ func (s *Service) registerContract(contractID string, c OmniLedgerContract) erro
 // Tries to load the configuration and updates the data in the service
 // if it finds a valid config-file.
 func (s *Service) tryLoad() error {
-	s.storage = &storage{
-		// Set the default to be the same as the default in
-		// skipchain/struct.go, which is 15 seconds.
-		PropTimeout: 15 * time.Second,
-	}
+	s.SetPropagationTimeout(120 * time.Second)
```
kc1212: I guess this is needed because we set the BFT timeout in `SetPropagationTimeout`. This will significantly slow down message propagation when there is a failure. My suggestion is that we keep it as is and set the two timeouts independently. There should be a relationship between the two timeouts but we don't exactly know what it is at the moment.

PR #1357, opened 2018-07-18
follow onet TestClose

ineiti:
Instead of having to call Close in all tests, use the new TestClose interface of onet


jeffallen:
Yay for red lines!

PR #1371, opened 2018-07-30
Store state changes in a batch

### LINKED ISSUE #1370: collectionDB.Store is extremely slow in Jenkins
`collectionDB.Store` sometimes takes about 90ms to store one state change, which is extremely slow and is causing the coin simulation to fail. Strangely, this behaviour is only observed in the Jenkins server and cannot be reproduced locally or on Travis. A workaround is to use a single boltdb transaction to store all the state changes.

kc1212:
- We have a lock on collection, so no need more locks in node.
- Closes #1370

### DIFF
--- a/omniledger/collection/node.go
+++ b/omniledger/collection/node.go
@@ -2,12 +2,10 @@ package collection
 
 import (
 	"crypto/sha256"
-	"sync"
 )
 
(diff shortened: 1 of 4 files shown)


kc1212:
Jenkins is confirmed to pass with this fix.

PR #1376, opened 2018-07-31
Remove getValueContract code duplication

### LINKED ISSUE #1363: Code duplication around getting values from collections
There are a couple of places where we have code duplication, e.g., `getValueContract` and `GetValues` are quite similar in https://github.com/dedis/cothority/blob/master/omniledger/service/struct.go. Consider merging these to one function.

kc1212:
Closes #1363

Further, we change collectionDb and CollectionView API to return `[]byte` for contract ID, because the lower-level API do not know about how the application is going to interpret the data, so it should not do any conversion and let the high level code handle it.

### DIFF
--- a/omniledger/service/struct.go
+++ b/omniledger/service/struct.go
@@ -53,11 +53,7 @@ func (r *roCollection) Get(key []byte) collection.Getter {
 // GetValues returns the value of the key and the contractID. If the key
 // does not exist, it returns an error.
 func (r *roCollection) GetValues(key []byte) (value []byte, contractID string, err error) {
-	record, err := r.c.Get(key).Record()
-	if err != nil {
-		return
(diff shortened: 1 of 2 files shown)


ineiti:
I'm not sure I understand why you're saying that we should keep it all `[]byte` for the `contractID`. In my understanding, if we give the `contractID` as `string` to the collection, that can be part of the 'contract' of the interface, no?

I propose we keep `string` for `contractID`.

kc1212:
After thinking a bit more, your proposal makes more sense. One of the reasons for changing to `[]byte` is because the public function `GetValueContract` returns `[]byte` for contract ID. But this function is very similar to `GetValues`, except the return type. So I've removed `GetValueContract`.

PR #1377, opened 2018-07-31
Use one DKG

### LINKED ISSUE #1256: Use one DKG
We have two DKG protocols, one in OCS and the other in evoting. They do the same thing, so we should combine them.

kc1212:
Before we had two very similar DKG implementations. So merge them into
one.

NOTE: this PR changes public API.

Closes #1256


ineiti:
1. Why did you create a sub-directory `protocol` in `dkg` if there is nothing else? Do you plan on adding other things there later? If not, I propose to move all things from `protocol` directly into `dkg`. Or we create a `distributed` directork and put `dkg` and `ocs` in there
2. for changing of public API: as far as I see, there is no change in anything that is in `stable/directories`, correct?

ineiti:
Last question: I have no memory if the evoting- or the ocs- version is more up-to-date. Did you have a look at the difference between the two? I remember having worked on one or both of the versions to remove some errors and would like to make sure that all improvements are merged in the 'official' version.

kc1212:
Yes I checked the diff.  Other than some naming differences, the only thing was the addition of `SharedSecret` type and `NewSharedSecret` in `dkg_struct.go`, in the ocs DKG. So I used the OCS dkg as the "official" version.

ineiti:
git-diff says this:
```
evoting/protocol/dkg.go → dkg/dkg.go
```
so does it think that the evoting-dkg is closer to the final version?

kc1212:
You can try to have a look at the result of `diff evoting/protocol/dkg_struct.go ocs/protocol/dkg_struct.go`, and `diff evoting/protocol/dkg.go ocs/protocol/dkg.go` in master. The only differences are the ones I mentioned above except naming differences. I think `git-diff` thinks the evoting one was used might be because I changed some struct names from `InitDKG` to `Init`, for example, which is the same as the names in evoting.

PR #1378, opened 2018-08-06
jjAdd state change cache

kc1212:
We re-compute state changes three times, one during the initial block
creation, one during block verification and finally one during
updateCollection. This PR introduces a cache to store successful state
change computation so that conodes spends less time running contracts.

### DIFF
--- a/omniledger/service/service.go
+++ b/omniledger/service/service.go
@@ -4,12 +4,11 @@ package service
 import (
 	"bytes"
 	"encoding/binary"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"sync"
 	"time"
-
-	"encoding/hex"
 
 	"github.com/dedis/cothority"
 	"github.com/dedis/cothority/messaging"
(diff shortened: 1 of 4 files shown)


kc1212:
I left the timing code in `startPolling` as is because @jeffallen most likely is replacing that with a different version of `createStateChanges`.

ineiti:
If I understand correctly:
1. this is not actually used yet in the service
2. the removal of the cache is missing

The second point probably comes when the block has been accepted, is that correct? But we'll also need a timeout.

kc1212:
1. It is used, just not in `startPolling` because it doesn't call `createStateChanges`. So the leader will compute the state changes twice, once in `startPolling` and the other in `createNewBlock`, instead of 4 times. Children only compute the state changes once instead of 3 times.
2. I implemented a first version with cache removal, but it was not nice and introduces more maintenance overhead. Namely, if `createStateChanges` succeeds and we add an entry to the cache, then there is no guarantee that it will be removed at `updateCollection` because something might fail in between. We'll have to either remove the cache at every place that returns an error after a successful `createStateChanges` (difficult to maintain) or, like you suggested, add a timeout (we'll need go-routines so we'll have to close them properly too). My feeling is that the current solution is cleaner. It uses more memory, but it does not grow per skipchain, so it's not going to cause a memory leak.

PR #1387, opened 2018-08-08
Add StoreSkipBlock callback

kc1212:
We add a callback, which is executed everytime when a skipblock gets
stored. This feature is to reduce the number of round trips in
omniledger needed to store and update its state.

### DIFF
--- a/omniledger/service/service.go
+++ b/omniledger/service/service.go
@@ -41,7 +41,7 @@ func init() {
 	var err error
 	OmniledgerID, err = onet.RegisterNewService(ServiceName, newService)
 	log.ErrFatal(err)
-	network.RegisterMessages(&omniStorage{}, &DataHeader{}, &updateCollection{})
+	network.RegisterMessages(&omniStorage{}, &DataHeader{})
 }
 
 // GenNonce returns a random nonce.
@@ -109,10 +109,6 @@ type omniStorage struct {
 	PropTimeout time.Duration
 
 	sync.Mutex
-}
-
-type updateCollection struct {
-	ID skipchain.SkipBlockID
 }
 
 // CreateGenesisBlock asks the service to create a new skipchain ready to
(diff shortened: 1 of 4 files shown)


omniledger/service/service.go
```diff
@@ -407,63 +403,57 @@ func (s *Service) createNewBlock(scID skipchain.SkipBlockID, r *onet.Roster, cts
...
 
-// updateCollection is called once a skipblock has been stored.
-// It is called by the leader, and every node will add the
-// transactions in the block to its collection.
-func (s *Service) updateCollection(msg network.Message) {
-	uc, ok := msg.(*updateCollection)
-	if !ok {
-		log.Error("updateCollection cast failure")
-		return
+// updateCollectionCB is registered in skipchain and is called after a
```
ineiti: s/updateColletionCB/updateCollectionCallback/

omniledger/service/service.go
```diff
@@ -407,63 +403,57 @@ func (s *Service) createNewBlock(scID skipchain.SkipBlockID, r *onet.Roster, cts
...
+// skipblock is updated. When this function is called, it is not always after
+// the addition of a new block, but an updates to forward links, for example.
+// Hence, we need to figure out when a new block is added. This can be done by
+// looking at the latest skipblock cache from Service.state.
+func (s *Service) updateCollectionCallback(sbID skipchain.SkipBlockID) error {
```
ineiti: I see some `log.Error` - will they stay in the final code?

kc1212: I was trying to decide whether to log the errors or return them or both, which is what I'm doing now but probably shouldn't. Now I think it might be better to just remove the return value from `updateCollectionCallback` because skipchain doesn't try to handle it anyway, it only logs it. What do you think?

omniledger/service/service.go
```diff
@@ -1181,17 +1173,12 @@ func newService(c *onet.Context) (onet.Service, error) {
 		return nil, err
 	}
 
-	var err error
```
ineiti: Good riddance...

omniledger/service/service.go
ineiti: I have a preference of returning errors and then printing them out if they are not nil...

ineiti:
@jeffallen asked me to not merge this PR until he has his subid in. He's tired of rebasing ;) So you'll have to rebase on his, he's 'nearly there'...

Issue #1360, opened 2018-07-24
need to use domain separation in omniledger hashing?


jeffallen:
When I was looking at DeriveID and also at darc.GetID(), I was reminded of the domain separation bytes that were used in kyber's Cipher AEAD before. I asked Phillipp about this and he said:

From Phillipp:
_Yeah domain separation is usually used to separate “cryptographic namespaces” so to say. The example you remembered is probably from sponge functions where domain separation is used to distinguish different types of data during processing. For example in NORX we use it to distinguish between associated data (which is not encrypted) and message data (which is encrypted), e.g., see the 0x01, 0x02, etc. in the following image:_ https://www.131002.net/data/talks/gtacs15/img/norx-D2.png

_Anyhow, I agree that it would be good to have a DARC type identifier that goes into the ID computation to ensure that IDs are always unique and thus prevent the issue you described. However, it seems that a description of the actions is going into the ID computation. Maybe that’s already enough to uniquely identify a DARC?_

We need to take a close look at the places we combine hashes and see if there are places where an attacker controls adjacent data and could try to cause a collision.

ineiti:
I remember vaguely that there is some random input to the darc, so that even two darcs with the same rules get another ID. Perhaps it's the base-ID of the darc that is random?

jeffallen:
See also [this comment](https://github.com/dedis/cothority/pull/1388#discussion_r208812995) from Linus.

jeffallen:
This needs to be fixed here:

```
func (instr Instruction) Hash() []byte {
	h := sha256.New()
...
	for _, a := range args {
		h.Write([]byte(a.Name)) // these should both have lengths before them
		h.Write(a.Value)
	}
```

PR #1388, opened 2018-08-08
Remove DarcID from InstanceID

jeffallen:
Remove DarcID from InstanceID. Change all the depending code to work correctly.


omniledger/service/service.go
```diff
@@ -669,32 +677,19 @@ func (s *Service) startPolling(scID skipchain.SkipBlockID, interval time.Duratio
...
-					}
+				cdb := s.getCollection(scID)
+				_, txOut, txBad, _, err := s.createStateChanges(cdb.coll, scID, txs, interval/2)
+				if err != nil {
+					log.Error("While choosing transactions, couldn't create state changes:", err)
+					continue
 				}
-				_, err = s.createNewBlock(scID, sb.Roster, txsCollect)
+
+				txs = txs[len(txOut)+len(txBad):]
```
kc1212: This is used to compute the unprocessed transactions, right? Maybe we can just have `createStateChanges` return that instead of `ctsBad`. I don't think `ctsBad` is used for anything else.

omniledger/service/service.go
```diff
@@ -799,6 +800,7 @@ clientTransactions:
 			scs, cout, err := s.executeInstruction(cdbI, cin, instr)
 			if err != nil {
 				log.Errorf("%s: Call to contract returned error: %s", s.ServerIdentity(), err)
+				ctsBad = append(ctsBad, ct)
```
kc1212: Shouldn't we also append `ctsBad` in the `scs` loop too? Again, it might be better to just return the remaining cts, probably look at the latest index and do something like `ctsRem := cts[latestIdx+1:]` after the outer for loop.

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/contracts/DarcInstance.java
```diff
@@ -166,22 +165,6 @@ public Instruction spawnContractInstruction(String contractID, Signer s, List<Ar
...
-    /**
-     * Like spawnContractInstruction, but creates a ClientTransaction with only this instruction and sends it
-     * to the omniledger.
-     *
-     * @param contractID the id of the contract to create
-     * @param s          the signer that is authorized to spawn this contract
-     * @param args       arguments to give to the contract
-     * @throws CothorityException
-     */
-    public TransactionId spawnContract(String contractID, Signer s, List<Argument> args) throws CothorityException {
```
ineiti: Why did you remove this? I can think of use-cases where you just want to send the spawn instruction without waiting for it to be included, no?

external/proto/omniledger.proto
```diff
@@ -122,42 +122,34 @@ message Proof {
...
-  // InstanceID holds the id of the existing object that can spawn new objects.
-  // It is composed of the Darc-ID + a random value generated by OmniLedger.
-  required InstanceID instanceid = 1;
-  // Nonce is monotonically increasing with regard to the darc in the instanceID
-  // and used to prevent replay attacks.
-  // The client has to track which is the current nonce of a darc-ID.
+  // InstanceID is either the object that can spawn a new object, or a the object
+  // that will be invoked or deleted.
+  required bytes instanceid = 1;
+  // Nonce is monotonically increasing with regard to the Darc controlling
```
ineiti: Is this used now? Or are we still ignoring it and hoping for a divine inspiration (I still don't have one ;)?

omniledger/collection/byteslice_test.go
```diff
@@ -13,15 +14,15 @@ func TestBytesliceEqual(test *testing.T) {
 	cutRho, _ := hex.DecodeString("85f46bd1ba19d1014b1179edd451ece95296e4a8c765ba8bba86c168939063")
 	alterRho, _ := hex.DecodeString("85f46bd1ba19d1014b1179edd452ece95296e4a8c765ba8bba86c16893906398")
 
-	if !(equal(lho, rho)) {
+	if !(bytes.Equal(lho, rho)) {
```
ineiti: why parentheses around `bytes.Equal`? Also for other occurrences.

omniledger/collection/field_test.go
```diff
@@ -1,12 +1,15 @@
...
+	"bytes"
+	"math/rand"
+	"testing"
+)
 
 func TestFieldData(test *testing.T) {
 	var data Data
 
-	if !equal(data.Encode([]byte("mydata")), []byte("mydata")) {
+	if !bytes.Equal(data.Encode([]byte("mydata")), []byte("mydata")) {
```
ineiti: Here it looks OK - no parentheses around `bytes.Equal`.

omniledger/contracts/coins.go
```diff
@@ -58,23 +59,28 @@ func (s safeUint64) sub(a uint64) (safeUint64, error) {
...
 	case omniledger.SpawnType:
-		// Spawn creates a new coin account as a separate instance. The subID is
-		// taken from the hash of the instruction.
-		ca := omniledger.InstanceID{
-			DarcID: inst.InstanceID.DarcID,
-			SubID:  omniledger.NewSubID(inst.Hash()),
-		}
-		log.Lvlf3("Spawing coin to %x", ca.Slice())
+		// Spawn creates a new coin account as a separate instance.
+		ca := omniledger.InstanceIDFromSlice(inst.Hash())
```
ineiti: Why is it `inst.Hash()` instead of `inst.DeriveID(ContractCoinID)`?

omniledger/contracts/coins.go
```diff
@@ -58,23 +59,28 @@ func (s safeUint64) sub(a uint64) (safeUint64, error) {
...
+		log.Lvlf3("Spawning coin to %x", ca.Slice())
 		sc = []omniledger.StateChange{
-			omniledger.NewStateChange(omniledger.Create, ca, ContractCoinID, make([]byte, 8)),
+			omniledger.NewStateChange(omniledger.Create, ca,
```
ineiti: Where is the `darcID` stored for this new coin instance?

omniledger/contracts/coins.go
```diff
@@ -58,23 +59,28 @@ func (s safeUint64) sub(a uint64) (safeUint64, error) {
 // You can only delete a contractCoin instance if the account is empty.
 func ContractCoin(cdb omniledger.CollectionView, inst omniledger.Instruction, c []omniledger.Coin) (sc []omniledger.StateChange, cOut []omniledger.Coin, err error) {
 	cOut = c
+
+	var darcID darc.ID
+	_, _, darcID, err = cdb.GetValues(inst.InstanceID.Slice())
```
ineiti: I propose you get both the `value` and the `darcID` here, so you don't have to re-get it in `case InvokeType` and `case DeleteType` again.

omniledger/contracts/coins_test.go
```diff
@@ -22,27 +23,29 @@ func TestCoin_Spawn(t *testing.T) {
...
 
+	zero := [32]byte{}
+	zerod := darc.ID(zero[:])
+
 	c := []omniledger.Coin{}
 	sc, co, err := ContractCoin(ct, inst, c)
 	require.Nil(t, err)
 	require.Equal(t, 1, len(sc))
-	ca := omniledger.InstanceID{DarcID: make([]byte, 32), SubID: omniledger.NewSubID(inst.Hash())}
+	ca := omniledger.InstanceIDFromSlice(inst.Hash())
```
ineiti: If you decide change to use `inst.DeriveID` in the `CoinContract`, then you'll also need to change it here.

omniledger/service/api_test.go
```diff
@@ -41,12 +41,13 @@ func TestClient_GetProof(t *testing.T) {
 	require.Nil(t, err)
 
 	// We should have a proof of our transaction in the skipchain.
+	newId := tx.Instructions[0].Hash()
```
ineiti: Like with the coin: why is it using `Hash()` instead of `DeriveID`?

omniledger/service/contracts.go
```diff
@@ -247,22 +238,31 @@ func (s *Service) spawnContractConfig(cdb CollectionView, inst Instruction, coin
...
+	id := d.GetBaseID()
 	return []StateChange{
-		NewStateChange(Create, GenesisReferenceID, ContractConfigID, inst.InstanceID.DarcID),
-		NewStateChange(Create, inst.InstanceID, ContractDarcID, darcBuf),
-		NewStateChange(Create,
-			InstanceID{
-				DarcID: inst.InstanceID.DarcID,
-				SubID:  oneSubID,
-			}, ContractConfigID, configBuf),
+		NewStateChange(Create, GenesisReferenceID, ContractConfigID, id, id),
```
ineiti: It looks strange (looked strange before, too, so this might be a new issue and not to be solved here) to have two `ContractConfigID` here. In the old system, I can see why we would want to have the `GenesisReferenceID` point to the `ContractConfigID`, but now I think we can do simpler:

```
GenesisReferenceID, ContractConfigID, configBuf, id
InstanceIDFromSlice(id), ContractDarcID, darcBuf, id
```

and then `s/GenesisReferenceID/GenesisConfigID/g`.

This would also remove the `DeriveConfigID`, which should've been `DeriveID(ContractConfigID)`.

omniledger/service/contracts.go
```diff
@@ -271,8 +271,9 @@ func (s *Service) ContractDarc(coll CollectionView, inst Instruction, coins []Co
 			if err != nil {
 				return nil, nil, errors.New("given darc could not be decoded: " + err.Error())
 			}
+			id := d.GetBaseID()
 			return []StateChange{
-				NewStateChange(Create, InstanceID{d.GetBaseID(), SubID{}}, ContractDarcID, darcBuf),
+				NewStateChange(Create, InstanceIDFromSlice(id), ContractDarcID, darcBuf, id),
 			}, coins, nil
 		}
 		// TODO The code below will never get called because this
```
ineiti: I don't know where this comment comes from, but it's wrong. Whenever we want to spawn a new instance, we need to send a `Spawn` instruction to an existing instance. This will often be a `ContractDarc` instance, which then will have to forward the spawn instruction to the appropriate contract.

omniledger/service/proto.go
```diff
@@ -138,42 +138,34 @@ type Proof struct {
 
 // Instruction holds only one of Spawn, Invoke, or Delete
 type Instruction struct {
-	// InstanceID holds the id of the existing object that can spawn new objects.
-	// It is composed of the Darc-ID + a random value generated by OmniLedger.
+	// InstanceID is either the object that can spawn a new object, or a the object
```
ineiti: s/a the/the/

omniledger/service/proto.go
```diff
@@ -138,42 +138,34 @@ type Proof struct {
...
+	// that will be invoked or deleted.
 	InstanceID InstanceID
-	// Nonce is monotonically increasing with regard to the darc in the instanceID
-	// and used to prevent replay attacks.
-	// The client has to track which is the current nonce of a darc-ID.
+	// Nonce is monotonically increasing with regard to the Darc controlling
+	// access to the instance. It is used to prevent replay attacks.
+	// The client has to track what the next nonce should be for a given instance.
```
ineiti: s/for a given instance/for a given darc/

omniledger/service/proto.go
```diff
@@ -138,42 +138,34 @@ type Proof struct {
...
 	Nonce Nonce
 	// Index and length prevent a leader from censoring specific instructions from
 	// a client and still keep the other instructions valid.
 	// Index is relative to the beginning of the clientTransaction.
 	Index int
 	// Length is the total number of instructions in this clientTransaction
 	Length int
-	// Spawn creates a new object
+	// Spawn creates a new object.
```
ineiti: s/object/instance/

also for following lines.

omniledger/service/proto.go
```diff
@@ -138,42 +138,34 @@ type Proof struct {
...
 	Spawn *Spawn
-	// Invoke calls a method of an existing object
+	// Invoke calls a method of an existing object.
 	Invoke *Invoke
-	// Delete removes the given object
+	// Delete removes the given object.
 	Delete *Delete
-	// Signatures that can be verified using the darc defined by the instanceID.
+	// Signatures that will be used while attempting to verify the Darc read request.
```
ineiti: `// Signatures that are verified using the darc referenced by InstanceID`

omniledger/service/service.go
```diff
@@ -669,32 +677,19 @@ func (s *Service) startPolling(scID skipchain.SkipBlockID, interval time.Duratio
...
+				_, txOut, txBad, _, err := s.createStateChanges(cdb.coll, scID, txs, interval/2)
+				if err != nil {
+					log.Error("While choosing transactions, couldn't create state changes:", err)
+					continue
 				}
-				_, err = s.createNewBlock(scID, sb.Roster, txsCollect)
+
+				txs = txs[len(txOut)+len(txBad):]
+				if len(txs) > 0 {
+					log.Lvlf3("Got more transactions than can be done in half the blockInterval. "+
```
ineiti: This should be at last `log.Lvl2` or even `log.Warn`, as it indicates that we're being swamped by requests.

omniledger/service/service_test.go
```diff
@@ -180,7 +183,7 @@ func TestService_GetProof(t *testing.T) {
 	s := newSer(t, 2, testInterval)
 	defer s.local.CloseAll()
 
-	serKey := s.tx.Instructions[0].InstanceID.Slice()
+	serKey := s.tx.Instructions[0].Hash()
```
ineiti: Might have to be replaced by `DeriveID`.

omniledger/service/struct.go
```diff
@@ -135,35 +146,53 @@ func (c *collectionDB) Get(key []byte) collection.Getter {
...
 	}
 	return c.db.Update(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket([]byte(c.bucketName))
 
-		// The contract type is stored in a key starting with C
-		keyC := make([]byte, 1+len(t.InstanceID))
-		keyC[0] = byte('C')
-		copy(keyC[1:], t.InstanceID)
+		// The 3 pieces of data (value, contract, darcid) are stored under three keys
+		// starting with 0, 1, 2.
```
ineiti: This could be
```
const (
  DBValue = iota
  DBContract
  DBDarcID
)
```

omniledger/service/transaction.go
```diff
@@ -36,41 +43,17 @@ func NewNonce(buf []byte) Nonce {
...
+// InstanceIDFromSlice converts the first 32 bytes of in into
+// an InstanceID.
+func InstanceIDFromSlice(in []byte) InstanceID {
+	var i InstanceID
+	copy(i[:], in)
+	return i
 }
 
-// Equal returns if both InstanceIDs point to the same instance.
-func (iID InstanceID) Equal(other InstanceID) bool {
```
ineiti: why remove `Equal` here? Most of our `*ID` do have `Equal`, which is convenient.

omniledger/service/transaction.go
```diff
@@ -123,25 +105,21 @@ func (instr Instruction) Hash() []byte {
 // InstanceID, the given string, and the hash of the Instruction.
 func (instr Instruction) DeriveID(what string) InstanceID {
 	h := sha256.New()
-	h.Write([]byte(what))
 	h.Write(instr.Hash())
+	h.Write([]byte{0})
```
ineiti: A 0-byte will not protect from any attack, as far as I can see. To prevent attacks we need to add the length of the coming part to be included in the hash, if it is variable-length. So it should be:
```
Write(instr.hash) // fixed length, no protection needed
sigBuf := ... // marshal all signatures
Write(VarInt(len(sigBuf))) // protect against somebody filling up using `what`
Write(sigBuf)
Write(what) // As this is the last element, no need to protect it
```
I hope that is correct. But adding `0` between two `Write` does not solve the problem. Except if we're sure that no `0` shows up in both `Signatures` and `what`, which is not the case, as far as I can tell.

proto.awk
```diff
@@ -5,7 +5,8 @@ BEGIN {
 	tr[ar"byte"]="bytes"
 	tr["abstract.Point"]="bytes"
 	tr["StateAction"]="int"
-	tr["SubID"]="bytes"
+	tr["omniledger.InstanceID"]="bytes"
+	tr["InstanceID"]="bytes"
```
ineiti: You can add

```
// type :InstanceID:bytes
```
in the `proto.go` just after `// PROTOSTART`.

omniledger/service/contracts.go
kc1212: I wrote the comment here some time ago, maybe it's fixed now. I basically put a panic here and none of the tests panic'd. My understanding is that in order for this contract to be executed, OL will look up `GetContractState`, which simply returns `inst.Spawn.ContractID`. THence, the first if statement in this function always gets called and returns.

eventlog/el/main_test.go
```diff
@@ -21,7 +21,7 @@ const blockInterval = 1 * time.Second
 
 func TestCli(t *testing.T) {
 	// TODO: Fix this test.
-	t.Skip("Disabled for now.")
+	t.Skip("Disabled for now; we would need to test ol and el at the same time?")
```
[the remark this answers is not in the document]
jeffallen: The Java tests do it by starting the service via Docker, then talking to it. I would like to maybe look into something like that for Go-based integration tests.

I would like very much to *not* write more shell code, I find shell scripts longer than 10 lines to be better in a typesafe language with a nice big library to use.

omniledger/service/service.go
```diff
@@ -809,14 +811,22 @@ clientTransactions:
...
+			}
 		}
 		cdbTemp = cdbI.c
 		ctsOK = append(ctsOK, ct)
 	}
 
 	// Store the result in the cache before returning.
 	merkleRoot = cdbTemp.GetRoot()
-	s.stateChangeCache.update(scID, cts.Hash(), merkleRoot, ctsOK, states)
+	s.stateChangeCache.update(scID, cts.Hash(), merkleRoot, ctsOK, ctsBad, states)
```
kc1212: I think something interesting is going on here when we return early. If the function times out, then `cts != ctsOK` and we store a state change cache keyed on `cts.Hash()`. Then when we do `createNewBlock`, this function gets called again but on `ctsOK` (or `txOut` in your code), so it'll look for a key on named `ctsOK.Hash()`. OL won't be able to use the data cached from the first call to `createNewBlock`. There won't be a wrong behaviour, there will simply be a cache miss if a timeout happens. We probably don't need to change any of the code, but put a comment to explain that this may happen.

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/Instruction.java
```diff
@@ -27,14 +28,13 @@
...
      * Use this constructor if it is a spawn instruction, i.e. you want to create a new object.
-     * @param instId The ID of the object, which must be unique.
      * @param nonce The nonce of the object.
      * @param index The index of the instruction in the atomic set.
      * @param length The length of the atomic set.
      * @param spawn The spawn object, which contains the value and the argument.
      */
-    public Instruction(InstanceId instId, byte[] nonce, int index, int length, Spawn spawn) {
-        this.instId = instId;
+    public Instruction(DarcId darcId, byte[] nonce, int index, int length, Spawn spawn) {
```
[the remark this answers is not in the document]
jeffallen: That's a missed cleanup. Fixed.

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/contracts/DarcInstance.java
jeffallen: I'm not sure it was on purpose. I'll add it back.

external/proto/omniledger.proto
jeffallen: Is what used? The nonce?

omniledger/service/service.go
kc1212: Actually, the cache will be invalid anyway because `createNewBlock` will re-order the transactions. I have a PR that removes the re-ordering - https://github.com/dedis/cothority/pull/1389, will rebase to this PR when it's in master.

omniledger/contracts/coins.go
jeffallen: Yeah, exactly. I was struggling with this too.

When I first realised we needed DeriveID it was to solve the case where you need to create a new object in the collection in order to help store/index the current item. I.e. while writing a new event, you need a new time bucket to store it in.

But you are proposing that new instance ID's should *always* come from DeriveID, which might be simpler to explain to people.

I do think it is regrettable that client side code which wants to predict what key a given object will be stored at needs to know the magic string to give to it's own invocation of DeriveId. To me that's kind of the internals of the contract and none of the client's business.

But in eventlog, we already violate that, because the instance ID of a newly spawned eventlog is not inst.Hash(), but is in fact inst.DeriveId("eventlog").

eventlog/el/main_test.go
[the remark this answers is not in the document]
ineiti: More discussion needed, as I think that shell scripts give:
- _real_ integration testing, where we use _all_ layers
- insight into how users will use the script - if something is awkward to do in `test.sh`, the users will also have their problems
- can integrate multiple 'commands' if needed

external/proto/omniledger.proto
ineiti: Yes, the nonce... But as far as I saw in the code, it's not used yet in a way that makes it secure. Anyway - that's not the purpose of this PR.

omniledger/contracts/coins.go
ineiti: If we say that `DeriveID` for a new instance should take the `ContractID` of that the new instance?

omniledger/service/contracts.go
ineiti: But for the coins we send a `Spawn` instruction to the `DarcContract`, no?

omniledger/contracts/coins.go
jeffallen: Fixed. That code was a messy leftover that was probably working by accident.

omniledger/service/proto.go
jeffallen: I reworded yours as well because "which darc?"
```
// Signatures that are verified using the Darc controlling access to the instance.
```

eventlog/el/main_test.go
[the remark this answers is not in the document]
kc1212: I do agree that testing using shell scripts have their use cases, but our home-brew framework isn't exactly easy to understand and use, I also struggle with it. In my previous job, we used the testing framework from FreeBSD - https://www.freebsd.org/cgi/man.cgi?query=atf-sh&sektion=3&apropos=0&manpath=FreeBSD+11.2-RELEASE+and+Ports, I'm sure there are many others. Perhaps we can think about using a shell testing framework, with nice documentation, for tests that must be done in shell scripts so that it's easier to learn/use and a bit more "industry standard".

omniledger/service/transaction.go
jeffallen: This is covered by #1360, so I'm going to ignore this for the moment, in order to get this thing checked in.

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/Instruction.java
```diff
@@ -186,13 +179,13 @@ public Request toDarcRequest() {
      * @param signers - the list of signers.
      * @throws CothorityCryptoException
      */
-    public void signBy(List<Signer> signers) throws CothorityCryptoException {
+    public void signBy(DarcId darcId, List<Signer> signers) throws CothorityCryptoException {
```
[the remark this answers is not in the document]
jeffallen: And then the poor programmer who comes along later has to understand why there are two methods that seem to do the same thing, and choose one. This is a cultural difference btw Go and Java. As always I lean towards the YAGNI side of things (You Are Not Going To Need It, so don't write more code now than you know you need now.)

But if you reply and say, "no really, in Java we always give them 3 constructors to choose from, please write 2 more" I'm good.

I can make a vanilla one, a chocolate one, a chocolate/vanilla mixed with an optional AbstractSprinkleFactoryBuilder argument, etc. 🍨

eventlog/el/main_test.go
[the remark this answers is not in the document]
ineiti: Sounds like a good idea. Let's have a look at those...

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/Instruction.java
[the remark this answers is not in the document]
ineiti: I like the AbstarctSprinkleFactoryBuilder, of course. If it could solve the climate catastrophy, too, that would be awesome!

For real (and not in this PR): we need to think how we can keep better track of the darc that is linked in an instance.

PR #1389, opened 2018-08-09
Do not re-order transactions

kc1212:
Transactions are re-ordered in createNewBlock, which means that if we
cache the state changes in startPolling, then the cache will be invalid
in createNewBlock.

Depends on https://github.com/dedis/cothority/pull/1388


jeffallen:
What was the attack sorting was intended to mitigate, and why is it no longer needed?

kc1212:
IIRC, it was to prevent the leader from re-ordering transactions in an order such that more transactions will fail to execute. But the leader can simply omit transactions so re-ordering does not help.

PR #1392, opened 2018-08-09
Fix race condition on pollChan and heartbeat

kc1212:
This was caught in Jenkins, I don't think it ever happened on Travis or
locally. It's nice that we have multiple testing infrastructures.


kc1212:
The long test in jenkins is confirmed to be passing with this fix.

PR #1393, opened 2018-08-09
(prototype) Docker for Go CLI tool tests

jeffallen:
A prototype of how it looks when we test CLI tools against a contained cothority like Java does.


jeffallen:
Interesting experiment, but we don't want this heavy mess in our Go tests.

PR #1411, opened 2018-08-15
Fix state change caching in the case of bad tx

jeffallen:
Previously, if a tx is removed because it is bad, the
state changes generated by the remaining tx's would be
cached under the wrong key. The result would be that
an attacker could disable the state change caching by
intentionally sending invalid transactions.

This was found by code inspection, but then I decided
to prove it was really happening by adding the hit/miss
monitoring mechanism.


jeffallen:
Yes, thanks for pointing that out. New version coming.

PR #1415, opened 2018-08-15
Database Versioning

ineiti:
First shot at adding a database versioning for the Identity and the OmniLedger
services. The idea is to add an additional bucket with an integer in it, that
can be used to get older versions of the database. In the Identity service,
`db.go` holds an example that does try different versions before giving up.

This depends on dedis/onet#464

### DIFF
--- a/evoting/service/service.go
+++ b/evoting/service/service.go
@@ -46,6 +46,7 @@ var serviceID onet.ServiceID
 
 // storageKey identifies the on-disk storage.
 var storageKey = []byte("storage")
+var dbVersion = 1
 
 // Service is the core structure of the application.
 type Service struct {
@@ -738,6 +739,9 @@ func (s *Service) save() {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
 	if err := s.Save(storageKey, s.storage); err != nil {
+		log.Error(err)
+	}
+	if err := s.SaveVersion(dbVersion); err != nil {
 		log.Error(err)
 	}
 }
--- a/omniledger/service/service.go
+++ b/omniledger/service/service.go
@@ -98,7 +98,7 @@ type Service struct {
 
 // storageID reflects the data we're storing - we could store more
 // than one structure.
-const storageID = "OmniLedger"
+var storageID = []byte("OmniLedger")
 
 // defaultInterval is used if the BlockInterval field in the genesis
 // transaction is not set.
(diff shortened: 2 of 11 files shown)


omniledger/service/service.go
```diff
@@ -1139,13 +1140,15 @@ func (s *Service) tryLoad() error {
...
 		if err != nil {
+			log.Print()
 			return err
 		}
 		s.state.setLast(sb)
 
 		// populate the darcID to skipchainID mapping
 		d, err := s.LoadGenesisDarc(gen)
 		if err != nil {
+			log.Print()
```
kc1212: probably didn't want this here

evoting/service/service.go
```diff
@@ -740,6 +741,7 @@ func (s *Service) save() {
 	if err := s.Save(storageKey, s.storage); err != nil {
 		log.Error(err)
 	}
+	s.SaveVersion(dbVersion)
```
kc1212: This might overwrite the existing version. Instead of just saving the version, I think we need to:
1. save the data only when current version is 0 or `dbVersion`, write `dbVersion` too if the version is 0
2. otherwise display an error

but I can imagine that this is the first PR out of many, so it's fine to leave it like this for now.

kc1212: actually, we should check for error

ineiti:
But for your other propositions, yes, this is more or less what I think of doing when updating the database. Your pseudo-code will not work, because not all versions will be able to fill in the correct `Storage` structure.

evoting/service/service.go
ineiti: checking for error is good - I don't see why you don't want to write the version every time we save.

ineiti:
Updated the names of `Storage` to `storage1` in the `Identity` example.

evoting/service/service.go
kc1212: It's just a small optimisation, if the version number is the same then there's no need to write it. What I'm imagining is that the upgrade process will store the correct version number after the upgrade. So, if the version number is something different then something went wrong and you don't want to overwrite it.

kc1212:
Regarding this comment https://github.com/dedis/cothority/pull/1415#issuecomment-413811892

Can you give me an example? Is it because new versions might introduces compulsory fields that the old version doesn't know about? I guess we'll have to specify default values when introducing new fields. In general, I think if the data is not corrupted, and the upgrade path exists, then there shouldn't be any errors, at least that's what we should aim to do.

ineiti:
For example if you merge two fields into one, then the previous versions will not be able to pass them to the next version.

kc1212:
Oh yes, the `s *Storage` needs to be changed indeed. It needs to be something generic.

PR #1417, opened 2018-08-15
Make libtest.sh part of cothority and less magical

jeffallen:


libtest.sh
```diff
@@ -0,0 +1,356 @@
...
+  if [ "$CLEANBUILD" ]; then
+    rm -f conode $APP
+  fi
+  build $APPDIR
+}
+
+test(){
+  echo "** DEPRECATED: please change to using runTest instead."
+  exit 1
+}
```
kc1212: I think it's best to just delete it. Keeping it still conflicts with the bash `test` command, which is what we're trying to fix.

libtest.sh
```diff
@@ -0,0 +1,356 @@
+#!/usr/bin/env bash
```
kc1212: This file that you copied over isn't from the latest onet. The latest version has comments on all the functions.

cisc/test.sh
```diff
@@ -29,29 +29,29 @@ main(){
...
-  test KeyAdd2
-  test KeyAddWeb
-  test KeyDel
-  test SSHAdd
-  test SSHDel
-  test Follow
-  test SymLink
-  test Revoke
-  test RosterEdit
+  runTest testBuild
```
ineiti: `govet` would say that this stutters...

libtest.sh
ineiti: This is an old version without the comments.

ineiti: Please take the latest `libtest.sh` from `onet/master`.

jeffallen: Arg, I don't know what happened here, thanks for catching this.

kc1212:
NOTE: using `run` like this is ok, but we might have change it again later because bats has a `run` command.

(Can't merge yet because it's blocked until everyone approves the changes.)

PR #1418, opened 2018-08-15
Make evalExpr() and evalExprWithSigns() public in the darc package

talhaparacha:
We, at LCA1, are using Omniledger to store access control rights for sensitive medical data as a part of the MedChain project. Given the nature of this project, we need to be able to evaluate multiple DARC expressions from a contract's code. 

So it would be great if the functions evalExpr() and evalExprWithSigns() were to be made public.


ineiti:
Currently fails because of dedis/protobuf#50...

PR #1419, opened 2018-08-16
Make invalid protobuf test more intelligent

ineiti:
With the latest protobuf changes, we discard unknown fields (like protobuf3)
and don't throw an error anymore. So we need to be more intelligent in
creating invalid input.


ineiti:
Superseded by dedis/protobuf#50

PR #1421, opened 2018-08-16
Remove DeriveID as much as possible

jeffallen:
In accordance with the (new) documentation on DeriveID, it should
ONLY be used inside a contract that needs to create new keys to
store additional information. All other instance ID's should come
from spawning new instances, and those should be inst.Hash().

The ida for DeriveID came from the legitimate need of eventlog
to get keys to store its time buckets. So it is still using
DeriveID.

Remove deriveId from Java, since if contracts are using
it correctly (i.e. only for their own use) then Omniledger
clients will never need to look up those keys.

### DIFF
--- a/eventlog/api.go
+++ b/eventlog/api.go
@@ -64,7 +64,7 @@ func (c *Client) Create() error {
 		return err
 	}
 
-	c.Instance = instr.DeriveID("eventlog")
+	c.Instance = omniledger.NewInstanceID(instr.Hash())
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


jeffallen:
This is apparently failing due to a flaky test: "db not open" in the coin simulation.

kc1212:
Maybe [this change](https://github.com/dedis/cothority/pull/1412/files) made the tests more flaky. I removed the final `Sleep` because we're not doing the `updateCollection` propagation anymore and didn't notice any failures in my testing. But maybe there are other things that are taking a long time to close when running with `-race`. You can try to add the final `Sleep` back in and see if that makes the flaky tests go away.

ineiti:
`db not open` sometimes also just means that the test did not abort properly and that one of the nodes is still trying to access the database.

PR #1422, opened 2018-08-17
Updating Omniledger to behave

ineiti:
There were a couple of things astray in Omniledger land:
- verification of signatures on instructions were done _after_ the contract has been run
  -> verification of signatures is done in contracts, so they can do whatever seems fit
- instructions sent to an instance didn't call that instance if it was a 'spawn' instruction
  -> all instructions call now the instance in the 'InstanceID' field
  -> fixed tests that sent instructions to non-existent instances
- followers didn't verify the signatures of instructions
  -> verification is done in contracts, so this is taken care of
- updated java and added `deriveID` again

### DIFF
--- a/external/java/src/main/java/ch/epfl/dedis/lib/omniledger/Instruction.java
+++ b/external/java/src/main/java/ch/epfl/dedis/lib/omniledger/Instruction.java
@@ -28,10 +28,11 @@ public class Instruction {
 
     /**
      * Use this constructor if it is a spawn instruction, i.e. you want to create a new object.
-     * @param nonce The nonce of the object.
-     * @param index The index of the instruction in the atomic set.
+     *
+     * @param nonce  The nonce of the object.
+     * @param index  The index of the instruction in the atomic set.
      * @param length The length of the atomic set.
-     * @param spawn The spawn object, which contains the value and the argument.
+     * @param spawn  The spawn object, which contains the value and the argument.
      */
     public Instruction(InstanceId instId, byte[] nonce, int index, int length, Spawn spawn) {
         this.instId = instId;
@@ -43,9 +44,10 @@ public class Instruction {
 
     /**
      * Use this constructor if it is an invoke instruction, i.e. you want to mutate an object.
+     *
      * @param instId The ID of the object, which must be unique.
-     * @param nonce The nonce of the object.
-     * @param index The index of the instruction in the atomic set.
+     * @param nonce  The nonce of the object.
+     * @param index  The index of the instruction in the atomic set.
      * @param length The length of the atomic set.
      * @param invoke The invoke object.
      */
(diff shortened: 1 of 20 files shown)


omniledger/service/service.go
```diff
@@ -768,10 +675,9 @@ func (s *Service) createStateChanges(coll *collection.Collection, scID skipchain
 	// ignore the error and compute the state changes.
 	merkleRoot, ctsOK, ctsBad, states, err = s.stateChangeCache.get(scID, cts.Hash())
 	if err == nil {
-		log.Lvl4(s.ServerIdentity(), "state changes from cache: HIT")
+		log.Lvl3(s.ServerIdentity(), "loaded state changes from cache")
 		return
 	}
-	log.Lvl4(s.ServerIdentity(), "state changes from cache: MISS")
```
kc1212: I think it's useful to keep these when debugging cache problems.

omniledger/service/service_test.go
```diff
@@ -464,57 +464,64 @@ func TestService_StateChange(t *testing.T) {
 			return nil, nil, err
 		}
 
+		zeroBuf := make([]byte, 8)
+		switch {
 		// create the object if it doesn't exist
-		if !rec.Match() {
-			if inst.Spawn == nil {
-				return nil, nil, errors.New("expected spawn")
+		case inst.Spawn != nil:
```
jeffallen: Why not convert this one to switch `inst.GetType()` as well?

omniledger/service/transaction.go
```diff
@@ -143,6 +136,11 @@ func (instr Instruction) GetContractState(coll CollectionView) (contractID strin
 	var cv []interface{}
 	cv, err = record.Values()
 	if err != nil {
+		if bytes.Compare(ConfigInstanceID[:], instr.InstanceID[:]) == 0 {
```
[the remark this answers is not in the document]
ineiti: I actually thought: "why doesn't jeff like `.Equal`?" when writing this. +1

jeffallen:
This might be failing for the same reason as #1421 , and the latest fix I made there worked. So once it is checked in, this might get fixed.

external/java/src/main/java/ch/epfl/dedis/lib/omniledger/Instruction.java
```diff
@@ -195,16 +203,43 @@ public void signBy(DarcId darcId, List<Signer> signers) throws CothorityCryptoEx
         }
     }
 
+    /**
+     * This function derives an instance ID from the instruction with the given string. The derived instance ID if
```
kc1212: `s/if/is`, or just say "is the key of this transaction"

omniledger/service/service_test.go
```diff
@@ -80,15 +80,18 @@ func TestService_AddTransaction_WithFailure(t *testing.T) {
 }
 
 func TestService_AddTransaction_WithFailure_OnFollower(t *testing.T) {
+	// for {
+	// 	log.Print("Starting test")
 	testAddTransaction(t, 1, true)
+	// }
```
kc1212: remove these?

PR #1424, opened 2018-08-19
Generalise the view-change test

kc1212:
This is to prepare for the upcoming work on improving viewchange to work
in more failures scenarios.

### DIFF
--- a/omniledger/service/service_test.go
+++ b/omniledger/service/service_test.go
@@ -756,63 +756,73 @@ func TestService_SetBadConfig(t *testing.T) {
 	}
 }
 
-// TestService_RotateLeader is an end-to-end test for view-change. We kill the
-// current leader, at index 0. Then the node at index 1 becomes the new leader.
-// Then, we try to send a transaction to a follower, at index 2. The new leader
-// should poll for new transactions and eventually make a new block containing
-// that transaction. The new transaction should be stored on all followers.
-func TestService_RotateLeader(t *testing.T) {
+// TestService_ViewChange is an end-to-end test for view-change. We kill the
+// first nFailures nodes, where the nodes at index 0 is the current leader. The
+// node at index nFailures should become the new leader. Then, we try to send a
+// transaction to a follower, at index nFailures+1. The new leader (at index
+// nFailures) should poll for new transactions and eventually make a new block
+// containing that transaction. The new transaction should be stored on all
+// followers. Finally, we bring the failed nodes back up and they should
+// contain the transactions that they missed (TODO this does not work yet
+// because the config contract fails).
+func TestService_ViewChange(t *testing.T) {
+	testViewChange(t, 4, 1)
+}
+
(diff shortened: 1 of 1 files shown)


omniledger/service/service_test.go
```diff
@@ -756,63 +756,77 @@ func TestService_SetBadConfig(t *testing.T) {
...
-func TestService_RotateLeader(t *testing.T) {
+// TestService_ViewChange is an end-to-end test for view-change. We kill the
+// first nFailures nodes, where the nodes at index 0 is the current leader. The
+// node at index nFailures should become the new leader. Then, we try to send a
+// transaction to a follower, at index nFailures+1. The new leader (at index
+// nFailures) should poll for new transactions and eventually make a new block
+// containing that transaction. The new transaction should be stored on all
+// followers. Finally, we bring the failed nodes back up and they should
+// contain the transactions that they missed (TODO this does not work yet
+// because we the config contract fails).
```
jeffallen: s/we//

omniledger/service/service_test.go
```diff
@@ -756,63 +756,77 @@ func TestService_SetBadConfig(t *testing.T) {
...
 	defer s.local.CloseAll()
 
 	for _, service := range s.services {
 		service.SetPropagationTimeout(interval / 2)
 	}
 
-	// Stop the leader, then the next node should take over.
-	s.service().TestClose()
-	s.hosts[0].Pause()
+	// The first nFailures hosts then the next node at index nFailures
```
jeffallen: // Stop the first nFailures hosts.

omniledger/service/service_test.go
```diff
@@ -756,63 +756,77 @@ func TestService_SetBadConfig(t *testing.T) {
...
+	// should take over.
+	for i := 0; i < nFailures; i++ {
+		s.services[i].TestClose()
+		s.hosts[i].Pause()
+	}
 
 	// wait for the new block
```
jeffallen: // Wait for proof that the new expected leader, s.services[nFailures], has taken over.

kc1212:
Thanks, fixed.

omniledger/service/service_test.go
```diff
@@ -756,63 +756,77 @@ func TestService_SetBadConfig(t *testing.T) {
...
+	// should take over.
+	for i := 0; i < nFailures; i++ {
+		s.services[i].TestClose()
+		s.hosts[i].Pause()
+	}
 
-	// wait for the new block
+	// Wait for proof that the new expected leader, s.services[nFailures], has taken over.
 	var ok bool
 	for i := 0; i < 5; i++ {
```
ineiti: Shouldn't this just

```
time.Sleep(2 * nFailures * s.interval)
```

and then verify right away if the new leader is active?

Looping here seems too fuzzy to me.

kc1212: nice catch! and you're right, we don't need looping.

PR #1429, opened 2018-08-20
Popcoins - WIP

ineiti:
This is a WIP PR for the pop-coin contract in OmniLedger.


ineiti:
It finally went in through a set of patches. Now I need to prepare the patches to the patches...

PR #1434, opened 2018-08-21
pop config path

yurunhub:
create config file path for pop if it does not exist

### DIFF
--- a/pop/app.go
+++ b/pop/app.go
@@ -613,10 +613,10 @@ func getConfigClient(c *cli.Context) (*Config, *service.Client) {
                 if os.IsNotExist(err) {
                         err := os.MkdirAll(dir, 0770)
                         if err != nil {
-                                return nil, err
+				log.ErrFatal(err)
                         }
                 } else {
-                        return nil, err
+			log.ErrFatal(err)
                 }
         }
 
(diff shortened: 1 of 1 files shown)


ineiti:
Thanks for the PR - some issues:
- it doesn't pass travis: https://travis-ci.org/dedis/cothority/jobs/418761888#L694
   (I think it's a formatting issue - please use `gofmt`)
- the directory should be created in the `write` method
- no need to check if the directory exists - `MkdirAll` will do that for you: https://golang.org/pkg/os/#MkdirAll

yurunhub:
added to write method with gofmt travis is happy

ineiti:
One last thing - it seems it's better to use `filepath` instead of `path`, because `filepath` abstracts the different separators. This article on stackoverflow goes into the details (read 'when do I use each'?). As we _do_ have some people using Windows, this difference is important to us.

https://stackoverflow.com/questions/39181790/whats-the-difference-between-path-path-filepath-packages-in-go

ineiti:
And one very last thing: if you will continue to work on `onet` / `cothority`, please write an email to linus.gasser at epfl dot ch, so that I can invite you into our slack channel.

yurunhub:
changed path->filepath and tested

ineiti:
Cool! Thank you very much.

Issue #1431, opened 2018-08-21
need test for InclusionWait > 0 && tx is refused


jeffallen:
This comment says that we probably don't notify correctly:

```
	for _, ct := range body.Transactions {
		// TODO: Where is the other call, for informWaitChannel(hash, false)?
		s.state.informWaitChannel(ct.Instructions.Hash(), true)
	}
```

Need a test to figure this out and solve it.

jeffallen:
This will be fixed in #1437.

jeffallen:
Fixed.

PR #1437, opened 2018-08-23
Added refused transactions to the created blocks

### LINKED ISSUE #1430: Add failed transaction hashes to payload and to DataHeader
In order to detect if a given transaction has been included in the ledger, a client can work backwards from the latest block until the first block with timestamp < the time they sent the transaction, looking for their transaction in the DataBody's of all the blocks.

However: This would not let them detect when their tx was refused (since it might be refused, or might still be in queue waiting for a future block). For that, they need to be able to find their failed transaction in some block. So the block Payload needs to have list of client txs that succeeded and that failed. (Or a per-tx bool.) Verifiers will run all the txs, even the ones that are marked invalid by the leader, and will indicate their agreement by giving the forward link (or not).

However (again): This gives an attacker the possibility of spamming the blockchain. Ethereum's solution is that tx that fail consume the gas allocated to them, costing the attacker cold hard cash. We need to consider making it cost something to propose a TX. Kelong says: In ethereum, there are actually two kinds of failure: failed to start verifying, and failed due to out of gas while verifying. The first kind never make it into the blockchain, the second kind do.

jeffallen:
Fixes #1430, #1431.


omniledger/service/service.go
```diff
@@ -301,31 +299,43 @@ func (s *Service) createNewBlock(scID skipchain.SkipBlockID, r *onet.Roster, cts
...
 		return nil, errors.New("Couldn't marshal data: " + err.Error())
 	}
 
 	// Store transactions in the body
-	body := &DataBody{Transactions: ctsOK}
-	sb.Payload, err = network.Marshal(body)
+	body := &DataBody{
+		Transactions:        ctsOK,
+		RefusedTransactions: refused,
+	}
```
kc1212: This part doesn't look right. Don't we want to store all the transactions but mark the ones that are bad? The follower should all execute every transaction and find the same transactions to be bad. I don't think the actual bad transactions are sent out for the followers to check.

omniledger/service/service.go
```diff
@@ -642,10 +659,19 @@ func (s *Service) verifySkipBlock(newID []byte, newSB *skipchain.SkipBlock) bool
...
+	err = protobuf.DecodeWithConstructors(newSB.Payload, &body, network.DefaultConstructors(cothority.Suite))
+	if err != nil {
+		log.Errorf("verifySkipblock: couldn't unmarshal body")
+		return false
+	}
+
+	rhash := sha256.New()
+	for _, x := range body.RefusedTransactions {
+		rhash.Write(x)
+	}
```
kc1212: Similar to my commend above, the follower isn't doing any real verification on the list of bad transactions, it's simply summing the hashes.

omniledger/service/service.go
jeffallen: During the design discussion for this we determined that allowing an attacker to fill the body with bad tx's was not a good plan, so we'd only store the hashes.

ineiti: On one of the first implementations we had a `bool` field for every `ClientTransaction` in the body of the block. That would make sense here.

jeffallen:
Moved this back to WIP while implementing the new specification in #1430.

omniledger/service/service.go
jeffallen: On further consideration we found that it is imperative to have the tx+bool setup, because that way the block verifiers get a chance to run the exact same txns in the same order as the leader and see if they also detect the same set of accepted/refused txns.

skipchain/api.go
```diff
@@ -201,6 +203,9 @@ func (c *Client) GetUpdateChain(roster *onet.Roster, latest SkipBlockID) (reply
 			if err == nil && len(r2.Update) != 0 {
 				break
 			}
+			time.Sleep(delay)
+			// exponential backoff
+			delay *= 2
```
kc1212: this will make the last retry very long without doing anything, can you an if condition when `i == retries - 1`?

omniledger/service/service.go
```diff
@@ -707,15 +714,17 @@ func (s *Service) verifySkipBlock(newID []byte, newSB *skipchain.SkipBlock) bool
 	return true
 }
 
-// createStateChanges goes through all ClientTransactions and creates
-// the appropriate StateChanges. If any of the transactions are invalid,
-// it returns an error. If timeout is not 0, createStateChanges will stop
-// running instructions after that long, in order for the caller to determine
-// how many instructions fit in a block interval.
-func (s *Service) createStateChanges(coll *collection.Collection, scID skipchain.SkipBlockID, cts ClientTransactions, timeout time.Duration) (merkleRoot []byte, ctsOK ClientTransactions, ctsBad ClientTransactions, states StateChanges, err error) {
+// createStateChanges goes through all the proposed ClientTransactions and
```
kc1212: the comment says `ClientTransactions` but the function uses `TxResults`. which makes me wonder whether it's possible to have `createStateChanges` take `ClientTransactions`, because it doesn't need the boolean flag to do the job.

kc1212:
@jeffallen got a couple of tests saying "too many retries; last error: websocket: close 4000: Couldn't find latest skipblock". I tried re-running and saw the same error

PR #1439, opened 2018-08-24
Update to our current understanding of how DeriveID works

jeffallen:
Also: Add el integration test to prove that it still works.

### DIFF
--- a/eventlog/el/test.sh
+++ b/eventlog/el/test.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+
+DBG_TEST=2
+DBG_SRV=0
+
+. "$(go env GOPATH)/src/github.com/dedis/cothority/libtest.sh"
+
+main(){
+	build $APPDIR/../../omniledger/ol
+    startTest
+    buildConode github.com/dedis/cothority/eventlog
+
+	# This must succeed before any others will work.
+    run testCreate
+	
+	run testLogging
+	
+    stopTest
+}
+
+testLogging(){
+	runCoBG 1 2 3
+	testOK ./el log -t test -c 'abc'
+	testOK ./el log -c 'def'
+	echo ghi | testOK ./el log 
+	seq 100 | testOK ./el log -t seq100
+
+	# two block intervals (2 * 0.5s)
+	sleep 1
+
+	testGrep "abc" ./el search -t test
+	testCountLines 103 ./el search
+
+	testCountLines 0 ./el search -t test -from '0s ago'
+	testCountLines 0 ./el search -t test -from '1h ago' -to `date -v -1d +%Y-%m-%d`
+	testCountLines 1 ./el search -t test -to `date -v +1d +%Y-%m-%d`
+}
+
+testCreate(){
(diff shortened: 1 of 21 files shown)


omniledger/ol/test.sh
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
-DBG_TEST=2
+DBG_TEST=1
 # Debug-level for app
-DBG_APP=2
-DBG_SRV=2
+DBG_APP=0
```
ineiti: I usually put `DBG_APP=1`, so you can at least see where the test failed in the jenkins output.

omniledger/service/transaction.go
```diff
@@ -110,19 +110,48 @@ func (instr Instruction) Hash() []byte {
...
+	// because an attacker could try to construct two messages resulting in the
+	// same hash by moving bytes from one neighboring input to the
+	// other. With fixed-length inputs, moving bytes is not possible, so
+	// no domain separation is needed.
+
 	h := sha256.New()
 	h.Write(instr.Hash())
-	h.Write([]byte{0})
+
+	// Putting the length of instr.Signatures into the hash does not seem to
```
ineiti: The reason we do this (according to Jeff ;) is because we might have OpenIDConnect signatures that have variable-length signatures and that might be prone to padding-attacks.

omniledger/simulation/coins.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/BurntSushi/toml"
 	"github.com/dedis/cothority/omniledger/contracts"
 	"github.com/dedis/cothority/omniledger/darc"
-	"github.com/dedis/cothority/omniledger/service"
+	omniledger "github.com/dedis/cothority/omniledger/service"
```
ineiti: should we standardise on `ol` or `omniledger`? Now we have both...

kc1212: I vote for `ol` because it's shorter

ineiti: :+1: here.

jeffallen: I see that I am outvoted, so ok. I like package names to be larger than 2 characters, because 2 characters are usually for local variables. See https://research.swtch.com/names

eventlog/el/test.sh
```diff
@@ -0,0 +1,63 @@
...
+	
+	runGrepSed "export EL=" "" ./el create
+	eval $SED
+	[ -z "$EL" ] && exit 1
+	
+	# We do not want cleanup to remove the db between each test.
+	export KEEP_DB=true
+}
+
+main
```
kc1212: Something's wrong with the indentation of this file, are we using tabs or spaces (how many) in shell scripts?

jeffallen: My editor defaults to tabs in shell files. As I am too lazy to modify it, I propose that we should use tabs in shell files. (I already lost a vote today, maybe I can do better on this one.)

I'm pushing a corrected version with all tabs.

PR #1441, opened 2018-08-24
Protocol fails when subleader of the same tree continuosly fail more than 2 times

swarali:
If the current subleader of a tree is unresponsive we choose the next subleader as the first child and append the current subleader to the last position.
Eg: [0,1,2,3,4]->[0,2,3,4,1](currently fails here)->[0,3,4,1,2]->[0,4,1,2,3](should fail here)

### DIFF
--- a/ftcosi/protocol/protocol.go
+++ b/ftcosi/protocol/protocol.go
@@ -334,7 +334,7 @@ func (p *FtCosi) startSubProtocol(tree *onet.Tree) (*SubFtCosi, error) {
 	cosiSubProtocol.Data = p.Data
 	// We allow for one subleader failure during the commit phase, and thus
 	// only allocate one third of the ftcosi budget to the subprotocol.
-	cosiSubProtocol.Timeout = p.Timeout / time.Duration(len(tree.List()))
+	cosiSubProtocol.Timeout = p.Timeout / 3
 
 	// the Threshold (minus root node) is divided evenly among the subtrees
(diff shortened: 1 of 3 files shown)


ineiti:
that might have to do with the timeout. Currently the timeouts are calculated to have two subleader failures (not sure this is tested). The third failure will most surely return an error.

ineiti:
In fact it only supports one subleader failure, the second one needs to pass. Else all our timeouts will be wrong...

jeffallen:
@Gilthoniel can you please check if this is applicable to blscosi? If not, close it, if yes, we need to decide if this is before v3 or after.

Gilthoniel:
@jeffallen https://github.com/dedis/cothority/blob/77d2e0fe7c1878ce8915fc8cb3647d70d163a89a/blscosi/protocol/protocol.go#L260

As of now, the protocol allows three subleaders before the global timeout is reached. It could also happen that an entire tree is failing but the protocol still ends with a signature if enough is received from the other trees (3f+1).

If your subtree has a lot of leaves and you assume that you want to let any member of the subtree to try to be a subleader, you will end with too small timeouts (or a very big global timeout depending how you compute the timeouts).

If you are ok with this behavior, we can close this.

jeffallen:
I'm am ok with this behaviour. (Or more precisely, by observation, our clients seem to be more or less ok with behaviour like this, and if it comes up after v3, we can adjust it.)

PR #1448, opened 2018-08-29
Keep block index with the database

### LINKED ISSUE #1425: Latest skipblock ID cache should be a part of collectionDB
We have a variable that stores the latest skipblock ID in the `olState` struct. However, it might become inconsistent with respect to the collectionDB. For example, a node that is just starting up from a failure will try to reply the missed transaction by running `updateCollectionCallback`. In the process, it will call `getCollection` which loads the state from boltdb, from . Since `olState` is only updated at the end of `updateCollectionCallback`, which is after the call to `getCollection`, there is a period where the state is inconsistent.

The suggestion is to keep the skipchain ID that represents the latest state in the collectionDB so that we do not run into this inconsistency.

jeffallen:
The last applied block index should be kept alongside the data
itself, not in a runtime state structure that will be lost on
server restart.

Fixes #1425.

### DIFF
--- a/omniledger/service/proof_test.go
+++ b/omniledger/service/proof_test.go
@@ -92,7 +92,7 @@ func createSC(t *testing.T) (s sc) {
 
 	s.key = []byte("key")
 	s.value = []byte("value")
-	s.c.Store(&StateChange{StateAction: Create, InstanceID: s.key, Value: s.value})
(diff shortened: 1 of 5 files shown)


jeffallen:
I don't understand "remove the inconsistency" and "derive the latest skipblock ID from the index". With this PR olState still exists, to hold the waitChannels and its mutex.

PR #1449, opened 2018-08-30
Fix Closing in Test

ineiti:
When Server.Close is called, the skipchain sometimes tried to finish
transactions and start new ones. This is fixed by adding a
TestClose method to the skipchain service.

The propagation protocol did not correctly handle errors, which is fixed
by always returning how many actual confirmations we get, even in the
case of errors.


jeffallen:
abool has an MIT license and is 63 lines of Go. Let's just copy and paste it into skipchain/abool.go.

ineiti:
Added it to its own sub-directory, because other services might also need it.

PR #1450, opened 2018-08-31
Cleanup cl* and co* files

kc1212:
I'm not sure why co* and cl* were removed, but bring them back fixes the pop integration test.


jeffallen:
I'm pretty sure I removed that because I thought they were targeting old files that no longer get created. I wonder how I missed that I broke things.

PR #1456, opened 2018-09-05
Adding GenesisDarc to Omniledger.Client

ineiti:
Retrieving the GenesisDarc directly when creating a DefaultGenesisMsg is useful when accessing the api.

Before merging: change commit-base in #1457

### DIFF
--- a/omniledger/service/api.go
+++ b/omniledger/service/api.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
 	"github.com/dedis/protobuf"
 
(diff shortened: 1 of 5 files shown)


omniledger/service/api.go
```diff
@@ -25,18 +26,29 @@ const ServiceName = "OmniLedger"
...
+func NewClientConfig(cfg Config) *Client {
+	return &Client{
+		Client:  onet.NewClient(cothority.Suite, ServiceName),
+		ID:      cfg.ID,
+		Roster:  cfg.Roster,
+		OwnerID: cfg.OwnerID,
+	}
+}
+
+// NewClient is like NewClient, but does not close the connection.
```
kc1212: "NewClient is like NewClient"?

omniledger/service/api.go
```diff
@@ -25,18 +26,29 @@ const ServiceName = "OmniLedger"
...
 func NewClient() *Client {
-	return &Client{Client: onet.NewClient(cothority.Suite, ServiceName)}
+	log.Warn("Deprecated - should use NewClientConfig")
```
jeffallen: We are in the unstable period, so don't deprecate it, just nuke it please.

kc1212: Given that OL is not stable, should we just remove these other ways of creating clients? Same for `NewClientKeep`.

omniledger/service/api.go
```diff
@@ -25,18 +26,29 @@ const ServiceName = "OmniLedger"
...
+	return &Client{Client: onet.NewClientKeep(cothority.Suite, ServiceName)}
 }
 
 // NewClientKeep is like NewClient, but does not close the connection.
 func NewClientKeep() *Client {
+	log.Warn("Deprecated - should use NewClientConfig")
```
jeffallen: Nuke it. (But then how will the caller know they should replace omniledger.Client.Client with a client that does connection caching?

omniledger/service/api.go
```diff
@@ -227,6 +239,8 @@ type Config struct {
 	Roster onet.Roster
 	// OwnerID is the identity that can sign evolutions of the genesis Darc.
 	OwnerID darc.Identity
+	// GenesisDarc is the first darc stored in omniledger.
+	GenesisDarc darc.Darc
```
kc1212: I'm wondering if this should be a darc ID, because the actual darc might change. But I don't know how this is going to be used, I guess it's in one of your other PRs.

ineiti: It's used later in the `pop`-app, where it stores this configuration and then uses it to communicate with omniledger. It saves two requests to get the config- and then the darc of the config. But the information might be stale.

PR #1457, opened 2018-09-05
Implementing coin-types

ineiti:
Coins can be of different types and we need to store that information.

Before merging: change commit-base in #1458
Please also squash this PR.

### DIFF
--- a/omniledger/contracts/coins.go
+++ b/omniledger/contracts/coins.go
@@ -1,7 +1,6 @@
 package contracts
 
 import (
-	"bytes"
 	"crypto/sha256"
 	"encoding/binary"
 	"errors"
@@ -9,6 +8,7 @@ import (
 	"github.com/dedis/cothority/omniledger/darc"
 	ol "github.com/dedis/cothority/omniledger/service"
 	"github.com/dedis/onet/log"
+	"github.com/dedis/protobuf"
 )
 
 // ContractCoinID denotes a contract that can store and transfer coins.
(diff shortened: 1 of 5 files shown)


omniledger/contracts/coins.go
```diff
@@ -71,19 +53,38 @@ func ContractCoin(cdb ol.CollectionView, inst ol.Instruction, c []ol.Coin) (sc [
 	if err != nil {
 		return
 	}
+	var ci CoinInstance
+	if inst.Spawn == nil {
+		// Only if its NOT a spawn instruction is ther data in the instance
```
jeffallen: s/ther/there/

omniledger/contracts/coins.go
```diff
@@ -123,35 +124,37 @@ func ContractCoin(cdb ol.CollectionView, inst ol.Instruction, c []ol.Coin) (sc [
...
 			}
 
-			targetCoin := newSafeUint64(v)
-			targetCoin, err = targetCoin.add(coinsArg)
+			var targetCI CoinInstance
+			err = protobuf.Decode(v, &targetCI)
+			if err != nil {
+				return nil, nil, errors.New("couldn't unmarshal target account: " + err.Error())
+			}
+			targetCI.Balance, err = safeAdd(targetCI.Balance, coinsArg)
```
jeffallen: I wonder if safeAdd should instead be a method on CoinInstance: `func (ci *CoinInstance)add(x int) error`?

omniledger/contracts/coins.go
```diff
@@ -123,35 +124,37 @@ func ContractCoin(cdb ol.CollectionView, inst ol.Instruction, c []ol.Coin) (sc [
...
-				err = errors.New("not enough coins in instance")
-				return
-			}
-			coinsCurrent, err = coinsCurrent.sub(coinsArg)
+			ci.Balance, err = safeSub(ci.Balance, coinsArg)
 			if err != nil {
 				return
 			}
-			cOut = append(cOut, ol.Coin{Name: CoinName, Value: coinsArg})
+			cOut = append(cOut, ol.Coin{Name: ol.NewInstanceID(ci.Type), Value: coinsArg})
```
jeffallen: NewInstanceID converts the first 32 bytes of in into an InstanceID. What if ci.Type is shorter than 32 bytes? I actually had a todo written down a long time ago to understand what this CoinName thing is.

omniledger/contracts/coins.go
```diff
@@ -123,35 +124,37 @@ func ContractCoin(cdb ol.CollectionView, inst ol.Instruction, c []ol.Coin) (sc [
...
 		case "store":
 			// store moves all coins from this instruction into the account.
 			cOut = []ol.Coin{}
 			for _, co := range c {
 				if bytes.Equal(co.Name.Slice(), CoinName.Slice()) {
```
jeffallen: Fetch is no longer using CoinName, but store is?

omniledger/contracts/coins.go
ineiti: Don't know what I've been thinking. It should simply store a `coin`, not invent a new structure and then convert between the two...

omniledger/contracts/coins.go
ineiti: Nice catch.

kc1212:
@ineiti could you remove the `t.Skip` and try again please? Perhaps 400ms is enough for travis.

ineiti:
as this PR passes, can we merge it like that, and I remove the t.Skip in the next PR, `ol_simplifying`? Else we'll never get through these...

PR #1458, opened 2018-09-05
Ol simplifying - depends on #1457

ineiti:
Added an option for ol to save to a specific configuration directory instead of
the default one. Also removed an unnecessary structure.


omniledger/ol/main.go
```diff
@@ -277,39 +292,31 @@ func readRoster(r io.Reader) (*onet.Roster, error) {
...
-	_, val, err := network.Unmarshal(buf, cothority.Suite)
+	_, msg, err := network.Unmarshal(buf, cothority.Suite)
 	if err != nil {
 		return nil, err
 	}
-	if cfg, ok := val.(*configPrivate); ok {
-		return cfg, nil
-	}
-	return nil, errors.New("unexpected private config format")
+	return msg.(*darc.Signer), nil
```
jeffallen: Why doesn't this type assertion need a check?

PR #1459, opened 2018-09-05
Adding pop-contract - depends on #1463

ineiti:
Adding pop-contract First step in having pop in omniledger by adding a simple contract that can save the party-proposition and the final statement. Once the final statement is accepted, the contract will create a new coin-account for every attendee of the party.

Before merging: change commit-base in #1460 
Please also squash this PR.

### DIFF
--- a/pop/service/api.go
+++ b/pop/service/api.go
@@ -170,6 +170,23 @@ func (c *Client) GetFinalStatements(dst network.Address) (map[string]*FinalState
 	return res.FinalStatements, nil
 }
 
+// StoreKeys asks the service to store public keys for a party.
+func (c *Client) StoreKeys(dst network.Address, partyID []byte, keys []kyber.Point) error {
+	si := &network.ServerIdentity{Address: dst}
+	ret := &StoreKeysReply{}
+
+	return c.SendProtobuf(si, &StoreKeys{partyID, keys, nil}, ret)
+}
+
+// GetKeys asks the service for the public keys for a party.
+func (c *Client) GetKeys(dst network.Address, partyID []byte) ([]kyber.Point, error) {
+	si := &network.ServerIdentity{Address: dst}
+	ret := &GetKeysReply{}
+
+	err := c.SendProtobuf(si, &GetKeys{partyID}, ret)
+	return ret.Keys, err
+}
+
 // StoreInstanceID asks the service to store an instanceID for a given party.
 func (c *Client) StoreInstanceID(dst network.Address, partyID []byte, instanceID ol.InstanceID) error {
 	si := &network.ServerIdentity{Address: dst}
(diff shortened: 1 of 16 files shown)


pop/service/api.go
```diff
@@ -144,6 +146,64 @@ func (c *Client) Merge(dst network.Address, p *PopDesc, priv kyber.Scalar) (
...
+}
+
+// GetSigner asks the service for an Signer for a given party.
+func (c *Client) GetSigner(dst network.Address, partyID []byte) (darc.Signer, error) {
+	si := &network.ServerIdentity{Address: dst}
+	ret := &GetSignerReply{}
+
+	err := c.SendProtobuf(si, &GetSigner{partyID}, ret)
+	return ret.Signer, err
+}
```
kc1212: These are new client APIs, but it seems they're not tested. Can we have some simple tests for these to make sure they work?

pop/service/contract.go
```diff
@@ -0,0 +1,215 @@
...
+//        This command has the following arguments:
+//       * "FinalStatement" - mandatory, to give the new final statement. It
+//         needs to be correctly finalized by the pop-service.
+//       * "Service" - when given, will create a darc and a coin-account for
+//         the service to use.
+func (s *Service) ContractPopParty(cdb ol.CollectionView, inst ol.Instruction, coins []ol.Coin) (scs []ol.StateChange, cOut []ol.Coin, err error) {
+	cOut = coins
+	var darcID darc.ID
+	var ppi PopPartyInstance
+	if inst.Spawn == nil {
```
kc1212: is there a missing signature verification? perhaps it's time to do the contract interface thing so that it'll at least do something by default.

pop/service/proto.go
```diff
@@ -5,6 +5,8 @@ This holds the messages used to communicate with the service over the network.
 */
 
 import (
+	"github.com/dedis/cothority/omniledger/darc"
+	omniledger "github.com/dedis/cothority/omniledger/service"
```
kc1212: we started using `ol`

skipchain/protocol.go
```diff
@@ -227,6 +203,46 @@ func (p *ExtendRoster) HandleExtendRosterReply(r ProtoStructExtendRosterReply) e
...
+	replies        int
+	closing        chan bool
+}
+
+// NewProtocolGetBlocks prepares for a protocol that fetches blocks.
+func NewProtocolGetBlocks(n *onet.TreeNodeInstance) (onet.ProtocolInstance, error) {
+	t := &GetBlocks{
+		TreeNodeInstance: n,
+		GetBlocksReply:   make(chan []*SkipBlock),
+		closing:          make(chan bool),
```
kc1212: I don't see any select statements using `closing` in this protocol, is it used?

ineiti:
Last action for the week-end. If by chance this passes, feel free to merge.

PR #1460, opened 2018-09-05
Adding personhood service - after #1459

ineiti:
The personhood service is a first example of how to use pop-tokens and pop-coins. For the first two services, we chose a messaging and a questionnaire service.

Only merge this AFTER #1459


personhood/db.go
```diff
@@ -0,0 +1,67 @@
...
+	network.RegisterMessage(&storage1{})
+}
+
+// saves all data.
+func (s *Service) save() error {
+	s.storage.Lock()
+	defer s.storage.Unlock()
+	err := s.Save(storageKey, s.storage)
+	if err != nil {
+		log.Error("Couldn't save data:", err)
```
kc1212: error is not returned

PR #1462, opened 2018-09-05
Identity hotfix - merge after #1461

ineiti:
Still not sure how to fix that the cleanest way - probably we will have to simply invalidate all old data-versions and start from scratch...

Please merge #1461 first!


ineiti:
will be replaced by a more appropriate identity-fix.

Issue #1409, opened 2018-08-14
Further ByzCoin optimizations


ineiti:
Currently, omniledger does the following in `startPolling`:

1. Ask all other nodes for new transactions
2. Calculate the next block during `blockInterval / 2` seconds
3. Send it off for verification
4. Wait for the block to be verified
5. Wait for `blockInterval` seconds

This is still very inefficient:
- point 1. should be done either in gossipping mode or independent of the block creation
- 2. & 4.: `startPolling` could assume that the block will be accepted and already start calculating the next block
- 5. this will make the actual blockinterval to be: 
`time(getTransactions) + blockInterval / 2 + time(verification) + blockInterval`
So it should make sure that the _tick_ gets in _every_ `blockInterval` instead of this formula.

This work will let us undertake another cleanup we know we need, which is that createStateChanges needs to be broken into selectTransactions and createStateChanges. selectTransactions should be able to drop transactions which are not possible to be dropped right now because of the convoluted API of createState changes. See this review conversation: https://github.com/dedis/cothority/pull/1464#discussion_r216937003
This same refactoring will allow us to correctly fix #1359.

ineiti:
See also #1311

kc1212:
I've been thinking about how to pipeline the transaction processing and block creation.

The rough idea is to use the pipeline approach. There will be one go-routine collecting transactions and sends the txs to a channel, this is the producer. A worker will pick up these txs from the channel and update some ephemeral statetrie to produce state changes. Finally, a consumer, who runs itself every blockinterval, will collect the new txs and state changes to create a block.

The most tricky part of this approach is that we still need to be stateful because the consumer needs the txs for the block, which is exactly what the pipeline pattern tries to avoid... Still thinking

ineiti:
Why do you separate the worker and the consumer? As far as I can see, there is one producer, one worker, and one consumer. This looks like a good indication that you don't need to go too far into producer/consumer land and can keep it much simpler...

PR #1464, opened 2018-09-05
Add a per-ledger block-size limit

jeffallen:
It can be modified via invoke update_contract as shown
in the TestService_SetConfig test.


omniledger/service/service.go
```diff
@@ -642,13 +652,20 @@ func (s *Service) LoadGenesisDarc(scID skipchain.SkipBlockID) (*darc.Darc, error
...
 }
 
-// LoadBlockInterval loads the block interval from the skipchain ID.
-func (s *Service) LoadBlockInterval(scID skipchain.SkipBlockID) (time.Duration, error) {
-	collDb := s.getCollection(scID)
-	if collDb == nil {
-		return defaultInterval, errors.New("nil collection DB")
+// LoadBlockInfo loads the block interval and the maximum size from the skipchain ID.
+func (s *Service) LoadBlockInfo(scID skipchain.SkipBlockID) (time.Duration, int, error) {
+	coll := s.getCollection(scID)
```
jeffallen: Should use a collection view here.

omniledger/service/service.go
```diff
@@ -735,12 +752,13 @@ func (s *Service) startPolling(scID skipchain.SkipBlockID, interval time.Duratio
...
 				cdb := s.getCollection(scID)
-				_, txOut, _ := s.createStateChanges(cdb.coll, scID, txIn, interval/2)
+				then := time.Now()
+				_, txOut, sc := s.createStateChanges(cdb.coll, scID, txIn, interval/2)
 
 				txs = txs[len(txOut):]
 				if len(txs) > 0 {
-					log.Warnf("Got more transactions than can be done in half the blockInterval. "+
-						"%d transactions left", len(txs))
+					sz := stateSize(sc)
```
jeffallen: We should be looking at the size of txOut, not sc.

omniledger/service/service.go
```diff
@@ -930,18 +963,36 @@ clientTransactions:
 			}
 			states = append(states, scs...)
 			cin = cout
+			txsz += stateSize(scs)
```
jeffallen: Calc size of tx, not scs.

omniledger/service/service_test.go
```diff
@@ -1247,6 +1329,7 @@ func newSerN(t *testing.T, step int, interval time.Duration, n int, viewchange b
 		switch i {
 		case 0:
 			resp, err := s.service().CreateGenesisBlock(genesisMsg)
+			// The problem here is that the config is being created with MaxBloclSize == 0. Why? Where?
```
jeffallen: what is this?

omniledger/service/service.go
```diff
@@ -930,18 +985,38 @@ clientTransactions:
 			}
 			states = append(states, scs...)
 			cin = cout
+		}
+		txsz += txrSize([]TxResult{tx})
 
+		// If the size of this one transaction alone is bigger than one block, it has to be rejected so it
+		// cannot sit on the front of the queue blocking all others.
+		if txsz > maxsz {
+			log.Errorf("%s transaction size %v is bigger than one block (%v), dropping it.", s.ServerIdentity(), txsz, maxsz)
```
ineiti: Dropping it means that the transaction will be removed from the list of available transactions? Or will it be included in the next block?

kc1212: Hmm, this is a good point. Where `createStateChanges` is called in `startPolling`, it assumes `createStateChanges` won't skip any client transactions, because it does `txs = txs[len(txOut):]`. If a transaction is skipped, it might be removed and never re-tried, and a later transaction might be replayed. Does this sound right, @jeffallen ?

jeffallen: Even refused transactions go into the block, so a tx that is so big it cannot fit into a block cannot be refused, it has to be dropped entirely, with no trace (except this message in the log).

It seems like the current createStateChanges api can't support that anymore. Looking and thinking still.

jeffallen: The immediate fix we can do is to prevent these from arriving at all in AddTransaction (and getTx). We can do a much bigger refactoring later, which I've added to the notes on #1409.

ineiti:
Just so I understand: now, if a transaction makes the block too big, it will be rescheduled for the next block, right?

jeffallen:
If a transaction that is too big makes it as input to createStateChanges, createStatechanges will exit by way of "stopping block creation when %v > %v, with len(txOut) of %v", every time, blocking the head of the queue until the block size is raised. Sicne block size changes require transactions, it is likely that the block size change would be trapped behind the large tx. :(

However, txns that are too big are stopped at the websocket-border of all conodes, and at the tx collection border of leaders, so that misbehaving followers cannot slide a big tx in.

I think you could probably construct pathological cases where you send through a block size reduction that is racing a transaction. It could maybe happen if the block size reduction arrived to the leader and the tx to a follower. Then the ordering of the arrival of the 2 txs across the root.TxsChan channel would be unpredictable.

PR #1465, opened 2018-09-06
Following fix_protocol_creation from onet

ineiti:
This fixes a break introduced by dedis/onet#471.


kc1212:
I'm a bit confused by this PR, why are we removing those lines? Are the protocols registered somewhere else?

PR #1466, opened 2018-09-06
Add time in wait blocks again

ineiti:
As we're not constantly creating new blocks, AddTransaction with InclusionWait > 0
doesn't work, because we're not getting any new blocks that could decrease the
counter. So adding a timer again.

### DIFF
--- a/omniledger/service/service.go
+++ b/omniledger/service/service.go
@@ -241,7 +241,8 @@ func (s *Service) AddTransaction(req *AddTxRequest) (*AddTxResponse, error) {
 
 		blocksLeft := req.InclusionWait
 
-		for found := false; !found; {
+		var found bool
+		for !found && blocksLeft > 0 {
 			select {
 			case success := <-ch:
 				if !success {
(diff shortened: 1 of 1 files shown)


omniledger/service/service.go
```diff
@@ -252,13 +253,17 @@ func (s *Service) AddTransaction(req *AddTxRequest) (*AddTxResponse, error) {
 				if id.Equal(req.SkipchainID) {
 					blocksLeft--
 				}
-				if blocksLeft == 0 {
-					return nil, fmt.Errorf("did not find transaction after %v blocks", req.InclusionWait)
-				}
+			case <-time.After(interval):
```
kc1212: the `tooLong` channel is already a timer. Do we need two timers? Maybe we can reduce `tooLongDur`?

ineiti: Of course, you're right. Didn't see that :(

ineiti: We could reduce `tooLongDur` to twice the expected, optimal time:

`tooLongDur := time.Duration(req.InclusionWait) * interval * 2`

ineiti: Will do so in `ol_simplifying`.

PR #1467, opened 2018-09-06
Identity db-loading - after #1461

ineiti:
Load db's similar to what the personhood-service is doing.

Please merge #1461 first!


ineiti:
Closed because it's included in the personhood-PR.

PR #1469, opened 2018-09-07
Use a new darc in calypso tests

kc1212:
Instead of the genesis darc for everything which is not realistic, we
create a new darc to test calypso.


omniledger/service/service.go
```diff
@@ -280,11 +280,36 @@ func (s *Service) GetProof(req *GetProof) (resp *GetProofResponse, err error) {
...
+	sb, err := s.db().GetLatestByID(req.ID)
+	if err != nil && sb == nil {
+		return
+	}
+	cdb := s.getCollection(req.ID)
+	if cdb.getIndex() > sb.Index {
+		err = errors.New("invalid index while getting proof")
+		return
+	}
+	// TODO can cdb.getIndex change in this loop?
```
jeffallen: Yes, it most certainly can. Probably the right answer is that NewProof has to return the index number it was working from when it created the proof. But that's not so easy, because the locking for the read/write of the index number is managed by boltdb, and the locking for the duration of the creation of the proof happens inside of c.coll.Get(key).Proof() in NewProof().

Ugh.

It is very much bad and wrong that we have a memory and a disk copy of one thing and that they are not accessed via one unified API which can manage the locking of them as an all or nothing unit.

kc1212: Yeah, I'm not too sure what we can do about it.. One thing is we store the first index and make sure it doesn't change during the looping, that will at least avoid inconsistency but clients will have to try again.

PR #1470, opened 2018-09-07
Fix view-change bug and enable it

kc1212:
There is a bug that might cause new-view messages not being sent when an
anomaly is the last to arrive. This is fixed and is covered by
regression tests. Further, we fix some flakiness in the omniledger
service tests by waiting for skipblocks to finish propagating before
trying to close.


kc1212:
The only test that's failing is on evoting, due to an issue on onet. Leaving this in WIP until the evoting problem is resolved.

PR #1478, opened 2018-09-13
Goodbye OmniLedger, hello ByzCoin

jeffallen:

PR #1480, opened 2018-09-14
Clean up randhound, timestamper. Also ByzCoin doc fixes

jeffallen:
Clean up randhound, timestamper. Also ByzCoin doc fixes.


kc1212:
Could you change this too please - https://github.com/dedis/cothority/pull/1478#discussion_r217678647

kc1212:
Do we want to keep an easily accessibly backup of randhound somewhere, e.g., in its own repo or a repo for nuked code? Someone reading the paper might come and look for the code.

ineiti:
The problem is that I don't know where we have a running version :( We really need a student assistant to take care of these things ;)

PR #1485, opened 2018-09-19
Transactions 1402

### LINKED ISSUE #1402: should be possible to confirm a transaction was committed
It should be possible for clients to ask for proof that a given ClientTransaction was committed.

So we can either:
1. create another merkle tree with all the transactionIDs in it, and then get proofs there
2. search through all latest blocks and see if the transaction has been included there

Option 1 is cleaner but needs more work, while 2 is easier to implement, but takes more time on the server side.

We should also consider if GetProof should do double duty, so as to avoid adding a new method to the service. Something like GetProof(which string, key []byte) so that you can ask for proof of inclusion of a transaction (GetProof("tx", txHash)) or proof of inclusion of data into the global state (GetProof("state", key)).

The key for a transaction will be ClientTransaction.Hash(), already implemented as:
```go
// Hash returns the sha256 hash of all client transactions.
func (cts ClientTransactions) Hash() []byte {
	h := sha256.New()
	for _, ct := range cts {
		h.Write(ct.Instructions.Hash())
	}
	return h.Sum(nil)
}
```

Linus' addition: we need to take care here as the `instructions.Hash` doesn't hash the signatures. And for the `ClientTransaction.Hash` we want to hash _everything_.

ineiti:
Implementation of the subscription to new SkipBlocks. Every SkipBlock can easily be transformed to a ByzCoin-`Block` and from there the `ClientTransactions` can be retrieved and looked if they are new.

Closes #1402

### DIFF
--- a/external/java/src/main/java/ch/epfl/dedis/lib/SkipBlock.java
+++ b/external/java/src/main/java/ch/epfl/dedis/lib/SkipBlock.java
@@ -2,10 +2,13 @@ package ch.epfl.dedis.lib;
 
 import ch.epfl.dedis.lib.exception.CothorityCryptoException;
 import ch.epfl.dedis.lib.exception.CothorityException;
+import ch.epfl.dedis.lib.skipchain.ForwardLink;
 import ch.epfl.dedis.proto.SkipchainProto;
 import com.google.protobuf.InvalidProtocolBufferException;
 
 import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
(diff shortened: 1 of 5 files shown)


kc1212:
In general, I'm ok with this implementation. But I'm not sure if it captures some other ways people might want to use subscription. In my opinion, this approach is more like event handling, where you register callbacks to events. Another type of API would be like the iterator or generator pattern, but it should be pretty easy to build from callbacks.

ineiti:
An iterator might in fact be a good idea...

PR #1487, opened 2018-09-26
Stream byzcoin blocks in Java/Go using new onet websocket

kc1212:
Add streaming Go/Java API to byzcoin. The Go API is new, clients give a callback and it is called whenever a new event/block arrives. The Java API is the same as the one in https://github.com/dedis/cothority/pull/1485 except it doesn't poll for new blocks but uses the new streaming API.


kc1212:
This is a very WIP java/go API for the byzcoin service that simply streams new transactions IDs to clients.

Nothing is tested yet and it'll be a miracle if it work on the very first, but you can have a look at the API. Namely, it's possible to use a java stream to represent the result. 

However, the java way of dealing with infinite/lazy streams isn't very nice in JDK 8. Namely it doesn't have `takeWhile`, so making the stream stop upon an error is going to be tricky (I'm not aware of a way to do it yet). JDK 9 has it but can we use it?

kc1212:
It was decided that we should stick to JDK 8 to support android use-cases. Unfortunately, JDK 8 doesn't have good support for Java streams, and streams doesn't fit well with the websocketclient. So we stick with the subscription interface for now. Nevertheless, it should be possible to implement a java stream API using the subscription/callback interface if necessary.

PR #1506, opened 2018-10-05
Proxy Authentication

jeffallen:
Proxy Authentication is a new mechanism for Byzcoin
clients to get their transaction signed. Before, they had
to posess (and protect) a secret key. Now they can instead
be authenticated by a third-party identity provider (for
example OpenID Connect). The Authentication Proxy
takes the third party auth info and a message as input
verifies the authentication info, and then creates a
partial signature that binds the user id to the message.
The client then assembles these partial signatures into
a final signature that can be included with the transaction.

On the Darc verification side, the user-id binding
is checked with respect to the public key that corresponds
to all the private key shares in the auth proxies.


authprox/apadmin/test.sh
```diff
@@ -0,0 +1,40 @@
...
+	runGrepSed "export BC=" "" ./bcadmin create --roster public.toml --interval 0.5s
+	eval $SED
+	[ -z "$BC" ] && exit 1
+
+	testOK ./bcadmin add spawn:authproxAdd -identity $ID
+}
+
+testAdd(){
+	runCoBG 1 2 3
+	testOK ./apadmin add --roster public.toml -issuer https://oauth.dedis.ch/
```
kc1212: Does this mean that this test will depend on an external server? it'll be better if we can run the test without external dependencies, but I'm not sure whether it's easy to do in this case.

authprox/oidc.go
```diff
@@ -0,0 +1,64 @@
...
+		is.p, err = oidc.NewProvider(is.ctx, is.Issuer)
+		if err != nil {
+			return
+		}
+		is.cfg.SkipClientIDCheck = true
+		is.v = is.p.Verifier(&is.cfg)
+		o.issuers = append(o.issuers, is)
+	}
+
+	idToken, err := is.v.Verify(is.ctx, string(input))
```
kc1212: I was reading the documentation of the `Verify` function, it says "Verify does NOT do nonce validation, which is the callers responsibility". I don't know where the nonce is stored or what exactly needs to be verified, but are we doing it here? If it's not needed, could you add a comment to explain why?

authprox/proto.go
```diff
@@ -0,0 +1,77 @@
...
+
+// PROTOSTART
+// package authprox;
+
+// EnrollRequest is the request sent to this service to enroll
+// a user, authenticated by a certain type of external authentication.
+type EnrollRequest struct {
+	Type         string
+	Issuer       string
+	Secret       kyber.Scalar
```
kc1212: Since it's still possible to run conodes without TLS, would it be better to encrypt the secret using the public key of the conode? Or print a warning if the authproxy services are running without TLS.

authprox/service.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+func (s *service) find(typ, issuer string) (*dssConfig, error) {
+	k, err := network.Marshal(&ti{T: typ, I: issuer})
+	if err != nil {
+		return nil, err
+	}
+	var buf []byte
+	err = s.db.View(func(tx *bolt.Tx) error {
+		v := tx.Bucket(s.bucket).Get(k)
```
kc1212: to be consistent with other services, it's best to put new buckets under a service specific namespace using `GetAdditionalBucket`.

authprox/service.go
```diff
@@ -0,0 +1,322 @@
...
+
+type dssConfig struct {
+	Secret       kyber.Scalar
+	Participants []kyber.Point
+	LongPri      share.PriShare
+	LongPubs     []kyber.Point
+}
+
+func (s *service) find(typ, issuer string) (*dssConfig, error) {
+	k, err := network.Marshal(&ti{T: typ, I: issuer})
```
kc1212: it's best to use `protobuf.Encode`, `network.Marshal` won't produce valid protobuf format because it has an additional header (which we probably don't need)

eventlog/el/main.go
```diff
@@ -347,3 +411,327 @@ func search(c *cli.Context) error {
 	}
 	return nil
 }
+
+func login(c *cli.Context) error {
```
kc1212: is this the part that will be moved?

eventlog/el/main.go
```diff
@@ -33,6 +44,16 @@ type bcConfig struct {
 	ByzCoinID skipchain.SkipBlockID
 }
 
+// Presets for -clientid and -clientsecret.
+var clientIDs = map[string]string{
```
kc1212: these should be in config files

eventlog/el/main.go
```diff
@@ -347,3 +411,327 @@ func search(c *cli.Context) error {
...
+
+		// Connect to servers, get sigs, stop when we achieve the threshold.
+		// TODO: This could be in parallel.
+		var partials []*share.PriShare
+		for _, idx := range shuffled {
+			s := r.List[idx]
+
+			if len(partials) >= T {
+				// Got enough, all done.
+				break
```
kc1212: breaking here means we're trusting all the proxies, which I don't think is in the threat model. we should collect all the partial signatures and then remove any that are not valid (not sure how, some kind of NIZK proof?). does DSS have this kind of functionality?

eventlog/el/main.go
kc1212: we should have an integration test for these (but it might be difficult because it relies on external servers, can we mock a server?)

authprox/service.go
kc1212: nevermind, this is fine

authprox/apadmin/test.sh
jeffallen: Enrolment does not contact the external service, so no, in this exact case.

An integration test that was fully end-to-end would need to talk to some kind of third-party auth server. I was thinking of making some kind of no-op kind of verification to compile in via build tags only during integration testing. (authprox/service_test.go is something like this, but in the unit test context)

authprox/oidc.go
jeffallen: I've updated the docs a bit. This is also discussed in el/main.go, at the call to AuthCodeURL.

authprox/proto.go
jeffallen: Given it is a one-time administrative operation, could we just document that it should be done across localhost? I know this is a very lazy solution, but I'm not sure this risk is one worth complicating the code for.

I could make apadmin refuse to send if !(tls || localhost).

authprox/service.go
jeffallen: This is done in newService.

authprox/service.go
jeffallen: Yeah, I always get confused about this. I converted it towards network.Marshal already once because I remembered this problem, but I remembered it backwards. Fixing.

authprox/proto.go
kc1212: I'm fine with either, as long as it's clear to the user, either from the docs or the code.

authprox/service.go
```diff
@@ -0,0 +1,322 @@
...
+
+	var out *dssConfig
+	var ok bool
+	if out, ok = x.(*dssConfig); !ok {
+		return nil, fmt.Errorf("unexpected type %T", x)
+	}
+	return out, nil
+}
+
+func (s *service) Enrollments(req *EnrollmentsRequest) (*EnrollmentsResponse, error) {
```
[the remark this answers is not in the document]
jeffallen: It is a private type, which is why lint did not catch it. I think I agree that Service entry points need to be documented, but am wondering now a bit about pub/priv on the type. (I'm happy it is private, but pondering if this will hurt us somehow?)

authprox/service.go
```diff
@@ -0,0 +1,322 @@
...
+	lpri := share.PriShare{
+		I: req.LongPri.I,
+		V: req.LongPri.V,
+	}
+	v, err := network.Marshal(&dssConfig{
+		Secret:       req.Secret,
+		Participants: req.Participants,
+		LongPri:      lpri,
+		LongPubs:     req.LongPubs,
+	})
```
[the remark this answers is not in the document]
jeffallen: Yes, the auth proxies can/should trust the input they get: The auth proxies are offering, one at a time, to give their opinion on which "claim" a given auth info maps to. The client gathers the results, and when it has a threshold, passes that on to byzcoin/darc who will ask, "for this username and this message, did a quorum of auth proxies attest that this user was authenticated?"

So if an admin sets up an auth proxy to give useless partial sigs, no one is hurt by this.

byzcoin/bcadmin/main.go
```diff
@@ -219,11 +225,23 @@ func add(c *cli.Context) error {
 	}
 	action := arg[0]
 
-	identity := c.String("identity")
-	if identity == "" {
+	ids := c.StringSlice("identity")
```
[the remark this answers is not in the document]
jeffallen: It defaults to OR, so we end up making an OR of one element.

external/java/src/main/java/authprox/Authprox.java
```diff
@@ -0,0 +1,7429 @@
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
+// source: authprox.proto
+
+package authprox;
```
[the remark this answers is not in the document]
jeffallen: No idea. Fixing.

eventlog/el/main.go
jeffallen: Honestly, my plan was to leave it until we had another Go caller want to do the same, and then use the opportunity of two callers to clean it up into a shared library. It is a bit hard to imagine how different callers will want to do it; login is an interactive process.

eventlog/el/main.go
jeffallen: You can override them from the commandline. They are just helpers so that if you don't put them on the commandline, the right thing will still happen.

eventlog/el/main.go
jeffallen: DSS callers are expected to check the signature on the PartialSignature, which we are not yet doing. But that does not address the case where a good signature on an incorrect PartialSignature makes the entire reconstruction fail.

I will put a TODO here for now, and discuss with cryptographers to see what they say.

eventlog/el/main.go
jeffallen: Filed issue on this: #1520

jeffallen:
The builds will fail until dedis/onet#480 is merged.

authprox/README.md
```diff
@@ -0,0 +1,50 @@
+# Authentication Proxy
```
ineiti: It would be nice to have that integrated into the rest of the Cothority's README system with the links. See `calypso/README.md` and others.

authprox/README.md
```diff
@@ -0,0 +1,50 @@
...
+The first user of this is byzcoin/darc/darc.go's type IdentityProxy, where the
+Verify function implements the verification side of this scheme.
+
+## Enrolling new authentication systems
+
+In order to start generating partial signatures attesting to some authentication
+information, a set of authentication proxies need to have shares of a secret
+stored into them. An administrator of the system will use the `apadmin` tool to do this.
+
+During the enrollment process, a secret is generated in the RAM of the `apadmin`
```
ineiti: I still think that we should use a DKG here, but I trust Lefteris that this is not necessary.

authprox/README.md
```diff
@@ -0,0 +1,50 @@
...
+tool, sharded, and sent to the authentication proxies. It then exits, causing the
+only copy of the complete secret to be lost. All users of the system need
+to trust that `apadmin` has discarded the original secret key. (Having the authentication
+proxies execute teh DKG protocol among them would remove this risk, but we
```
ineiti: s/teh/the/

authprox/README.md
```diff
@@ -0,0 +1,50 @@
...
+decided that was overkill for this job.)
+
+The roster of the authentication proxies can be disjoint from the roster of
+the distributed system that will consume (i.e. verify) the generated signatures.
+(However, the only example client for the moment is `el`, which assumes that
+the Authentication Proxy roster is one and the same as the Byzcoin roster.)
+
+The threshold for reassembling full signatures from the partial signatures
+is fixed during enrollment. For n servers, the threshold is set at
+n - (n-1)/3, i.e. for 7 servers, 5 signatures are required.
```
ineiti: Using the new DKG @nikkolasg and @Daeinar are putting up, we could even change that during runtime.

authprox/README.md
```diff
@@ -0,0 +1,50 @@
...
+their name is. The clients then present this information, and a message to be
+signed, to the Authentication Proxies one after another (or in parallel).  When
+the client has gathered the threshold of these partial signatures, it
+reassembles them into a final signature. It can then present that final
+signature to a system that wants proof of the ID of the human driving the
+client.
+
+The first (and likely only) system that consumes these user/message binding
+signatures is the `proxy` Darc Identity type. Thus, Byzcoin instances which are
+protected by a Darc with a rule like "proxy$pubkey:user@example.com" will only
```
ineiti: Is that the format of the proxy? I thought that others are `ed25519:abcde`, so I would've thought we have `proxy:abcde/user@example.com` here.

authprox/apadmin/main.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	// Get all the key material ready.
+
+	// n: how many auth proxies will be holding shares
+	n := len(roster.List)
+	T := threshold(n)
+
+	pubs := make([]kyber.Point, n)
+	for i := 0; i < n; i++ {
```
ineiti: Nobody likes `for i := range pubs`? I already had this fight with Philipp... Seems I lost.

authprox/apadmin/test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+. "$(go env GOPATH)/src/github.com/dedis/cothority/libtest.sh"
+
+main(){
+	build $APPDIR/../../byzcoin/bcadmin
+	startTest
+	buildConode github.com/dedis/cothority/byzcoin github.com/dedis/cothority/authprox
+
+	# This must succeed before the others will work.
+	run BCSetup
```
ineiti: Usually it's nice to have every test being able to run independently. This makes it easier to run them if something goes wrong...

authprox/service.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+	network.RegisterMessages(
+		&EnrollRequest{}, &EnrollResponse{},
+		&SignatureRequest{}, &SignatureResponse{},
+		&EnrollmentsRequest{}, &EnrollmentsResponse{}, &EnrollmentInfo{},
+		&ti{}, &dssConfig{},
+	)
+}
+
+// A Validator is able to check the provided authInfo with respect to
```
ineiti: Another detail (nothing needs to be changed): I like starting service-files with constants, `init`, structures, `NewStruct`, public methods, then private methods and the `newService` at the bottom. This helps me navigate the file.

authprox/service.go
```diff
@@ -0,0 +1,354 @@
...
+		return nil, err
+	}
+
+	// Use the validator to extract a claim from the auth info.
+	claim, _, err := validator.FindClaim(req.Issuer, req.AuthInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: check that if extraData != "", then extraData == msg
```
ineiti: Might be more important - leftover TODO

authprox/service.go
```diff
@@ -0,0 +1,354 @@
...
+	msg2 := h.Sum(nil)
+
+	// Find the config that is associated with the issuer.
+	dsscfg, err := s.find(req.Type, req.Issuer)
+	if err != nil {
+		return nil, fmt.Errorf("cannot find key: %v", err)
+	}
+
+	// Make the signature.
+	suite := suites.MustFind("ed25519")
```
ineiti: Why don't you use `cothority.Suite` here?

byzcoin/bcadmin/main.go
```diff
@@ -65,13 +64,28 @@ var cmds = cli.Commands{
 				EnvVar: "BC",
 				Usage:  "the ByzCoin config to use",
 			},
-			cli.StringFlag{
+			cli.StringSliceFlag{
 				Name:  "identity",
-				Usage: "the identity of the signer who will be allowed to access the contract (e.g. ed25519:a35020c70b8d735...0357))",
+				Usage: "the identity(ies) of the signer(s) who will be allowed to access the contract (e.g. ed25519:a35020c70b8d735...0357))",
+			},
+			cli.StringFlag{
```
ineiti: I wrote to one of the students also working on this to simply use a `string` here, as the user can then do whatever he wants with regard to the identity.

authprox/README.md
ineiti: In fact the actual implementation is `proxy:abcde:user@example.com`, which looks good.

authprox/apadmin/main.go
jeffallen: Yes, that's better. When you do it that way, you give the compiler enough info to be able to remove the bounds checks, because it can prove that 0 <= i  < len(pubs).

authprox/apadmin/test.sh
jeffallen: I will let the next person that's hacking on this decide on that for themselves. At this point, yagni.

authprox/service.go
jeffallen: Oops, thanks. Tomorrow, first thing.

authprox/service.go
jeffallen: Because kyber/sign/dss makes EdDSA compatible signatures, and thus only works with ed25519. That's the same reason the rule format is `proxy:abcde:user@example.com` and not `proxy:ed25519:abcde:user@example.com`.

byzcoin/bcadmin/main.go
jeffallen: I don't understand this comment...

PR #1513, opened 2018-10-10
Implement Ephemeral Key Re-encryption

### LINKED ISSUE #1498: Add Ephemeral re-encryption to Calypso
The new Calypso is missing re-encryption to an ephemeral key. But this is needed for:
- X509 keys
- when complying with the review-recommendations

ineiti:
Add the missing java classes for the Ephemeral Key Re-encryption.

Closes #1498 

I wish all PRs would be as easy as this one ;)


jeffallen:
One build failed because of flakiness, I have restarted. The other failed because of this new code, so we need to figure that out.

PR #1523, opened 2018-10-12
Add configuration contract

ineiti:
Adds the possibility to change the configuration instance of ByzCoin. There are three configuration options:
- maximum blocksize - fully supported
- block interval - fully supported
- change roster - only permutations allowed - if the set of nodes changes, bad things will happen


jeffallen:
@ineiti Should this be closed?

PR #1525, opened 2018-10-12
Fix small name issue

decltypeme:
Fixing a small typo in the plotting python script by changing the name of a file from `service.csv` to `coins.csv`


kc1212:
Looks good, please add a small description so that we can see it in the git history and then I'll merge.

decltypeme:
Thanks, added description.

Issue #1522, opened 2018-10-11
Create a go API for Calypso


decltypeme:
To be able to use in simulations, we need to create a go client api for the Calypso service.

Needed API functions:
1. `CreateLTS`
2. `DecryptKey`
3. `AddWrite`
4. `AddRead`
5. `SpawnDarc`

PR #1526, opened 2018-10-12
Create Calypso API

decltypeme:
This implements the API requested in #1522 .

- Added message structs
- Added go api functions for `CreateLTS`, `AddWrite`, `AddRead`, `DecryptKey`, `SpawnDarc`
- Added Unit Tests


calypso/api.go
```diff
@@ -0,0 +1,128 @@
...
+	read := &Read{
+		Write: byzcoin.NewInstanceID(write.InclusionProof.Key),
+		Xc:    signer.Ed25519.Point,
+	}
+	reply = &ReadReply{}
+	readBuf, err = protobuf.Encode(read)
+	if err != nil {
+		return nil, err
+	}
+	gDarc, err := c.byzcoin.GetGenDarc()
```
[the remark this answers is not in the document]
decltypeme: Good call, will this still work if the data provider change the darc for that piece of data?

calypso/api_test.go
```diff
@@ -0,0 +1,126 @@
...
+// waitInstID this is a utility testing function to wait for the proof
+// of the write transaction
+// TODO(islam): This function is the same procedure as in calypso/service_test
+// We want to create an interface for both ts and client and have them both have
+// waitInstID once
+func (c *Client) waitInstID(t *testing.T, instID byzcoin.InstanceID,
+	wait time.Duration) *byzcoin.Proof {
+	var err error
+	var pr *byzcoin.Proof
+	for i := 0; i < 10; i++ {
```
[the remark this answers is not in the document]
decltypeme: good call, thanks for noting this, would have been painful to maintian

calypso/service.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 // Used for tests
-var calypsoID onet.ServiceID
+var CalypsoID onet.ServiceID
```
kc1212: should be kept private if it's only used for test, the test is in the same package, so that should be fine, right?

calypso/api.go
```diff
@@ -0,0 +1,193 @@
...
+
+// SpawnDarc spawns a Darc Instance by adding a transaction on the byzcoin client.
+// Input:
+//   - signer - The signer authorizing the spawn of this darc (calypso "admin")
+//   - controlDarc - The darc governing this spawning
+//	 - spawnDarc - The darc to be spawned
+//   - wait - The number of blocks to wait -- 0 means no wait
+//
+// Output:
+//   - reply - WriteReply containing the transaction response and instance id
```
kc1212: this comment is wrong

calypso/proto.go
```diff
@@ -46,13 +46,25 @@ type Write struct {
...
 type Read struct {
 	Write byzcoin.InstanceID
 	Xc    kyber.Point
 }
 
+// ReadReply is is returned upon successfully spawning a Read instance.
+type ReadReply struct {
+	*byzcoin.AddTxResponse
+	byzcoin.InstanceID
+}
```
kc1212: the structures in `proto.go` are used to generate protobuf files and java implementation. but `WriteReply` and `ReadReply` are not sent over the wire, so they don't need to be here. just keeping these in `api.go` is enough

calypso/api.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/dedis/cothority/byzcoin"
+	"github.com/dedis/cothority/darc"
+	"github.com/dedis/onet"
+	"github.com/dedis/protobuf"
+)
+
+// Client is a class to communicate to the calypso service.
+type Client struct {
+	byzcoin  *byzcoin.Client
+	onet     *onet.Client
```
kc1212: you don't need a onet.Client, just use the byzcoin Client (it should have SendProtobuf too). actually, it might be better if you just embed the byzcoin client to calypso. e.g.,
```
 type Client struct { 
  *byzcoin.Client
  ltsReply *CreateLTSReply
}

calypso/api.go
decltypeme: good call, thanks, fixing it

calypso/proto.go
decltypeme: thanks, moving them

calypso/api.go
decltypeme: I was thinking if someone wants to separate them and not use both on the same client, it may be good to allow for this, what do you think?

kc1212: We usually keep the features minimal until we know something is necessary for sure. Further, the Java API is quite similar - https://github.com/dedis/cothority/blob/master/external/java/src/main/java/ch/epfl/dedis/calypso/CalypsoRPC.java#L24 to what I described, the byzcoin client is the super class of the calypso client

PR #1529, opened 2018-10-16
Add streaming draft for ByzCoinRPC cothority-1516

### LINKED ISSUE #1516: Implement Java stream for streaming blocks
Instead of registering a callback, use the API to return a java-stream.

Keep both interfaces, so that the user can chose.

Gilthoniel:
Fixes #1516

### DIFF
--- a/external/java/src/test/java/ch/epfl/dedis/byzcoin/ByzCoinRPCTest.java
+++ b/external/java/src/test/java/ch/epfl/dedis/byzcoin/ByzCoinRPCTest.java
@@ -125,7 +125,7 @@ public class ByzCoinRPCTest {
         Stream<SkipBlock> stream = bc.subscribeSkipBlock();
 
         // create one block
-        bc.getGenesisDarcInstance().evolveDarcAndWait(bc.getGenesisDarc(), admin, 10);
+        bc.getGenesisDarcInstance().evolveDarcAndWait(bc.getGenesisDarc(), admin, 0);
 
         // no need to wait as it will hang until one block is accepted
         assertEquals(1, stream.limit(1).count());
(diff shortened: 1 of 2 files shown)


external/java/src/test/java/ch/epfl/dedis/byzcoin/ByzCoinRPCTest.java
```diff
@@ -13,15 +13,19 @@
 import ch.epfl.dedis.lib.darc.SignerEd25519;
 import ch.epfl.dedis.lib.exception.CothorityCommunicationException;
 import ch.epfl.dedis.lib.exception.CothorityCryptoException;
+import org.junit.Rule;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.omg.CORBA.TIMEOUT;
```
kc1212: is this one, and Rule used?

external/java/src/test/java/ch/epfl/dedis/byzcoin/ByzCoinRPCTest.java
```diff
@@ -119,6 +123,22 @@ void subscribeSkipBlocks() throws Exception {
...
 
+    /**
+     * Subscribe to new blocks using a stream
+     */
+    @Test
+    void subscribeSkipBlockStream() throws Exception {
+        Stream<SkipBlock> stream = bc.subscribeSkipBlock();
+
+        // create one block
+        bc.getGenesisDarcInstance().evolveDarcAndWait(bc.getGenesisDarc(), admin, 10);
```
kc1212: what's more interesting to do is use 0 instead of 10 in `evolveDarcAndWait`. it'll return immediately but the stream should block until a new block is accepted.

PR #1533, opened 2018-10-18
allow view change flag per skipchain cothority-1474

### LINKED ISSUE #1474: enableViewChange should be per-skipchain
From skipchain.go:

```
// If this flag is set, then we relax our forward-link signature verification
// to accept the aggregate signature by the public keys of the rotated roster.
// View-change should only be enabled if there are no security implications
// when the forward-links are signed by the rotated roster instead of the
// roster in the skipblock that points to it. The security implication depends
// on the services that use skipchain, so it is disabled by default. The option
// can be changed from outside of the package by calling EnableViewChange. This
// flag should ideally be the service configuration, but other structs depend
// on it too, e.g., SkipBlock, so we keep it in a global.
var enableViewChange bool
var enableViewChangeOnce sync.Once
```
The task is to make this be a per-skipchain flag, and not a global for the entire service.

Gilthoniel:
This PR changes the enableViewChange flag to be set per skipchain and to be in the scope of the service instead of a global variable.

The service must be pass through specific functions of the different structs so it doesn't seem very clean but that's the only way I found to avoid the global declaration and keep the verification functions at the struct level.

Let me know if I can improve this

Fixes #1474

### DIFF
--- a/byzcoin/proof.go
+++ b/byzcoin/proof.go
@@ -81,7 +81,7 @@ func (p Proof) Verify(scID skipchain.SkipBlockID) error {
 			publics = l.NewRoster.Publics()
 			continue
 		}
-		if err = l.Verify(cothority.Suite, publics); err != nil {
+		if err = l.Verify(cothority.Suite, publics, true); err != nil {
 			return ErrorVerifySkipchain
 		}
 		if !l.From.Equal(sbID) {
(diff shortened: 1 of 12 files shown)


kc1212:
In general I'm ok with this approach. But is the cost (somewhat unclean code) worth the benefit? I don't have a good answer to that question. We should discuss it with the other engineers when they're back.

Gilthoniel:
Close #1474

Gilthoniel:
Close #1474

PR #1534, opened 2018-10-18
Implementation of the coniks trie

kc1212:
This is an implementation of the Merkle Prefix Tree from CONIKS. Which will eventually replace the collections library.


kc1212:
You can find performance results here - https://gist.github.com/kc1212/ec7f62bf73deee559c676b9ca61cc95c

In summary `collectionDB` is quicker for smaller workloads (58ms for 50x10 transactions), but its performance becomes very bad when the workload increases (19 seconds for 200x100 transactions). On the other hand, disk-based `Trie` is slightly slower for smaller workloads (75ms for 50x10) but it does not slow down as fast as `collectionDB` (7 seconds for 200x100).

byzcoin/trie/node.go
```diff
@@ -0,0 +1,165 @@
...
+func decodeInteriorNode(buf []byte) (interiorNode, error) {
+	if len(buf) == 0 {
+		return interiorNode{}, errors.New("empty buffer")
+	}
+	if nodeType(buf[0]) != typeInterior {
+		return interiorNode{}, errors.New("wrong node type")
+	}
+	var node interiorNode
+	if err := protobuf.Decode(buf[1:], &node); err != nil {
+		return interiorNode{}, err
```
Gilthoniel: Can't you return `nil` as there is an error anyway ?

ineiti:
`Merge.Lock()`

jeffallen:
Am reviewing now... 🕐

byzcoin/trie/node.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+}
+
+type interiorNode struct {
+	Left  []byte
+	Right []byte
+}
+
+func (n *interiorNode) hash() []byte {
+	h := sha256.New()
```
ineiti: Why does this hash not include the `typeInteriorNode`?

byzcoin/trie/node.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (n *emptyNode) hash(nonce []byte) []byte {
+	h := n.hashInterface(nonce)
+	return h.Sum(nil)
+}
+
+func (n *emptyNode) hashInterface(nonce []byte) hash.Hash {
+	h := sha256.New()
+	h.Write(typeEmpty.toBytes())
```
ineiti: Style: Isn't `[]byte{typeEmpty}` also good here? I suppose we'll never have multi-byte identifiers...

byzcoin/trie/node.go
ineiti: Same question for `leafNode`

byzcoin/trie/proof.go
```diff
@@ -0,0 +1,181 @@
...
+			return false, errors.New("invalid prefix in empty node")
+		}
+		return false, nil
+	} else {
+		return false, errors.New("invalid edge node")
+	}
+}
+
+// Match is a convenience function that ignores the error of Exists and just
+// returns false.
```
ineiti:
```suggestion
// returns false in the case of an error.
```

kc1212:
> `Merge.Lock()`

Is this related to invalidating staging tries if there is more than one of them getting created from the same (source) trie? for example,
```
var t Trie
staging1 := t.MakeStagingTrie()
staging2 := t.MakeStagingTrie()
// do some operations on both
staging1.Commit()
staging2.Commit() // this should fail because the source trie t is now different
```
I think eventually we will have this feature. But `Commit` isn't used in `byzcoin` yet so I haven't implemented a way to invalidate a staging trie. I'm thinking of replacing the `stateChangeCache` to store a staging trie instead of the state changes (probably in a future PR). When we have that, `Commit` will be used and we should make sure that the commitments are always happening on the correct trie.

Nevertheless, it shouldn't be difficult to add this check. It's just a matter of storing the initial root in the staging trie and check that the root is still the same when commit is called.

ineiti:
`Merge.Unlock` - hoping that Jeff is carrying on well ;)

byzcoin/trie/trie.go
```diff
@@ -0,0 +1,648 @@
...
+		val = clone(buf)
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return val, nil
+}
+
+func (t *Trie) binSlice(buf []byte) []bool {
```
jeffallen: I'm unclear why this code to change []byte into []bool exists; it seems like the only thing we do with the resulting []bool is to do lookups into it, i.e. `if bits[i] ...etc...`. It seems to me like checking if bit #n in []byte x is set should not require a []bool.

byzcoin/trie/trie.go
```diff
@@ -0,0 +1,648 @@
+package trie
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"errors"
+)
+
+const entryKey = "dedis_trie"
```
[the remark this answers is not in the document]
jeffallen: What these are actually doing is capturing metadata about the trie. And there is one other piece of metadata that probably needs to be tightly integrated (i.e.atomically) with the trie, which is the latest index (see getIndex in byzcoin/struct.go)

byzcoin/trie/diskdb.go
```diff
@@ -0,0 +1,84 @@
...
+	dryRunErr := errors.New("this is a dry-run")
+	err := r.db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket(r.bucket)
+		if b == nil {
+			return errors.New("bucket does not exist")
+		}
+		if err := f(&diskBucket{b}); err != nil {
+			return err
+		}
+		return dryRunErr
```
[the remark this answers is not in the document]
kc1212: the documentation for Update says "Attempting to manually commit or rollback within the function will cause a panic. "

byzcoin/trie/diskdb.go
```diff
@@ -0,0 +1,84 @@
...
+type diskBucket struct {
+	b *bolt.Bucket
+}
+
+func (r *diskBucket) Delete(k []byte) error {
+	return r.b.Delete(k)
+}
+
+func (r *diskBucket) Put(k, v []byte) error {
+	return r.b.Put(k, v)
```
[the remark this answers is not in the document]
kc1212: I think it's better to keep it the same as the boltdb API, so I'm not copying those values. but for `memBucket`, we need to copy during `Put` to make sure it doesn't get modified unintentionally, which I will fix

byzcoin/trie/node.go
kc1212: actually, there is an implementation error, will be fixed as a part of - https://github.com/dedis/cothority/pull/1534#discussion_r227795381

byzcoin/trie/node.go
kc1212: `return _, err` doesn't seem to compile, what I can do to simplify things is to use named return values, like `func decodeInteriorNode(buf []byte) (node interiorNode, err error) `

Gilthoniel: My bad I meant `nil, err`

byzcoin/trie/trie.go
kc1212: I initially thought it was possible to remove the `binSlice` functions, but then I realised they're not only used for iteration, it's also important to store and use `[]bool` that are not on "byte-boundaries". that's because the way we make new nodes is when their prefix becomes different, so two nodes may have the same prefix for 5 bits, then the 6th one is different. in that case we have to store a `[]bool` of length 6 in the leaf node, see `extendLeaf`. it wouldn't be possible to do with just a `[]byte` without additional information. bool in go is one byte (https://play.golang.org/p/jkfkv5LJSqU), hopefully that shouldn't be too bad on the storage

byzcoin/trie/trie.go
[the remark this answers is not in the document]
kc1212: see https://github.com/dedis/cothority/pull/1534#discussion_r227803225

byzcoin/trie/memdb.go
```diff
@@ -0,0 +1,106 @@
...
+func (r *memDB) View(f func(bucket) error) error {
+	r.Lock()
+	defer r.Unlock()
+
+	r.bucket.writable = false
+	return f(r.bucket)
+}
+
+func (r *memDB) UpdateDryRun(f func(bucket) error) error {
+	r.Lock()
```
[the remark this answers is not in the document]
jeffallen: yes a general explanation of the way dry run is implemented, the tradeoff and the consequences (i.e. cannot detect if underlying was changed, etc)

jeffallen:
I'm not sure if this PR is waiting on more doc updates or not; if not @kc1212 can/should just merge it now.

kc1212:
I was integrating changes here into the `trie_byzcoin` to make sure that that i'm not missing anything. i'm at the point of fixing some tests now (and writing new ones), so this PR should be pretty feature/doc-complete.

PR #1543, opened 2018-10-24
Personhood verification

ineiti:
PoP-service:
* Verify the party is really finalized when it is written to the
ByzCoin.
* Re-enabled the test in pop/test.sh.
* Store organizers' keys

Personhood service:
* Correctly calculate service-darc

### DIFF
--- a/pop/test.sh
+++ b/pop/test.sh
@@ -11,518 +11,547 @@ NBR_SERVERS_GROUP=$NBR_SERVERS
 . "$(go env GOPATH)/src/github.com/dedis/cothority/libtest.sh"
 
 MERGE_FILE=""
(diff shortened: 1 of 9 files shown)


pop/test.sh
```diff
@@ -11,518 +11,547 @@ NBR_SERVERS_GROUP=$NBR_SERVERS
...
+  run testAuthStore
+  run testAtVerify
+  run testAtMultipleKey
+  run testMerge
+  run testPropagateConfig
+  run testOlStore
+  stopTest
+}
+
+testOlStore() {
```
jeffallen: s/ol/bc/    gee!

PR #1544, opened 2018-10-24
Byzcoin interval 1514

### LINKED ISSUE #1514: Follow new interval in ByzCoin
Currently the ByzCoin service always keeps the same block interval, even if the configuration changes.

Wanted: follow new intervals stored in the `Configuration`.

ineiti:
Adds the configuration-contract to java and implements interval changes when the configuration instance is changed.

Closes #1514


byzcoin/heartbeat.go
```diff
@@ -17,17 +17,18 @@ type heartbeat struct {
 	getTimeChan chan chan time.Time
 	timeout     time.Duration
 	timeoutChan chan string
+	updateTO    chan time.Duration
```
jeffallen: s/updateTO/updateTimeoutChan/ please. It's too hard to know if to is like to/From or to is timeout.

byzcoin/heartbeat.go
```diff
@@ -17,17 +17,18 @@ type heartbeat struct {
...
 }
 
 type heartbeats struct {
 	sync.Mutex
 	wg           sync.WaitGroup
-	heartbeatMap map[string]heartbeat
```
jeffallen: What problem does changing this to a pointer solve? I've not got a real opinion one way or the other, but Kelong probably had a reason for not using a pointer there...

byzcoin/heartbeat.go
```diff
@@ -70,49 +71,71 @@ func (r *heartbeats) exists(key string) bool {
...
 
 	r.wg.Add(1)
-	go func() {
+	go func(h *heartbeat) {
 		defer r.wg.Done()
 		currTime := time.Now()
-		to := time.After(timeout)
+		to := time.NewTimer(h.timeout)
 		for {
 			select {
```
jeffallen: This is tricky, how about a small blcok comment here to explain the interaction between to.Stop() and t.C for the next guy. :)

byzcoin/heartbeat.go
ineiti: Reverted to value.

ineiti: Reverted to pointer, so that the timeout update gets stored correctly.

byzcoin/service_test.go
```diff
@@ -1017,36 +1019,54 @@ func TestService_SetConfig(t *testing.T) {
...
+	}
+
+	lastInterval := testInterval
+	for i := 0; i < len(intervals); i++ {
+		for _, interval := range intervals {
+			// The next block should now be in the range of testInterval.
+			log.Lvl1("Setting interval to", interval)
+			start := time.Now()
+			ctx, _ := createConfigTx(t, interval, *s.roster, defaultMaxBlockSize, s)
+			s.sendTxAndWait(t, ctx, 10)
```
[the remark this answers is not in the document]
ineiti: Added comment.

PR #1545, opened 2018-10-24
Remove view change flag

### LINKED ISSUE #1538: FL generation: remove view change flag
From skipchain.go:
```
// If this flag is set, then we relax our forward-link signature verification
// to accept the aggregate signature by the public keys of the rotated roster.
// View-change should only be enabled if there are no security implications
// when the forward-links are signed by the rotated roster instead of the
// roster in the skipblock that points to it. The security implication depends
// on the services that use skipchain, so it is disabled by default. The option
// can be changed from outside of the package by calling EnableViewChange. This
// flag should ideally be the service configuration, but other structs depend
// on it too, e.g., SkipBlock, so we keep it in a global.
var enableViewChange bool
var enableViewChangeOnce sync.Once
```

This flag should be removed and the view change should be allowed by default. That means the generation of forward links needs to be adapted to accept the rotation.

Gilthoniel:
This PR removes the view change flag because we use the roster of the source block which means that we cannot allow a different roster to sign the forward link even if it's a rotation. No fix to the mask during the verification is necessary as the failing nodes are already ignored in the signature mask, in the correct order.

Fixes #1538

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -41,18 +41,6 @@ const bftFollowBlock = "SkipchainBFTFollow"
 var storageKey = []byte("skipchainconfig")
 var dbVersion = 1
 
-// If this flag is set, then we relax our forward-link signature verification
-// to accept the aggregate signature by the public keys of the rotated roster.
-// View-change should only be enabled if there are no security implications
-// when the forward-links are signed by the rotated roster instead of the
-// roster in the skipblock that points to it. The security implication depends
-// on the services that use skipchain, so it is disabled by default. The option
-// can be changed from outside of the package by calling EnableViewChange. This
-// flag should ideally be the service configuration, but other structs depend
-// on it too, e.g., SkipBlock, so we keep it in a global.
-var enableViewChange bool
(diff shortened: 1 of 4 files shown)


skipchain/skipchain.go
```diff
@@ -371,12 +359,15 @@ func (s *Service) StoreSkipBlock(psbd *StoreSkipBlock) (*StoreSkipBlockReply, er
...
-			s.SendRaw(back.Roster.List[0], &ForwardSignature{
+			err := s.SendRaw(back.Roster.List[0], &ForwardSignature{
 				TargetHeight: i + 1,
 				Previous:     back.Hash,
 				Newest:       prop.Copy(),
 			})
-			// time.Sleep(time.Second)
+
+			if err != nil {
+				log.Lvlf2("%v", err)
```
ineiti: It's worth to put a warning here.

```suggestion
				log.Warn("%v", err)
```

Gilthoniel: I changed it and also removed the unnecessary formatting

PR #1549, opened 2018-10-28
Fix evoting-admin commands in docs

omidraha:


ineiti:
Hi @omidraha - thanks for the PR. We're currently changing the blockchain-part to follow our latest byzcoin developments. If you're interested, don't hesitate to contact us at dedis@epfl.ch

PR #1550, opened 2018-10-29
Instance versioning

### LINKED ISSUE #1537: Information stored in state
A requirement stated by ByzGen is to have the possibility of following the history of an instance. For this, we would need the following:

- version of instance (also needed for #1442)
- link to previous instance, but:
  - if we link to the block, what to do when there were multiple updates?
  - can we give a link to the clientTransaction that updated the instance? Or even the instruction in the clientTransaction?
- other information needed?
- how do we re-create the state of #n-1, #n-2, ...? Do we need to store some intermediate states? Store the `StateChange`s?
- who is responsible for updating the version of the State? Is it the contract? Or the database?

Depends on the new trie that @kc1212 is working: #1534 and #1542

Gilthoniel:
This PR adds a cache for the state changes that allows to iterate over the history of an instance. BoltDB is used and the key is generated so that the byte-sorted ordering will sort first by instance and then by version. The block index is appended at the end so that it doesn't change the order and the information is accessible without decoding the value.

When a block is confirmed and added to the chain, the conodes will append the new state changes created from the transactions of the block and not before that. It can also happen that a node looses the DB or enters in the cothority that's why it will recreate the state changes by restoring from the last known block.

The storage gives two way to prevent the size from growing indefinitely. You can either specify the maximum size that must be store (not strict and computed from the data size without DB metadata) or the history of a maximum number of blocks. Both can be enabled at the same time. In both cases the oldest versions are removed.

Closes #1537


byzcoin/DataStructures.md
```diff
@@ -115,6 +115,8 @@ message StateChange {
   required bytes value = 4;
   // DarcID is the Darc controlling access to this key.
   required bytes darcid = 5;
+  // Version is the instance version for this particular state change
+  required int version = 7;
```
ineiti: Where is `6`?

byzcoin/InstanceVersioning.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Versioning
+
+Each state change has a `Version` field populated with the version of the instance
+after having applied the change. It is monotonically increasing by one every time
+an instruction is performed on an instance.
+The global state also stores the version at the moment of the block creation because
+we need to know what is the latest version to create new transaction asking the right
+version. This prevents replay attacks (see #1442).
+The Byzcoin service takes care of the versioning by modifying the results of a
```
ineiti: If the byzcoin service takes care of the versioning, then you don't need to put the version in the `StateChange`. `StateChange` is created by the contract.

byzcoin/InstanceVersioning.md
```diff
@@ -0,0 +1,52 @@
...
+skipchains created in that roster. After this step is successful, the service will
+start a routine to populate the cache by reading the skipchain from the beginning
+or one of the snapshot of the global state (not yet available #????). Note that
+the history will be as old as the snapshot.
+
+## Size management
+
+This feature needs a size control as the space taken will grow with the skipchain and
+that's why you can specify the size of the cache or the maximum number of blocks that
+need to be kept. When a size is specified, the storage will delete all state changes
```
ineiti: It should delete the oldest state changes, not the ones from the last block stored.

Gilthoniel: I see, my goal was to prevent multiple small removals.

byzcoin/service.go
```diff
@@ -1107,7 +1116,27 @@ func (s *Service) executeInstruction(cdbI CollectionView, cin []Coin, instr Inst
 	}
 	// Now we call the contract function with the data of the key.
 	log.Lvlf3("%s Calling contract %s", s.ServerIdentity(), contractID)
-	return contract(cdbI, instr, cin)
+	scs, cout, err = contract(cdbI, instr, cin)
+	if err != nil || instr.InstanceID == [32]byte{} {
+		// The config contract is ignore for the versioning
```
ineiti: should it read "the config contract ignores versioning"? or "the config contract doesn't have versioning"? But a contract doesn't have versions, its instance(s) have.
Why should the config-contract not have versions?

byzcoin/InstanceVersioning.md
Gilthoniel: I am fine with storing the version alongside the state change and the block index.

external/java/src/test/java/ch/epfl/dedis/calypso/CalypsoTest.java
```diff
@@ -358,9 +358,16 @@ void multiLTS() throws CothorityException{
     }
 
     @Test
-    void reConnect() throws CothorityException {
+    void reConnect() throws CothorityException, InterruptedException, IOException {
         WriteInstance wr = doc.spawnWrite(calypso, publisherDarc.getBaseId(), publisher);
 
+        for (int i=0; i<3; i++){
```
ineiti: That looks like `git merge` instead of `git rebase`. Please use the latter for your PRs, it makes it easier to follow what you changed.

byzcoin/service.go
Gilthoniel: I was under the impression that this particular contract was not stored but I just got confused myself. This part needed to handle unknown state changes when it has not been stored yet.

Gilthoniel:
Right now the storage is using one bucket and that means conflicts are possible. For instance the config instance always uses 0 as instance ID. Should I use different bucket per skipchain ?

byzcoin/struct.go
```diff
@@ -1,12 +1,531 @@
...
+				if err != nil {
+					return err
+				}
+				s.sortedKeys[0].time = sce.Timestamp
+				copy(s.sortedKeys[0].key, k)
+
+				sort.Sort(s.sortedKeys)
+			} else {
+				// if none, that means it was the last
+				s.sortedKeys = s.sortedKeys[1:]
```
kc1212: it's better to update the state after the transaction succeeds

byzcoin/struct.go
```diff
@@ -1,12 +1,531 @@
...
+		return nil
+	}
+
+	err := s.db.Update(func(tx *bolt.Tx) error {
+		// We make enough space to not have to do it everytime
+		// we append state changes
+		thres := int(float64(s.maxSize) * cleanThreshold)
+
+		for s.size > thres {
+			if len(s.sortedKeys) == 0 {
```
kc1212: missing a mutex?

byzcoin/transaction.go
```diff
@@ -423,8 +425,35 @@ func decodeStateChangeBody(buf []byte) (StateChangeBody, error) {
...
+
 // Hash returns the sha256 of all stateChanges
 func (scs StateChanges) Hash() []byte {
+	if !sort.IsSorted(scs) {
+		// An order needs to be defined so that we can compute
+		// a hash from an arbitrary list of state changes
+		// (i.e. the cache and the history storage have different
+		// ordering which is producing a different hash)
+		sort.Sort(scs)
+	}
```
kc1212: depending on how you use this function, i have a suspicion that sorting StateChange will have a conflict with the SignerCounter change that just got merged. because if one user submits a lot of instructions, which results in a lot of state changes, then they must happen in the same order, otherwise the SignerCounter won't be monotonically increasing. I guess we'll tell only when we try to merge.

byzcoin/struct.go
```diff
@@ -1,12 +1,531 @@
...
+	bolt "github.com/coreos/bbolt"
 	"github.com/dedis/cothority/skipchain"
+	"github.com/dedis/onet"
+	"github.com/dedis/protobuf"
 )
 
+const defaultMaxSize = 2 * 1024 * 1024 * 1024 // 2GB
+const versionLength = 64 / 8                  // bytes
+const prefixLength = 32                       // bytes
+const cleanThreshold = 0.8
```
kc1212: why do we need both a max size and a threshold? isn't maxsize by itself enough?

byzcoin/struct.go
```diff
@@ -1,12 +1,531 @@
...
+				err := binary.Write(&buf, binary.BigEndian, thres)
+				if err != nil {
+					return err
+				}
+				index := buf.Bytes()
+
+				c := b.Cursor()
+				for k, v := c.Seek(sc.InstanceID); k != nil && bytes.HasPrefix(k, sc.InstanceID); k, v = c.Next() {
+					if bytes.Compare(k[len(k)-len(index):], index) <= 0 {
+						c.Delete()
```
kc1212: will this return an error? we should check it if it does

byzcoin/struct.go
```diff
@@ -1,12 +1,531 @@
...
+						s.size -= len(v)
```
kc1212: modify state only when transaction is successful

byzcoin/struct.go
```diff
@@ -1,12 +1,531 @@
...
+				})
+			}
+
+			err = b.Put(key, buf)
+			if err != nil {
+				return err
+			}
+
+			// optimization for cleaning to avoir recomputing the size
+			s.size += len(buf)
```
kc1212: wait for transaction to complete successfully then change the size

external/java/src/main/java/ch/epfl/dedis/byzcoin/StateChange.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Represents the state change of an instance and thus contains all the
+ * information related.
+ */
+public class StateChange {
+    private StateAction stateAction;
+    private ByteString instanceId;
+    private ByteString contractId;
+    private ByteString darcId;
+    private ByteString value;
```
kc1212: some of these types can be improved, I think we have an InstanceID class, contractID is usually a string in other classes, DarcID probably has its own class too.

byzcoin/struct.go
```diff
@@ -1,12 +1,531 @@
...
+
+// this generates a storage key using the instance ID and the version
+func (s *stateChangeStorage) key(iid []byte, ver uint64, idx int64) ([]byte, error) {
+	b := bytes.Buffer{}
+	_, err := b.Write(iid)
+	if err != nil {
+		return nil, err
+	}
+
+	err = binary.Write(&b, binary.BigEndian, ver)
```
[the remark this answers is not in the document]
Gilthoniel: Yes exactly. I will.

byzcoin/transaction.go
Gilthoniel: Ok I will have a particular look at that when rebasing

byzcoin/struct.go
Gilthoniel: Max size is the limit and the threshold tells how much will be cleaned if the storage goes above max size. My goal was to prevent multiple small rounds of cleaning.

byzcoin/struct.go
Gilthoniel: Yes it will, I need to fix that

byzcoin/struct.go
```diff
@@ -1,12 +1,531 @@
...
+				return err
+			}
+
+			// Check if the instance has already a value
+			// and add the timestamp if not
+			c := b.Cursor()
+			k, _ := c.Seek(sc.InstanceID)
+			if !bytes.HasPrefix(k, sc.InstanceID) {
+				// no need to sort here as it's necessarily the newest
+				sortedKeys = append(sortedKeys, keyTime{
```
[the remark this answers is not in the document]
Gilthoniel: I actually created a copy line 305 that I populate and I change the global one only after the transaction.

byzcoin/struct.go
Gilthoniel: ... but you're right on this one

byzcoin/struct.go
Gilthoniel: No because the function is used inside the append that is already holding the lock. But I gonna change that for a read/write lock.

Gilthoniel: No but I'm going to change that so that if you use the cleaning elsewhere it will hold the lock anyway, that will be less confusing.

byzcoin/struct.go
```diff
@@ -254,15 +265,21 @@ func (s *stateChangeStorage) cleanByBlock(scs StateChanges, sb *skipchain.SkipBl
...
+						size -= len(v)
 					}
 				}
 			}
 		}
 
 		return nil
 	})
+
+	s.size = size
```
kc1212: should it be `if err == nil { s.size = size }`?

Gilthoniel:
@kc1212 Your comments should be fixed, I also had to make some change after rebasing:
1. Catching up counter state changes
2. Store the state change index when appending so that the order can be reproduced later on

ineiti:
having one last go through it...

PR #1553, opened 2018-10-29
Implementing DARC editing in bcadmin CLI

### LINKED ISSUE #1552: Improve bcadmin
To actually use darcs in byzcoin from the command line, we need additional commands in the `bcadmin` binary. The following commands should be added:

- darc_add
- darc_show
- key
- darc_rule_add
- darc_rule_update
- darc_rule_delete

LMeynent:
Closes #1552

### DIFF
--- a/byzcoin/bcadmin/README.md
+++ b/byzcoin/bcadmin/README.md
@@ -51,3 +51,61 @@ transactions.
 
 You can set the environment variable BC to the config file for the ByzCoin
 you are currently working with. (Client apps should follow this same standard.)
+
+## Generating a new keypair
+
+```
+$ bcadmin key
+```
+
+Generates a new keypair and prints the result in the console
+
+Optional flags:
+
+-save file.txt            Outputs the key in file.txt instead of stdout
+## Managing DARCS
+
+```
+$ bcadmin darc add -bc $file
+```
+
+Adds a new darc with a random keypair for both signing and evolving it.
+
+Optional flags:
+
+ * -out_id file.txt          Outputs the ID of the DARC in file.txt
+ * -out_key file.txt         Outputs the key of the DARC in file.txt
+ * -out file.txt             Outputs the full description of the DARC in file.txt
+ * -owner key:%x             Creates the DARC with the mentioned key as owner (sign & evolve)
+ * -darc darc:%x             Creates the DARC using the mentioned DARC for creation (uses Genesis DARC by default)
+ * -sign key:%x              Uses this key to sign the transaction (AdminIdentity by default)
+
+```
+$ bcadmin darc show -bc $file
+```
+
+Shows a DARC either in stdout or in a given file
+
+Optional flags:
+
+ * -out file.txt             Outputs the description of the DARC in file.txt instead of stdout
+ * -darc darc:%x             Shows the DARC with provided ID, Genesis DARC by default
+
+```
+$ bcadmin darc rule -bc $file -rule $action
+```
+
+Manages rules for a DARC
+
+Optional flags:
+ * -darc darc:%x             Modifies the rules of this DARC (uses Genesis DARC by default)
+ * -sign key:%x              Uses this key to sign the transaction (AdminIdentity by default)
+ * -delete                   Deletes the specified rule if it exists
+ * -identity:%x              The expression that will determine the necessary signatures to perform the action (mandatory if -delete is not used)
+ * -replace                  Overwrites the expression for the necessary signatures to perform the action (if not provided and action already exists in Rules the action will fail)
+
+ ```
+ $ bcadmin darc
+ ```
+
+ is equivalent to show
--- a/byzcoin/bcadmin/main.go
+++ b/byzcoin/bcadmin/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 	"time"
@@ -12,10 +13,13 @@ import (
 	"github.com/dedis/cothority/byzcoin/bcadmin/lib"
 	"github.com/dedis/cothority/darc"
 	"github.com/dedis/cothority/darc/expression"
+	"github.com/dedis/kyber/util/random"
 	"github.com/dedis/onet"
+	"github.com/dedis/onet/app"
 	"github.com/dedis/onet/cfgpath"
 	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
+
 	cli "gopkg.in/urfave/cli.v1"
 )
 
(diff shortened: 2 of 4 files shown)


byzcoin/bcadmin/main.go
```diff
@@ -295,9 +434,458 @@ func add(c *cli.Context) error {
...
+	  if err != nil {
+	      panic(err)
+	  }
+
+		fo.Write([]byte(newSigner.Identity().String()))
+
+		fo.Close()
+	}
+
+	/*Saving description in special file*/
```
ineiti:
```suggestion
	// Saving description in special file
```

byzcoin/bcadmin/main.go
```diff
@@ -295,9 +434,458 @@ func add(c *cli.Context) error {
...
+
```
ineiti: No line between comment and code.

byzcoin/bcadmin/README.md
```diff
@@ -45,3 +45,68 @@ transactions.
...
+
+```
+$ bcadmin key
+```
+
+Generates a new keypair and prints the result in the console
+
+Optional flags:
+
+-save file.txt            Outputs the key in file.txt instead of stdin
```
jeffallen:
```suggestion
-save file.txt            Outputs the key in file.txt instead of stdout
```

byzcoin/bcadmin/README.md
```diff
@@ -45,3 +45,68 @@ transactions.
...
+
+Optional flags:
+
+-out_id file.txt          Outputs the ID of the darc in file.txt
+
+-out_key file.txt         Outputs the key of the darc in file.txt
+
+-out_desc file.txt        Outputs the full description of the darc in file.txt
+
+-owner key:%x             Creates the darc the the mentioned key as owner (sign & evolve)
```
jeffallen:
```suggestion
-owner key:%x             Creates the darc with the mentioned key as owner (sign & evolve)
```

byzcoin/bcadmin/README.md
```diff
@@ -45,3 +45,68 @@ transactions.
...
+Shows the darc for which BaseID is the one specified after the -id flags
+
+Optional flags:
+
+-save file.txt            Outputs the description of the darc in file.txt instead of stdin
+
+## Editing the rules of a darc
+
+```
+$ bcadmin darc_rule_add $action -bc $file -identity $expr -darc darc:%x -signer key:%x
```
jeffallen: Sorry, you haven't seen this yet, but the latest behaviour of "bcadmin add" (as of commit 4ea5faa42d03fd) is that it refuses to overwrite, unless it receives the "-replace" argument.

I would like to see darc_rule_add implemented the same way, and thus no darc_rule_update.

It seems that with these new functions the "bcadmin add" command could be removed. It is hardcoded to work on the genesis darc. Perhaps if these commands were missing the -darc argument, then they would default to operating on the genesis darc?

byzcoin/bcadmin/README.md
```diff
@@ -45,3 +45,68 @@ transactions.
...
+-owner key:%x             Creates the darc the the mentioned key as owner (sign & evolve)
+
+-darc darc:%x             Creates the darc using the mentioned darc for creation (uses Genedis node by default)
+
+-sign key:%x              Uses this key to sign the transaction (mandatory if -darc is used)
+
+## Showing a darc
+
+```
+$ bcadmin darc_show -bc $file -id darc:%x
```
jeffallen: I am not really too happy with darc_show, darc_add_rule, etc. I would like darc to be a [sub-command](https://godoc.org/gopkg.in/urfave/cli.v1#example-App-Run-Subcommand).

If Linus specified it this way, then we'll go with his plan.

byzcoin/bcadmin/main.go
```diff
@@ -13,12 +14,15 @@ import (
...
 	"github.com/dedis/cothority/darc/expression"
 	"github.com/dedis/onet"
+	"github.com/dedis/onet/app"
 	"github.com/dedis/onet/cfgpath"
 	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
 	cli "gopkg.in/urfave/cli.v1"
 )
 
+
```
jeffallen: Remove extra lines, please.

byzcoin/bcadmin/main.go
```diff
@@ -80,10 +84,145 @@ var cmds = cli.Commands{
 		Action: add,
 	},
 	{
-		Name:    "keys",
-		Usage:   "creates a new key pair, which may be used with add",
+		Name:    "key",
+		Usage:   "generates a new keypair and prints the public key in the stdin",
```
jeffallen:
```suggestion
		Usage:   "generates a new keypair and prints the public key on stdout",
```

byzcoin/bcadmin/main.go
```diff
@@ -295,9 +434,458 @@ func add(c *cli.Context) error {
...
 
-func keys(c *cli.Context) error {
-	s := darc.NewSignerEd25519(nil, nil)
-	fmt.Println("Identity:", s.Identity())
-	fmt.Printf("export PRIVATE_KEY=%v\n", s.Ed25519.Secret)
+func key(c *cli.Context) error {
+	newSigner := darc.NewSignerEd25519(nil, nil)
+	lib.SaveKey(newSigner)
+
+	save := c.String("save")
```
jeffallen: Something is wrong with the formatting here. Look into configuring gofmt-on-save for your editor.

byzcoin/bcadmin/main.go
```diff
@@ -295,9 +434,458 @@ func add(c *cli.Context) error {
...
+	if save == "" {
```
jeffallen: Don't repeat what goes into the file twice, instead set fo correctly before and then call fmt.Println(fd,...) one time:
```go
var fo io.Writer
var err error
if save == "" {
  fo = os.Stdout
} else {
  fo, err = os.Create(save)
  // check err
  defer fo.Close()
}
fmt.Fprintln(fo, newSigner.Identity.String())
```

Another way to go about it would be to use some kind of writeToFileOrStdout(fn, w) wrapper which would take care of this for you all in one place. Then use fmt.Fprintln to write into a *bytes.Buffer (which is an io.ReadWriter) and then give the buf to your writeToFileOrStdout where it would use io.Copy to send the output to the right file descriptor.

byzcoin/bcadmin/test.sh
```diff
@@ -28,4 +33,80 @@ testCreateStoreRead(){
     testFail ./bcadmin add spawn:xxx
 }
 
+testAddDarc(){
+  runCoBG 1 2 3
+    runGrepSed "export BC=" "" ./"$APP" create --roster public.toml --interval .5s
```
jeffallen: Something is wrong with formatting in this file, if you can find an auto-formatter for shell, please tell us about it, so we all get the benefit!

byzcoin/bcadmin/test.sh
```diff
@@ -28,4 +33,80 @@ testCreateStoreRead(){
...
+  [ -z "$BC" ] && exit 1
+
+  testOK ./"$APP" darc_add -out_id ./darc_id.txt -out_key ./darc_key.txt
+  ID=`cat ./darc_id.txt`
+  KEY=`cat ./darc_key.txt`
+  testOK ./"$APP" key -save ./key.txt
+  KEY2=`cat ./key.txt`
+
+  testOK ./"$APP" darc_rule_add spawn:darc -identity "$KEY | $KEY2" -darc "$ID" -signer "$KEY"
+  testOK ./"$APP" darc_show -id "$ID"
```
jeffallen: Testing darc_show should actually grep for something, or else it could succeed without working correctly.

byzcoin/bcadmin/test.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-DBG_TEST=1
+DBG_TEST=0
```
jeffallen: I think Linus usually asks me to check these in at level 1, so that you can see what it is testing. Not sure about this, though.

byzcoin/bcadmin/main.go
```diff
@@ -80,10 +84,145 @@ var cmds = cli.Commands{
...
+				Name:  "identity",
+				Usage: "the identity of the signer who will be allowed to access the contract (e.g. ed25519:a35020c70b8d735...0357))",
+			},
+			cli.StringFlag{
+				Name:  "darc",
+				Usage: "the darc to which we want to add a rule and signer",
+			},
+			cli.StringFlag{
+				Name:  "signer",
+				Usage: "the signer of the operation",
```
jeffallen: Make it clearer that this is the filename of the private key file. (I think; is that what this is? If not, where does the private key get set?)

byzcoin/bcadmin/test.sh
ineiti: yes, `DBG_TEST=1` rulez!

byzcoin/bcadmin/README.md
ineiti: sub-command would be better, definitely.

byzcoin/bcadmin/README.md
ineiti: I like the idea of working on the genesis-darc if `-darc` is not present.

For `darc_rule_add` and `darc_rule_update`, I would replace both by `darc_rule_set`, or `darc rule set` when using sub commands. I think separating `add` and `update` doesn't really make sense with regards to darc-rules.

LMeynent:
@jeffallen , @ineiti :

I have taken your remarks into account and I changed the code accordingly.
I wait for other comments/advice/remarks as you see fit.

jeffallen:
Thanks for this. Now please get the tests to turn green again. i.e. https://travis-ci.org/dedis/cothority/jobs/448372279

LMeynent:
Hello @ineiti @jeffallen ,

I have started working on making the tests go green.
However, I'm having the two following issues :

 * I can't figure out why this last test fails
 * As I've had to git pull some work that had been done on other files, I now have a problem with the generation of pseudorandom keypairs: using the function NewSignerEd25519() yields multiple times the same public key. As an example, in my first bash test, I create 3 DARCs in a row to test that the darc show command yields the good one, but an error occurs as the first and third DARC are created with the same public key as an owner - this creates a key conflict as the BaseID of the DARC is a digest of the content of the DARC, and thus if two DARCs are created empty with the same owner they have the same BaseID. Do I have a way to get a different keypair each time?

Thanks

PR #1557, opened 2018-10-30
Refuse to start ByzCoin with an old-style database

jeffallen:
Because we were not using database versions until now, this is
a bit hacky. But future migrations on boot will be easier since
there is now a clear way to detect them and a place to put them.

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -1504,7 +1504,7 @@ func (s *Service) trySyncAll() {
 	}
 }
 
-var existingDB = regexp.MustCompile(`^ByzCoin_[0-9a-f]+`)
+var existingDB = regexp.MustCompile(`^ByzCoin_[0-9a-f]+$`)
 
 // newService receives the context that holds information about the node it's
 // running on. Saving and loading can be done using the context. The data will
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -9,10 +9,12 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
 
+	bolt "github.com/coreos/bbolt"
 	"github.com/dedis/cothority"
 	"github.com/dedis/cothority/byzcoin/viewchange"
 	"github.com/dedis/cothority/darc"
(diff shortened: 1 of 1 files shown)


byzcoin/service.go
```diff
@@ -1553,6 +1557,43 @@ func newService(c *onet.Context) (onet.Service, error) {
...
+		// Version 0 means it hasn't been set yet. If there are any ByzCoin_[0-9af]+
+		// buckets, then they must be old format.
+		db, _ := s.GetAdditionalBucket([]byte("check-db-version"))
+
+		// Look for a bucket that has a byzcoin database in it.
+		err := db.View(func(tx *bolt.Tx) error {
+			c := tx.Cursor()
+			for k, _ := c.First(); k != nil; k, _ = c.Next() {
+				log.Lvlf4("looking for old ByzCoin data in bucket %v", string(k))
+				if existingDB.Match(k) {
```
kc1212: My understanding is that we're looping over bucket names. But here's you're checking if the bucket name matches `^ByzCoin_[0-9af]+`. But isn't that true for the new database too? In `createStateTrie` we name buckets based on the hex skipchain ID.

kc1212: also, why isn't it `[0-9a-f]`?

byzcoin/service.go
```diff
@@ -1553,6 +1557,43 @@ func newService(c *onet.Context) (onet.Service, error) {
...
 
+	ver, err := s.LoadVersion()
+	if err != nil {
+		return nil, err
+	}
+	switch ver {
+	case 0:
+		// Version 0 means it hasn't been set yet. If there are any ByzCoin_[0-9af]+
+		// buckets, then they must be old format.
+		db, _ := s.GetAdditionalBucket([]byte("check-db-version"))
```
kc1212: we're using `s.GetAdditionalBucket` is because we cannot get a database pointer with any other way?

jeffallen: yes.

byzcoin/service.go
jeffallen: On the first start of a db that will be new-style, it has no buckets that match ByzCoin_[0-9a-f]+ yet. So we get to the end of the loop, set the version to 1, and never do the scan again. On an attempted start with a database that has been used with the collections code, we match the first one, throw the error, and refuse to start.

jeffallen:
Somehow now JS testing is not working. I will come back to this when I finish another thing with a deadline tomorrow.

PR #1558, opened 2018-10-30
Increase block interval for failure tests

kc1212:
Doing TestClose and then startAllChains may take longer than the block
interval, which will cause an unintentional view-change in tests that do
not expect it. So increasing the block interval for those tests.

In addition, we add some more delay for the calypso benchmark to make
Jenkins happy.


kc1212:
This PR should fix two unit test failures on Jenkins - `TestContract_Write_Benchmark` and `TestService_AddTransaction_WithFailure`.

PR #1561, opened 2018-10-31
Return an error in GetReadOnlyStateTrie 

### LINKED ISSUE #1560: GetReadOnlyStateTrie probably should return an interface
I merged https://github.com/dedis/cothority/pull/1559

Later, while working on byzcoin, using the new `GetReadOnlyStateTrie` doesn't feel too good because it gives you the full trie. So I have to do something like 
```
	var st ReadOnlyStateTrie = s.GetReadOnlyStateTrie(scID)
```
to make sure I only use the read-only functions.

Returning a real `StateTrie` also allows other packages to change the byzcoin trie in anyway they want. 

I propose that `GetReadOnlyStateTrie` should still return an interface, but maybe with an error if the trie is nil.

kc1212:
Closes #1560

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -316,7 +316,11 @@ func (s *Service) GetProof(req *GetProof) (resp *GetProofResponse, err error) {
 		err = errors.New("cannot find skipblock while getting proof")
 		return
 	}
-	proof, err := NewProof(s.GetReadOnlyStateTrie(sb.SkipChainID()), s.db(), req.ID, req.Key)
+	st, err := s.GetReadOnlyStateTrie(sb.SkipChainID())
+	if err != nil {
+		return nil, err
+	}
+	proof, err := NewProof(st, s.db(), req.ID, req.Key)
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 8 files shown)


byzcoin/service.go
```diff
@@ -684,13 +698,13 @@ func isViewChangeTx(txs TxResults) *viewchange.View {
 
 // GetReadOnlyStateTrie returns a read-only accessor to the trie for the given
 // skipchain.
-func (s *Service) GetReadOnlyStateTrie(scID skipchain.SkipBlockID) ReadOnlyStateTrie {
+func (s *Service) GetReadOnlyStateTrie(scID skipchain.SkipBlockID) (ReadOnlyStateTrie, error) {
```
sarkortrantor: quick thought: What about moving getStateTrie content here (or if used for other purposes make getStateTrie return Interface instead of pointer to actual struct
**or last and maybe best option check in GetReadOnlyStateTrie that return value of getStateTrie call not nil and if nil return nil <= not same nil ^^ )

by doing so nil will be the nil interface => no problem later 
(the problem was that the nil returned by getStateTrie "was of type (*stateTrie)(nil)" and stored without check into an interface variable => returned interface not nil but holding a nil

kc1212: That works also. But I've been wanting to add an error to `{get,load}StateTrie` when I was working on another PR. Because sometimes we forget to check whether the returned trie is nil. If it also returns an error, then it's less likely that we'll forget. The same goes for `GetReadOnlyStateTrie`, for example https://github.com/dedis/cothority/pull/1561/files#diff-f6ebb21b8df740800fa497e7e19179a4L49

ineiti: I learnt this: `Accept interfaces, return concrete types`. It's easier for testing and more go-like:

http://idiomaticgo.com/post/best-practice/accept-interfaces-return-structs/

sarkortrantor: probably a good advice, (but the additional benefit of returning interface (sometime) is like here if you don't want to export datastructures, encapsulation mechanism)

PR #1564, opened 2018-11-01
Proto files deprecation script

### LINKED ISSUE #1563: Proto files generation doesn't work anymore on Linux
`clean_proto_deprecation.sh` needs to be fixed to work on both Linux and MacOS (let's forget about Windows)

Gilthoniel:
This fixes the bash script that replaces the deprecation
in the Java proto files to work on Linux and MacOS

Fixes #1563


Gilthoniel:
@kc1212 or @jeffallen if one of you could confirm this works on Mac, the space is important under Linux

kc1212:
yep, it works

PR #1571, opened 2018-11-07
Fix state trie already exists error

### LINKED ISSUE #1567: Error: "state trie already exists" is not always an error
In the unit tests, we always get error logs such as 
```
Error while adding block 96edfc87de8504faaf1d586d0f880583bbd0896c82a6bfab2d5e78a4dd40cec0: state trie already exists
Error while adding block 96edfc87de8504faaf1d586d0f880583bbd0896c82a6bfab2d5e78a4dd40cec0: state trie already exists
Error while adding block 96edfc87de8504faaf1d586d0f880583bbd0896c82a6bfab2d5e78a4dd40cec0: state trie already exists
```
But these are not real errors because it's from`updateTrieCallback` when updating forward links for existing blocks. If a genesis block gets added again, and we get to the `createStateTrie` branch, it'll return an error.

Gilthoniel:
This adds a check to prevent from creating a state trie for a
genesis block if it has already been created.

Fixes #1567

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -57,7 +57,7 @@ func init() {
 	var err error
 	ByzCoinID, err = onet.RegisterNewService(ServiceName, newService)
 	log.ErrFatal(err)
-	network.RegisterMessages(&omniStorage{}, &DataHeader{}, &DataBody{})
+	network.RegisterMessages(&bcStorage{}, &DataHeader{}, &DataBody{})
 	viewChangeMsgID = network.RegisterMessage(&viewchange.InitReq{})
 }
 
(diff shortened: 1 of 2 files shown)


byzcoin/service.go
```diff
@@ -704,6 +705,13 @@ func (s *Service) GetReadOnlyStateTrie(scID skipchain.SkipBlockID) (ReadOnlyStat
 	return s.getStateTrie(scID)
 }
 
+func (s *Service) hasStateTrie(id skipchain.SkipBlockID) bool {
```
jeffallen: On line 719 below, it seems like s.storage.Mutex guard access to this map. This is not well documented, and messy. Let's fix it.

So: s/omniStorage/bcStorage/, use less generic name, s.storage.Mutex.Lock -> s.storage.Lock (same for unlock). Make a new mutex for stateTrie, put it NEXT TO stateTrie in the type defn.

PR #1572, opened 2018-11-08
Make the set interval test work for Jenkins

kc1212:
Jenkins fails on the TestService_SetConfigInterval test. Probably because we may send the transaction just before the root decides to create a block. As a result, the block doesn't take very long to be created. So we relax the constraint a bit.


kc1212:
The test passes on a custom jenkins job.

PR #1576, opened 2018-11-13
Fix skipblock ID string representation

### LINKED ISSUE #1575: SkipBlockID string formatting
As the String function is defined for this struct, it yelds wrong output when using `%x` in a formatting function. Jenkins tests are failing because it cannot extract the ID correctly due to the bad formatting.

Gilthoniel:
This removes the String function that was causing bad
formatting when using "%x".

It also fixes a data race in the state change storage.

Fixes #1575


jeffallen:
👍

ineiti:
I'm downvoted - dang that democracy ;)

PR #1577, opened 2018-11-14
Helm chart

leafty:
Provides a [helm](https://helm.sh/) chart for running a conode on kubernetes.


jeffallen:
Thank you for this, we'll review it today or tomorrow. Sorry for the delay.

Issue #1536, opened 2018-10-22
Add LTS update


ineiti:
Currently the LTS is fixed and cannot be changed. Once we have the possibility to change the set of nodes in the roster, LTS should be able to change, too.

For this, we need dedis/kyber#326 and then add an api-call in the `calypso` service.

kc1212:
For resharing the LTS, how are the conodes going to agree on the new roster? Leader could propose a new roster but the other nodes have no way to verify. Then the leader can just add anybody in the roster.

Is it reasonable to do something like this:
1. All conodes must obtain the new roster out-of-band, e.g., the administrator must add a new toml roster file to the config directory, or we can use the database. 
2. Roster starts the resharing protocol for the new roster.
3. All nodes verify that the roster is the same one as the one stored in config.
4. All nodes proceed to execute the protocol and creates new shares.

ineiti:
With regard to the first one - this looks like an excellent use-case for something called a 'blockchain'. You create an `LTS`-contract and spawn an instance by giving it the roster and configuration of the nodes to be used in the sharing algorithm. Then this instance can be used to prove you have the right to do the DKG.

When you want to reshare the LTS, you update the instance (by respecting the DARC linked to it), then you get a proof for this instance and you can use it to do the reshare.

ineiti:
With regard to Rabin / Pedersen - Nicolas told me already a long time ago I should switch, but never did it :( Is it complicated to change from one to another? As far as I understand, from an engineering point of view, there is no difference. Besides the fact that Pedersen has resharing...

So I'd say to create a Pedersen DKG and use that in `CreateLTS`.

jeffallen:
I recommend against using this so-called 'blockchain'. I read on HN that it is a communist conspiracy and anyway it "doesn't scale".

kc1212:
> With regard to the first one - this looks like an excellent use-case for something called a 'blockchain'. You create an `LTS`-contract and spawn an instance by giving it the roster and configuration of the nodes to be used in the sharing algorithm. Then this instance can be used to prove you have the right to do the DKG.
> 
> When you want to reshare the LTS, you update the instance (by respecting the DARC linked to it), then you get a proof for this instance and you can use it to do the reshare.

I was thinking that the secret-management cothority shouldn't need access to a blockchain system, at least that was the concept behind Calypso. But you're right that it's not easy to evolve the LTS group members without using byzcoin and Darc.

[...]

PR #1579, opened 2018-11-14
Use Pedersen DKG in CreateLTS

kc1212:
This is the first step for #1536 where we implement the Pedersen DKG in an onet protocol and use it in CreateLTS. We need to use this DKG because key-resharing is not available for the Rabin DKG.


dkg/pedersen/dkg.go
```diff
@@ -0,0 +1,216 @@
...
+	}, nil
+}
+
+// Start sends the Announce-message to all children
+func (o *Setup) Start() error {
+	log.Lvl3("Starting Protocol")
+	// 1a - root asks children to send their public key
+	errs := o.Broadcast(&Init{Wait: o.Wait})
+	if len(errs) != 0 {
+		return fmt.Errorf("boradcast failed with error(s): %v", errs)
```
Gilthoniel: broadcast

jeffallen:
```suggestion
		return fmt.Errorf("broadcast failed with error(s): %v", errs)
```

jeffallen: Perhaps like that you can just click and accept in Github? It's a new feature, unclear how it works...

kc1212: wow that's a useful feature

dkg/DKG.md
```diff
@@ -5,20 +5,32 @@ Distributed Key Generation
...
-Based Cryptosystems" by R. Gennaro, S. Jarecki, H. Krawczyk, and T. Rabin.
-DKG enables a group of participants to generate a distributed key
-with each participants holding only a share of the key. The key is also
-never computed locally but generated distributively whereas the public part
-of the key is known by every participants.
+Under this folder, we implement two distributed key generation protocols:
+Pedersen DKG and Rabin DKG.
+
+DKG enables a group of participants to generate a distributed key with each
+participants holding only a share of the key. The key is also never computed
```
jeffallen:
```suggestion
participant holding only a share of the key. The key is also never computed
```

Issue #1213, opened 2018-04-26
Skipchain cleanup


ineiti:
The current skipchain is not optimal. Some cleanup functions to do:
- make sure the correct blocks are available when changing roster
- upon completion of byzcoinX, don't send the block again but only the forward-links

As this issue would change the messages between conodes, we first need a way to share capacities between conodes, and then proceed with this.

Issue #1201, opened 2018-04-23
Skipchain storing of accepted blocks


ineiti:
For the moment a block to be accepted is sent twice:
1. for signature
2. when the forward signature is finished

in `skipchain/skipchain.go:StoreSkipBlock` the `changed` slice of `SkipBlocks` should be a slice of `ForwardLink`s.  Then the verification function could cache the received block, and when it receives a valid forwardlink, store that cached block in the database.

PR #1580, opened 2018-11-15
Skipchain propagation optimization

### LINKED ISSUE #1395: Reduce the number of skipchain.startPropagation
Linked issues:
- #1213 
- #1201 

Taken from https://github.com/dedis/cothority/issues/1352
> avoid sending the block around multiple times in skipchain.startPropagation - the block is already known from the byzcoinx call and only the forwardlink needs to be propagated.

Gilthoniel:
This changes the way the skipchain is propagating new block and
forward links so that it sends minimal information and let conodes
ask for missing data.

Fixes #1395 #1213 #1201

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -1378,7 +1378,7 @@ func (s *Service) propagateGenesisHandler(msg network.Message) {
 	}
 
 	if !s.blockIsFriendly(pg.Genesis) {
-		log.Error("Block is not friendly")
+		log.Error("Conode doesn't want to follow that skipchain")
 		return
 	}
 
@@ -1433,12 +1433,7 @@ func (s *Service) propagateForwardLinkHandler(msg network.Message) {
 	}
 
 	// Add the block if available
-	if sb, ok := s.blockBuffer.get(sb.SkipChainID(), pfl.ForwardLink.To); ok {
-		if !s.blockIsFriendly(sb) {
-			log.Error("Trying to add a block that is not friendly")
-			return
-		}
-
+	if sb := s.blockBuffer.get(sb.SkipChainID(), pfl.ForwardLink.To); sb != nil {
 		id := s.db.Store(sb)
 		if id == nil {
 			// error already logged
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -52,8 +52,10 @@ func init() {
 type Service struct {
 	*onet.ServiceProcessor
 	db                      *SkipBlockDB
+	blockBuffer             *skipBlockBuffer
 	propagateGenesis        messaging.PropagationFunc
 	propagateForwardLink    messaging.PropagationFunc
+	propagateProof          messaging.PropagationFunc
 	verifiers               map[VerifierID]SkipBlockVerifier
 	storageMutex            sync.Mutex
 	Storage                 *Storage
@@ -958,16 +960,30 @@ func (s *Service) forwardLinkLevel0(src, dst *SkipBlock) error {
 		return errors.New("Wrong BFT-signature: " + err.Error())
 	}
 
-	ro := roster.Concat(dst.Roster.List...)
+	// We send the new forward link to the previous roster only
+	err = s.startPropagation(s.propagateForwardLink, roster, &PropagateForwardLink{fwd, 0})
+	if err != nil {
+		return err
+	}
 
-	// Propagate the forward link and the new block as the roster
-	// could have changed, meaning one node cannot know the block
-	return s.startForwardLinkPropagation(ro, &PropagateForwardLink{
-		ForwardLink: fwd,
-		Newest:      dst,
-		Roster:      roster,
-		Height:      0,
-	})
+	// We send the shortest chain to the new conodes to let
+	// them know they joined the cothority
+	proof, err := s.db.GetProof(src.SkipChainID())
+	if err != nil {
+		return err
+	}
+
+	newRoster := []*network.ServerIdentity{}
+	for _, si := range dst.Roster.List {
+		if i, _ := src.Roster.Search(si.ID); i < 0 {
+			newRoster = append(newRoster, si)
+		}
+	}
+	log.Lvlf3("%v is propagating %d blocks to %v", s.ServerIdentity(), len(proof), newRoster)
+
+	// current conode needs to be in the propagation roster
+	newRoster = append(newRoster, s.ServerIdentity())
+	return s.startPropagation(s.propagateProof, onet.NewRoster(newRoster), &PropagateProof{proof})
 }
 
 func mapMask(origRoster *onet.Roster, newRoster *onet.Roster, sig []byte) ([]byte, error) {
@@ -1082,6 +1098,10 @@ func (s *Service) bftForwardLinkLevel0(msg, data []byte) bool {
 	}()
 	if ok {
 		s.verifyNewBlockBuffer.Store(sliceToArr(msg), true)
+		// We can cache the block because it has been verified by this conode
+		// but it will be added to the DB later on after the protocol
+		// has succeeded
+		s.blockBuffer.add(fs.Newest)
 	}
 	return ok
 }
@@ -1171,12 +1191,7 @@ func (s *Service) forwardLink(req *network.Envelope) {
 		// BackLinks are already stored in the new block so only
 		// the roster from the previous block needs to add the
 		// forward link
-		return s.startForwardLinkPropagation(from.Roster, &PropagateForwardLink{
-			ForwardLink: fl,
-			Newest:      nil,
-			Roster:      from.Roster,
-			Height:      fs.TargetHeight,
-		})
+		return s.startPropagation(s.propagateForwardLink, from.Roster, &PropagateForwardLink{fl, fs.TargetHeight})
 	}()
 	if err != nil {
 		log.Error(s.ServerIdentity(), "couldn't create forwardLink:", err, "requested by", req.ServerIdentity)
@@ -1357,7 +1372,16 @@ func (s *Service) propagateGenesisHandler(msg network.Message) {
 		return
 	}
 
-	s.db.Store(pg.Genesis)
+	if !s.blockIsFriendly(pg.Genesis) {
+		log.Error("Block is not friendly")
+		return
+	}
+
+	id := s.db.Store(pg.Genesis)
+	if id == nil {
+		// the error is logged in the store function
+		return
+	}
 
 	if !pg.Genesis.ParentBlockID.IsNull() {
 		sb := s.db.GetByID(pg.Genesis.ParentBlockID)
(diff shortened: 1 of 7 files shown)


skipchain/skipchain.go
```diff
@@ -1360,22 +1349,76 @@ func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg
...
+		if sb == nil {
+			log.Error("Couldn't get the block to attach the forward link")
+			return
+		}
+	}
+
+	// Because of the sync, the forward link could already be added
+	// but we need to catch the case where the height is much above
+	// the current one which means we missed previous propagation
+	if len(sb.ForwardLink) <= pfl.Height {
```
jeffallen: Did we have a test case for this before? Did you have to make one?

skipchain/skipchain.go
```diff
@@ -1360,22 +1349,76 @@ func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg
...
 	}
-	_, err := s.db.StoreBlocks(sbs.SkipBlocks)
-	if err != nil {
-		log.Error(err)
+}
+
+// PropagateForwardLinkHandler will update the latest block with
+// the new forward link and the new block when given
+func (s *Service) propagateForwardLinkHandler(msg network.Message) {
+	pfl, ok := msg.(*PropagateForwardLink)
```
jeffallen: We need to think about security here... pfl is essentially untrusted input from the network at this point. Anyone who manages to send the right message into onet can get us to execute this code with their input. What bad stuff can they do? (This was already true before, but I'm wondering if somehow this change opens us up to trying SyncChains we shouldn't be trying?)

skipchain/skipchain.go
```diff
@@ -1360,22 +1349,76 @@ func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg
...
+	if !ok {
+		log.Error("Couldn't convert to slice of ForwardLinks")
+		return
+	}
+
+	// Get the block to update the list of FLs
+	sb := s.db.GetByID(pfl.ForwardLink.From)
+	if sb == nil {
+		// Happen when the roster has changed
+		err := s.SyncChain(pfl.Roster, pfl.ForwardLink.From)
```
jeffallen: pfl.Roster is controlled by the attacker here, that doesn't seem right; they could ask us to try to go talk to whitehouse.gov!

skipchain/skipchain.go
```diff
@@ -1360,22 +1349,76 @@ func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg
...
+	if len(sb.ForwardLink) <= pfl.Height {
+		err := sb.AddForwardLink(pfl.ForwardLink, pfl.Height)
+		if err != nil {
+			log.Error(err)
+			return
+		}
+	}
+
+	// Update the forward link list of the block
+	s.db.Store(sb)
```
jeffallen: I'm not clear where this block was validated before we accept it. Does that happen in some other layer of the code? Or is forgotten here?

skipchain/skipchain_test.go
```diff
@@ -609,12 +609,44 @@ func TestService_Propagation(t *testing.T) {
 	}
 	service := genService.(*Service)
 
-	sbRoot, err := makeGenesisRosterArgs(service, ro, nil, VerificationNone,
+	roster := onet.NewRoster(ro.List[:5])
```
jeffallen: it would be easier to understand this test if these were renamed as: ro -> fullRoster, roster -> firstFiveRoster

skipchain/skipchain.go
```diff
@@ -975,27 +958,16 @@ func (s *Service) forwardLinkLevel0(src, dst *SkipBlock) error {
...
-	}
-	s.startPropagation(proof)
-	return nil
+	ro := roster.Concat(dst.Roster.List...)
+
+	// Propagate the forward link and the new block as the roster
+	// could have changed, meaning one node cannot know the block
+	return s.startForwardLinkPropagation(ro, &PropagateForwardLink{
+		ForwardLink: fwd,
+		Newest:      dst,
```
ineiti: Needed: For optimizing it correctly, you have to store the block in the `bftForwardLinkLevel0`, so this will not send it again.

skipchain/skipchain.go
```diff
@@ -975,27 +958,16 @@ func (s *Service) forwardLinkLevel0(src, dst *SkipBlock) error {
...
+		Roster:      roster,
```
ineiti: Question: I still don't see why you send the roster here.

skipchain/skipchain.go
```diff
@@ -1360,22 +1349,76 @@ func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg
...
+	pg, ok := msg.(*PropagateGenesis)
 	if !ok {
 		log.Error("Couldn't convert to slice of SkipBlocks")
 		return
 	}
-	for _, sb := range sbs.SkipBlocks {
-		if !s.blockIsFriendly(sb) {
-			log.Lvlf2("%s: block is not friendly: %x", s.ServerIdentity(), sb.Hash)
+
+	s.db.Store(pg.Genesis)
```
ineiti: Please use `blockIsFriendly` here to make sure we should accept the block. It is a very simple authorization mechanism that allows to set up following of chains and nodes.

skipchain/skipchain.go
ineiti: I like the reference:
- what is the worst that could happen in this code?
- DDoS attack on whitehouse.gov!
- oh dang - the world is going to end!

Besides the funny reference you're of course completely right.

skipchain/skipchain.go
```diff
@@ -1360,22 +1349,76 @@ func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg
...
+func (s *Service) propagateForwardLinkHandler(msg network.Message) {
+	pfl, ok := msg.(*PropagateForwardLink)
+	if !ok {
+		log.Error("Couldn't convert to slice of ForwardLinks")
+		return
+	}
+
+	// Get the block to update the list of FLs
+	sb := s.db.GetByID(pfl.ForwardLink.From)
+	if sb == nil {
```
ineiti: As far as I understand, if `sb == nil`, then we can return here. What usually happens is this (for level >0 link):
1. leader node asks earlier leader with `ForwardSignature` to create a forward signature
2. earlier leader creates forward signature using `forwardLink`
3. all earlier nodes verify the forward-link and _TODO_: store the block
4. the earlier leader sends `propagateForwardLinkHandler` with the forwardlink

So if a node receives a forward-link
- pointing to/from a block it doesn't has, it can ignore it
- it cannot verify, it can ignore it

If you're in doubt, write me on slack, and we'll whiteboard it.

skipchain/skipchain.go
ineiti: As written above, the block should not even be propagated here. Only taken from the cache (populated by `bftForwardLink`) and then stored if the forward link works out.

skipchain/skipchain.go
Gilthoniel: The roster could actually be taken from the previous blocks. That would fix the three points you pointed out.

skipchain/skipchain.go
Gilthoniel: the skipchain service Store function updates existing blocks with the new forward links after verifying them. If the block as been tempered, the hash is different and it won't be found in the db, then it won't be stored because it can be linked.

That could actually be improved by checking if the id is returned. If not the data is corrupted.

skipchain/skipchain.go
jeffallen: Honestly, right now a DOS on whitehouse.gov's access to twitter.com might be an improvement.

skipchain/skipchain.go
Gilthoniel: I had errors after I changed the propagation test which are actually not errors. As the comment says, if you don't have the previous block and then try to sync, you might end up with the block already updated by another member of the cothority. In my case the leader was sending the block with the forward link already updated.

skipchain/struct.go
```diff
@@ -485,6 +485,58 @@ func (sb *SkipBlock) updateHash() SkipBlockID {
...
+				return fmt.Errorf("Wrong backlink %v %v", sb.BackLinkIDs, sbs[i-1].Hash)
+			}
+		}
+		if i < len(sbs)-1 {
+			// Check if there is a forward link to the next block
+			if len(sb.ForwardLink) == 0 {
+				return errors.New("Missing forward links")
+			}
+
+			fl := sb.ForwardLink[len(sb.ForwardLink)-1]
```
kc1212: are you not checking all the forward links?

Gilthoniel: Only the highest one because the block for the lower ones are not propagated

Gilthoniel:
Addressing the comments required some major changes in order to cache the newest block and store it after it has been verified. i guess you should go over it once again.

skipchain/skipchain.go
```diff
@@ -1360,23 +1369,117 @@ func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg
...
 	if !ok {
 		log.Error("Couldn't convert to slice of SkipBlocks")
 		return
 	}
-	for _, sb := range sbs.SkipBlocks {
-		if !s.blockIsFriendly(sb) {
-			log.Lvlf2("%s: block is not friendly: %x", s.ServerIdentity(), sb.Hash)
+
+	if !s.blockIsFriendly(pg.Genesis) {
+		log.Error("Block is not friendly")
```
ineiti:
```suggestion
		log.Error("Conode doesn't want to follow that skipchain")
```

skipchain/skipchain.go
```diff
@@ -1360,23 +1369,117 @@ func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg
...
+	// before the update.
+	id := s.db.Store(sb)
+	if id == nil {
+		// error already logged
+		return
+	}
+
+	// Add the block if available
+	if sb, ok := s.blockBuffer.get(sb.SkipChainID(), pfl.ForwardLink.To); ok {
+		if !s.blockIsFriendly(sb) {
```
ineiti: Is that possible? We have the genesis block, so I think `blockIsFriendly` will return true anyway.

skipchain/skipchain.go
```diff
@@ -1360,23 +1369,117 @@ func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg
...
+		if id == nil {
+			// error already logged
+			return
+		}
+		s.blockBuffer.clear(sb.SkipChainID())
+	}
+}
+
+// TODO: change messaging package to allow to return an error
+// it makes the function easier to test
```
ineiti: Problem here is that the messages are asynchronous - we don't have a way to return an error to the sender.

skipchain/struct.go
```diff
@@ -1027,3 +1112,57 @@ func (db *SkipBlockDB) getAllSkipchains() (map[string]*SkipBlock, error) {
...
+// Blocks are stored per skipchain to prevent the cache to grow and it
+// is cleared after a new block is added (again per skipchain)
+type skipBlockBuffer struct {
+	buffer map[[32]byte]*SkipBlock
+	sync.Mutex
+}
+
+func newSkipBlockBuffer() *skipBlockBuffer {
+	return &skipBlockBuffer{
+		buffer: make(map[[32]byte]*SkipBlock),
```
ineiti: We're using `string` as key in other places - @jeffallen can surely tell us which one is better: using `string` as key or array of bytes?

skipchain/struct.go
```diff
@@ -1027,3 +1112,57 @@ func (db *SkipBlockDB) getAllSkipchains() (map[string]*SkipBlock, error) {
...
+	defer sbb.Unlock()
+
+	key := [32]byte{}
+	copy(key[:], block.SkipChainID())
+	sbb.buffer[key] = block
+}
+
+// get returns the block if the skipchain ID hits with the
+// correct block ID
+func (sbb *skipBlockBuffer) get(sid SkipBlockID, id SkipBlockID) (*SkipBlock, bool) {
```
ineiti: Why do you return a `bool` to indicate if the block is available or not? Isn't a `== nil` the same?

skipchain/skipchain.go
ineiti: OK - sounds good. Opened dedis/onet#501

PR #1588, opened 2018-11-16
Change contract format to use interfaces and embedding for defaults

### LINKED ISSUE #1294: check contract type in a more organized manner
Right now, we have a proliferation of these:

```go
func (s *Service) ContractConfig(cdb CollectionView, tx Instruction, coins []Coin) (sc []StateChange, c []Coin, err error) {
	if tx.getType() != spawnType {
		return nil, nil, errors.New("Config can only be spawned")
	}
```

or in eventlog:

```go
func (s *Service) contractFunction(v omniledger.CollectionView, tx omniledger.Instruction, c []omniledger.Coin) ([]omniledger.StateChange, []omniledger.Coin, error) {
	if tx.Delete != nil {
		return nil, nil, errors.New("delete tx not allowed")
	}
	if tx.Invoke != nil {
		return nil, nil, errors.New("invoke tx not allowed")
	}
	if tx.Spawn == nil {
		return nil, nil, errors.New("expected a spawn tx")
	}
```

There should be a neater way to do this. For instance, contracts should be objects that fulfil an interface, including the "execution" function, and also a "requires" function which tells omniledger what tx type limits it expects.

Otherwise, there should at least be a public function that does what we are doing in an ad-hoc way now.

jeffallen:
Fixes #1294.


jeffallen:
Having big problems debugging Java side of things, so I'm pushing this for now and will try again on Monday.

The interface-based API for contracts turned out quite nicely, I think.

byzcoin/contracts.go
```diff
@@ -29,9 +29,23 @@ var ConfigInstanceID = InstanceID{}
 // CmdDarcEvolve is needed to evolve a darc.
 var CmdDarcEvolve = "evolve"
 
-// ContractFn is the type signature of the class functions which can be
+// ContractFn is the type signature of the instance factory functions which can be
```
ineiti: Cool! Looks much better, I like it. For the file `contracts.go`, can we either:
1. split it in 
  - `contracts.go` with the interface and the `BasicContract`
  - `contractDarc.go` and `contractConfig.go` (or `contractDarcConfig.go` or better name)
2. change to order in the file (I'm very much used to navigate the file without searching) to have:
  - the interface on top
  - followed by the `BasicContract`
  - followed by `contractConfig` and `contractDarc`

byzcoin/contracts/coins.go
```diff
@@ -32,154 +32,162 @@ var CoinName = iid("olCoin")
...
 	}
-	err = errors.New("instruction type not allowed")
+
+	// Finally update the coin value.
+	var ciBuf []byte
+	ciBuf, err = protobuf.Encode(&c.Coin)
+	sc = append(sc, byzcoin.NewStateChange(byzcoin.Update, inst.InstanceID,
+		ContractCoinID, ciBuf, darcID))
 	return
+	/*	case byzcoin.DeleteType:
```
ineiti: Why is that commented out?

pop/app.go
```diff
@@ -214,6 +214,25 @@ func orgConfig(c *cli.Context) error {
 	return cfg.write()
 }
 
+// rosterEqual checks if the first list contains the second
+func rosterEqual(r1, r2 *onet.Roster) bool {
+	if len(r1.List) != len(r2.List) {
+		return false
+	}
+	for _, p := range r2.List {
```
ineiti: you can use `r1.Contains` here

pop/app.go
```diff
@@ -180,7 +180,7 @@ func orgConfig(c *cli.Context) error {
 		// Check that current party is included in merge config
 		found := false
 		for _, party := range desc.Parties {
-			if service.Equal(desc.Roster, party.Roster) {
+			if rosterEqual(desc.Roster, party.Roster) {
```
ineiti:
```suggestion
			if desc.IsRotation(party) {
```

pop/service/api.go
```diff
@@ -506,8 +506,8 @@ func (desc *PopDesc) Hash() []byte {
 	return hash.Sum(nil)
 }
 
-// Equal checks if the first list contains the second
-func Equal(r1, r2 *onet.Roster) bool {
+// rosterEqual checks if the first list contains the second
+func rosterEqual(r1, r2 *onet.Roster) bool {
```
ineiti: replace it with `r1.IsRotation(r2)`.

pop/service/contract.go
```diff
@@ -37,149 +37,147 @@ func init() {
...
-func (s *Service) ContractPopParty(cdb byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruction, ctxHash []byte, coins []byzcoin.Coin) (scs []byzcoin.StateChange, cOut []byzcoin.Coin, err error) {
-	cOut = coins
-
-	err = inst.Verify(cdb, ctxHash)
+type contract struct {
+	byzcoin.BasicContract
+	PopPartyInstance
+}
+
+func contractPopParty(in []byte) (byzcoin.Contract, error) {
```
ineiti: the other methods were called `contractPopPartyFromBytes` - I propose to use the same naming.

pop/service/contract.go
```diff
@@ -37,149 +37,147 @@ func init() {
...
-		if err != nil {
-			return nil, nil, errors.New("couldn't get instance data: " + err.Error())
-		}
-		err = protobuf.DecodeWithConstructors(ppiBuf, &ppi, network.DefaultConstructors(cothority.Suite))
-		if err != nil {
-			return nil, nil, errors.New("couldn't unmarshal existing PopPartyInstance: " + err.Error())
+func (c *contract) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruction, coins []byzcoin.Coin) (scs []byzcoin.StateChange, cout []byzcoin.Coin, err error) {
+	cout = coins
+
+	/*
```
ineiti: Can this be removed?

pop/service/api.go
jeffallen: That causes an error. But r1.Contains(r2.Publics()) doesn't, so I'll go with that.

PR #1600, opened 2018-11-26
Correctly updating roster from config-instance to new block

### LINKED ISSUE #1578: Strange behaviour of byzcoin.createNewBlock
When a new block is proposed in `byzcoin.createNewBlock`, there is a strange case when the roster is changed:

1. if it is a view-change, then the new leader will already show up in the roster of the new block, and the global state at the time of the block is in synch with the roster in the skipblock
2. if the chainConfig has been updated with a new roster, the global state at the time of the block is not in synch anymore with the roster in the skipblock, and only the next skipblock will hold the new roster (`byzcoin.startPolling:{1269,1360}`)

I propose we make sure that even in 2., the new roster is immediately applied to the new block. However this will require some trick to get the new chainConfig out of the statechanges, as we don't have an updated trie yet.

ineiti:
Closes #1578

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -618,8 +618,10 @@ func (s *Service) createNewBlock(scID skipchain.SkipBlockID, r *onet.Roster, tx 
 		// For a genesis block, we create a throwaway staging trie.
 		// There is no need to verify the darc because the caller does
 		// it.
+		if r == nil {
+			return nil, errors.New("need roster for genesis block")
+		}
 		sb = skipchain.NewSkipBlock()
-		sb.Roster = r
 		sb.MaximumHeight = 32
 		sb.BaseHeight = 4
 		// We have to register the verification functions in the genesis block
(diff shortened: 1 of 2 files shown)


byzcoin/service.go
```diff
@@ -686,12 +682,22 @@ func (s *Service) createNewBlock(scID skipchain.SkipBlockID, r *onet.Roster, tx
...
+		sb.Roster = r
+	}
 	var ssb = skipchain.StoreSkipBlock{
 		NewBlock:          sb,
 		TargetSkipChainID: scID,
 	}
 	log.Lvlf3("Storing skipblock with %d transactions.", len(txRes))
-	ssbReply, err := s.skService().StoreSkipBlock(&ssb)
+	var ssbReply *skipchain.StoreSkipBlockReply
+	if sb.Roster.List[0].Equal(s.ServerIdentity()) {
```
Gilthoniel: According to line 685, the roster could be nil. Either we check the function parameter to force it won't or this needs to be checked against nil pointer access.

ineiti: Added test in genesis-block creation.

PR #1604, opened 2018-11-28
Cleanup Locking and Errors

ineiti:
The locking and waitgroup counting in Byzcoin was a bit off - fixed.
The errors returned when loading a trie were uncomprehensible when printed on the command line - fixed.


byzcoin/service.go
```diff
@@ -1250,17 +1248,18 @@ func (s *Service) LoadBlockInfo(scID skipchain.SkipBlockID) (time.Duration, int,
 }
 
 func (s *Service) startPolling(scID skipchain.SkipBlockID) chan bool {
+	s.pollChanWG.Add(1)
 	closeSignal := make(chan bool)
 	go func() {
 		s.closedMutex.Lock()
+		s.pollChanWG.Done()
```
Gilthoniel: I might have missed something but I think you can remove the wait group. You removed the different places where it was incremented and this one will be done right after the increment and the poll channel is closed when cleaning goroutines then it's not a missing defer or something.

ineiti: Thanks, that should be `defer`red

Issue #1589, opened 2018-11-19
integrate blscosi into cothority


jeffallen:
The task is to take the code from here: https://github.com/dedis/student_18_blsftcosi
(check with @swarali which branch)

You will also need to change the kyber/suites/pairing/bn256 suite to not require the vartime build tag (modifications will also be necessary in the suite registry, kyber/suites)

This task will be done when:
* blscosi is available in the cothority repo, with approximately matching code quality, documentation and test coverage to the byzcoinx directory.
* jenkins and travis are happy with the mix of constant time and vartime kyber code

PR #1614, opened 2018-12-07
Blsftcosi 1562

### LINKED ISSUE #1562: Replace CoSi with BLSCoSi
from Lefteris:

_I spoke with the main authors of this paper https://eprint.iacr.org/2018/417.pdf
They told that there is follow-up work breaking schnorr base CoSi (able to forge signatures if there are a few dozen CoSi rounds running concurrently)
As a result I would propose to bump the priority of swapping to BLSCoSi that is provable secure
@swarali is currently working on testing the implementation we have but she has trouble when trying to run to more than a dozen nodes_

Gilthoniel:
The purpose of the this PR is to use the BLS signature algorithm in the cosi protocol to fix the security concern about the current implementation.

First of all, the blsftcosi package has been brought for the student project (@svarali) and cleaned to respect DEDIS rules. Then the ByzcoinX protocol has been fixed to work with this new algorithm and the same for the skipchain and other services that are using byzcoinx.

Fixes #1562 #1589


blsftcosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,431 @@
...
+			}
+		case reply := <-p.ChannelRefusal:
+			_, ok := responses[reply.ID]
+			serviceName := onet.ServiceFactory.Name(p.Token().ServiceID)
+
+			if !ok {
+				log.Warnf("Got a message from an unknown node %v", reply.ServerIdentity.ID)
+			} else if err := bls.Verify(p.suite, reply.ServerIdentity.ServicePublic(serviceName), reply.Nonce, reply.Signature); err == nil {
+				// The child gives an empty signature as a mark of refusal
+				responses[reply.ID] = &Response{}
```
kc1212: I have a feeling that using the treeID as the key might be a security risk. Is the treeID verified by onet before it gets to this point? In order words, can malicious nodes forget the treeID? It might be safer to use the public key as the key to `responses` because at you can verify that they actually hold the private key by checking the signature.

blsftcosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,431 @@
...
+				done++
```
kc1212: What happens if the same node sends more than one refusal message? Will the done counter be incremented twice and make the loop terminate early?

blsftcosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,431 @@
...
+	"crypto/rand"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/dedis/kyber"
+	"github.com/dedis/kyber/pairing"
+	"github.com/dedis/kyber/sign/bls"
+	"github.com/dedis/kyber/sign/cosi"
```
kc1212: i see that we need to use the mask so we need to import cosi here. but is the cosi mask independent from cosi? if that's the case we probably should move it into its own package like `kyber/sign/mask`?

blsftcosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+
+	sig, err := bls.Sign(p.suite, p.Private(), nonce)
+
+	return &Refusal{Signature: sig, Nonce: nonce}, err
+}
+
+// makeAggregateResponse takes all the responses from the children and the subleader to
+// aggregate the signature and the mask
+func makeAggregateResponse(suite pairing.Suite, publics []kyber.Point, responses ResponseMap) (*Response, error) {
```
kc1212: is it possible to simplify this method using what is in https://github.com/dedis/kyber/pull/338/files ?

blsftcosi/protocol/sub_protocol.go
Gilthoniel: That would make sense yes, the mask should work for any kind of public keys

blsftcosi/protocol/sub_protocol.go
Gilthoniel: I don't think it's worth it. You still need to aggregate the mask so the loop can't be avoided and thus you save maybe one line.

.travis.yml
```diff
@@ -5,20 +5,20 @@ dist: trusty
...
-  - git clone . `go env GOPATH`/src/github.com/dedis/cothority
-  - (cd `go env GOPATH`/src/github.com/dedis/cothority && go get -t ./... )
+# install:
+#   - gimme 1.11
+#   - . $HOME/.gimme/envs/go1.11.env
+#   - go get github.com/dedis/Coding || true
+#   # Because we are using "language: node_js" the "git clone" is not in the
+#   # GOPATH. So make a copy of it over where it is supposed to be.
+#   - git clone . `go env GOPATH`/src/github.com/dedis/cothority
+#   - (cd `go env GOPATH`/src/github.com/dedis/cothority && go get -t ./... )
```
kc1212: why is this part commented out?

blsftcosi/README.md
```diff
@@ -0,0 +1,73 @@
...
+[Cothority](../README.md) ::
+[Building Blocks](../doc/BuildingBlocks.md) ::
+Fault Tolerant Collective Signing
+
+# Fault Tolerant Collective Signing
+
+This package provides functionality to request and verify collective signatures
+as well as run a standalone server for handling collective signing requests.
+It is a fault tolerant version of CoSi, implemented in
+[cosi](../cosi/README.md) package.
```
kc1212: This part needs to be updated to mention BLS.

blsftcosi/README.md
```diff
@@ -0,0 +1,73 @@
...
+You can read more about BLS signature by refering to
+[the paper](https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html).
+
+## Description
+
+The purpose of this work is to implement a robust and scalable consensus
+algorithm using ftCoSi protocol and handling some exceptions. The ftCoSi tree
+is a three level tree to make a compromise between the two-level tree, making
+the root-node vulnerable to DoS, and a more than three level tree, slowing the
+algorithm because of the RTT between the root node and the leaves.
```
kc1212: also needs updating

blsftcosi/README.md
```diff
@@ -0,0 +1,73 @@
...
+      /       \
+     /         \
+  leaf_1 ... leaf_n
+```
+
+Namely, if there are m sub-leaders, the root will run m sub-protocols. The
+sub- protocols do bulk of the work (collective signatures) and communicates
+the result to the main protocol via channels.
+
+- [BlsFtCosi CLI](blscosi/README.md) is a command line interface for interacting with ftCoSi
```
kc1212: why is the package/folder name called `blscosi` and not `blsftcosi`? maybe it's better to name it to `blsftcosicli` (it's a bit of a mouthful.. i'm open to suggestions). Also, "ftCoSi" need to be updated.

blsftcosi/blscosi/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+# BLS CoSi CLI
+
+To use the code of this package you need to:
+
+- Install [Golang](https://golang.org/doc/install)
+- Optional: Set [`$GOPATH`](https://golang.org/doc/code.html#GOPATH) to point to your workspace directory
+- Put $GOPATH/bin in your PATH: `export PATH=$PATH:$(go env GOPATH)/bin`
+
+To build and install the ftCoSi application, execute:
```
kc1212: bltFtCoSi

blsftcosi/blscosi/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+ftcosi verify -g $COTHORITY -s file.sig file
+```
+
+To check the status of a collective signing group, use:
+
+```
+ftcosi check -g $COTHORITY
+```
```
kc1212: these need updating

blsftcosi/blscosi/README.md
```diff
@@ -0,0 +1,87 @@
...
+## References
+- OmniLedger: A Secure, Scale-Out, Decentralized Ledger via Sharding: https://eprint.iacr.org/2017/406.pdf part 4 A & B
+- (CoSi) Keeping Authorities "Honest or Bust" with Decentralized Witness Cosigning: https://arxiv.org/abs/1503.08768
+- (ByzCoin) Enhancing Bitcoin Security and Performance with Strong Consistency via Collective Signing: https://arxiv.org/abs/1602.06997
+
+
+## Further Information
+
+For more details, e.g., to learn how you can run your own CoSi server or cothority, see the [wiki](https://github.com/dedis/cothority/wiki/CoSi).
+The same applies to ftCoSi.
```
kc1212: and this

blsftcosi/README.md
kc1212: but if we can't think of a better name then let's keep it as `blscosi` for the CLI

external/java/src/main/java/ch/epfl/dedis/lib/crypto/Bn256G2Point.java
```diff
@@ -0,0 +1,131 @@
+package ch.epfl.dedis.lib.crypto;
+
+import ch.epfl.dedis.lib.Hex;
+import com.google.protobuf.ByteString;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
```
kc1212: I've had troubles with the sun standard libraries before. I think it's not in all the JDKs. Could you change it to https://docs.oracle.com/javase/8/docs/api/java/lang/UnsupportedOperationException.html ?

blsftcosi/protocol/sub_protocol.go
```diff
@@ -0,0 +1,444 @@
...
+	// If an error happens when sending the announcement, we can assume there
+	// will be a timeout from this node
+	done := len(errs)
+	for done < len(p.Children()) {
+		select {
+		case <-p.closeChan:
+			return nil
+		case reply, ok := <-p.ChannelResponse:
+			public := searchPublicKey(p.TreeNodeInstance, reply.ServerIdentity)
+			r, ok := responses[public.String()]
```
kc1212: would this panic if `public` is nil?

.travis.yml
Gilthoniel: Mostly to keep the "hack" to use the correct repos in the before_install to have everything at the same place

blsftcosi/README.md
Gilthoniel: Yeah I wasn't sure which name to use. I remove the _ft_  to make it shorter as it's not fundamental and wanted to avoid cli in the name as the package name is used as the executable name.

Gilthoniel: Let's take feedback from @jeffallen and @ineiti for that

ineiti: Wondering whether we actually need a CLI for that. Probably for demo purposes it would be nice to have one. Some more suggestions:
- `blssign`
- `bfcsign` - BlsFtCosiSIGN
- `blscosi`

we could even remove the `ft` part entirely from the protocol, as this was mostly used to distinguish between the `cosi` and the `ftcosi`. But as there is no _non-fault-tolerant_ bls-cosi, we can remove it altogether and make the name a bit shorter.

My 500 Satoshis (2 cents)

external/java/src/main/java/ch/epfl/dedis/lib/crypto/Bn256G2Point.java
jeffallen: Yeah, this makes problems on Android too. Thanks for fixing.

PR #1615, opened 2018-12-10
Byzcoin.TestService_AddTransaction_WithFailure fix

ineiti:
The 'WithFailure' test is failing only in Jenkins due to a timing-problem.
This PR is passing (currently) in Jenkins, and also locally.

### DIFF
--- a/skipchain/api.go
+++ b/skipchain/api.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"strings"
 	"time"
 
 	"github.com/dedis/cothority"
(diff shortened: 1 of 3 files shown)


skipchain/api.go
```diff
@@ -335,9 +336,17 @@ func (c *Client) GetSingleBlock(roster *onet.Roster, id SkipBlockID) (reply *Ski
 // It returns that block, or an error if that block is not found.
 func (c *Client) GetSingleBlockByIndex(roster *onet.Roster, genesis SkipBlockID, index int) (reply *GetSingleBlockByIndexReply, err error) {
 	reply = &GetSingleBlockByIndexReply{}
-	err = c.SendProtobuf(roster.RandomServerIdentity(),
-		&GetSingleBlockByIndex{genesis, index}, reply)
-	return
+	perms := rand.Perm(len(roster.List))
```
Gilthoniel: btw we never set the seed, so it's not really random

ineiti: true - but all that can happen is that the access pattern can be known. Is that bad?

Gilthoniel: Not really bad. It's more for the unlikely scenario that several clients do the same at the same time so you contact the same nodes in a short period of time. Not critical though.

PR #1623, opened 2018-12-11
Eventlog update

jeffallen:
Fix the integration test, fix other things.

### DIFF
--- a/eventlog/el/test.sh
+++ b/eventlog/el/test.sh
@@ -3,6 +3,10 @@
 DBG_TEST=2
 DBG_SRV=0
 
+# Use 3 servers, use all of them, don't leave one down.
+NBR=3
+NBR_SERVERS_GROUP=$NBR
+
 . "$(go env GOPATH)/src/github.com/dedis/cothority/libtest.sh"
 
 # Use a local config dir.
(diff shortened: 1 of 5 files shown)


eventlog/el/test.sh
```diff
@@ -21,15 +25,27 @@ main(){
 	stopTest
 }
 
+# This is necessary becaues el does not remember the current signer counter from invocation
+# to invocation, it always gets it from the server. And so if you call el twice
+# in one block, the second one will use the same counter as the first one did.
+waitBlock(){
```
Gilthoniel: Isn't it bad to have to rely on that to make the test pass ? I mean, what if a external person uses a script to send multiple logs ? He/She will have the same issue. The error returned might be clear enough but isn't it on our side to have something to deal with that ? The log command could wait for a confirmation or the refresh signer counter could be more clever, I don't know. It might be worth to work on that but that's the kind of test that could fail for no obvious reason if for some reason the delay is above 0.5 sec.

jeffallen: I agree, 1000%. The problem is that the el tool was designed before we understood that it would be helpful/necessary to remember the signer counter with the private key. Re-engineering el right now is not my highest priority. The el tool is a bit a testing toy we use, the actual target client for eventlog is SecureKG, who will call it from Scala. (But we absolutely need to update the Scala example to demonstrate how to track the counters on the client side.)

PR #1636, opened 2018-12-17
Implement bn256 in Java

### LINKED ISSUE #1626: bn256 for Java
Since we're switching to bn256 for skipblock signatures, we need the same on java (and later js) in order to verify the blocks.

I'm making this issue to note down my investigation on the possible bn256 implementations that we could use on the java side.

The first I tested is https://github.com/herumi/mcl it's mostly written in assembly but there's a Java FFI. I cloned the repo and tried the tests. Other than changing some gcc include paths the tests seem to work. The license is BSD so that's fine too. The biggest downside is that java is only a wrapper, so the building and packaging process is more going to become cumbersome. At the moment the developer do not provide a package release.

kc1212:
This is an implementation ported directly from https://github.com/golang/crypto/tree/505ab145d0a99da450461ae2c1a9f6cd10d1f447/bn256 (note the revision number). The goal is to keep it as close to the golang implementation as possible so that we can easily compare and keep it up-to-date. Which means some parts might not be idiomatic in Java, which is the intention. 

Of course, there are differences that are impossible to avoid. The main one is that we do not use `bnPool` because the Java `BigInteger` is immutable, so there is no reason to keep the pool to save allocation. Further, the random point/field generation is also a bit different due to differences in standard libraries.

This PR is intended to be the first step in a series of PRs that will add pairing based crypto to Java. The next step will be to add a wrapper class, e.g., `ch.epfl.dedis.crypto.lib.crypto.{G1Scalar,G1Point,...}` that uses the `bn256` package (this one) but wraps it in a kyber-like way. Following that, the skipblock verification.

Finally, some bench testing confirmed that there is no different in the byte representation between this implementation, the golang standard library one and the kyber/pairing one, even when the two are doing the the marshalling/unmarshalling in very different ways.

Closes #1626


external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/BN.java
```diff
@@ -0,0 +1,455 @@
...
+         * @return the marshalled bytes.
+         */
+        public byte[] marshal() {
+            final int numBytes = 256/8;
+
+            if (this.p.isInfinity()) {
+                return new byte[numBytes*2];
+            }
+
+            this.p.makeAffine();
```
ineiti: I prefer calling the `makeAffine` on a copy, else you 'change' the underlying point. Even though it should only be changed in a way that the point is 'identical', it still changes it.

Or perhaps that's just my experience with go-races in this code...

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/CurvePoint.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private CurvePoint(BigInteger x, BigInteger y, BigInteger z, BigInteger t) {
+        this.x = x;
+        this.y = y;
+        this.z = z;
+        this.t = t;
+    }
+
+    public String toString() {
+        this.makeAffine();
```
ineiti: same comment as in `G1.marshal`

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/CurvePoint.java
```diff
@@ -0,0 +1,195 @@
...
+                sum.set(t);
+            }
+        }
+
+        this.set(sum);
+        return this;
+    }
+
+
+    CurvePoint makeAffine() {
```
ineiti: Just an idea: I wonder whether `makeAffine` could return a new point, then the caller can decide if he wants to actually set the point to the affine point or not.

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/GFp2.java
```diff
@@ -0,0 +1,179 @@
...
+
+    GFp2 exp(GFp2 a, BigInteger power) {
+        GFp2 sum = new GFp2();
+        sum.setOne();
+        GFp2 t = new GFp2();
+
+        for (int i = power.bitLength() - 1; i >= 0; i--) {
+            t.square(sum);
+            if (power.testBit(i)) {
+                sum.mul(t, a);
```
ineiti: Is this correct? No, just kidding ;)

external/java/src/test/java/ch/epfl/dedis/lib/crypto/bn256/BNTest.java
```diff
@@ -0,0 +1,238 @@
...
+        k3.scalarMul(k3, c);
+        byte[] k3Bytes = k3.marshal();
+        assertTrue(Arrays.equals(new BN.GT().unmarshal(k3Bytes).marshal(), k3Bytes), "failed to unmarshal GT k3");
+
+        assertFalse(!Arrays.equals(k1Bytes, k2Bytes) || !Arrays.equals(k2Bytes, k3Bytes), "keys didn't agree");
+    }
+
+    @Test
+    void benchmarkPairing() {
+        for (int i = 0; i < 10; i++) {
```
ineiti: For a benchmark, that's short... Perhaps should also benchmark multiplications and additions.

kc1212:
bn256 doesn't have a standard as far as I'm aware, so there are no test vectors unfortunately. I ported all the tests in from golang/crypto/bn256 to this package. there are also no test vectors in the cloudflare's implementation https://github.com/cloudflare/bn256/blob/master/bn256.go

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/BN.java
kc1212: Good point, this actually touches on a bigger problem. If we want to make it fully thread-safe, we need to make sure no operation can change the underlying point (at the moment every operation changes the inner points/elements). Effectively, making the classes immutable (final classes), just like `BigInteger`.

I feel there are pros/cons to this approach. Making it thread-safe will make the package look and feel quite different from the one in golang/crypto and it'll be harder to port future changes. Keeping the classes mutable as the one in golang/crypto will be easier to maintain. We can write a comment and say all the classes in this package are not thread-safe and users should take appropriate measures (e.g., locking/copying).

On the other hand, if our clients are very multi-threaded, then we definitely need to make the classes final. What are your opinions @jeffallen @Gilthoniel ?

external/java/src/test/java/ch/epfl/dedis/lib/crypto/bn256/BNTest.java
kc1212: This is ported from the golang benchmark test. I don't think junit lets you write benchmark tests the same way as go, so my idea was to simply let the use change the value "10" to something else when they actually need to do a benchmark. I can write a comment to explain.

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/BN.java
Gilthoniel: I'm facing race conditions because of this approach in the go code so I would go in Linus' direction. I'm not sure we need to have the whole class immutable, in my opinion that's better but if you say that it will be hard to maintain, it's probably not worth. The real issue is this makeAffine that is used in stuff you're not expecting writes from it, like the marshal function. I'd change only this to return the affine result instead of mutate the class if that makes sense.

kc1212: 1. Probably no big changes
2. I can't use `dedis/kyber` as reference because it's using assembly, so the internal implementation is very different and you cannot use it as a reference for porting to java. `golang/crypto` only uses `big.Int` so it's much easier to port.
3. I'm sure it will be superseded, but I doubt that will be soon. Right now we don't have a candidate curve as far as I know. When there is one, we cannot use it immediately because it might not be secure. We'll have to wait for the security research community to scrutinize it.

kc1212: > I'm facing race conditions because of this approach in the go code so I would go in Linus' direction. I'm not sure we need to have the whole class immutable, in my opinion that's better but if you say that it will be hard to maintain, it's probably not worth. The real issue is this makeAffine that is used in stuff you're not expecting writes from it, like the marshal function. I'd change only this to return the affine result instead of mutate the class if that makes sense.

I will look at whether it's easy to make the class final.

kc1212:
Moving it back to WIP while I investigate the possibility of making the classes final.

kc1212:
Correction: in java, a final class is a class that cannot be extended (not exactly what we want), we need final fields/attributes.

kc1212:
Moving back to R4M: added some simple checks to test the implementation and copy G1/G2 before marshalling or converting to string.

It's quite involved if we want to mark everything final because the type signatures would all change. To make it nice, the signatures would become similar to `BigInteger`, e.g., `p.negate() vs p.negate(q)` so pretty much every line has to be revisited...

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/BN.java
```diff
@@ -0,0 +1,521 @@
...
+         * @param rnd is the random source.
+         * @return a pair of a point and a scalar.
+         */
+        public static PairG1 rand(Random rnd) {
+            BigInteger k = randPosBigInt(rnd, Constants.order);
+            G1 p = new G1().scalarBaseMul(k);
+            return new PairG1(k, p);
+        }
+
+        public String toString() {
```
Gilthoniel: Override (cannot add the @ without pinging someone..)

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/BN.java
```diff
@@ -0,0 +1,521 @@
...
+         * @param rnd is the random source.
+         * @return a pair of a point and a scalar.
+         */
+        public static PairG2 rand(Random rnd) {
+            BigInteger k = randPosBigInt(rnd, Constants.order);
+            G2 p = new G2().scalarBaseMul(k);
+            return new PairG2(k, p);
+        }
+
+        public String toString() {
```
Gilthoniel: Override

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/BN.java
```diff
@@ -0,0 +1,521 @@
...
+        /**
+         * Construct a new GT object from a GFp12 object.
+         *
+         * @param p is the GFp12 object.
+         */
+        public GT(GFp12 p) {
+            this.p = p;
+        }
+
+        public String toString() {
```
Gilthoniel: Override

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/CurvePoint.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    private CurvePoint(BigInteger x, BigInteger y, BigInteger z, BigInteger t) {
+        this.x = x;
+        this.y = y;
+        this.z = z;
+        this.t = t;
+    }
+
+    public String toString() {
```
Gilthoniel: Override

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/GFp12.java
```diff
@@ -0,0 +1,170 @@
...
+        this.x = new GFp6();
+        this.y = new GFp6();
+    }
+
+    GFp12(GFp12 a) {
+        this.x = new GFp6(a.x);
+        this.y = new GFp6(a.y);
+    }
+
+    public String toString() {
```
Gilthoniel: Override

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/GFp6.java
```diff
@@ -0,0 +1,242 @@
...
+        this.z = new GFp2();
+    }
+
+    GFp6(GFp6 a) {
+        this.x = new GFp2(a.x);
+        this.y = new GFp2(a.y);
+        this.z = new GFp2(a.z);
+    }
+
+    public String toString() {
```
Gilthoniel: Override

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/TwistPoint.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    private TwistPoint(GFp2 x, GFp2 y, GFp2 z, GFp2 t) {
+        this.x = x;
+        this.y = y;
+        this.z = z;
+        this.t = t;
+    }
+
+    public String toString() {
```
Gilthoniel: Override

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/BN.java
```diff
@@ -0,0 +1,521 @@
...
+import java.util.Arrays;
+import java.util.Random;
+
+public class BN {
+    /**
+     * The pair of a G1 point and a scalar that is used to generate it.
+     */
+    public static class PairG1 {
+        public BigInteger k;
+        public G1 p;
```
Gilthoniel: I know you want to keep stuff as the go implementation but it might be worth to make that private if it doesn't require to many changes (honestly I didn't check). It's just that you can't really change the attributes as they're bounded together. I leave a comment only here so you decide if it's useful or not for here and other places.

PR #1637, opened 2018-12-18
Implemented QR Code generation in bcadmin for Config export

### LINKED ISSUE #1603: Implement functionalities related to PopCoins in bcadmin
Implementing new features in bcadmin dedicated to use it along with PopCoins.

 * Showing QR Code of BCConfig with or without the admin keypair

LMeynent:
Closes #1603 

Adds the possibility to generate QR Codes from bcadmin, both with and without the admin credentials.
The QR code is designed to be able to be scanned by the PopCoins app


byzcoin/bcadmin/lib/config.go
```diff
@@ -30,6 +30,19 @@ type Config struct {
...
 }
 
+//BaseConfig is a smaller version of Config containing only the necessary material
+//to export the Config using, for example, a QR code
+type BaseConfig struct {
+	ByzCoinID skipchain.SkipBlockID
+}
+
+//AdminConfig is a smaller version of Config containing only the necessary material
+//to export the Config with its admin credentials using, for example, a QR code
```
jeffallen:
```suggestion
//to export the Config with its admin credentials using, for example, a QR code.
```

byzcoin/bcadmin/main.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
 
+	"encoding/json"
+	qrgo "github.com/skip2/go-qrcode"
```
jeffallen: CISC uses github.com/qantik/qrgo. Why use a different generator here? Or can we change CISC to use the same one? Or are they actually forks of the same code?

Anyway, please find a way to keep the number of dependencies the same.

byzcoin/bcadmin/main.go
```diff
@@ -151,6 +153,27 @@ var cmds = cli.Commands{
...
 	},
+	{
+		Name:    "qrcode",
+		Usage:   "generates a QRCode containing the description of the BC Config",
+		Aliases: []string{"qr"},
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:   "bc",
+				EnvVar: "BC",
+				Usage:  "the ByzCoin config to use (always use)",
```
jeffallen:
```suggestion
				Usage:  "the ByzCoin config to use (required)",
```

byzcoin/bcadmin/main.go
```diff
@@ -692,6 +715,49 @@ func darcRuleDel(c *cli.Context, d *darc.Darc, action string, signer *darc.Signe
...
 
+func qrcode(c *cli.Context) error {
+	bcArg := c.String("bc")
+	if bcArg == "" {
+		return errors.New("--bc flag is required")
+	}
+
+	out := c.String("out")
+	if out == "" {
+		return errors.New("--out flag is required")
```
jeffallen: Please make --out not required, and if not provided, write to os.Stdout (you will need to use qrgo.Write, which takes an io.Writer as it's second parameter). This makes it more Unixy, which always ends up having unexpected benefits down the road. (see http://www.tldp.org/LDP/GNU-Linux-Tools-Summary/html/x12429.htm#UNIX-TOOLS-FURTHER-READING)

Or another interesting behavior with n --out would be, like CISC, to print the QR code in the terminal directly...

byzcoin/bcadmin/test.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-DBG_TEST=1
+DBG_TEST=2
```
jeffallen: Put this back to 1.

Also, there's no new integration test for qrcode. There should be.

byzcoin/bcadmin/main.go
```diff
@@ -151,6 +153,27 @@ var cmds = cli.Commands{
 		},
 		Action: darcCli,
 	},
+	{
+		Name:    "qrcode",
```
jeffallen: Please call the sub-command 'qr', which is more concise.

darc/darc.go
```diff
@@ -706,6 +706,11 @@ func (s Signer) GetPrivate() (kyber.Scalar, error) {
 	}
 }
 
+func (s Signer) String() (string, error) {
+	prk, err := s.GetPrivate()
+	return "{Pub:" + s.Identity().String() + ", Priv:" + prk.String() + "}", err
```
jeffallen: I'm not comfortable with making it quite so easy to print out a private key by mistake. Can you explain what problem this new code solves?

byzcoin/bcadmin/lib/config.go
```diff
@@ -30,6 +30,19 @@ type Config struct {
 	AdminIdentity darc.Identity
 }
 
+//BaseConfig is a smaller version of Config containing only the necessary material
```
jeffallen: I think these structures are only used in the QR generator to tell the JSON encoder what to do. If so, then move them inside of the qr() function. (Types can be defined anywhere variables can, and have the same limited scope. Smaller scope means less for the reader of your code to think about. see https://research.swtch.com/names for a meditation on how the scope of something in a program effects the brainpower needed to process it.)

Also: their type name should be lower case, since they do not need to be exported.

byzcoin/bcadmin/main.go
LMeynent: Done, qrcode still usable as an alias

byzcoin/bcadmin/main.go
LMeynent: Fixed

byzcoin/bcadmin/main.go
LMeynent: I do not see any other exported methods in qrgo than NewQR and OutputTerminal.

For now, the bcadmin will thus only be able to print the result into terminal, and the user can save the output using the ">" and "cat" commands

byzcoin/bcadmin/test.sh
LMeynent: Fixed for DBG_TEST, forgot it once more...

Put a test for the QR Code, but the tester will still have do scan the QR Code manually

byzcoin/bcadmin/lib/config.go
LMeynent: Fixed

darc/darc.go
jeffallen: I am modifying this code so that putting the private and public keys into the JSON message happens in bcadmin.

kc1212:
Might need to be rebased to make sure the java tests run correctly.

byzcoin/bcadmin/main.go
```diff
@@ -13,12 +13,17 @@ import (
...
 	"github.com/dedis/kyber/util/random"
 	"github.com/dedis/onet"
 	"github.com/dedis/onet/app"
 	"github.com/dedis/onet/cfgpath"
 	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
+
+	"encoding/json"
+
+	"github.com/qantik/qrgo"
```
kc1212: what's the reason for using this library instead of something that's more well-known like https://github.com/skip2/go-qrcode

jeffallen: We were already using it in cisc, so I asked @LMeynent to not introduce another dependency. Also, qrgo is interesting because it writes to the terminal, instead of requiring you to open a PNG.

PR #1639, opened 2018-12-19
Route logs from exec commands to java logger

### LINKED ISSUE #1426: Correctly log container when restarting new conode
Now: https://github.com/dedis/cothority/blob/master/external/java/src/test/java/ch/epfl/dedis/integration/DockerTestServerController.java#L103
ignores stdout and stderr, which makes it really hard to debug conodes that are being restarted. One simple way out is to add:

```
        dockerClient.execStartCmd(execCreateCmdResponse.getId())
                .exec(new ExecStartResultCallback(System.out, System.err)).awaitStarted();
```

But that writes everything to stdout instead of the logging-service. Fiddling around with Java for 15 minutes didn't show me the correct method to call. Any ideas welcome!

jeffallen:
The logs from the container as a whole are routed to
the logger via withLogConsumer. But for a conode
restarted via exec, the logs need to be fetched by a
callback.

Fixes #1426.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ node_js:
   - "lts/*"
 
 env:
-#   - TEST_DIR=cothority
   - TEST_DIR=kyber
 
 install:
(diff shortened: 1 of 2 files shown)


kc1212:
The commit history looks a bit messy. The changes in those two new commits are not big, was rebase not possible (or cherry-pick)?

jeffallen:
I have no idea what git did, I'll figure it out and push it again.

PR #1642, opened 2018-12-21
Fix eventlog integration test

Gilthoniel:
This fixes the eventlog integration test by adding
an option to wait for block inclusion.

The wait block routine doesn't seem to work well on Jenkins and you should anyway be able to know when the event has been sent through because of the signer counter.

### DIFF
--- a/eventlog/el/main.go
+++ b/eventlog/el/main.go
@@ -135,7 +135,7 @@ var cmds = cli.Commands{
 			},
 			cli.IntFlag{
 				Name:  "wait, w",
-				Usage: "wait for block inclusion",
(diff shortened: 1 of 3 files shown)


eventlog/el/main.go
```diff
@@ -133,6 +133,11 @@ var cmds = cli.Commands{
 				Name:  "content, c",
 				Usage: "the text of the log",
 			},
+			cli.IntFlag{
+				Name:  "wait, w",
+				Usage: "wait for block inclusion",
```
jeffallen:
```suggestion
				Usage: "wait for block inclusion (default: do not wait)",
```

PR #1647, opened 2019-01-07
Renegerate tomls using the new way to marshal G2

kc1212:
The way to marshal G2 is change, so the toml files need to regenerate
because they use the old way of marshalling.


kc1212:
Depends on https://github.com/dedis/kyber/pull/349

PR #1648, opened 2019-01-07
Java mask

kc1212:
The consequence of implementing the mask is that we need to add a
getZero method to the Point interface. Because we need to start from the
zero point when doing the aggregate.

The isZero function of ed25519 is updated to match what is in Kyber, not
completely sure what the consequences are but the tests are passing.

Waiting for https://github.com/dedis/cothority/pull/1640

### DIFF
--- a/external/java/src/main/java/ch/epfl/dedis/lib/crypto/Mask.java
+++ b/external/java/src/main/java/ch/epfl/dedis/lib/crypto/Mask.java
@@ -1,0 +1,91 @@
+package ch.epfl.dedis.lib.crypto;
+
+import ch.epfl.dedis.lib.Hex;
+import ch.epfl.dedis.lib.exception.CothorityCryptoException;
+
+public class Mask {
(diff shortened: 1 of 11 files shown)


external/java/src/main/java/ch/epfl/dedis/lib/crypto/Mask.java
```diff
@@ -0,0 +1,125 @@
...
+public class Mask {
+    private final byte[] mask;
+    private final Point[] publics;
+    private Point aggregate;
+
+    /**
+     * Create a read-only mask from a set of public keys and the given bitmask.
+     *
+     * @param publics is the set of public keys.
+     * @param mask is the bit mask, the number of bits must be greater or equals to the number of public keys.
```
jeffallen:
```suggestion
     * @param mask is the bit mask, the number of bits must be greater than or equal to the number of public keys.
```

PR #1649, opened 2019-01-09
Java forwardlink verification

kc1212:
Implement skipblock signature verification.


external/java/src/main/java/ch/epfl/dedis/lib/SkipBlock.java
```diff
@@ -94,6 +103,42 @@ public int getIndex(){
...
+    /**
+     * This function checks whether all signatures in the forward-links
+     * are correctly signed by the aggregate public key of the roster
+     *
+     * @return true if the signature is ok.
+     */
+    public boolean verifyForwardSignatures() {
+        List<Point> publics;
+        try {
+            publics = new Roster(this.skipBlock.getRoster()).getServicePublics("Skipchain");
```
Gilthoniel: Do we have a constant or something for the "Skipchain" service name? Just in case the service name changes.

external/java/src/main/java/ch/epfl/dedis/lib/SkipBlock.java
```diff
@@ -94,6 +103,42 @@ public int getIndex(){
...
+        } catch (URISyntaxException e) {
+            return false;
+        }
+
+        for (ForwardLink fl : this.getForwardLinks()) {
```
Gilthoniel: Should we have a check that at list one is not empty if the length is above 0 ?

kc1212: I think it's ok if there are no forward links. It might happens when the block is the latest.

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
```diff
@@ -74,39 +84,114 @@ public SkipBlock getSkipblock(SkipblockId id) throws CothorityCommunicationExcep
...
 
-            if (reply.getUpdateCount() == 0){
-                logger.info("didn't find any updates to {}", scID);
-                return null;
+            // Step through the returned blocks one at a time, verifying
+            // the forward links, and that they link correctly backwards.
+            for (int j = 0; j < r2.getUpdateCount(); j++) {
+                SkipBlock b = new SkipBlock(r2.getUpdateList().get(j));
+                if (j == 0 && update.size() > 0) {
+                    if (Arrays.equals(update.get(update.size()-1).getHash(), b.getHash())) {
```
jeffallen: I think this line says, "if we are processing the first block, and we've already processed some blocks, then make sure this first block is the same as the last block we've already accepted".

If I'm correct, this is what we want to be doing, so good on ya mate. So add a comment saying so.

If I'm not correct, it means this code is too confusing and needs to be either fixed or documented.

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
```diff
@@ -74,39 +84,114 @@ public SkipBlock getSkipblock(SkipblockId id) throws CothorityCommunicationExcep
...
+     * @throws CothorityCommunicationException if something goes wrong with communication
+     * @throws CothorityCryptoException if the verification goes wrong
+     */
+    public SkipBlock getLatestSkipblock() throws CothorityCommunicationException, CothorityCryptoException {
+        return this.getLatestSkipblock(this.scID);
+    }
 
-        ByteString msg = roster.sendMessage("Skipchain/GetUpdateChain",
-                request);
+    /**
```
jeffallen: I'm wondering why this code has to be written to work only with the level-0 links. It we can find a path back to the trustedLatest via higher level links, that's better right?

I think I understand that this is the simplest thing that works, but it feels like we are losing the benefits of skipchain right here....

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
```diff
@@ -74,39 +84,114 @@ public SkipBlock getSkipblock(SkipblockId id) throws CothorityCommunicationExcep
...
-            logger.info("Got the following latest skipblock: {}", sb);
 
-            return sb;
-        } catch (InvalidProtocolBufferException e) {
-            throw new CothorityCommunicationException(e);
+            SkipBlock last = update.get(update.size()-1);
+
+            // If they updated us to the end of the chain, return.
+            if (last.getForwardLinks().size() == 0) {
+                logger.info("Got the following latest skipblock: {}", Hex.printHexBinary(last.getId().getId()));
```
jeffallen: is this left over from debugging?

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
```diff
@@ -74,39 +84,114 @@ public SkipBlock getSkipblock(SkipblockId id) throws CothorityCommunicationExcep
...
+     */
+    public SkipBlock getLatestSkipblock(SkipblockId trustedLatest) throws CothorityCommunicationException, CothorityCryptoException {
+        List<SkipBlock> update = new ArrayList<>();
+        for (;;) {
+            // make the request
+            SkipchainProto.GetUpdateChainReply r2;
+            try {
+                SkipchainProto.GetUpdateChain request =
+                        SkipchainProto.GetUpdateChain.newBuilder()
+                                .setLatestID(ByteString.copyFrom(trustedLatest.getId()))
```
jeffallen: A get update chain request with no MaxHeight will traverse the highest links possible (see comments on skipchain/msgs.go, line 97). But the code below seems to assume it is working on level 0 links only?

external/java/src/test/java/ch/epfl/dedis/byzcoin/ByzCoinRPCTest.java
```diff
@@ -56,6 +56,32 @@ void ping() {
         assertTrue(bc.checkLiveness());
     }
 
+    @Test
+    void getBlocks() throws Exception {
```
jeffallen: I don't see how this code will exercise the part of the code about doing multiple get update chains, and linking from the last known update to the first block of the second get update chain call.

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/GFp2.java
```diff
@@ -179,8 +179,8 @@ GFp2 square(GFp2 a) {
     }
 
     GFp2 invert(GFp2 a) {
-        BigInteger t = a.y.multiply(a.y);
-        BigInteger t2 = a.x.multiply(a.x);
+        BigInteger t = a.y.multiply(a.y).mod(p);
```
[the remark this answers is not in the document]
kc1212: This doesn't make a difference because everything is in mod p. We use these values later so keeping them small is a minor performance improvement (but I didn't do a benchmark)

external/java/src/main/java/ch/epfl/dedis/skipchain/ByzcoinSig.java
```diff
@@ -25,11 +34,48 @@ public ByzcoinSig(ByteString bs) throws InvalidProtocolBufferException{
...
+
+        if (lenCom >= this.getSignature().length) {
+            // mask is missing
+            return false;
+        }
+
+        byte[] maskBits = Arrays.copyOfRange(this.getSignature(), lenCom, this.getSignature().length);
+        Mask mask;
+        try {
+            mask = new Mask(publics.toArray(new Point[0]), maskBits);
```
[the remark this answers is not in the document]
kc1212: I think that's a good idea.

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/GFp2.java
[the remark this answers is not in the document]
kc1212: Changing it back because there's no evidence that this helps.

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
kc1212: this part is a copy/paste from the go code and it's what I understood too. will add a comment to explain

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
kc1212: this is actually left from the old code - https://github.com/dedis/cothority/pull/1649/files#diff-e1b7d9d6ed776d2c287512dce30b3a63L105

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
kc1212: I'm a bit confused by this comment. I don't think this code only works for level-0 links. The request doesn't specify the maximum height. So what is returned to us are all the blocks that are linked by the highest forward link. Then we only check the hash on those highest links (`https://github.com/dedis/cothority/pull/1649/files#diff-e1b7d9d6ed776d2c287512dce30b3a63L105`)

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
kc1212: See comment here https://github.com/dedis/cothority/pull/1649#discussion_r247058325

external/java/src/test/java/ch/epfl/dedis/byzcoin/ByzCoinRPCTest.java
kc1212: you're right, this doesn't exercise the part the multiple get update chains. trying to write a test that will do it, probably something with roster change

kc1212: testing the getUpdateChain loop turned out to be quite tricky. I had to implement some new skipchain APIs to create skipblocks the same way as the test in go. the only way to make skipblocks before was using the byzcoin API

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
```diff
@@ -203,4 +212,19 @@ public SkipblockId getID() {
     public Roster getRoster() {
         return roster;
     }
+
+    // for testing, none of the blocks are signed
```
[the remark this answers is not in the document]
kc1212: moved to test code

PR #1650, opened 2019-01-09
KyberJS converted to TypeScript

Gilthoniel:
This changes the Kyber javascript library so that the source is using TypeScript and the module is generated with the types definition. Tests have also been rewritten in TypeScript.

Note the change of name of the package from _kyber-js_ to _kyber_.

Note that the transpiled files are included in github because it is what is supposed to be published in npm.

Note also the change of document generator because of typescript


external/js/kyber/src/curve/nist/curve.ts
```diff
@@ -0,0 +1,69 @@
...
+                continue;
+            }
+            options[k] = new BN(options[k], 16, "le");
+        }
+        this.curve = new elliptic.curve.short(options);
+        this.bitSize = bitSize;
+        this.redN = BN.red(options.n);
+    }
+    
+    coordLen() {
```
kc1212: I'm just randomly browsing the code. How come some functions don't have a type? Like this one or `string()` and `inspect()` in some of the other files.

external/js/kyber/src/curve/nist/curve.ts
```diff
@@ -0,0 +1,69 @@
...
+import NistPoint from "./point";
+import NistScalar from "./scalar"
+import elliptic = require("elliptic");
+import BN = require("bn.js");
+
+export type BNType = string | Buffer | BN
+
+export default class Weierstrass implements Group {
+    curve: any;
+    redN: any;
```
kc1212: how come these have the type `any`? is it because those are objects from an external library? maybe some of those types can be found here https://github.com/DefinitelyTyped/DefinitelyTyped ?

Gilthoniel: If you check the package.json file, you can note that I used this repo for the typing but for a reason I can't explain, they decided not to export the Reduction type... I'm investigating, the pull request was mostly to check the tests but I'm not done yet ;)

kc1212: ah yes, it's still WIP. but it already looks really good!

external/js/kyber/package.json
```diff
@@ -1,16 +1,17 @@
 {
-  "name": "@dedis/kyber-js",
+  "name": "@dedis/kyber",
   "version": "0.0.10",
```
jeffallen: since the name is changing, please reset the version to 0.0.1. We will set it to 3.0.0 soon enough anyway.

external/js/kyber/package.json
```diff
@@ -25,33 +26,54 @@
   "author": "DEDIS",
   "license": "AGPL-3.0",
   "homepage": "https://github.com/dedis/cothority",
+  "files": [
+    "dist/",
+    "doc/",
+    "index.html"
+  ],
   "dependencies": {
     "bn.js": "^4.11.8",
```
jeffallen: Should these be updated now?

external/js/kyber/webpack.config.js
```diff
@@ -39,18 +12,30 @@ const browserConfig = {
...
           loader: "babel-loader",
           options: {
-            presets: ["env"],
-	    plugins: [ require("babel-plugin-transform-object-rest-spread") ],
+            presets: ["env"]
           }
         }
+      },
+      {
+        test: /\.tsx?$/,
```
jeffallen: I wondered (and lost some time last time I did TS) about this .tsx extension. Can we remove it? Seems like some windows thing. So here, change to /\.ts$/ and below remove it.

external/js/kyber/dist/constants.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
```
jeffallen: I wonder if you can use .gitattributes to mark everything in dist/... as binary, so that we don't see diffs of it.

external/js/kyber/src/@types/elliptic/index.d.ts
```diff
@@ -0,0 +1,62 @@
+// types declaration for version 6.4.1 for Kyber usage
+//
+// The definition contains only a subset of the library
```
jeffallen: Not a criticism, just to understand: do you expect these local @types files to go away as you manage to get the upstreams to take your fixes to their .d files?

Gilthoniel: So that's the main reason why I don't like TypeScript, most of the library you need don't have a types definition. You can find decent definitions on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) but to take the example of bn.js, their version was totally wrong regarding how the library works and _elliptic_ is incomplete.
I decided to write the definitions for the project because fixing the ones on DefinitelyTyped would require some time. Up to you if you want me to do that to use them eventually.

external/js/kyber/package.json
Gilthoniel: No, _dist_ contain the actual library (transpiled js files and tyoes definition files), _doc_ the documentation and _index.html_ I guess it's good to have the example in npm.

jeffallen: I meant the dependencies, should they be updated?

external/js/kyber/src/@types/elliptic/index.d.ts
jeffallen: Yes, we always try to contribute to projects which we depend on and we notice a problem that we have to work around. It makes everything better, for the other users, and our own code has fewer weird things in it that we have to maintain.

external/js/kyber/package.json
Gilthoniel: Erf my bad. _bn.js_ is on the last version and you're right for _elliptic_, there is a minor update.

kc1212:
we had tests under `external/js/kyber/test`, i saw those files got removed in the PR, where are the ts equivalents?

PR #1653, opened 2019-01-11
Update the ByzCoin diagram to match reality

jeffallen:
Also: Remove old references to collection.


kc1212:
should "sharding" go under byzcoin?

Issue #1440, opened 2018-08-24
java: Omniledger.Proof verifications


ineiti:
`Proof.verify` lacks a proper verification to know if the proof is actually valid, given a `SkipChainID`.

ineiti:
Will change once @kc1212 finishes the new collection library.

kc1212:
Since blscosi is in and we're trying to push cothority to v3, it's a good time to add all the missing verification functions in Java. After some investigation, the missing functions are as follows, the order is the suggested implementation order.

1. Implement BLS G2 point and BLS G1 point, these are needed for computing BLS signatures.
2. Implement BlsSig class (similar to SchnorrSig)
3. Implement a mask class that produces the correct aggregate for BlsSig and SchnorrSig.
3. Implement a verify function in ForwardLink that uses BlsSig to verify blocks
4. Implement the Trie class (also EmptyNode, LeafNode and InteriorNode), but only enough that we can do existence/absence proofs.
5. Add the existance/absence proofs to the Proof class

PR #1654, opened 2019-01-14
Java trie verification

kc1212:
Port the proof verification code from go to java. There are no tests that are specific for proofs because we don't have the code to construct those proofs. The proof verification is performed implicitly in EventLog.get and ByzCoinRPC.fromByzCoin.

Fixes https://github.com/dedis/cothority/issues/1440

### DIFF
--- a/external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
+++ b/external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
@@ -6,7 +6,6 @@ import ch.epfl.dedis.lib.crypto.Bn256G2Point;
 import ch.epfl.dedis.lib.crypto.Point;
 import ch.epfl.dedis.lib.darc.DarcId;
 import ch.epfl.dedis.lib.exception.CothorityCryptoException;
-import ch.epfl.dedis.lib.exception.CothorityException;
 import ch.epfl.dedis.lib.exception.CothorityNotFoundException;
 import ch.epfl.dedis.lib.network.ServerIdentity;
 import ch.epfl.dedis.lib.proto.NetworkProto;
@@ -31,24 +30,29 @@ import java.util.stream.Collectors;
  * root node.
  */
 public class Proof {
-    private TrieProto.Proof proof;
-    private List<SkipchainProto.ForwardLink> links;
-    private SkipBlock latest;
-
-    private StateChangeBody finalStateChangeBody;
+    private final TrieProto.Proof proof;
+    private final List<SkipchainProto.ForwardLink> links;
+    private final SkipBlock latest;
+    private final StateChangeBody finalStateChangeBody;
 
     /**
-     * Creates a new proof given a protobuf-representation.
+     * Creates a new proof given a protobuf-representation and a trusted skipchain ID.
      *
-     * @param p the protobuf-representation of the proof
+     * @param p    is the protobuf representation
+     * @param scID is the skipchain ID
+     * @throws InvalidProtocolBufferException if the protobuf representation is invalid
+     * @throws CothorityCryptoException       if the verification of the forward links are wrong
      */
-    public Proof(ByzCoinProto.Proof p) throws InvalidProtocolBufferException {
+    public Proof(ByzCoinProto.Proof p, SkipblockId scID) throws InvalidProtocolBufferException, CothorityCryptoException {
         proof = p.getInclusionproof();
         latest = new SkipBlock(p.getLatest());
         links = p.getLinksList();
         if (!proof.getLeaf().getKey().isEmpty()) {
             finalStateChangeBody = new StateChangeBody(ByzCoinProto.StateChangeBody.parseFrom(proof.getLeaf().getValue()));
+        } else {
+            finalStateChangeBody = null;
         }
+        this.verify(scID);
     }
 
     /**
@@ -75,7 +79,7 @@ public class Proof {
     }
 
     /**
-     * Getter for the skipblock assocaited with this proof.
+     * Getter for the skipblock associated with this proof.
      *
      * @return SkipBlock
      */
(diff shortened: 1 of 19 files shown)


external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
```diff
@@ -61,38 +75,71 @@ public Instance getInstance() throws CothorityNotFoundException{
...
+            if (!l.verify(publics)) {
+                throw new CothorityCryptoException("stored skipblock is not properly evolved from genesis block");
+            }
+            if (!Arrays.equals(l.getFrom().getId(), sbID.getId())) {
+                throw new CothorityCryptoException("stored skipblock is not properly evolved from genesis block");
+            }
+            sbID = l.getTo();
+            try {
+                if (l.getNewRoster() != null) {
+                    publics = getPoints(this.links.get(i).getNewRoster().getListList());
```
jeffallen: What on earth is getListList?

external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
```diff
@@ -112,64 +159,183 @@ public StateChangeBody getValues() {
...
+     * @param id       the Byzcoin id to verify the proof against
+     * @return true if the proof is correct with regard to that Byzcoin id and the contract is of the expected type.
+     * @throws CothorityException if something goes wrong
      */
-    public boolean isByzCoinProof(){
-        if (!matches()) {
-            return false;
-        }
-        return true;
+    public boolean isContract(String expected, SkipblockId id) throws CothorityException {
```
Gilthoniel: I was confused with the name at first. Should this be renamed to something like _isContractOk_ or better ?

Gilthoniel: Also the comment says the exception is thrown if something goes wrong but verify will throw if the byzcoin id is bad, so nothing went wrong, it's just the id that is wrong. Is it the expected behaviour to have either a boolean or an error ?

external/java/src/main/java/ch/epfl/dedis/calypso/CalypsoRPC.java
```diff
@@ -42,8 +42,7 @@ public CalypsoRPC(ByzCoinRPC byzcoin, DarcId darcId, Roster ltsRoster, List<Sign
         LTSInstance inst = new LTSInstance(this, darcId, ltsRoster, signers, signerCtrs);
         Proof proof = inst.getProof();
         // Start the LTS/DKG protocol.
-        CreateLTSReply lts = createLTS(proof);
-        this.lts = lts;
+        this.lts = createLTS(proof);
```
jeffallen: I'm not sure, but I don't see where proof.verify() is ever called here.

external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
```diff
@@ -112,64 +159,183 @@ public StateChangeBody getValues() {
...
+                expectedHash = this.proof.getInteriors(i).getLeft().toByteArray();
+            } else {
+                expectedHash = this.proof.getInteriors(i).getRight().toByteArray();
+            }
         }
-        if (!isContract(expected)){
+        // we use i below instead of i+1 (like the go code) because the final i is one more than the i used in the final iteration
+        if (Arrays.equals(expectedHash, hashLeaf(this.proof.getLeaf(), this.proof.getNonce().toByteArray()))) {
+            if (!Arrays.equals(Arrays.copyOf(bits, i), this.proof.getLeaf().getPrefixList().toArray(new Boolean[0]))) {
+                throw new CothorityCryptoException("invalid prefix in leaf node");
```
jeffallen: Is it important to test this? It seems to me like we are only testing the happy path through here. What if a nasty conode is sending out Proofs to clients that are invalid? Are we sure they will catch it?

I wonder if it would be enough to take a correct proof and then fiddle with some of it's internals, then call verify() again on it, in order to check for the expected exceptions?

external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
kc1212: it's from the generated protobuf code. there's an attribute called `List` and if you want a java `List` of then you gotta call `getListList`

external/java/src/main/java/ch/epfl/dedis/calypso/CalypsoRPC.java
kc1212: You're right, it's not just at this place, the proof is not checked in some other places too. I'm looking for places that calls `getProof` and adding the verification for those that need it.

external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
kc1212: yes, the name could be improved, maybe to something like `contractIsType` because it's checking whether the contract matches the expected type. regarding the second comment, `verify` should never fail when the system is working correctly (the forward links should always checkout), but it might return false because the client used a wrong contract name, so it's two different types of errors. I will update the comment.

external/java/src/main/java/ch/epfl/dedis/calypso/CalypsoRPC.java
kc1212: hopefully this commit https://github.com/dedis/cothority/commit/fed20e7684a893f30702ee2de2d0ecbebb6aa94f addresses this issue in general. in the new code, the proof is always verified during construction because there is no reason (under normal operation) that the forwardlinks are wrong or the Merkle-root is wrong. the `exists` and `match` functions are staying the same. they are only called when it's necessary.

external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
kc1212: moving this back to WIP while I think about how we can test something like this

kc1212: Added better tests - https://github.com/dedis/cothority/pull/1654/commits/6d8d31d9ee8893b59163edfab865ade938da290b

PR #1655, opened 2019-01-15
Use AES-GCM to encrypt data on Calypso

### LINKED ISSUE #1488: Move to AES-GCM
Our example implementation of doing the symmetric encryption part for Calypso uses CBC-mode, which is bad.

Needed: use AES-GCM in `Encryption.java`.

kc1212:
Previously the encryption was done using AES-CBC, which is not
authenticated so that's bad. This PR changes is to use AES-GCM.

Fixes #1488

### DIFF
--- a/external/java/src/main/java/ch/epfl/dedis/calypso/Document.java
+++ b/external/java/src/main/java/ch/epfl/dedis/calypso/Document.java
@@ -34,9 +34,8 @@ public class Document {
      *                    This will be encrypted before the transmission to the skipchain.
      * @param extraData   any public data that will be stored unencrypted on the skipchain.
      * @param publisherId the publisher darc with the rules to create a WriteInstance and a ReadInstance.
-     * @throws CothorityCryptoException if there's a problem with the cryptography
      */
-    public Document(byte[] data, byte[] keyMaterial, byte[] extraData, DarcId publisherId) throws CothorityCryptoException {
+    public Document(byte[] data, byte[] keyMaterial, byte[] extraData, DarcId publisherId) {
         this.data = data;
         this.keyMaterial = keyMaterial;
         this.publisherId = publisherId;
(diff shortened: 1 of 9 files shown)


external/java/src/main/java/ch/epfl/dedis/calypso/Document.java
```diff
@@ -50,14 +49,12 @@ public Document(byte[] data, byte[] keyMaterial, byte[] extraData, DarcId publis
      * @param data        any data that will be stored encrypted on the skipchain.
      *                    There is a 10MB-limit on how much data can be stored. If you
```
jeffallen: The limit is based on the block size, which defaults to 4 megs (see byzcoin/service.go, defaultMaxBlockSize). If a given ledger has a higher block size, then more data could be stored in a Document.

It would be helpful to actually test this, because there is likely some overhead meaning that the real max size is 4mb - X, where X = ??.

external/java/src/main/java/ch/epfl/dedis/calypso/Encryption.java
```diff
@@ -1,78 +1,90 @@
...
+        final byte[] iv;
+        final GCMParameterSpec gcmSpec;
+        final SecretKeySpec keySpec;
 
-        public keyIv(byte[] keyMaterial) throws CothorityCryptoException{
-            int symmetricLength = keyMaterial.length - ivLength;
-            if (symmetricLength <= 0){
-                throw new CothorityCryptoException("too short symmetricKey material");
+        /**
+         * Construct KeyIV from some keyMaterial.
```
jeffallen: Need to remind them that "For a given key, the IV must never repeat." Thus, encourage them to use the 0-argument constructor to get a new ephemeral key with random IV.

Or maybe: do not allow them to pass in IV at all, have this constructor take the key material and choose the IV.

kc1212: I updated the comment. I'm not changing the way how IV is passed in because that constructor also needs to be used when decrypting the secret. We have a constructor for `Document` which uses the default constructor to create the key material.

Issue #1489, opened 2018-09-27
Don't re-use C


ineiti:
In `WriteRequest.java` the point `C` is re-used as a _mask_ to encrypt parts of the symmetric key. In order to not re-use `C`, we need to:

- throw an exception if we have a bigger symmetric key than 29 bytes (the maximum we can embed)
- limit the symmetric key material to 16 bytes by putting the IV in the extra data from the write request

kc1212:
More information: the file name is change to `WriteData.java`, the function we're interested in is `encryptKey`, and the problem is that `C` is reused inside the for loop. This is a backward incompatible change but not an API change.

PR #1659, opened 2019-01-16
Do not reuse C in Calypso

kc1212:
Fixes https://github.com/dedis/cothority/issues/1489


kc1212:
> I'm having trouble understanding the crypto side of this (what was being reused?) but I'm confident this is implementing the original change requested by @ineiti . Just these stupid little documentation thing to fix first...

The `C` is being reused in the for loop, in the old code, see `Point Ckey = C.add(keyEd25519Point);`.

PR #1664, opened 2019-01-17
Simulation updates

jeffallen:
Make it easier to gather tx/sec. Record measurements so
that we can monitor it as we work on improving performance.

Also: tiny doc update to scmgr


jeffallen:
Forgotten.

PR #1669, opened 2019-01-22
Send messages to a random node instead of node 0

### LINKED ISSUE #1630: java client Roster.sendMessage always chooses node 0
There's a TODO reminding us that we should choose a random server from the Roster.

After fixing this, we need to check that Java unit tests are still stable. If not, we need to solve the underlying problems.

kc1212:
Fixes #1630


kc1212:
Closing because we'll do this after v3.

PR #1670, opened 2019-01-22
Add BLS support to KyberJS

Gilthoniel:
This adds BLS signatures support by porting the Go code to typescript and by implementing the existing interfaces.

It's worth noting that the point factory for instance must be implemented but this will be done in the scope of moving the cothority typescript version from personhood.online.


Gilthoniel:
@kc1212 also note that in JS is _not_ a compiled language. Typescript is a layer above javascript but you still import some js code into your Typescript code. The import actually imports the bn.js library. The definition file is there only to help typescript checking everything is correct.

external/js/kyber/spec/pairing/bn.spec.ts
```diff
@@ -0,0 +1,253 @@
+import { randomBytes } from 'crypto';
+import BN from 'bn.js';
+import { G1, G2, GT } from '../../src/pairing/bn';
+import { p } from '../../src/pairing/constants';
+
+/**
+ * Unsecure pseudo random generator
+ * @returns a big number between 0 and p
+ */
+function getRandomInt(): BN {
```
jeffallen: I do not see why we should ever be shipping what we call an unsecure generator, even in test code. In test code, if we generate keys from some static seed, that's insecure, but there's a reason for it (test repeatability).

I propose that in the one test where there are used that either you use hardcoded scalars (i.e. 1, 2, 3) or you get real cryptographically random input.

It is important, especially because students use our code, that we do not teach bad habits.

PR #1674, opened 2019-01-24
Omniledger sharding

qmng:


jeffallen:
FYI, we have no plans of including this code before the end of Feb, when we release Cothority v3. We'll take a look at it later, but right now we are trying to get a stable release out.

PR #1677, opened 2019-01-29
Cothority TypeScript

Gilthoniel:
This pull request migrates the current cothority javascript library from the typescript version done for personhood.online. It cleans up the code, makes it more comparable with the Java library and adds support for testing, documentation and coverage.


external/js/cothority/spec/support/public.toml
```diff
@@ -0,0 +1,134 @@
+
```
jeffallen: Are these exactly the same as the file in external/docker/public.toml? I can see why it need to be repeated here, but it's a bummer. Wonder if we could copy it over via some magic in package.json's build/test rules? For later...

jeffallen:
@kc1212 could you please review the part in proof.ts where proofs are checked? That seems like a critical part of the client, and I don't feel qualified to notice traps there.

external/js/cothority/spec/support/public.toml
Gilthoniel: oh yeah thanks for the reminder, I need to do something about that

kc1212:
`proof.ts` looks good to me, hopefully you didn't waste too much time on the use of index @Gilthoniel (it's `i+1` in Go and `i` in Java because Go uses something like "foreach" and java just uses the normal for-loop, so the final `i` is different. It's correctly implemented in the TS code coz it also uses the normal for-loop)

Gilthoniel:
A few heads up for the R4M:
- Generated files in Kyber (dist, doc) have been removed and those will be generated on demand like before publishing. That allows much cleaner and easier PRs.
- Proof checking and block validation has been added
- Travis test has been enabled but PoP tests have been disabled until the personhood.online branch has been merged
- TSLint has been added and is running before the tests

PR #1678, opened 2019-01-29
Use the new module import path

jeffallen:


byzcoin/heartbeat_test.go
```diff
@@ -28,7 +28,7 @@ func TestHeartbeat_Start(t *testing.T) {
 }
 
 func TestHeartbeat_Timeout(t *testing.T) {
-	t.Skip("https://github.com/dedis/cothority/issues/1473")
+	t.Skip("https://go.dedis.ch/cothority/v3/issues/1473")
```
kc1212: this should stay on github

doc.go
```diff
@@ -11,7 +11,7 @@
 // protocols who make conodes work together.
 //
 // For further documentation, please go to
-// https://github.com/dedis/cothority/blob/master/README.md.
+// https://go.dedis.ch/cothority/v3/blob/master/README.md.
```
kc1212: this should stay on github too

byzcoin/heartbeat_test.go
Gilthoniel: I guess we should have a redirection anyway

jeffallen: The go,dedis.ch domain is just for the meta data that the go build tool uses. More redirects would just be more confusing.

jeffallen:
A fix for the comments from @kc1212 are in #1679.

PR #1680, opened 2019-01-30
Restrict darcs

### LINKED ISSUE #1625: Too much freedom in darcs
At the moment, anybody with the spawn permission can spawn a darc with arbitrary rules. Similarly, anybody that's allowed to evolve the darc can evolve it without restrictions.

For example, in a hierarchical scenario, where we have admins, managers and users. We need a way to restrict the managers such that they cannot spawn other managers (this right is reserved for admins). Similarly, users should be allowed to evolve some of its rules but they should not have the possibility of adding a rule that lets them spawn manager darcs.

A high level idea to implement this feature is to create attach a white list (or black list) that specified what kind of evolution and spawn every instance is allowed to do. Since darcs are not useful when they're not attached to instances. We still allow anybody to spawn darcs, but when they try to spawn an instance, the contract will check that the darc rules satisfy the white/black list. Putting restrictions in evolution is trickier, ByzCoin must have a way to find which instance/contract that the darc is attached to, and then make sure the new rules satisfy the restrictions. When restrictions are used, it may be trick to attached multiple instances (from different contracts) to the same darc, but this is a inconvenience that we have to live with.

kc1212:
The old DARC contract has too much freedom. We need to limit this
freedom in two ways to improve security. First, the owner of any DARC
should not be allowed to change its rules arbitrarily, he/she is only
allowed to change the existing rules. Nevertheless, we still allow a set
of authorised parties to add rules. Second, we want to let certain DARCs
create new DARCs, but those new DARCs cannot have arbitrary rules.

These features are implemented in a new contract called "secure_darc".
For the first restriction, we changed the "evolve" command to reject
requests that tries to add new rules. We also added a new command
"evolve_unrestricted" to let authorised users add rules. For the second
restriction, we give an example of how to implement it statically in the
contract, which is to reject spawn requests that contains "spawn:darc".
Additionally, we explain in the example how to impelement it the same
feature dynamically.

Since we have more than one contract that can be decoded into a DARC, we
add a new parameter into ChainConfig to keep track of these contracts.

Fixes #1625


kc1212:
Looking for early feedback regarding this approach. I describe the two issues that we're trying to solve and how we solve it. Basically it boils down to "if you want to do these extra stuff, make your own darc contract and register it on the blockchain". This part might also be interesting - https://github.com/dedis/cothority/pull/1680/files#diff-2aebfa975a350d95cf74ad351d4a008eR12

Gilthoniel:
I'm not sure why you need to statically refuse spawn:darc requests. Couldn't this be also manage as a permission ?

kc1212:
> I'm not sure why you need to statically refuse spawn:darc requests. Couldn't this be also manage as a permission ?

Suppose you give somebody the rule `spawn:secure_darc`, then they're allowed to make their own DARC with their own rules. Since we're in the secure-darc domain, we should prevent privilege escalation, which means owners should not be allowed to spawn a regular darc.

Nevertheless, I added that as an example. In practice the access control rules could be quite complicated. The users are encouraged to make their own contract to fit their needs.

byzcoin/Contracts.md
```diff
@@ -178,37 +178,60 @@ are available to ByzCoin.
...
+### Secure Darc Customization
+
+Sometimes the user needs fine-grained control of all aspects of their access
+control policy. For these purposes we encourage the user to write their own
+Darc contracts.
+
+Suppose we are in a hierarchical situation where the boss is allowed to do
+anything, the managers are allowed to do a certain set of tasks including
+spawning user Darcs and the users are allowed to do another set of tasks but
+they are not allowed to do anything. The SecureDarc contract already achieves
```
jeffallen:
```suggestion
they are not allowed to spawn any new Darcs. The SecureDarc contract already achieves
```

byzcoin/secure_darc_test.go
```diff
@@ -0,0 +1,195 @@
...
+				},
+				SignerCounter: []uint64{1},
+			}},
+		}
+		require.Nil(t, ctx2.SignWith(restrictedSigner))
+		_, err = cl.AddTransactionAndWait(ctx2, 10)
+		require.Error(t, err)
+	}
+
+	log.Info("evolve to modify existing rules - pass")
```
jeffallen: Is there a test for "I am restricted, but I want to remove the restriction on myself -> fail"?
Well, in writing it down, it seems covered by other tests and probably not required, but please see if my question has shaken lose any ideas for you...

darc/darc.go
```diff
@@ -293,6 +293,20 @@ func (r Rules) Copy() Rules {
 	return rCopy
 }
 
+// IsSubset checks whether the receiver is a subset of the given actions set.
```
jeffallen: Seems like this code should be unit tested. Is there a reason you didn't/couldn't?

external/java/src/main/java/ch/epfl/dedis/byzcoin/contracts/SecureDarcInstance.java
```diff
@@ -102,10 +101,30 @@ public void update() throws CothorityException {
      * @return Instruction to be sent to byzcoin
      */
     public Instruction evolveDarcInstruction(Darc newDarc, Long signerCtr) {
+        return evolveDarcInstruction(newDarc, signerCtr, false);
+    }
+
+    /**
+     * Creates an instruction to (unrestricted) evolve the darc in byzcoin. The signer must have its identity in the current
+     * darc as "Invoke_Evolve" rule.
```
jeffallen: Is this comment right?

kc1212: no, it was from a bad copy paste

byzcoin/secure_darc_test.go
kc1212: it's a good idea, I added a test for it and the test caught a bug

darc/darc.go
kc1212: added a unit test for this one too

PR #1682, opened 2019-01-31
Fix kyberjs test for g2 points

### LINKED ISSUE #1681: BN256 Point Tests should add and subtract g2 points failure
saw this test failure:
```
Failures:
1) BN256 Point Tests should add and subtract g2 points
  Message:
    Expected property to hold. Counterexample found: []; [1]
  Stack:
    Error: Expected property to hold. Counterexample found: []; [1]
        at <Jasmine>
        at UserContext.it (/home/travis/build/dedis/cothority/external/js/kyber/spec/pairing/point.spec.ts:86:22)
        at <Jasmine>
Pending:
```
https://travis-ci.org/dedis/cothority/jobs/486523352

Gilthoniel:
This fixes the test by changing the negative function of
the point to not set t. The _t_ is actually never used so I'm not really sure why it is there at the first place but the issue was caused because it was set to 0 for the negative operation.

Fixes #1681

### DIFF
--- a/external/js/kyber/spec/pairing/point.spec.ts
+++ b/external/js/kyber/spec/pairing/point.spec.ts
@@ -86,6 +86,18 @@ describe('BN256 Point Tests', () => {
         expect(prop).toHold();
     });
 
+    it('should add and subtract 0 and 1', () => {
+        const p1 = new BN256G2Point([]);
+        const p2 = new BN256G2Point([1]);
+
+        expect(p1.getElement().isInfinity());
+
+        const aa = new BN256G2Point().add(p1, p2);
+        const bb = new BN256G2Point().sub(p1, p2.clone().neg(p2));
+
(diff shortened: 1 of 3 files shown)


kc1212:
I'm generally fine with this approach because I also don't see how `t` is used.

There is another approach that might fix the problem: I don't know how the equal function is implements in ts. In java, I override `equals` to do a `MakeAffine` so that the comparison happens on the same representation.

I have a preference for the second approach because it follows the other implementations we use.

Gilthoniel:
Same here, the equals function is using the makeAffine function but in that specific case, z is equal to one and the makeAffine thinks it's already affine. But this t element is the only different one, x and y are indeed affine. I first fixed that by setting t to one when z is one but I eventually changed to that because that was more a hack than an actual fix. _negative_ was the only algorithmic operation to set t to something, that's why I took this approach.

PR #1684, opened 2019-01-31
Make unit and integration tests run outside of the GOPATH.

jeffallen:
Together with matching changes on the Jenkins side, this will fix the Jenkins build.

### DIFF
--- a/eventlog/el/test.sh
+++ b/eventlog/el/test.sh
@@ -7,7 +7,7 @@ DBG_SRV=0
 NBR=3
 NBR_SERVERS_GROUP=$NBR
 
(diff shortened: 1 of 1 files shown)


eventlog/el/test.sh
```diff
@@ -7,15 +7,15 @@ DBG_SRV=0
 NBR=3
 NBR_SERVERS_GROUP=$NBR
 
-. "$(go env GOPATH)/src/github.com/dedis/cothority/libtest.sh"
+. ../..//libtest.sh
```
Gilthoniel: one / more

PR #1694, opened 2019-02-04
Remove parent/child relationship in skipchains

### LINKED ISSUE #1683: Coin contract panic for bad arguments
https://github.com/dedis/cothority/blob/b67ce34a83c82ee942c7f04231620ed3131156b6/byzcoin/contracts/coins.go#L98

When testing the typescript cothority library, I went through a panic from the conodes because the buffer supposedly containing the value of coins was not encoded correctly. This should return an error instead of a panic.

jeffallen:
Also: Travis cleanup.
Also: Fix #1683.


kc1212:
I think the java tests are failing because `SkipBlockFix` is not fixed in the (manually written) proto file.

byzcoin/contracts/coins.go
```diff
@@ -95,6 +95,10 @@ func (c *contractCoin) Invoke(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instru
 			err = errors.New("argument \"coins\" is missing")
 			return
 		}
+		if len(coinsBuf) != 8 {
```
ineiti: Nitpick: is this the correct place to improve the coins behaviour?

jeffallen: Where else would it be? I agree some sort of nice "coin" abstract type would be fun to write, but it seems too big of an investment for this little problem.

jeffallen:
Travis is confused, and refuses to build the correct commit. Closing and trying again.

PR #1697, opened 2019-02-05
debugging improvements

ineiti:
When working on running conodes, I like having _good_ debugging outputs. So this one improves some of those messagers.


byzcoin/heartbeat.go
```diff
@@ -72,14 +72,18 @@ func (r *heartbeats) exists(key string) bool {
...
-func (r *heartbeats) updateTimeout(key string, timeout time.Duration) {
+func (r *heartbeats) updateTimeout(key string, timeout time.Duration) (changed bool) {
 	r.Lock()
 	defer r.Unlock()
 	h, ok := r.heartbeatMap[key]
 	if !ok {
 		return
 	}
-	h.updateTimeoutChan <- timeout
+	if h.timeout != timeout {
```
jeffallen: This is a data race, I just removed it in c84be53ad8bd7.

byzcoin/service.go
```diff
@@ -1127,8 +1127,9 @@ func (s *Service) updateTrieCallback(sbID skipchain.SkipBlockID) error {
 	if nodeInNew {
 		// Update or start heartbeats
 		if s.heartbeats.exists(string(sb.SkipChainID())) {
-			log.Lvlf2("%s updating heartbeat monitor for %x with window %v", s.ServerIdentity(), sb.SkipChainID(), interval*rotationWindow)
-			s.heartbeats.updateTimeout(string(sb.SkipChainID()), interval*rotationWindow)
+			if s.heartbeats.updateTimeout(string(sb.SkipChainID()), interval*rotationWindow) {
```
jeffallen: this can't be conditional, or else it creates a data race (see above). Instead, make the message something about "sending a heartbeat update to for %v", I guess.

byzcoin/transaction.go
```diff
@@ -345,11 +345,12 @@ func (instrs Instructions) Hash() []byte {
 type TxResults []TxResult
 
 // NewTxResults takes a list of client transactions and wraps them up
-// in a TxResults with Accepted set to false for each.
+// in a TxResults with Accepted set to true for each.
 func NewTxResults(ct ...ClientTransaction) TxResults {
 	out := make([]TxResult, len(ct))
 	for i := range ct {
 		out[i].ClientTransaction = ct[i]
+		out[i].Accepted = true
```
jeffallen: Why? The default of false was more conservative.

kc1212: Why does this need to be true? It seems to be more natural to default these flags to false because the moment that `NewTxResults` is used is before we verify the transactions `createStateChanges`.

byzcoin/service.go
```diff
@@ -1612,7 +1615,7 @@ func (s *Service) createStateChanges(sst *stagingStateTrie, scID skipchain.SkipB
 	var err error
 	merkleRoot, txOut, states, err = s.stateChangeCache.get(scID, txIn.Hash())
 	if err == nil {
-		log.Lvlf3("%s: loaded state changes %x from cache", s.ServerIdentity(), scID)
+		log.LLvlf3("%s: loaded state changes %x from cache", s.ServerIdentity(), scID)
```
kc1212: Lvlf3

byzcoin/service.go
ineiti: OK, but then I have to put the debugging-level to 3, else it spams too much...

byzcoin/transaction.go
ineiti: I'm not sure anymore why I put this to `true` :( I looked at all occurrences of `NewTxResults`, and it's always when calling `createStateChanges`, which sets the `Accepted` field anyway, so it doesn't change whether it's `true` or `false` here...

ineiti: as written above: can't remember anymore, and it doesn't change anything.

PR #1698, opened 2019-02-05
simplifying trySyncAll by using catchup methods

ineiti:


byzcoin/service.go
```diff
@@ -862,6 +862,47 @@ func (s *Service) downloadDB(sb *skipchain.SkipBlock) error {
 	return errors.New("none of the non-leader and non-subleader nodes were able to give us a copy of the state")
 }
 
+// catchupAll calls catchup for every byzcoin instance stored in this system.
```
jeffallen: Were the existing tests enough to check this code? It seems like a substantial, and tricky change, to not have any (new/modified) testing.

ineiti: There is one existing test, 

https://github.com/dedis/cothority/pull/1698/files/10c089ae55fa33c5896f761d13f3488950f01a72#diff-a47d422bdf18ea247341c7f28cea3d87R2158

But I can think of some edge-cases that would really need to be tested to make sure this is correctly handled. But definitely less edge-cases in this PR than on master ;)

kc1212: I think it's also tested in the view-change tests, because we make the node go down and then bring it back up to see if it can download the missed transactions.

PR #1699, opened 2019-02-05
Various v3 tasks

jeffallen:
A combined PR with many small jobs, and one large one. See commits for more details.


kc1212:
Travis is failing because of this line
https://github.com/dedis/cothority/blob/5947a772281e53b633c7e0850f9a8516bbd1e856/byzcoin/secure_darc_test.go#L95

It should be `FillSignersAndSignWith` instead of `SignWith` if the signers are not already in the instruction.

Travis is catching an error for once...

jeffallen:
So strange, I feel like I fixed that just this morning.

PR #1707, opened 2019-02-08
Domain separation for byzcoin Instruction hashing

### LINKED ISSUE #1360: need to use domain separation in omniledger hashing?

kc1212:
Fixes #1360

### DIFF
--- a/byzcoin/transaction.go
+++ b/byzcoin/transaction.go
@@ -121,16 +121,28 @@ func (instr Instruction) Hash() []byte {
 		h.Write([]byte(instr.Delete.ContractID))
 	}
 	for _, a := range args {
-		h.Write([]byte(a.Name))
+		nameBuf := []byte(a.Name)
+		nameLenBuf := make([]byte, 8)
+		binary.LittleEndian.PutUint64(nameLenBuf, uint64(len(nameBuf)))
+		h.Write(nameLenBuf)
+		h.Write(nameBuf)
+
+		valueLenBuf := make([]byte, 8)
+		binary.LittleEndian.PutUint64(valueLenBuf, uint64(len(a.Value)))
+		h.Write(valueLenBuf)
 		h.Write(a.Value)
 	}
-	for _, ver := range instr.SignerCounter {
-		verBuf := make([]byte, 8)
-		binary.LittleEndian.PutUint64(verBuf, ver)
-		h.Write(verBuf)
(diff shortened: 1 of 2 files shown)


ineiti:
I don't see the attack vector here - seems to me `premature attack protection`.

The classical attack you want to avoid is this:

- Alice has the private key to evolve darcs
- Alice also has an automatic publicly available signing service, _using the same private key as to evolve the darcs_ (which is bad, but for the example is needed)
- Eve manages to get Alices automatic signing service to sign a new darc with Eves public keys in it

Here we would have to add `darc` to the hash that would be signed for the darc evolution, and `public-signature` to the hash that would be signed for the public signing service, thus creating a domain-separation between the two signatures.

But I don't see any possible attack in case the darc-attributes are not domain-separated. Only Alice has the private key, so she can do whatever she wants anyway. Whether you protect the attributes or not. And we suppose no attacker can get the same hash with a different darc anyway.

So, unless you see an attack here, I'd propose to keep the simpler, non-`18 exa`-length protected version...

kc1212:
I believe an attack is possible when there are multiple signers.

Suppose Eve creates an important instruction that is valid only when 2 or more (e.g., Alice and Eve) signers sign it. With the old way of hashing, Eve can create two transactions that result in the same hash (because there is no domain separation in the name/value pair). Then Eve sends one form of the instruction to Alice to get her signature, but sends the other version to the blockchain.

PR #1723, opened 2019-02-18
Publish cothority and kyber to npm

Gilthoniel:
This PR makes the dist folder the one that will be published so that the imports of the library are more clean. It also adds sub-module exports to make life easier when importing basic classes.

Interface only files have been either removed or excluded from the coverage.


external/js/cothority/src/darc/identity-wrapper.ts
```diff
@@ -24,3 +24,35 @@ export default class IdentityWrapper extends Message<IdentityWrapper> {
     return Buffer.from([]);
   }
 }
+
+/**
+ * Identitiy is an abstract class for all the Darcs's identities
```
jeffallen:
```suggestion
 * Identity is an abstract class for all the Darcs's identities
```

Search... I think this typo is other places maybe.

external/js/kyber/publish.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
jeffallen: Need better comments in here to explain what's so weird about this package that requires it has a special way of publishing it.

.travis.yml
```diff
@@ -72,7 +74,7 @@ matrix:
         - cd $TRAVIS_BUILD_DIR/external/js/kyber
         - npm install
         - npm run build
-        - npm link
+        - npm run link
```
jeffallen: If someone types "npm link" out of habit, will private give them a useful error?

I'm really not happy about this non-standard package format, but I don't know what you could have done to avoid it...

Gilthoniel: Well the only other way is to import everything from the root (and this export everything up to the root). For historical reasons, the relative path when importing is the node_modules folder and you can't change that.

PR #1727, opened 2019-02-19
Various tasks to prepare for v3

jeffallen:


external/js/kyber/package.json
```diff
@@ -38,6 +38,7 @@
     "@types/elliptic": "^6.4.2",
     "@types/jasmine": "^3.3.5",
     "@types/node": "^9.6.6",
+    "ajv": "^6.9.1",
```
Gilthoniel: Is this because of a missing dependency or it shouldn't be there ? I don't remember to have used this

jeffallen: NPM told me it was missing, so I added it.

PR #1733, opened 2019-02-22
Adding latest personhood service and contracts

ineiti:
Update the personhood service and the contracts to the version used in the latest personhood
mobile app available at github.com/dedis/personhood. Most of the code is only tested in
the javascript-modules in external/js/cothority.


personhood/db.go
```diff
@@ -48,15 +48,22 @@ func (s *Service) tryLoad() error {
 }
 
 type storage1 struct {
```
Gilthoniel: because _storage_ is not available ?

personhood/phapp/test.sh
```diff
@@ -0,0 +1,69 @@
...
+    build $APPDIR/../../byzcoin/bcadmin
+    run testSpawner
+    run testWipe
+    run testRegister
+    stopTest
+}
+
+testSpawner(){
+  rm -f config/*
+  runCoBG 1 2 3
```
Gilthoniel: might be worth to have some sleep here and the other same lines to let the conodes start. It seems slow sometimes on Jenkins which makes the tests fail.

personhood/service.go
```diff
@@ -265,26 +530,50 @@ func (s *Service) TopupMessage(tm *TopupMessage) (*StringReply, error) {
 	return &StringReply{}, nil
 }
 
+// TestStore allows easier testing of the mobile apps by giving an endpoint
+// where current testing data can be stored.
```
Gilthoniel: It must have a better way to do that. It sounds really wrong to have to maintain a test endpoint for tests not even in the repo.

personhood/api.go
```diff
@@ -20,12 +23,48 @@ func NewClient() *Client {
...
 
-// LinkPoP sends a party description to the message server for further
-// reference in messages.
-func (c *Client) LinkPoP(si *network.ServerIdentity, p Party) error {
-	err := c.SendProtobuf(si, &LinkPoP{p}, nil)
-	if err != nil {
-		return err
+// TestData stores a byzcoin-ID and a spawner-IID in the system for the other clients to pick up. This is a
+// test-endpoint, not used in the final personhood.online app.
+func (c *Client) TestData(r onet.Roster, bcID skipchain.SkipBlockID, spawnIID byzcoin.InstanceID) (errs []error) {
```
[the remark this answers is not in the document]
ineiti: I use it in the personhood.online app to do UI testing. When you start the app in testing-mode, you can create your own byzcoin blockchain, including an admin account, and do some mining of coins. This makes it much easier to quickly launch three emulators, create a chain, and onboard the other two emulators.

If you have a better idea how this could be done, I'm all ears.

personhood/db.go
ineiti: For a new version of the db, it will be called `storage2` and will evolve from `storage1`.

personhood/phapp/test.sh
ineiti: That's an error in `libtest.sh` then, as it's supposed to wait for the conodes to have started up:

https://github.com/dedis/cothority/blob/master/libtest.sh#L370

BTW: `sleep 10` there seems really long. It should correctly `ping` the servers or something like that.

personhood/service.go
ineiti: I can maintain my own service for that...

ineiti:
Removed a lot of stuff that is unused and for testing, so I'll use a special branch for local testing, which is not ideal, but probably better than having `TestData` in `master`. Also fixed `phapp` to run correctly.

With regard to the `Wipe` endpoint, the best would be to have a separate API endpoint and protect it with a signature. But currently all of this is very experimental anyway, so I'm not sure it's really worth it. What we would need to discuss is, whether the personhood-service as such should in fact be separated from the contracts and run on different servers than the nodes. Currently it's just for convenience that the two run together. But I don't know how to do a decentralized untrusted service...

So my current stance is: having `wipe`, `add`, and `list` in one endpoint is the least of the bad design choice here, and changing it now doesn't really clean up a lot. We need a third opinion - 

@kc1212 - do you think it's worth making separate `(Party|Poll)(Add|List|Wipe)` endpoints?

external/proto/personhood.proto
```diff
@@ -2,29 +2,56 @@ syntax = "proto2";
...
   // InstanceID is where to find the party in the ledger.
-  required bytes instanceid = 2;
-  // FinalStatement describes the party and the signature of the organizers.
-  required pop.FinalStatement finalstatement = 3;
-  // Darc being responsible for the PartyInstance.
-  required darc.Darc darc = 4;
-  // Signer can call Invoke on the PartyInstance.
-  required darc.Signer signer = 5;
+  required bytes instanceid = 3;
+  // TODO: remove signer
```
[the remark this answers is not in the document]
ineiti: removed `Signer` in this branch

external/proto/personhood.proto
```diff
@@ -186,3 +213,200 @@ message TopupMessage {
   // Amount to coins to put in the message
   required uint64 amount = 2;
 }
+
+// TestStore is used to store test-structures. If it is called
+// with null pointers, nothing is stored, and only the currently
+// stored data is returned.
+// This will not be saved to disk.
+message TestStore {
```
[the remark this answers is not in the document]
ineiti: Removed in this branch.

personhood/api.go
```diff
@@ -20,12 +23,48 @@ func NewClient() *Client {
...
+}
+
+// WipeParties removes all parties stored in the system.
+func (c *Client) WipeParties(r onet.Roster) (errs []error) {
+	t := true
+	pl := PartyList{
+		WipeParties: &t,
+	}
+	for _, si := range r.List {
+		err := c.SendProtobuf(si, &pl, nil)
```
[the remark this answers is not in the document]
ineiti: See comment below.

Gilthoniel:
@ineiti could this service that you mentioned be included using a special build tag ? I totally got why this is not ideal and my goal is not to make your life miserable so we can have it on master but separate from a production build, that would be great.

personhood/phapp/test.sh
Gilthoniel: Fixed by #1741

ineiti:
@Gilthoniel the main goal of having everything included in `master` is that I'd get free update to the latest changes in the code - but then you'd have to include my build-tag in the tests, and I don't know if this is ideal. Also, with `v3` being just out there, it shouldn't need too many updates anyway.

Currently I created a directory `personhoodBeta` for me, and I'll test new things in there, before moving them over to `personhood`. That should work nicely...

external/js/cothority/src/byzcoin/contracts/credentials-instance.ts
```diff
@@ -124,16 +123,20 @@ export class CredentialStruct extends Message<CredentialStruct> {
...
-    public credentials: Credential[];
+    /**
+     * Returns a new CredentialStruct class
+     * @param d
+     */
+    static fromData(d: Buffer): CredentialStruct {
+        return CredentialStruct.decode(d);
+    }
+
+    readonly credentials: Credential[];
```
ineiti: I don't understand how that works - how can `credentials` be readonly, and then we do this?

https://github.com/dedis/cothority/blob/831f24277d778cafc78611ba27985c4e8f00d0d5/external/js/cothority/src/byzcoin/contracts/credentials-instance.ts#L86

Gilthoniel: _readonly_ only prevents you from doing `this.credentials = []` but it doesn't prevent you from updating the array. I want to avoid people from removing an entire array accidentally and then force them to use specific API like push/splice so you know that you have previous data to deal with. If we want something more strict we can freeze the array so push will be forbidden and you would need to instantiate a new struct with the new credentials.

personhood/contract_popparty.go
```diff
@@ -0,0 +1,492 @@
...
+	var sigStrs []string
+	for _, sig := range signers {
+		sigStrs = append(sigStrs, sig.Identity().String())
+	}
+	signerCtrs, err := cl.GetSignerCounters(sigStrs...)
+	if err != nil {
+		return
+	}
+
+	fsBuf, err := protobuf.Encode(&fs)
```
jeffallen: Pretty confused how it can be that this var of type FinalStatement is encoded then sent in via an arg called Description, the unmarshalled into c.Description. I can imagine how it might work, but there's either some code smell here, or maybe doc smell... or naming smell. (At least there's no caching to get wrong.)

personhood/contract_popparty.go
```diff
@@ -0,0 +1,492 @@
...
+	return
+}
+
+// PopPartyMine collects the reward. If 'atts' is nil, the party will be fetched from
+// byzcoin. One of 'coinIID' or 'd' must be set to either mine the coin, or spawn
+// a new darc and coin.
+func PopPartyMine(cl *byzcoin.Client, popIID byzcoin.InstanceID, kp key.Pair,
+	atts *Attendees, coinIID *byzcoin.InstanceID, d *darc.Darc) (err error) {
+	if atts == nil {
+		popProof, err := cl.GetProof(popIID.Slice())
```
jeffallen: popProof is not checked here for being an inclusion proof of the requested key. (Or am I mis-remembering how to safely use GetProof?)

personhood/contract_popparty.go
```diff
@@ -0,0 +1,492 @@
...
+	err = ctx.FillSignersAndSignWith(signers...)
+	if err != nil {
+		return
+	}
+	_, err = cl.AddTransactionAndWait(ctx, 5)
+	return
+}
+
+// PopPartyMine collects the reward. If 'atts' is nil, the party will be fetched from
+// byzcoin. One of 'coinIID' or 'd' must be set to either mine the coin, or spawn
```
jeffallen: I'm not clear why this whole coinIID or DARC thing exists. I think it's because I've lost track of the big picture here. If I can't find that big picture doc someplace else in this PR, perhaps you can explain it better here?

personhood/contract_popparty.go
```diff
@@ -0,0 +1,492 @@
...
+			},
+		}},
+	}
+	_, err = cl.AddTransactionAndWait(ctx, 5)
+	return
+}
+
+// PopPartyMineDarcToCoin calculates the coin given a darc and returns the coin instance.
+func PopPartyMineDarcToCoin(cl *byzcoin.Client, d *darc.Darc) (coinIID byzcoin.InstanceID, coin byzcoin.Coin, err error) {
+	h := sha256.New()
```
jeffallen: This code is duplicated up a few hundred lines in this file.

personhood/contract_spawner.go
```diff
@@ -0,0 +1,241 @@
...
+// Spawn creates a new spawner. Depending on what the client wants to spawn, this method will check the
+// price for that instance, and spawn a new instance if the price is correct. The following instances
+// and their arguments can be spawned:
+//   - ContractSpawnerID only with a valid darc. The arguments will be parsed for the costs:
+//     cost(Darc|Coin|Credential|Party|RoPaSci)
+//   - ContractDarcID takes the 'darc' argument and puts the calling darc as the parent darc
+//   - ContractCoinID takes 'coinName' and 'darcID' as arguments creates a coin using all inputs from
+//     'coinName' to the new coin, protected by the darcID. The IID of the coin is defined by
+//     sha256( "coin" | darcID )
+//     TODO: avoid that the payment is put in the newly created coin
```
jeffallen: Done?

personhood/contract_spawner.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (c ContractSpawner) getCoins(coins []byzcoin.Coin, cost byzcoin.Coin) error {
+	if cost.Value == 0 {
+		return nil
+	}
+	for i := range coins {
+		if coins[i].Name.Equal(cost.Name) {
+			if coins[i].Value >= cost.Value {
+				coins[i].SafeSub(cost.Value)
```
jeffallen: please don't ignore the err SafeSub returns. I see why you can get away with it here, but it's dangerous for a future reader/editor.

personhood/contract_spawner.go
```diff
@@ -0,0 +1,241 @@
...
+func (c ContractSpawner) VerifyInstruction(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruction, ctxHash []byte) error {
+	if inst.GetType() != byzcoin.SpawnType {
+		if err := inst.Verify(rst, ctxHash); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Spawn creates a new spawner. Depending on what the client wants to spawn, this method will check the
```
jeffallen: Where do the coins go? Are they just burned? Is that on purpose? Documented someplace?

personhood/contract_credential.go
```diff
@@ -0,0 +1,110 @@
...
+
+	var darcID darc.ID
+	_, _, _, darcID, err = rst.GetValues(inst.InstanceID.Slice())
+	if err != nil {
+		return
+	}
+
+	// Spawn creates a new credential as a separate instance.
+	ca := inst.DeriveID("")
+	if caBuf := inst.Spawn.Args.Search("instID"); caBuf != nil {
```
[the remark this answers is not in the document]
ineiti: In fact it is finally not used. So I cannot even make a good use-case ;)

Removed

personhood/contract_credential.go
```diff
@@ -0,0 +1,110 @@
...
+		darcID = darc.ID(darcIDBuf)
+	}
+	log.Lvlf3("Spawning Credential to %x", ca.Slice())
+	var ciBuf []byte
+	if ciBuf = inst.Spawn.Args.Search("credential"); ciBuf == nil {
+		ciBuf, err = protobuf.Encode(&c.CredentialStruct)
+		if err != nil {
+			return nil, nil, errors.New("couldn't encode CredentialInstance: " + err.Error())
+		}
+	}
```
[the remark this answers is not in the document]
ineiti: That's a trick-question: verifying is always good!

personhood/contract_popparty.go
ineiti: Thanks for catching this artifact - previously it was indeed a `FinalStatement` being passed in here. But when cleaning up it morphed into the `Description`.

Fun fact: it worked, because the `FinalStatement`'s first argument is a `Description`, so the `protobuf.Decode` produced a valid `Description` in the `Spawn`. Even though it looks really funny that it should've worked.

personhood/contract_popparty.go
ineiti: added `Verify`.

personhood/contract_popparty.go
ineiti: Updated comment:

```
// PopPartyMine is a method to be called by an outside client. It collects the reward for a given
// attendee of the party. For convenience, this can be called with some of the arguments being 'nil'.
// 
//   - atts - the list of the public keys of the attendees. If it is nil, the party will be fetched from
//     byzcoin. 
//   - coinIID - if set, 'd' must be nil. coinIID points to the coin InstanceID where the reward will be stored.
//   - d - if set, 'coinIID' must be nil. d is the darc that will be used to create a new coinInstance.
```

personhood/contract_popparty.go
ineiti: Yes, but no. This method is to be used from a client, the `sha256("coin" | darc.baseID)` further up doesn't have a `byzcoin.Client`. 

Perhaps `PopPartyMineDarcToCoin` and `PopPartyMine` should go into `api.go`?

personhood/contract_spawner.go
ineiti: Yes, removed TODO.

personhood/contract_spawner.go
ineiti: Added comment:
```
// Currently the coins are simply burned and will never be seen again. In future versions, the coins will
// be part of the mining reward for the nodes participating in the consensus.
```

Issue #1731, opened 2019-02-21
Fix issues discovered when porting the skipchain explorer


Gilthoniel:
- [x] model.json is not included in the build
- [x] protobufjs creates Uint8Array instead of Buffer in a browser environment
- [x] better default values when creating protobuf-based classes

PR #1734, opened 2019-02-22
Fix issues discovered when porting the skipchain explorer

Gilthoniel:
This PR fixes the issue #1731


external/js/cothority/src/byzcoin/contracts/ro-pa-sci-instance.ts
```diff
@@ -177,25 +177,52 @@ export default class RoPaSciInstance {
...
+    readonly secondPlayer: number;
+    readonly secondPlayerAccount: Buffer;
+
+    constructor(props?: Properties<RoPaSciStruct>) {
+        super(props);
+
+        this.firstPlayerHash = Buffer.from(this.firstPlayerHash || EMPTY_BUFFER);
+        this.secondPlayerAccount = Buffer.from(this.secondPlayerAccount || EMPTY_BUFFER);
+
+        Object.defineProperty(this, "firstplayer", {
```
kc1212: does the capitalization matter? `Object.defineProperty(this, "firstplayer"` vs `this.firstPlayer` and `Object.defineProperty(this, "firstplayerhash",` vs `this.firstPlayerHash`

Gilthoniel: Yes it is important because the defineProperty will create the necessary fields for protobuf to work. Thus yes the _secondPlayer_ is wrong, I will fix it, thanks!

Gilthoniel: It has not been caught because those tests are disabled until personhood is merged.

PR #1735, opened 2019-02-24
Fix one lint problem and add lint as a direction on Makefile

claucece:
I don't know if this is of your interest; but, just in case, I added a direction to the Makefile as well.

Additionally, there are still some golint problems:

```
calypso/protocol/ocs_struct.go:45:2: struct field Ui should be UI
calypso/protocol/onchain_test.go:52:2: var Ui should be UI
pop/service/service_test.go:447:3: don't use underscores in Go names; var hash_fr should be hashFr
pop/service/service_test.go:561:2: don't use underscores in Go names; var copy_descs should be copyDescs
```

Idk if this is something you might want to check :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,3 +29,5 @@ docker:
 test_java: docker
 	cd external/java; mvn test
 
+lint:
+	golint ./...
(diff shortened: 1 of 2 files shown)


jeffallen:
Thank you for this, but we use `make test` to run golint, and the wrapper that does that (https://github.com/dedis/Coding/blob/master/bin/Makefile.base#L19) excludes some hits, including the ones you fixed.

I agree that a simpler "golint" would be better, but for historical reasons, it is like this and we'll keep it that way.

PR #1739, opened 2019-02-25
Hash byte array to point securely

kc1212:
Depends on dedis/kyber#365

For travis to pass, we need to merge kyber, make a new pre-release and then update `go.mod` in onet and cothority to use it.


Gilthoniel:
I've added the JS implementation. As for Java, the tests will pass after Kyber has been updated.

kc1212:
TS part looks good to me.

PR #1741, opened 2019-02-26
Improve the speed of the integration tests

Gilthoniel:
This replaces the 10 seconds sleep by a ping check to verify when all the conodes have started.


Gilthoniel:
It would be nice if someone can confirm this works on MacOS aswell.

kc1212:
`authprox/apadmin/test.sh:BCSetup` is failing for me with the message "instruction verification failed: schnorr: invalid signature", does that happen to you too?

Gilthoniel:
Hm no, it's working both locally and on Jenkins for me. Could it be you have out-of-date dependencies ? Like if you have a build folder it won't compile so you end up with a wrong hash for the instructions (or similar)

kc1212:
You're right, deleting the build folder fixed it.

PR #1748, opened 2019-03-01
Use the roster to fetch blocks

Gilthoniel:
This PR changes the skipchain RPC to poll more than the leader to fetch blocks.

### DIFF
--- a/external/js/cothority/src/skipchain/skipchain-rpc.ts
+++ b/external/js/cothority/src/skipchain/skipchain-rpc.ts
@@ -114,10 +114,11 @@ export default class SkipchainRPC {
     /**
      * Get the shortest path to the more recent block starting from latestID
      *
-     * @param latestID ID of the block
+     * @param latestID  ID of the block
+     * @param verify    Verify the integrity of the chain when true
      * @returns a promise that resolves with the list of blocks
      */
-    async getUpdateChain(latestID: Buffer): Promise<SkipBlock[]> {
+    async getUpdateChain(latestID: Buffer, verify = true): Promise<SkipBlock[]> {
         const req = new GetUpdateChain({ latestID });
         const ret = await this.pool.send<GetUpdateChainReply>(req, GetUpdateChainReply);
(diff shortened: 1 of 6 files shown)


external/js/cothority/src/skipchain/skipchain-rpc.ts
```diff
@@ -104,68 +106,54 @@ export default class SkipchainRPC {
...
+    async getUpdateChain(latestID: Buffer, verify = true): Promise<SkipBlock[]> {
         const req = new GetUpdateChain({ latestID });
-        const ret = await this.conn[0].send<GetUpdateChainReply>(req, GetUpdateChainReply);
+        const ret = await this.pool.send<GetUpdateChainReply>(req, GetUpdateChainReply);
+        const blocks = ret.update;
 
-        const err = this.verifyChain(ret.update, latestID);
-        if (err) {
-            throw new Error(`invalid chain received: ${err.message}`);
+        const last = blocks[blocks.length - 1];
```
kc1212: what would happen if blocks is empty?

Gilthoniel: You can check in your browser: in JS you get _undefined_ for undefined indices

external/js/cothority/src/skipchain/skipchain-rpc.ts
```diff
@@ -104,68 +106,54 @@ export default class SkipchainRPC {
...
+        if (last && last.forwardLinks.length > 0) {
+            // more blocks exist but typically the roster has changed
+            const rpc = new SkipchainRPC(last.roster);
```
kc1212: will this part create a lot of new connections? maybe some connections will be duplicated?

i don't know if this is a serious issue or if we need to do something about it because roster changes are infrequent.

Gilthoniel: The connection is actually created only when you send something. Nothing is opened at this point.

PR #1762, opened 2019-03-12
Fix protobuf parsing of the instruction in CothorityJS

Gilthoniel:
Some of the messages were not actually instantiated with the right
class as there was no registration. It also forces Long to be used
in a browser environment.

### DIFF
--- a/external/js/cothority/README.md
+++ b/external/js/cothority/README.md
@@ -49,6 +49,37 @@ You must use the given script instead of `npm publish` because we need to publis
 the _dist_ folder instead. If you try to use the official command, you will get
 an error on purpose.
 
+### Message classes
+
+You can write a class that will be used when decoding protobuf messages by using
+this template:
+```javascript
+class MyMessage extends Message<MyMessage> {
+    /**
+     * @see README#Message classes
+     */
+    static register() {
+        registerMessage("abc.MyMessage", MyMessage, MyMessageDependency);
+    }
+
+    readonly myField: MyMessageDependency;
+
+    constructor(props?: Properties<MyMessage>) {
+        super(props);
+
+        // whatever you need to do for initialization
+    }
+}
+
+MyMessage.register();
+```
+
+Note that protobuf will instantiate with an empty object and then fill the fields
+so this happens after the constructor has been called.
+The _register_ is used to register the dependencies of the message but you also
+have to use it as a side effect of the package so that as soon as the class is
+imported, the message will be known by protobuf and used during decoding.
+
 ### Side note on Buffer
 
 Protobuf definition and classes implemented expect a _Buffer_ for _bytes_ but
--- a/external/js/cothority/package-lock.json
+++ b/external/js/cothority/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@dedis/cothority",
-  "version": "3.0.1",
+  "version": "3.0.2",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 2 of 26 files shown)


external/js/cothority/src/byzcoin/client-transaction.ts
```diff
@@ -6,6 +6,9 @@ import Signer from "../darc/signer";
 import { EMPTY_BUFFER, registerMessage } from "../protobuf";
 import { InstanceID } from "./instance";
 
+// messages registration
```
jeffallen: This feels fragile. How do we know which imports with which registration side-effects are needed in which files? Is there a way to be more systematic about this?

Gilthoniel: It is definitely. I'm looking for a way to test this.

Gilthoniel: So after investigating ways to force an import when there are side effects, it turns out TS guys don't want that to happen (related to static checking) so I changed the registration to declare the internal dependencies of a message so that it will correctly import every type used in the message.

I think this is better than using the previous force import because you know exactly which type is used in your message and then you simply declare it. No need to check the modules if they have side-effects.

The only confusing part is that you still need to register the messages as a side-effect because any file is an independent module and we don't want people to have to register manually when they import specific messages. I will add a paragraph in the README about that.

external/js/cothority/README.md
```diff
@@ -49,6 +49,37 @@ You must use the given script instead of `npm publish` because we need to publis
...
 an error on purpose.
 
+### Message classes
+
+You can write a class that will be used when decoding protobuf messages by using
+this template:
+```javascript
+class MyMessage extends Message<MyMessage> {
+    /**
+     * @see README#Message classes
```
jeffallen: I don't know if you expected this to work as a link, but it doesn't.

Gilthoniel: No I don't think there is an actual way to do that. I'm open to suggestions.

external/js/cothority/package-lock.json
```diff
@@ -4836,13 +4836,13 @@
       "dependencies": {
         "ansi-regex": {
           "version": "3.0.0",
-          "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-3.0.0.tgz",
```
jeffallen: I don't understand why all of these resolved lines were changed. Here is a bug that mentions it might be due to using npm link, which i know you use during development. https://npm.community/t/npm-install-or-npm-update-turns-a-bunch-of-resolved-in-package-lock-json-from-real-values-to-false/3308

external/js/cothority/README.md
jeffallen: I don't know if you expected this to work as a link, but it doesn't. (And it appears all over this PR too.)

Gilthoniel: I know it doesn't work as a link but the goal was to centralize the comment to one place and not having this function laying there and there with no explanations. I wanted it to be lightweight but if you prefer me to use an actual link I'll change it.

jeffallen: No, it's fine.

external/js/cothority/package-lock.json
Gilthoniel: Ok so I cleaned my npm cache and stuff and it turns out to fix the resolved. It still removes some optional fields but those are less worrying.

PR #1764, opened 2019-03-14
Check block integrity when necessary

### LINKED ISSUE #1703: Java library doesn't check the hash of received blocks
Data of the block is accessed but the hash is never calculated to make sure it hasn't been forged. One simple example is the proof verification that only checks the forward links.

Gilthoniel:
The proof verification was not checking that the last forward link was pointing on the actual latest block provided by the proof which that a malicious peer can in some extend modify the block. This check has been added in the Java and the Go code.

Another issue was that only the forward links of a request skipblock were checked so any other fields could be modified as we were not checking the received skipblock hash. This has been fixed in the Java library.

Fixes #1703


Gilthoniel:
I think this covers the different request of blocks we do in the Java library. @kc1212 Please let me know if I forgot something.

I'm investigating the skipchain service on the Go side..

kc1212:
The java side looks good

Gilthoniel:
So the conclusion of this PR is that mainly the byzcoin and skipchain API were assuming data coming from the service is safe and then I added integrity checks where it was possible to verify. Sometimes you lack the context and then the verification must be done upstream, like when you ask for a single block (maybe we should always return a proof alongside?).

Aside of that the main goal of this PR was to fix the proof verification to insure the last forward link points to the correct returned latest block.

byzcoin/api.go
```diff
@@ -417,3 +450,54 @@ func DefaultGenesisMsg(v Version, r *onet.Roster, rules []string, ids ...darc.Id
 	}
 	return &m, nil
 }
+
+func verifyGenesisBlock(sb *skipchain.SkipBlock, req *CreateGenesisBlock) error {
```
kc1212: nit: I would name those parameters like: `actual *skipchain.SkipBlock, expected *CreateGenesisBlock`, so it's obvious which input is trusted.

byzcoin/api.go
```diff
@@ -417,3 +450,54 @@ func DefaultGenesisMsg(v Version, r *onet.Roster, rules []string, ids ...darc.Id
...
+
+func extractDarcID(sb *skipchain.SkipBlock) (darc.ID, error) {
+	var data DataBody
+	err := protobuf.Decode(sb.Payload, &data)
+	if err != nil {
+		return nil, fmt.Errorf("fail to decode data: %v", err)
+	}
+
+	if len(data.TxResults) != 1 || len(data.TxResults[0].ClientTransaction.Instructions) != 1 {
+		return nil, errors.New("didn't get only one instruction")
```
kc1212:
```suggestion
		return nil, errors.New("genesis darc tx should only have one instruction")
```

byzcoin/api.go
```diff
@@ -60,7 +64,18 @@ func NewLedger(msg *CreateGenesisBlock, keep bool) (*Client, *CreateGenesisBlock
 	if err := c.SendProtobuf(msg.Roster.List[0], msg, reply); err != nil {
 		return nil, nil, err
 	}
-	c.ID = reply.Skipblock.CalculateHash()
+
+	// for testing purpose
+	if testCorruptNewLedgerResponse != nil {
+		reply = testCorruptNewLedgerResponse
+	}
```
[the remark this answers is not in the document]
kc1212: another possibility: can we register a corrupted byzcoin service that always just returns a wrong genesis block (e.g., embed `byzcoin.Service` inside `byzcoin.BadService` and overwrite the create genesis block function), it doesn't even have to do consensus or anything like that, just returns the bad block? just a few ideas to think about

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
```diff
@@ -86,6 +86,12 @@ public SkipBlock getSkipblock(SkipblockId id) throws CothorityCommunicationExcep
             SkipchainProto.SkipBlock sb = SkipchainProto.SkipBlock.parseFrom(msg);
             SkipBlock ret = new SkipBlock(sb);
 
+            if (!Arrays.equals(id.getId(), ret.getHash())) {
+                // we don't want a malicious peer to send a block with tempered
+                // data like a different roster with different public keys
+                throw new CothorityCommunicationException("invalid block integrity");
+            }
```
kc1212: the function comment says the block is not verified, is that still the case?

skipchain/api.go
```diff
@@ -135,9 +158,56 @@ func (c *Client) CreateGenesisSignature(ro *onet.Roster, baseH, maxH int, ver []
 	if err != nil {
 		return nil, err
 	}
+
+	if testCorruptSSBResponse != nil {
+		sb = testCorruptSSBResponse
+	}
```
kc1212: see comment above regarding the possibility of creating a bad service

skipchain/api.go
```diff
@@ -135,9 +158,56 @@ func (c *Client) CreateGenesisSignature(ro *onet.Roster, baseH, maxH int, ver []
...
+	if err != nil {
+		return nil, err
+	}
+
 	return sb.Latest, nil
 }
 
+// CompareGenesisBlocks compares the content of two blocks and returns an error
+// if there is any difference
+func compareGenesisBlocks(prop *SkipBlock, ret *SkipBlock) error {
```
kc1212: why can't you re-calculate the hash? if the hash is different then the block content should also be different right?

Gilthoniel: I wish I could but the service is adding information to the genesis like a random backlink

byzcoin/api.go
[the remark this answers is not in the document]
Gilthoniel: Yes for the `verifyGenesisBlock` but I really want to make sure the request is checking this, and that no one removed this check accidentally.

I tried that first, but the service name is a constant (which is fine) so you can't change which service is going to be used, or am I missing something ? I would love to do it that way..

external/java/src/main/java/ch/epfl/dedis/skipchain/SkipchainRPC.java
Gilthoniel: I'll change the comment. The integrity is verified but the caller must verify if it belongs to the chain.

byzcoin/api.go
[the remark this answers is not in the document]
Gilthoniel: Nevermind, I found a way to go around without too many complex mocking

PR #1770, opened 2019-03-18
Implement commit for byzcoin staging trie

kc1212:
The the commit feature was never used before because the staging trie is always thrown away after we propose the block. We need it for the pipeline feature because it can be used as a checkpoint. That is, everything that's not committed could be "rolled back" when an error happens in the pipeline.

### DIFF
--- a/byzcoin/statetrie_test.go
+++ b/byzcoin/statetrie_test.go
@@ -1,6 +1,7 @@
 package byzcoin
 
 import (
+	"bytes"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


byzcoin/statetrie_test.go
```diff
@@ -51,4 +52,20 @@ func TestStateTrie(t *testing.T) {
...
+	require.NoError(t, sst.Set(key2, val2))
+	newRoot := sst.GetRoot()
+	require.False(t, bytes.Equal(oldRoot, newRoot))
+	candidateVal2, err := sst.Get(key2)
+	require.NoError(t, err)
+	require.True(t, bytes.Equal(val2, candidateVal2))
+
+	// test the commit of staging state trie, root should be computed differently now, but it should be the same
+	require.NoError(t, sst.Commit())
+	require.True(t, bytes.Equal(sst.GetRoot(), newRoot))
```
Gilthoniel: I'm not sure to get this test because it's like doing 1==1. The `Commit` isn't supposed to commit the changes to `st` (in the contrary of `sst`) ? Looking at the test, it would pass without the commit.

kc1212: Yes, it would pass without `Commit`. It's just making sure that sst.StagingTrie.Commit() gets called, the actual tests are in the trie package.

kc1212: I added this test because I mistakenly implemented it as:

```
// Commit commits the staged data to the source trie.
func (t *stagingStateTrie) Commit() error {
	return t.Commit()
}
```

which caused a stack overflow

Gilthoniel: I see, now I got it

PR #1776, opened 2019-03-21
Improve bcadmin usability; fix Calypso docs

jeffallen:


byzcoin/api.go
```diff
@@ -450,6 +452,15 @@ func DefaultGenesisMsg(v Version, r *onet.Roster, rules []string, ids ...darc.Id
 	return &m, nil
 }
 
+// getServer returns a server from the roster, observing the ServerNumber selection.
+func (c *Client) getServer() *network.ServerIdentity {
```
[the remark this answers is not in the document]
jeffallen: I dunno, but Travis is happy, so I'm happy.

go.sum
```diff
@@ -49,18 +49,8 @@ go.dedis.ch/fixbuf v1.0.3 h1:hGcV9Cd/znUxlusJ64eAlExS+5cJDIyTyEG+otu5wQs=
 go.dedis.ch/fixbuf v1.0.3/go.mod h1:yzJMt34Wa5xD37V5RTdmp38cz3QhMagdGoem9anUalw=
 go.dedis.ch/kyber/v3 v3.0.0-pre2 h1:ezviD36AEOytXJn91tsvQeR+rEzo3UOh75P/PzSisRo=
```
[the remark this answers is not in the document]
jeffallen: I'll use `go mod tidy` on it and see what happens.

calypso/api_test.go
```diff
@@ -158,8 +160,48 @@ func TestClient_Calypso(t *testing.T) {
...
+//   - suite - the cryptographic suite to use
+//   - X - the aggregate public key of the DKG
+//   - C - the encrypted key-slices
+//   - XhatEnc - the re-encrypted schnorr-commit
+//   - xc - the private key of the reader
+//
+// Output:
+//   - key - the re-assembled key
+//   - err - an eventual error when trying to recover the data from the points
+func decodeKey(suite kyber.Group, X kyber.Point, C kyber.Point, XhatEnc kyber.Point,
```
kc1212: is it better to be conservative and keep `DecodeKey` and `EncodeKey` for compatibility?

ineiti: It's in `_test`, so I'd keep it private.

kc1212: what I mean is the old `DecodeKey` and `EncodeKey` are public, they're in `struct.go`, so removing those will be, in theory, breaking compatibility

libtest.sh
```diff
@@ -366,9 +366,9 @@ runCoBG(){
         ./conode -d $DBG_SRV -c co$nb/private.toml server >& "$COLOG$nb.log"
       fi
       touch "$COLOG$nb.log.dead"
-    ) &
+    ) 2>/dev/null &
     # This makes `pkill conode` not outputting errors here
-    disown
+    disown %1
```
ineiti: Seems I missed how to properly disown the process here - what is the %1 doing?

jeffallen: I thought it was required by the disown docs, but not I see `If no jobspec is present, and neither the -a nor the -r option is supplied,  the  current job is used` so perhaps it was always working correctly.

The problem I was trying to solve was that bash was complaining about the child process dying, but that was solved by throwing the stderr from the subshell away (the stderr from the conode is already handled above). So I will remove the %1.

calypso/api_test.go
jeffallen: According to our most recent definition of stability ("we are not willing to accept chagnes right now that destabilise current student projects based on cothority v3") then if these APIs are not useful, and not used, then removing them is ok. This was my reasoning for removing them; they seem to be only useful in test, and should never have been exported, and should not be in use now.

jeffallen:
Man, there's so much gunk in this API, it is clear that we were figuring this out as we wrote it. I'm sending through another commit that will simplify things further.

PR #1778, opened 2019-03-21
Pipeline transaction processing

### LINKED ISSUE #1409: Further ByzCoin optimizations

kc1212:
This is an optimisation for the leader. While the leader is waiting to create
the next block, he can already start collecting and processing transactions. 

Fixes #1409

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -1533,8 +1533,8 @@ func (s *Service) startPolling2(scID skipchain.SkipBlockID) chan bool {
 		processor: &defaultTxProcessor{
 			stopCollect: make(chan bool, 1),
 			stopProcess: make(chan bool, 1),
-			scID: scID,
-			Service: s,
+			scID:        scID,
+			Service:     s,
 		},
 	}
 	st, err := s.getStateTrie(scID)
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -1531,7 +1531,7 @@ func (s *Service) LoadBlockInfo(scID skipchain.SkipBlockID) (time.Duration, int,
 func (s *Service) startPolling(scID skipchain.SkipBlockID) chan bool {
 	pipeline := txPipeline{
 		processor: &defaultTxProcessor{
-			stopCollect: make(chan bool, 1),
+			stopCollect: make(chan bool),
 			scID:        scID,
 			Service:     s,
 		},
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -1107,6 +1107,7 @@ func (s *Service) catchUp(sb *skipchain.SkipBlock) {
 		s.catchingUp = false
 		s.updateTrieLock.Unlock()
 	}()
+
 	log.Lvlf2("%v Catching up %x / %d", s.ServerIdentity(), sb.SkipChainID(), sb.Index)
 
 	// Load the trie.
@@ -1160,6 +1161,7 @@ func (s *Service) catchUp(sb *skipchain.SkipBlock) {
 		latest = updates[len(updates)-1]
 		trieIndex = latest.Index
 	}
+	log.Lvlf2("%v Done catch up %x / %d", s.ServerIdentity(), sb.SkipChainID(), trieIndex)
 }
 
 // updateTrieCallback is registered in skipchain and is called after a
@@ -1993,7 +1995,7 @@ func (s *Service) getTxs(leader *network.ServerIdentity, roster *onet.Roster, sc
 		// The function will prevent multiple request to catch up so we can securely call it here
 		err := s.catchupFromID(roster, scID)
 		if err != nil {
-			log.Error(err)
+			log.Error(s.ServerIdentity(), err)
 		}
 		// Give up the current request and wait for the next one, and keep skipping requests
 		// until the catching up is done
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -1113,7 +1113,7 @@ func (s *Service) catchUp(sb *skipchain.SkipBlock) {
 	download := false
 	st, err := s.getStateTrie(sb.SkipChainID())
 	if err != nil {
-		log.Warn("problem with trie:", err)
+		log.Warn(s.ServerIdentity(), "problem with trie:", err)
 		download = true
 	} else {
 		download = sb.Index-st.GetIndex() > catchupDownloadAll
@@ -1121,7 +1121,7 @@ func (s *Service) catchUp(sb *skipchain.SkipBlock) {
 
 	// Check if we are updating the right index.
 	if download {
-		log.Lvl2("Downloading whole DB for catching up")
+		log.Lvl2(s.ServerIdentity(), "Downloading whole DB for catching up")
 		err := s.downloadDB(sb)
 		if err != nil {
 			log.Error("Error while downloading trie:", err)
@@ -1529,7 +1529,6 @@ func (s *Service) LoadBlockInfo(scID skipchain.SkipBlockID) (time.Duration, int,
 }
 
 func (s *Service) startPolling(scID skipchain.SkipBlockID) chan bool {
-	// TODO is closing??
 	pipeline := txPipeline{
 		processor: &defaultTxProcessor{
 			stopCollect: make(chan bool, 1),
@@ -1549,6 +1548,17 @@ func (s *Service) startPolling(scID skipchain.SkipBlockID) chan bool {
 	go func() {
 		s.pollChanWG.Add(1)
 		defer s.pollChanWG.Done()
+
+		s.closedMutex.Lock()
+		if s.closed {
+			s.closedMutex.Unlock()
+			return
+		}
+
+		s.working.Add(1)
+		defer s.working.Done()
+		s.closedMutex.Unlock()
+
 		pipeline.start(&initialState, stopChan)
 	}()
 
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -1322,7 +1322,7 @@ func (s *Service) updateTrieCallback(sbID skipchain.SkipBlockID) error {
 	if nodeIsLeader {
 		if _, ok := s.pollChan[scIDstr]; !ok {
 			log.Lvlf2("%s new leader started polling for %x", s.ServerIdentity(), sb.SkipChainID())
-			s.pollChan[scIDstr] = s.startPolling2(sb.SkipChainID())
+			s.pollChan[scIDstr] = s.startPolling(sb.SkipChainID())
 		}
 	} else {
 		if c, ok := s.pollChan[scIDstr]; ok {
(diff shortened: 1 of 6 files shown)


kc1212:
The performance test depends on the workload. For the pipeline feature to shine, it needs to run contracts that take time. To simulate it, I added a `time.Sleep(10 * time.Millisecond)` in the coin contract and plotted the figures below. The first one is on master, the second one is on this branch (both have the 10 ms delay in the contract). 

master

pipeline

The performance doubled because the leader doesn't execute the transactions at the block interval boundaries. Instead, it executes them while waiting.

On the other hand, if the contract doesn't take a long time to run, then there is no difference in performance.

For reference, this is the code I changed to run this experiment, it can be applied on master and this branch.

```diff
diff --git a/byzcoin/collect_tx.go b/byzcoin/collect_tx.go
index 7fb2b1cd..b56ef64b 100644
--- a/byzcoin/collect_tx.go
+++ b/byzcoin/collect_tx.go
@@ -118,6 +118,7 @@ func (p *CollectTxProtocol) Dispatch() error {
 	}
 
 	// send the result of the callback to the root
+	time.Sleep(200 * time.Millisecond) // simulate slow getTxs
 	resp := &CollectTxResponse{
 		Txs: p.getTxs(req.ServerIdentity, p.Roster(), req.SkipchainID, req.LatestID),
 	}
diff --git a/byzcoin/contracts/coins.go b/byzcoin/contracts/coins.go
index 0fe62248..a1949a46 100644
--- a/byzcoin/contracts/coins.go
+++ b/byzcoin/contracts/coins.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"errors"
+	"time"
 
 	"go.dedis.ch/cothority/v3/byzcoin"
 	"go.dedis.ch/cothority/v3/darc"
@@ -111,6 +112,7 @@ func (c *contractCoin) Invoke(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instru
 		coinsArg = binary.LittleEndian.Uint64(coinsBuf)
 	}
 
+	time.Sleep(10 * time.Millisecond)
 	switch inst.Invoke.Command {
 	case "mint":
 		// mint simply adds this amount of coins to the account.
diff --git a/byzcoin/simulation/coins.go b/byzcoin/simulation/coins.go
index 4c140962..7dd28a78 100644
--- a/byzcoin/simulation/coins.go
+++ b/byzcoin/simulation/coins.go
@@ -169,6 +169,12 @@ func (s *SimulationService) Run(config *onet.SimulationConfig) error {
 		txs := s.Transactions / s.BatchSize
 		insts := s.BatchSize
 		log.Lvlf1("Sending %d transactions with %d instructions each", txs, insts)
+
+		// Confirm the transaction by sending the last transaction using
+		// AddTransactionAndWait. There is a small error in measurement,
+		// as we're missing one of the AddTransaction call in the measurements.
+		confirm := monitor.NewTimeMeasure("confirm")
+
 		tx := byzcoin.ClientTransaction{}
 		// Inverse the prepare/send loop, so that the last transaction is not sent,
 		// but can be sent in the 'confirm' phase using 'AddTransactionAndWait'.
@@ -213,11 +219,6 @@ func (s *SimulationService) Run(config *onet.SimulationConfig) error {
 			}
 			prepare.Record()
 		}
-
-		// Confirm the transaction by sending the last transaction using
-		// AddTransactionAndWait. There is a small error in measurement,
-		// as we're missing one of the AddTransaction call in the measurements.
-		confirm := monitor.NewTimeMeasure("confirm")
 		log.Lvl1("Sending last transaction and waiting")
 		_, err = c.AddTransactionAndWait(tx, 20)
 		if err != nil {
@@ -242,12 +243,6 @@ func (s *SimulationService) Run(config *onet.SimulationConfig) error {
 		}
 		confirm.Record()
 		roundM.Record()
-
-		// This sleep is needed to wait for the propagation to finish
-		// on all the nodes. Otherwise the simulation manager
-		// (runsimul.go in onet) might close some nodes and cause
-		// skipblock propagation to fail.
-		time.Sleep(blockInterval)
 	}
 	// We wait a bit before closing because c.GetProof is sent to the
 	// leader, but at this point some of the children might still be doing
diff --git a/byzcoin/simulation/coins.toml b/byzcoin/simulation/coins.toml
index 0cb64914..472cfcb3 100644
--- a/byzcoin/simulation/coins.toml
+++ b/byzcoin/simulation/coins.toml
@@ -1,13 +1,16 @@
 Simulation = "TransferCoins"
 Servers = 2
 Bf = 4
-Rounds = 1
+Rounds = 4
 RunWait = "6000s"
 Suite = "Ed25519"
 # Keep the different columns in case someboday wants to run another battery
 # of tests
 
 Keep,   Transactions, BatchSize,  Hosts,  Delay, BlockInterval
-true,   200,          5,          5,     10,    "1s"
+true,   400,          5,          4,     10,    "1s"
+true,   400,          5,          5,     10,    "1s"
+true,   400,          5,          6,     10,    "1s"
+true,   400,          5,          7,     10,    "1s"
 # true,   200,          10,         10,     10,    "2s"
 # true,   200,          20,         10,     10,    "2s"
diff --git a/byzcoin/simulation/plot_sim.py b/byzcoin/simulation/plot_sim.py
index 8a9cf531..d9ab02db 100755
--- a/byzcoin/simulation/plot_sim.py
+++ b/byzcoin/simulation/plot_sim.py
@@ -53,6 +53,7 @@ for delay in delays:
             plt.xlabel('number of hosts')
             plt.ylabel('time in seconds')
             plt.title(titlestring)
+            plt.grid()
             plt.savefig(data_dir + 'barplot_' + namestring + '.png')
             plt.close()
 
@@ -68,5 +69,6 @@ for delay in delays:
             plt.xlabel('number of hosts')
             plt.ylabel('logarithm of time in seconds')
             plt.title(titlestring)
+            plt.grid()
             plt.savefig(data_dir + 'barplot_log_delay_' + namestring + '.png')
             plt.close()
```

byzcoin/tx_pipeline.go
```diff
@@ -0,0 +1,394 @@
...
+	<-stopChan
+	close(stopChan1)
+	close(stopChan2)
+	p.processor.Stop()
+	p.wg.Wait()
+}
+
+func (p *txPipeline) collectTx() (<-chan ClientTransaction, chan<- bool) {
+	stopChan := make(chan bool, 1)
+	outChan := make(chan ClientTransaction, 200)
```
Gilthoniel: what's the reasoning behind 200 ?

byzcoin/tx_pipeline.go
```diff
@@ -0,0 +1,394 @@
...
+	// wait for signal and then stop
+	<-stopChan
+	close(stopChan1)
+	close(stopChan2)
+	p.processor.Stop()
+	p.wg.Wait()
+}
+
+func (p *txPipeline) collectTx() (<-chan ClientTransaction, chan<- bool) {
+	stopChan := make(chan bool, 1)
```
Gilthoniel: Same as the other stop channel, is the buffering necessary ?

byzcoin/tx_pipeline.go
Gilthoniel: I'm thinking about a crazy scenario where you get tons of TXs (more than 200) which means that `outChan` will block at some point (a lot of small TXs but a slow contract) and then you stop everything. I might be wrong but I don't think that close can hang so `processTxs` will stop and `collectTx` will be stuck in line 287.

byzcoin/service.go
```diff
@@ -1529,138 +1529,40 @@ func (s *Service) LoadBlockInfo(scID skipchain.SkipBlockID) (time.Duration, int,
 }
 
 func (s *Service) startPolling(scID skipchain.SkipBlockID) chan bool {
-	s.pollChanWG.Add(1)
-	closeSignal := make(chan bool)
+	pipeline := txPipeline{
+		processor: &defaultTxProcessor{
+			stopCollect: make(chan bool, 1),
```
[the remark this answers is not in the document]
kc1212: i think i was sending stuff down the channel but not anymore, so it doesn't need to be buffered

byzcoin/tx_pipeline.go
kc1212: 200 is kinda arbitrary, as we do more testing hopefully we'll settle down on a suitable value. yes, `collectTx` could get stuck. I'm not too sure how we should handle the situation when clients flood the ledger. Perhaps we set a large channel size, then do a non-blocking write with a select such that if clients try to overflow the transaction queue, then we just drop the transactions.

byzcoin/tx_pipeline.go
kc1212: nope

byzcoin/tx_pipeline.go
```diff
@@ -0,0 +1,394 @@
...
+	sst *stagingStateTrie
+
+	// Below are changes that were made that led up to the state in sst
+	// from the starting point.
+	scs StateChanges
+	txs TxResults
+}
+
+func (s *txProcessorState) size() int {
+	// TODO encoding again is wasting CPU, is there a better way?
```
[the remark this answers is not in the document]
kc1212: good idea, we could cache that in `txProcessorState`

byzcoin/tx_pipeline.go
Gilthoniel: Ok so let's resume:
1. CollectTx is sending TX one by one in the outChan inside a goroutine
2. ProcessTx is getting the TX one by one coming from outChan inside a goroutine

What if `processTx` doesn't rely on a stop channel but only on the outChan so that when this channel closes, it stops the processing go routine because you know for sure none are coming anymore ?

kc1212: yes, that could be a simplification. then we don't need the `stopChan` for `processTxs` anymore

kc1212:
https://github.com/dedis/cothority/issues/1795 is still a problem, but it doesn't manifest in this PR anymore due to "good" randomness

PR #1780, opened 2019-03-22
Remove BLS service from conodes

Gilthoniel:
This removes the bls service as it allows anyone to ask for any
given bytes to sign which actually an oracle to an attacker. It
then also cleans the toml files used for testing. The check is
replaced by fetching the status of the conodes.

While working on the rogue public-key attack, I noticed that this service lying around is not just a side-effect but it actually opens an oracle..


kc1212:
This PR looks good but I'm not sure whether it is classified as a breaking change.

Gilthoniel:
we might wait on Jeff's opinion. The check command is still present but with less details.

ineiti:
A nicer solution, IMO, would be to only sign the `sha256(message)` or even `sha256("bls" | message)`, then you can keep the bls signing service, but nobody can use it as an oracle, as you're not supposed to know the pre-image of the hash... The latter gives even more security, as you might know the message, but using `sha256("bls" | message)`, your message would have to start with "bls".

jeffallen:
I think our backwards compatibility commitment is "we will not destabilise student's work during the semester". Because practically speaking for the moment, we do not have industrial users that are sensitive to strict backwards compatibility. So, I'm ok with this.

PR #1781, opened 2019-03-22
Fix out of bound leader index in view change

Gilthoniel:
This fixes the potential crash due to a panic when a view change request is sent with an invalid leader index. It changes the protocol so that negative indices are refused and bigger are reduced modulo the roster length.


Gilthoniel:
The PR also includes a fix for the flaky tests due to a bug in Go 1.11

PR #1784, opened 2019-03-27
Fix catch up 1777

Gilthoniel:
This PR moves the restoration of the state change storage to start on boot to recreate the state of the storage (used to clean). It was previously executed after the blocks catch up with the given state trie which is bad. The blocks catch up procedure takes care of creating the state changes for new blocks so we don't need to redo it again after it.

Note: it also cleans the `DecodeWithConstructor` usage to be only `Decode`

### DIFF
--- a/byzcoin/service_test.go
+++ b/byzcoin/service_test.go
@@ -2170,7 +2170,7 @@ func TestService_StateChangeStorage(t *testing.T) {
 			sb, err := service.skService().GetSingleBlock(&skipchain.GetSingleBlock{ID: res.BlockID})
 			require.Nil(t, err)
 			var header DataHeader
-			err = protobuf.DecodeWithConstructors(sb.Data, &header, network.DefaultConstructors(cothority.Suite))
+			err = protobuf.Decode(sb.Data, &header)
 			require.Nil(t, err)
 			require.Equal(t, StateChanges(res.StateChanges).Hash(), header.StateChangesHash)
 		}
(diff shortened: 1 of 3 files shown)


Gilthoniel:
This PR also includes a fix where a new leader is elected but one of the conode misses the propagation and then continues to assume the old leader. It then refuses to catch up because it assumes the current leader is not legit (basically it's the block 10 bug we found recently).

The catch up was previously protected by the leader check and then we need another way to prevent someone malicious to send forged requests to start multiple catch up. There is now a timestamp that will limit the requests by skipchain and the catch up is started only for *known* skipchains which means that if the genesis block propagation fails, the new conode will learn only if a new block is added.

Gilthoniel:
For the TODO about optimizing the state change storage, I'm working on another PR that will be a following of this one. I noticed some potential issues..

byzcoin/service_test.go
```diff
@@ -2194,102 +2193,74 @@ func TestService_StateChangeStorage(t *testing.T) {
...
 
-	// Remove some entries to check it will recreate them
-	err := s.service().stateChangeStorage.db.Update(func(tx *bbolt.Tx) error {
-		b := s.service().stateChangeStorage.getBucket(tx, s.genesis.SkipChainID())
-		if b == nil {
-			return errors.New("missing bucket")
-		}
+	for i := 0; i < 10; i++ {
+		log.Lvl1("Sleeping for 1 second...")
+		time.Sleep(time.Second)
```
kc1212: is this because we're waiting for the nodes to catch up?

Gilthoniel: Yes for the new one added

PR #1786, opened 2019-03-28
State change storage initialization and cleaning

Gilthoniel:
This fixes some potential issues with the cleaning by size because
an array of sorted timestamp is used and its growth is not limited
and the maintainability depends on the insertion order. Cleaning is
now based on the block index.
It also removes the state changes reconstruction at conode boot by
a calculation of the current size as the state changes are already
built when a block is propagated and persisted in the db.

Follow up of #1784

### DIFF
--- a/byzcoin/struct.go
+++ b/byzcoin/struct.go
@@ -62,49 +62,21 @@ func (sc StateChange) Copy() StateChange {
 	return c
 }
 
-// keyTime stores information to keep track of the age of the
-// oldest version of an instance for cleaning purposes.
-type keyTime struct {
-	key  []byte
-	scid []byte
-	time time.Time
-}
-
-type keyTimeArray []keyTime
-
-// Len returns the number of keys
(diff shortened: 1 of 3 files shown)


byzcoin/struct.go
```diff
@@ -194,77 +120,131 @@ func (s *stateChangeStorage) setMaxNbrBlock(nbr int) {
...
-	defer s.sortedKeysLock.Unlock()
-
 	if s.size < s.maxSize || s.maxSize == 0 {
 		// nothing to clean
 		return nil
 	}
 
-	sortedKeys := make(keyTimeArray, len(s.sortedKeys))
-	copy(sortedKeys, s.sortedKeys)
+	size := s.size
```
kc1212: do you need a mutex lock here?

Gilthoniel: The lock is already held by the append function

PR #1789, opened 2019-04-01
Fix BCAdmin to work for non-admins

ineiti:
The previous version of bcadmin supposed that only the genesis-admin(s)
work on darcs. This PR makes it possible for non-genesis-admins
to create new darcs and to mint their money!

### DIFF
--- a/byzcoin/bcadmin/README.md
+++ b/byzcoin/bcadmin/README.md
@@ -6,7 +6,76 @@ bcadmin
 
 # bcadmin - the CLI to configure ByzCoin ledgers
 
-## Create a new ByzCoin, saving the config
+## Example usage
+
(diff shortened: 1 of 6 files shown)


ineiti:
This is WIP, as the created `bc-xxx.cfg` file is not complete!

jeffallen:
Oops, I just reviewed this and was ready to check it in. I've put it back to WIP.

ineiti:
I'm currently looking at what to do with the `wallet`, `phadmin`, and an eventual new `credadmin` tool. Not sure yet if I will do one or two PRs.

ineiti:
OK, will do (at least) two PRs. Here is a first one to allow `bcadmin` work for a non-genesis-admin. Updated the `README.md` with an example.

Issue #1785, opened 2019-03-27
BLSCoSi warning and fix


ineiti:
As @Gilthoniel wrote in #1780, the BLSCoSi service can be used by a malicious user to lead an attack by getting the service to sign a malicious block, for example.

Currently it has been disabled, but there should be a message in the README.md showing that it's a problem.

To solve it, the BLSCoSi service should only sign the `sha256("bls" | message)`, then nobody can use it as an oracle, as an attacker cannot know the pre-image of the hash...

ineiti:
In fact it seems that there is no need to use a specific signature, as the new version of the cothority has one separate keypair per service. So the `BLSCoSi` service cannot sign a message that will be accepted by the `ByzCoin` service.

Problem solved.

If somebody from DEDIS thinks this is correct, please close this issue.

nkcr:
I let @jeffallen close this issue and PR #1790 if he agrees.

jeffallen:
I agree.

PR #1790, opened 2019-04-02
Prevents the bls service from beeing used as an oracle 

nkcr:


blscosi/blscosi/README.md
```diff
@@ -83,5 +83,7 @@ This will first contact each server individually and then check a few random col
 
 ## Further Information
 
+During the sign and verify operations, each message is prefixed with a constant token. Withtout it, the BLSCoSi service could be used as an oracle, allowing a malicious user to lead an attack by getting the service to sign a malicious block, for example.
```
jeffallen:
```suggestion
During the sign and verify operations, each message is prefixed with a constant token. Without it, the BLSCoSi service could be used as an oracle, allowing a malicious user to lead an attack by getting the service to sign a malicious block, for example.
```

jeffallen: Also, please keep these lines to 80 characters, so that even if someone is not reading them via GitHub's markdown formatter, they are beautiful.

blscosi/protocol/struct.go
```diff
@@ -23,6 +23,12 @@ const DefaultProtocolName = "blsCoSiProtoDefault"
 // main protocol.
 const DefaultSubProtocolName = "blsSubCoSiProtoDefault"
 
+// AddPrefix adds a prefix to msg in order to prevent the use of
+// BLSCoSi service as an oracle.
+func AddPrefix(msg *[]byte) {
```
jeffallen: Use lowercase i.e. `addPrefix` because there's no value in exporting this for use outside of the package.

In general, we default to private because that preserves our ability to evolve the internals of packages without disrupting callers.

blscosi/protocol/protocol.go
```diff
@@ -396,7 +398,9 @@ func (p *BlsCosi) generateSignature(responses ResponseMap) (kyber.Point, *cosi.M
 	}
 
 	// generate personal signature and append to other sigs
-	personalSig, err := bls.Sign(p.suite, p.Private(), p.Msg)
+	msg := p.Msg
+	AddPrefix(&msg)
+	personalSig, err := bls.Sign(p.suite, p.Private(), msg)
```
jeffallen: The spec called for signing `sha256(bls | msg)`. That further reduces the attacker's margin for trying to get a certain thing she wants to be signed. Is there a reason you decide not to do that?

Also: If the existing code you are modifying has historical weirdness in it, like "personal signature", you can change it (comment and variable name). We want that code reads like a simple explanation of what's happening and it is hard to see what "personal" has to do with what's happening here. Perhaps `mySig` or `localSig`?

conode/conode.go
```diff
@@ -209,7 +210,7 @@ func checkConfig(c *cli.Context) error {
 		}
 	}
 
-	return nil
+	return check.CothorityCheck(tomlFileName, c.Bool("detail"))
```
jeffallen: So this doubles up the behaviour of the blscosi check function (which is really kind of example code) with the "collect status from all nodes" behaviour @Gilthoniel put there before.

I think what we want is that the go.dedis.ch/cothority/blscosi service is imported into conode.go. Then if someone wants to use the blscosi client CLI (see blscosi/blscosi/blscosi.go) to run the check command, they can.

You should be sure that the test.sh in blscosi/blscosi still works after you make your changes. This is the integration test, and it could show you that something needed to be changed in the client side to make this change. test.sh is not run by travis, but is run by Jenkins.

Gilthoniel: One of my goal was also to eject the key pair of this service from the public toml file. The question here is: is it worth to have it by default imported for the conodes or can we keep it optional ?

blscosi/protocol/struct.go
```diff
@@ -23,6 +23,12 @@ const DefaultProtocolName = "blsCoSiProtoDefault"
...
+	*msg = append([]byte{'b', 'l', 's'}, *msg...)
```
ineiti:
```suggestion
	*msg = append([]byte("bls"), *msg...)
```

ineiti:
Do I understand correctly that only the leader prepends the `bls` string? Because in that case you could have a malicious leader who sends out a non-prepended message, and the oracle would be back.

So, if it's not already done, every node needs to prepend the `bls` string to the message, and then commit to this message.

nkcr:
> Do I understand correctly that only the leader prepends the `bls` string? Because in that case you could have a malicious leader who sends out a non-prepended message, and the oracle would be back.
> 
> So, if it's not already done, every node needs to prepend the `bls` string to the message, and then commit to this message.

As far as I understand every node prepend the `bls` string, because I implemented it in the `startSubProtocol(...)` and `generateSignature(...)`. Maybe @Gilthoniel can confirm.

blscosi/protocol/struct.go
nkcr: done in a6b6462f5b7b3738cd3402ffeea9922a07af7a25

blscosi/protocol/protocol.go
nkcr: * a6b6462f5b7b3738cd3402ffeea9922a07af7a25 uses now "localSig" instead.
* The "not using sha256" is resolved with @ineiti 's comment on slack: "I think `bls.Sign("domain" | msg)` should be enough then." The reason comes from the `bls.Sign(..)` function that already performs a hashing of the message.

conode/conode.go
nkcr: * 9bff07d47ef88ca700363abf48a092fe8350b192 reverts to not-calling-by-default the check command. The package is imported however for anyone to manually run the check command.
* @Gilthoniel 's observation should be raised as a new issue.

ineiti:
The `BLSCoSi` protocol is in fact a collection of two protocols (unfortunately not very nicely presented here: https://github.com/dedis/student_18/blob/master/omniledger2/report-2018_1-raphael_dunant-omniledger.pdf )

- `BLSCoSi` protocol - which is created by the caller
- `SubProtocol` - which is created for every sub-leader and has the following tree-structure:
_leader - subleader_x - leafs_x_

So it's in fact in the `SubBlsCosi.makeResponse` that the prefix and the message should be merged to avoid that the service gets to be an oracle. Where the message is concatenated now does not protect from a malicious leader who decides to broadcast a non-prefixed message.

So you'll also have to transmit the prefix to the other nodes... Which brings the problem of backward-compatibility... But I let Jeff handle that one ;)

Gilthoniel:
@ineiti the protocol itself is not an oracle as far as the verification function is correctly implemented (i.e. the skipchain verifies that msg matches the hash)

Gilthoniel:
You can't prefix something to the message during the protocol because underlying services expect the message they sent through to be signed (ok yes you could make them learn about the prefix but it sounds like the wrong way to do it).

The protocol is secured through the verification function and it's up to the implementation of the service to not open an oracle, like the **service** (not the **protocol**) was doing. My point here is that the service should be fixed, not the protocol.

ineiti:
@Gilthoniel - I think there is a profound misunderstanding between us what this PR should be doing and what it is trying to protect. The way I understand, it you're preparing a band-aid that will not prevent a malicious node of abusing the blscosi-protocol as an oracle, and he will be able to create for example a valid forward-link to a block that has never been verified by any node.

If you want to know why or convince me I'm wrong, don't hesitate to pass by my office...

ineiti:
Convincing myself I'm wrong: as each service has its own keypair, in fact we don't even need to prepend a hash to the message. Even a rogue conode with a malicious `blscosi` service cannot create a signature that will pass for the `byzcoin` service, as they will be signed by different keypairs.

So in fact there is not even an oracle-attack if we suppose that the `blscosi` service signs with a different keypair than the `byzcoin` service.

So we don't even need to add a prefix to the message, as all messages are only signed with the `blscosi` keypair, and a rogue conode cannot convince correct conodes to sign with the `blscosi` keypair.

Does that make sense?

Gilthoniel:
Yes I know, my goal when I removed the service from production conodes was mainly to remove the key pair from the toml files because the service is not essential. (not saying it's not useful).

For me the prefix is not necessary, I just wanted to point out that if we use wrongly (like using another service key pair) it couldn't have big consequences but as is it's not harmful. In my opinion, the service should be disabled by default and if someone needs it, he can still import it and then use the CLI.

ineiti:
OK - I let you guys decide if the `BLSCoSi` service is useful or not. The initial idea of a conode was that he was available freely for all kinds of services, but perhaps that doesn't hold anymore...

Anyway - seems we all agree that after reflection, this PR is in fact not necessary - correct? So we can close #1785, as I forgot that there is a separate keypair for every service...

jeffallen:
Turns out this was not required. Thank you for your engagement anyway, @nkcr .

PR #1794, opened 2019-04-03
Rogue public key attack

### LINKED ISSUE #1675: rogue public-key attack on BLS
@kc1212 pointed out that the attack described here:

https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html

could be an issue. We are safe for rosters generated from scratch but an attacker could join the cothority after learning about the current public keys. The attack is limited because one bad node can only forge one signature but still.

We need some help from the crypto guys to make sure we're safe.

Gilthoniel:
This PR first adds a service that will act a very simple PKI so that we can ask for proofs to conodes and the service will store them to avoid further requests (see the README).

Then it fixes the rogue public-key attack by making sure the roster of a block is honest. It *doesn't* fix the protocol by itself. There is no way to be backwards compatible and make it secured. This requires a breaking change that should be planned for v4. The robust scheme could also be implemented now so that new services could use it.

Concerning the backwards compatibility, an admin can use the environment variable `ACCEPT_UNVERIFIED_PK` to migrate a cothority step by step. When all of the conode have a version higher enough, it can be turned off to make the protection efficient.

Fixes #1675

### DIFF
--- a/conode/README.md
+++ b/conode/README.md
@@ -54,3 +54,13 @@ For creating a new docker image, there are two commands:
 on your machine.
 * `make docker BUILD_TAG=v3.0.0-pre1` - creates a docker image from source at tag
 BUILD_TAG.
+
+# Version 3
+
+## BLS security for version <= 3.0.2
+
+Prior to version 3.0.3, the BLS signature scheme was open to rogue public-key
+attacks. A verification has been added using Proof of Possession but conodes
+lacking the update won't be able to provide it. In order to migrate a cothority
+incrementally, the environment variable `ACCEPT_UNVERIFIED_PK` can be set and
+then missing proofs will be ignored.
--- a/pki/README.md
+++ b/pki/README.md
@@ -1,0 +1,52 @@
+Navigation: [DEDIS](https://github.com/dedis/doc/tree/master/README.md) ::
+[Cothority](../README.md) :: PKI
+
+# Public Key Infrastructure
+
+This service has the sole purpose of providing an API to get proofs that public
+keys hold by the conodes have a known pair secret key. In the context of the
+BLS CoSi scheme, a rogue public-key could be use to forge a correct signature
+using a single malicious node registered with a very specific public key.
+
+Even if the usage is very specific, it has been implemented as a service open
+to external requests so that it is possible for a client to ask for the Proofs
+of Possession of a given conode. It's not harmful because the API will only
+sign the public key appended to a nonce, thus avoiding to open an oracle.
+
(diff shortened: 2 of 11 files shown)


Gilthoniel:
One question is about the robust BLS signature scheme, should we implement it now so later usages won't be unsecured ?

ineiti:
I'm still not convinced that using a service for that functionality is better than adding a field to the `ServiceIdentity` structure. Setting up and using a service is much heavier, also, calling it `PKI` is not really correct - it's more a `POK` - Proof Of Knowledge.

Adding a field to the `ServiceIdentity` would remove the need of additional calls between the nodes to get the `POK` and would make the code much simpler without losing any of the benefits.

Gilthoniel:
The Proof of Possession/Knowledge can be stored once and forever then the difference is that conodes store them instead of extracting them from the Roster. The request for proofs is then made once (per conode) so the overhead is fine in my opinion. I still insist it's better not to increase the size of a roster object, it's already big enough.

ineiti:
Well here you increase the size of the Forward-links when a new roster arrives. 

I'm more concerned about having a new service for this Proof-of-Knowledge (it is not a `PKI`), and the fact that now for every signature you need to have the `Roster`, the `Signature`, and the `Proofs` if you want to verify it without access to the nodes.

Please have a look at #1797...

Gilthoniel:
1. Forward links have the exact same size
2. Not **every** signature, only the ones with a change in the roster

ineiti:
Also regarding the size of the roster - perhaps the roster stored in the `contractConfig` should be stripped from all `ServerIdentity`s except the `byzcoin` one... But then we couldn't use a byzcoin anymore as a `PKI` for other services (this time a 'real' one ;)

ineiti:
> * Forward links have the exact same size

Sorry - `ForwardSignature`s...

> * Not **every** signature, only the ones with a change in the roster

Yes, if you replace my `ForwardLink` with `ForwardSignature`, that's what I wrote.

Gilthoniel:
Again only when a change in the roster (actually only if a new conode is added) happens, the array won't be empty. Otherwise there is no need for proofs.

ineiti:
OK, here you go:
```
s/Well here you increase the size of the Forward-links when a new roster arrives./
Well here you increase the size of the Forward-signature when a new node arrives./
```

jeffallen:
Closing for the moment because we suspect a better path forward is non-interactive, i.e. a new signing protocol.

Issue #1791, opened 2019-04-02
Use URL from ServerIdentity struct (JS only)


jeffallen:
The task is to use the URL field in the ServerIdentity struct so that a roster can indicate how to contact the server from the client side.

Ideally, we'd do this work on all three client libraries (Go, Java, JavaScript). But to keep this task from being too big, let's deliver this feature on JS first, and then move on to other client libraries.

Tasks:
- [x] get dedis/student_18_explorer app working for dev mode, use the yarn version of `npm link` to work on a development version of cothority TS client.
- [x] fix cothority TS client to respect the URL field (see getWebSocketAddress). Hopefully callers will not notice the change at all (i.e. no corresponding change in the skipchain explorer?)
- [x] figure out how this can be tested in the context of the TypeScript tests -- perhaps it is enough to include one URL in the external/js/cothority/spec/support/public.toml file?
- [x] deploy it, and see it working for real
- [x] evolve one production conode to have URL in the ServerIdentity in it's roster (get help from @jeffallen )

nkcr:
#1800

nkcr:
#1804

PR #1800, opened 2019-04-05
Uses URL from ServerIdentity struct (JS only) issue #1791

nkcr:
Solves issue #1791. 
If an `Url` param is found in the toml file, `getWebSocketAddress()` will use it instead of simply transforming the current server address with a port increment of one.

This PR also adds a test case to check if a docker service is available and changes a strange condition in `spec/support/conondes.ts` that was failing the tests. It also documents how to use `js/cothority` as a local module, which was used in order to check the service against `dedis/student_18_explorer`.

### DIFF
--- a/external/js/cothority/README.md
+++ b/external/js/cothority/README.md
@@ -94,22 +94,9 @@ As this is a [polyfill](https://remysharp.com/2010/10/08/what-is-a-polyfill), pl
 check that what you need is implemented or you will need to use a different approach. Of
 course for NodeJS, you will always get a [Buffer](https://nodejs.org/api/buffer.html).
 
-### Use a developpment version from an external app
+## Use a developpment version from an external app
 
-**Note**: Using the official command `yarn link` doesn't seem to work. We leave
-the steps at the end for informational purpose, but we found out that using symlink
-is simpler, clearer, and actually works.
-
-Build and Link the local `@dedis/cothority` module to a local app:
-
-```bash
-(cothority/external/js/cothority) $ npm run build
-$ ln -s /<cothority root path>/cothority/external/js/cothority/dist /<local app root path>/node_modules/@dedis/cothority
-```
-
-**For information purpose**: The "clean" way to symlink a local package is
-with the `yarn link` command. However it happens to link the root folder
-instead of the `root/dist` one. Hence this is not likely to work:
+Steps to use `js/cothority` as a local module for a local external app:
 
 1) Build a version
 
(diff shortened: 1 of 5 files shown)


external/js/cothority/README.md
```diff
@@ -91,3 +93,45 @@ will be provided thanks to the [buffer package](https://www.npmjs.com/package/bu
...
+
+**Note**: Using the official command `yarn link` doesn't seem to work. We leave
+the steps at the end for informational purpose, but we found out that using symlink
+is simpler, clearer, and actually works.
+
+Build and Link the local `@dedis/cothority` module to a local app:
+
+```bash
+(cothority/external/js/cothority) $ npm run build
+$ ln -s /<cothority root path>/cothority/external/js/cothority/dist /<local app root path>/node_modules/@dedis/cothority
```
Gilthoniel: It seems easier to me to use `npm link @dedis/cothority` inside the local app root

external/js/cothority/README.md
```diff
@@ -91,3 +93,45 @@ will be provided thanks to the [buffer package](https://www.npmjs.com/package/bu
 As this is a [polyfill](https://remysharp.com/2010/10/08/what-is-a-polyfill), please
 check that what you need is implemented or you will need to use a different approach. Of
 course for NodeJS, you will always get a [Buffer](https://nodejs.org/api/buffer.html).
+
+### Use a developpment version from an external app
```
Gilthoniel: `##` as it is not related to the Protobuf generation chapter

external/js/cothority/README.md
```diff
@@ -91,3 +93,45 @@ will be provided thanks to the [buffer package](https://www.npmjs.com/package/bu
...
+3) From the root folder of the external app, link the new package
+
+```bash
+(external_app) $ yarn link "@dedis/cothority"
+```
+
+4) If needed, install the new linked pachage
+
+```bash
+(external_app) $ npm install --save @dedis/cothority
```
Gilthoniel: why using npm in that case ? I guess you mean `yarn add --save` ?

Gilthoniel: Also I'm not sure we should encourage using yarn as cothority and kyber are using npm. Or we should switch to yarn everywhere but it seems confusing and frustrating to mix the usage of both.

nkcr: I totally agree! Yarn is used in the student_18_explorer project while cothority uses npm. Thats something we should agree on some guidelines.

external/js/cothority/README.md
nkcr: That doesn't work in my case. It links the `cothority` folder instead of the `cothority/dist` one.

Gilthoniel: I mean you run `npm run link` in the cothority library (the script will link the dist folder) and you navigate to the external project and use `npm link @dedis/cothority`.

external/js/cothority/README.md
```diff
@@ -91,3 +93,32 @@ will be provided thanks to the [buffer package](https://www.npmjs.com/package/bu
 As this is a [polyfill](https://remysharp.com/2010/10/08/what-is-a-polyfill), please
 check that what you need is implemented or you will need to use a different approach. Of
 course for NodeJS, you will always get a [Buffer](https://nodejs.org/api/buffer.html).
+
+## Use a developpment version from an external app
```
jeffallen:
```suggestion
## Use a development version from an external app
```

external/js/cothority/src/network/proto.ts
```diff
@@ -213,11 +214,16 @@ export class ServerIdentity extends Message<ServerIdentity> {
     }
 
     /**
-     * Convert the address of the server to match the websocket format
+     * Returns this.url if wset, otherwise converts the server
```
jeffallen:
```suggestion
     * Returns this.url if set, otherwise converts the server
```

PR #1802, opened 2019-04-09
Adding calypso_v4 API

ineiti:
This is a proposed API to support non-byzcoin verified calypso deployments. The first use-case is ByzGen's need to have a HyperLedger/Fabric backend for the authorization, but Calypso for the re-encryption.


ineiti:
Closed in favour of #1803

PR #1804, opened 2019-04-09
fix #1791, work in progress

### LINKED ISSUE #1791: Use URL from ServerIdentity struct (JS only)

nkcr:
This PR closes #1791


external/js/cothority/src/network/proto.ts
```diff
@@ -178,6 +178,42 @@ export class ServerIdentity extends Message<ServerIdentity> {
...
+            case "http:": {
+                urlParser.protocol = "ws:";
+                break;
+            }
+            case "https:": {
+                urlParser.protocol = "wss:";
+                break;
+            }
+            default : {
+                throw new Error("The url field should use either 'http:' or 'https:', but We found "
```
jeffallen:
```suggestion
                throw new Error("The url field should use either 'http:' or 'https:', but we found "
```

external/proto/network.proto
```diff
@@ -14,7 +14,9 @@ message ServerIdentity {
     required bytes id = 3;
     required string address = 4;
     required string description = 5;
-    optional string url = 6;
+	// Tag number 6 is used by go.dedis.ch/protobuf for the private
```
jeffallen:
```suggestion
    // Tag number 6 is used by go.dedis.ch/protobuf for the private
```

external/proto/network.proto
```diff
@@ -14,7 +14,9 @@ message ServerIdentity {
...
+	// key field, which is not exported.
```
jeffallen:
```suggestion
    // key field, which is not exported.
```

nkcr:
This PR closes #1791

conode/Operating.md
```diff
@@ -11,19 +11,77 @@ line examples, please refer to:
...
+```
+
 ## Reverse proxy
 
 Conode should only be run as a non-root user.
 
 The current recommended way to add HTTPS to the websocket port is to use a web
 server like Apache or nginx in reverse proxy mode to forward connections from
-port 443 to port 6880, the default websocket port.
+port 443 the websocket port, which is the conode's port plus 1.
```
Gilthoniel: `port 443 to the websocket port` ?

kc1212:
Please update the PR description to say which issue this PR closes. The PR description goes into the commit logs but not the PR comments.

PR #1807, opened 2019-04-11
Improve docs about Proof

### LINKED ISSUE #1769: Add Proof to Byzcoin-SVG
As the proofs are something you don't find in other blockchains, we should make them better visible in the Byzcoin-SVG overview.

jeffallen:
Fixes #1769.


jeffallen:
Maybe you can fix those the next time you make a cothority commit. :)

PR #1809, opened 2019-04-12
Robust BLSCoSi protocol

Gilthoniel:
This adds a new protocol in the blscosi package that is using the BDN signature scheme which is the robust version of the BLS scheme. It only implements but a later PR will switch the new skipchains to it.


Gilthoniel:
Next step is to get a working implementation of the signature scheme for the Java and Javascript libraries

blscosi/protocol/sub_protocol.go
```diff
@@ -239,15 +248,16 @@ func (p *SubBlsCosi) dispatchSubLeader() error {
 
 	responses := make(ResponseMap)
 	for _, c := range p.Children() {
-		public := p.NodePublic(c.ServerIdentity)
+		_, index := searchPublicKey(p.TreeNodeInstance, c.ServerIdentity)
```
jeffallen: If key is not found, index will be -1 and line 253 will panic. (Same for most other uses of searchPublicKey.)

blscosi/bdnproto/protocol.go
```diff
@@ -0,0 +1,82 @@
...
+	GlobalRegisterBdnProtocols()
+}
+
+// GlobalRegisterBdnProtocols registers both protocol to the global register
+func GlobalRegisterBdnProtocols() {
+	onet.GlobalProtocolRegister(BdnProtocolName, NewBdnProtocol)
+	onet.GlobalProtocolRegister(BdnSubProtocolName, NewSubBdnProtocol)
+}
+
+// NewBdnProtocol is used to register the protocol with an always-true verification
```
jeffallen: I'm unclear what this API is for; it seems like we know we'll only be using this in the context where we use a real verifier, and anyway a caller who wants "always verify" will be able to easily provider the no-op vf.

blscosi/bdnproto/protocol.go
```diff
@@ -0,0 +1,82 @@
...
+func NewBdnProtocol(n *onet.TreeNodeInstance) (onet.ProtocolInstance, error) {
+	vf := func(a, b []byte) bool { return true }
+	return NewBdnCosi(n, vf, BdnSubProtocolName, pairing.NewSuiteBn256())
+}
+
+// NewBdnCosi makes a protocol instance for the modified BLS CoSi protocol
```
jeffallen: Please make doc strings complete sentences.

blscosi/bdnproto/protocol_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	err = cosiProtocol.Start()
+	if err != nil {
+		return err
+	}
+
+	select {
+	case sig := <-cosiProtocol.FinalSignature:
+		pubs := roster.ServicePublics(testServiceName)
```
jeffallen: Please verify that `BdnSignature(sig).Verify(testSuite, theWrongMessage, pubs)` returns an error. You might need to pass the *testing.T into runProtocol to call `require.Error(t, ...)`.

blscosi/protocol/struct.go
```diff
@@ -28,15 +29,15 @@ func init() {
...
 
 // ResponseMap is the container used to store responses coming from the children
-type ResponseMap map[string]*Response
+type ResponseMap map[int]*Response
 
 // BlsSignature contains the message and its aggregated signature
 type BlsSignature []byte
 
 // GetMask creates and returns the mask associated with the signature. If
 // no mask has been appended, mask with every bit enabled is assumed
```
jeffallen:
```suggestion
// no mask has been appended, a mask with every bit enabled is returned.
```

blscosi/bdnproto/struct.go
```diff
@@ -0,0 +1,67 @@
...
+	"go.dedis.ch/kyber/v3/sign"
+	"go.dedis.ch/kyber/v3/sign/bdn"
+	"go.dedis.ch/kyber/v3/sign/cosi"
+)
+
+// BdnSignature is a signature that must be verified using coefficients
+// derived from the public keys
+type BdnSignature []byte
+
+// GetMask returns the bytes representing the mask
```
jeffallen: GetMask does not return bytes. Maybe just copy and paste the other GetMask doc string, which I also corrected slightly (see below).

blscosi/bdnproto/struct.go
```diff
@@ -0,0 +1,67 @@
...
+func (sig BdnSignature) GetMask(suite pairing.Suite, pubkeys []kyber.Point) (*sign.Mask, error) {
+	return protocol.BlsSignature(sig).GetMask(suite, pubkeys)
+}
+
+// Point returns the point associated with the signature
+func (sig BdnSignature) Point(suite pairing.Suite) (kyber.Point, error) {
+	return protocol.BlsSignature(sig).Point(suite)
+}
+
+// Verify returns an error if the signature can't be verified or nil if it matches
```
jeffallen: Can you mention in the doc string what the threshold policy will be for the verification, please?

blscosi/bdnproto/struct.go
```diff
@@ -0,0 +1,67 @@
...
+func (sig BdnSignature) Verify(suite pairing.Suite, msg []byte, pubkeys []kyber.Point) error {
+	policy := cosi.NewThresholdPolicy(protocol.DefaultThreshold(len(pubkeys)))
+
+	return sig.VerifyWithPolicy(suite, msg, pubkeys, policy)
+}
+
+// VerifyWithPolicy checks that the signature is correct and that the number of signers
+// matches the policy
+func (sig BdnSignature) VerifyWithPolicy(suite pairing.Suite, msg []byte, pubkeys []kyber.Point, policy cosi.Policy) error {
```
jeffallen: I wonder if cosi.Policy still makes sense here, or if that type should be someplace else. Is this our only dependency on package cosi now? https://go-proverbs.github.io -> "A little copying is better than a little dependency."

blscosi/bdnproto/struct_test.go
```diff
@@ -0,0 +1,45 @@
...
+	asig, err := bdn.AggregateSignatures(suite, [][]byte{sig1, sig2}, mask)
+	require.NoError(t, err)
+
+	buf, err := asig.MarshalBinary()
+	require.NoError(t, err)
+
+	sig := BdnSignature(append(buf, mask.Mask()...))
+	pubkeys := []kyber.Point{pk1, pk2, pk3}
+
+	require.Error(t, sig.Verify(suite, msg, pubkeys))
```
jeffallen: A comment explaining why this one expects an error, please.

blscosi/bdnproto/struct_test.go
```diff
@@ -0,0 +1,45 @@
...
+	policy := cosi.NewThresholdPolicy(2)
+	require.NoError(t, sig.VerifyWithPolicy(suite, msg, pubkeys, policy))
+	require.Error(t, sig.VerifyWithPolicy(suite, []byte{}, pubkeys, policy))
```
jeffallen: Ah, this was the test I was talking about in runProtocol. So maybe no need there. For self-documenting code, please use `wrongMsg := []byte("cba")` here instead of []byte{}, which doesn't say what it is testing. Thanks.

blscosi/protocol/sub_protocol.go
Gilthoniel: I should have verified but note that this is a map and then `-1` is a valid key. I changed into integers because I need an ordering for the coefficients but it is still a map to filter messages and take only the ones from known distant peers.

Gilthoniel: Also if the public key is returned, the index will always be positive so other uses are safe with the nil check.

blscosi/bdnproto/protocol.go
Gilthoniel: It seems that this is the historical way to provide the protocol with an empty verification function, mainly for the tests. I think it's safe to say this is only used for testing so the registration could be effective only for the tests.

Gilthoniel: This is true for all CoSi-related packages

blscosi/bdnproto/struct.go
Gilthoniel: I agree it could/should be in the same package as the mask. Now if we move entirely, changes will be required for previous cosi packages. I guess we can do what you suggest a copy it to the new location with a deprecation notice ?

PR #1810, opened 2019-04-12
Fixes #1756 (store new skipblocks when skipchain service is linked)

### LINKED ISSUE #1756: byzcoin cannot store new skipblocks when skipchain service is linked
Once you have run `scmgr link add private.toml`, further calls to StoreSkipblock will fail, even if they are coming internally (i.e. from sister service Byzcoin on the same conode).

The workaround is to link to the conode while you are doing what you need to do, then unlink again (`scmgr link del $host:$port`).

We need to find a more workable solution. One possibility would be to have a setter on type StoreSkipblock, making it possible to set a private field named `internal`. Callers on the network would never be able to set that, but sister services would be able to.

A more general solution would be for onet to be able to signal somehow, that requests are originating locally instead of over the network. This would apply to all services, and then Skipchain could use this to decide to allow unsigned block writes from internal services.

nkcr:
This PR implements a new function `StoreSkipBlockInternal` which is meant to be used in the case a service is used on the same host as the Skipchain service.~~I also removed a condition in `skipchain/skipchain.go` as discussed.~~

~~Note: I had to update 2 tests in `byzcoin/bcadmin/test.sh` line 95 and 98 that were supposed to pass. This is due to the removal of the condition. I am not 100% confident about the expected behavior now that I removed the condition, so please double check.~~ (Reverted it)

Fixes #1756

### DIFF
--- a/byzcoin/bcadmin/test.sh
+++ b/byzcoin/bcadmin/test.sh
@@ -97,13 +97,15 @@ testRoster(){
   testGrep "Roster: tls://localhost:2006" runBA latest -server 2 $bc
 }
 
-# When a conode is linked to a client (`scmgr link add ...`), it removes the possibility
-# for 3rd parties to create a new skipchain on that conode. In the case a Bizcoin service
-# hosted on a linked conode wants to adds a new skipchain, we have to bypass this 
-# authorization process and allow a local service be able to send requests on the same
-# local linked conode. This process is handled with the `StoreSkipBlockInternal` method,
-# and this is what this method checks.
-# Note: this methods relies on the `scmgr` and the ability to create/update Byzcoin.
+# When a conode is linked to a client (`scmgr link add ...`), it removes the
+# possibility for 3rd parties to create a new skipchain on that conode. In the
+# case a Bizcoin service hosted on a linked conode wants to adds a new
+# skipchain, we have to bypass this authorization process and allow a local
+# service be able to send requests on the same local linked conode. This process
+# is handled with the `StoreSkipBlockInternal` method, and this is what this
+# method checks. 
+# Note: this methods relies on the `scmgr` and the ability to create/update 
+#       Byzcoin.
 testLinkPermission() {
   rm -f config/*
   runCoBG 1 2 3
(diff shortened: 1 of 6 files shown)


byzcoin/bcadmin/test.sh
```diff
@@ -66,23 +77,55 @@ testRoster(){
...
+  [ -z "$BC" ] && exit 1
+  bc=config/bc*cfg
+  key=config/key*cfg
+  testOK runBA latest $bc
+  if [ ! -x ../../../scmgr/scmgr ]; then
+    echo "Didn't find \"cothority/scmgr/smgr\" executable"
+    echo "You may want to run \"go build\" in that directory"
+    exit 1
+  fi
+  scmgr link add co1/private.toml
```
jeffallen: Look at runBA, it is doing something like "./bcadmin", because bcadmin was built by libtest.sh. You need to do the same for scmgr.

See eventlog/el/test.sh for an example of an integration test that needs to build a second binary to use.

byzcoin/bcadmin/test.sh
```diff
@@ -66,23 +77,55 @@ testRoster(){
...
   testGrep "Roster: tls://localhost:2006" runBA latest -server 2 $bc
 }
 
+# When a conode is linked to a client (`scmgr link add ...`), it removes the possibility
+# for 3rd parties to create a new skipchain on that conode. In the case a Bizcoin service
+# hosted on a linked conode wants to adds a new skipchain, we have to bypass this 
+# authorization process and allow a local service be able to send requests on the same
+# local linked conode. This process is handled with the `StoreSkipBlockInternal` method,
+# and this is what this method checks.
+# Note: this methods relies on the `scmgr` and the ability to create/update Byzcoin.
```
jeffallen:
```suggestion
# Note: this test relies on the `scmgr` and the ability to create/update Byzcoin.
```

libtest.sh
```diff
@@ -353,6 +353,7 @@ setupConode(){
   DBG_TEST=$DBG_OLD
 }
 
+# A few words about what this function does wouldn't hurt
```
jeffallen:
```
# runCoBG: Run a conode in the background. It runs a conode under a subshell
# so that when it exits, it can make the .dead file once the conode dies. It then checks
# that they started listening on the expected port, and finally reports if one did not
# start as expected.
```

PR #1814, opened 2019-04-12
Fix authprox integration test

Gilthoniel:
Fix the usage of bcadmin in one of the integration test of authprox to reflect the recent changes

### DIFF
--- a/authprox/apadmin/test.sh
+++ b/authprox/apadmin/test.sh
@@ -34,7 +34,7 @@ BCSetup(){
 	eval $SED
 	[ -z "$BC" ] && exit 1
 
-	testOK ./bcadmin add spawn:authproxAdd -identity $ID
+	testOK ./bcadmin darc rule --rule spawn:authproxAdd --identity $ID
 }
 
 testAdd(){


authprox/apadmin/test.sh
```diff
@@ -34,7 +34,7 @@ BCSetup(){
 	eval $SED
 	[ -z "$BC" ] && exit 1
 
-	testOK ./bcadmin add spawn:authproxAdd -identity $ID
+	testOK ./bcadmin darc rule --rule spawn:authproxAdd --identity $ID
```
kc1212: does that work? it seems the other tests use `-rule` and `-identity` (one minus sign)

Gilthoniel: both forms seem to work

Gilthoniel: but `bcadmin darc rule -help` is saying double minus, that's why I went for it

Gilthoniel: From the cli doc:
```
If you enable the *UseShortOptionHandling, then you must not use any flags that have a single leading - or this will result in failures. For example, -option can no longer be used. Flags with two leading dashes (such as --options) are still valid.
```

Issue #1813, opened 2019-04-12
Prevent catching up when one block away


Gilthoniel:
```
1 : (                   byzcoin.waitInclusion: 653) - Create correct transaction and wait
2 : (       byzcoin.(*Service).AddTransaction: 311) - Instruction[0]: spawn:dummy
1 : (        byzcoin.(*Service).catchupFromID: 1081) - local://127.0.0.1:2010: catching up with chain 4dea6fd988f9ae66540ffe38bdcd0acf55560167d9bdf8fa17453e063b029e41
1 : (        byzcoin.(*Service).catchupFromID: 1081) - local://127.0.0.1:2030: catching up with chain 4dea6fd988f9ae66540ffe38bdcd0acf55560167d9bdf8fa17453e063b029e41
1 : (        byzcoin.(*Service).catchupFromID: 1081) - local://127.0.0.1:2020: catching up with chain 4dea6fd988f9ae66540ffe38bdcd0acf55560167d9bdf8fa17453e063b029e41
2 : (                  v3.wsHandler.ServeHTTP: 199) - ws request from 127.0.0.1:57284: Skipchain/GetSingleBlock
E : (                  v3.wsHandler.ServeHTTP: 236) - Got an error while executing Skipchain/GetSingleBlock: No such block
2 : (            v3.wsHandler.ServeHTTP.func1: 163) - ws close 127.0.0.1:57284 n 1 rx 34 tx 0
2 : (                  v3.wsHandler.ServeHTTP: 199) - ws request from 127.0.0.1:40042: Skipchain/GetSingleBlock
2 : (                  v3.wsHandler.ServeHTTP: 199) - ws request from 127.0.0.1:57288: Skipchain/GetSingleBlock
E : (                  v3.wsHandler.ServeHTTP: 236) - Got an error while executing Skipchain/GetSingleBlock: No such block
2 : (            v3.wsHandler.ServeHTTP.func1: 163) - ws close 127.0.0.1:57288 n 1 rx 34 tx 0
2 : (                  v3.wsHandler.ServeHTTP: 199) - ws request from 127.0.0.1:40046: Skipchain/GetSingleBlock
2 : (                  v3.wsHandler.ServeHTTP: 199) - ws request from 127.0.0.1:40048: Skipchain/GetSingleBlock
2 : (            v3.wsHandler.ServeHTTP.func1: 163) - ws close 127.0.0.1:40042 n 1 rx 34 tx 2893
[...]

PR #1815, opened 2019-04-12
Fix jenkins failures

Gilthoniel:
This PR includes a bunch of fixes for Jenkins where tests are failing.


Gilthoniel:
I will fix #1813 after this PR and I will try to use a global test interval smaller than 500 ms. That should help with the running time and it's actually helpful to test with small intervals, some edge cases can appear like that..

PR #1820, opened 2019-04-26
add error handling to bcadmin and extend 'key' and 'dump' command

ineiti:
byzcoin/bcadmin was missing a lot of error handling, which has been updated.
The 'key' command got '--print' to show the private part of the key.
The 'dump' command got '--verbose' to print some details of the printed instances.

### DIFF
--- a/byzcoin/bcadmin/main.go
+++ b/byzcoin/bcadmin/main.go
@@ -36,7 +36,7 @@ import (
 	"go.dedis.ch/onet/v3/log"
 	"go.dedis.ch/onet/v3/network"
 	"go.dedis.ch/protobuf"
-	cli "gopkg.in/urfave/cli.v1"
+	"gopkg.in/urfave/cli.v1"
 )
 
 func init() {
@@ -70,12 +70,12 @@ var cmds = cli.Commands{
 		ArgsUsage: "roster.toml [bcid]",
 		Flags: []cli.Flag{
 			cli.StringFlag{
-				Name:  "darc, ad",
-				Usage: "a darc that has 'evolve_unrestricted'",
+				Name:  "admindarc, ad",
+				Usage: "the admin darc that has 'evolve_unrestricted'",
 			},
 			cli.StringFlag{
-				Name:  "pub, ap",
-				Usage: "the public key with rights to sign on the darc",
+				Name:  "adminpub, ap",
+				Usage: "the public key of the admin to use",
 			},
 		},
 		Action: link,
@@ -117,8 +117,14 @@ var cmds = cli.Commands{
 				ArgsUsage: "ip:port",
 			},
 			{
-				Name:      "dump",
-				Usage:     "dumps a given byzcoin instance",
+				Name:  "dump",
+				Usage: "dumps a given byzcoin instance",
+				Flags: []cli.Flag{
+					cli.BoolFlag{
+						Name:  "verbose, v",
+						Usage: "print more information of the instances",
+					},
+				},
 				Action:    debugDump,
 				ArgsUsage: "ip:port byzcoin-id",
 			},
@@ -190,6 +196,10 @@ var cmds = cli.Commands{
 				Name:  "save",
 				Usage: "file in which the user wants to save the public key instead of printing it",
 			},
+			cli.StringFlag{
+				Name:  "print",
+				Usage: "print the private and public key",
+			},
 		},
 		Action: key,
 	},
@@ -211,7 +221,7 @@ var cmds = cli.Commands{
 					},
 					cli.StringFlag{
 						Name:  "darc",
-						Usage: "the darc to show (the admin darc from the BC config)",
+						Usage: "the darc to show (no default)",
 					},
 				},
 			},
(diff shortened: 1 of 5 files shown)


byzcoin/bcadmin/main.go
```diff
@@ -477,20 +493,24 @@ func link(c *cli.Context) error {
 		}
 		ad := &darc.Darc{}
 		adPub := cothority.Suite.Point()
-		if adIDStr := c.String("darc"); len(adIDStr) == 69 {
-			adID, err := hex.DecodeString(adIDStr[5:])
+		// Accept both plain-darcs, as well as "darc:...." darcs
```
jeffallen: getDarcByString is already doing "strings.HasPrefix("darc:")" and then trimming it. Is there any way we can keep this in one place?

byzcoin/bcadmin/main.go
```diff
@@ -477,20 +493,24 @@ func link(c *cli.Context) error {
...
-			if pubStr := c.String("pub"); len(pubStr) == 72 {
-				adPubBuf, err = hex.DecodeString(pubStr[8:])
-				if len(adPubBuf) != 32 || err != nil {
-					return errors.New("please give valid admin public key in hex: " + err.Error())
-				}
-			} else {
-				return errors.New("please give a key in the following format: ed25519:public_key_in_hex")
+			adPubStr := c.String("adminpub")
+			// Accept both plain-public keys and "ed25519:..." keys
+			if (len(adPubStr) != 64 && len(adPubStr) != 72) || err != nil {
```
jeffallen: Am really unconvinced on this length-based hacking around. If we are going to hack around on the input, strings.HasPrefix is much more self-documenting.

byzcoin/bcadmin/main.go
```diff
@@ -399,8 +409,14 @@ func create(c *cli.Context) error {
 		return err
 	}
 
-	fmt.Fprintf(c.App.Writer, "Created ByzCoin with ID %x.\n", cfg.ByzCoinID)
-	fmt.Fprintf(c.App.Writer, "export BC=\"%v\"\n", fn)
+	_, err = fmt.Fprintf(c.App.Writer, "Created ByzCoin with ID %x.\n", cfg.ByzCoinID)
+	if err != nil {
+		return err
```
jeffallen: What problem are you trying to solve here? Yes, we know that Printf can fail, but was it really failing? It just seems a little excessive, and wondering how you decided these needed to be checked...

I'm not saying to remove these checks, just curious...

byzcoin/proto.go
```diff
@@ -57,7 +57,7 @@ type CreateGenesisBlock struct {
 	Version Version
 	// Roster defines which nodes participate in the skipchain.
 	Roster onet.Roster
-	// GenesisDarc defines who is allowed to write to this skipchain.
+	// AdminDarc defines who is allowed to write to this skipchain.
```
jeffallen: What?

byzcoin/service.go
```diff
@@ -1875,6 +1875,7 @@ func (s *Service) processOneTx(sst *stagingStateTrie, tx ClientTransaction) (Sta
 				}
 				return nil, nil, fmt.Errorf("%s: contract %s %s", s.ServerIdentity(), contractID, reason)
 			}
+			log.Lvlf2("StateChange %s for id %x - contract: %s", sc.StateAction, sc.InstanceID, sc.ContractID)
```
jeffallen: Are you sure level 2 is correct here? Seems like a 3 thing?

byzcoin/bcadmin/main.go
ineiti: Just following IntelliJ's yellow lines...

byzcoin/proto.go
ineiti: When I wanted to connect to the DEDIS-ByzCoin, you created a DARC for me, which is now _my_ admin darc. So the darc stored in the config is not necessarily the genesis-darc, but only any admin darc that is allowed to do things in this ByzCoin instance.

byzcoin/service.go
ineiti: I would say that until we have a lot of transactions, it's really a useful debugging output. If we start having multi-transaction blocks regularly, we can change it to `Lvl3`. What do you think?

byzcoin/bcadmin/main.go
ineiti: OK, split `getDarcByString` to a string-to-id and a `getDarc` part. Also added `stringToEd25519Buf`.

byzcoin/proto.go
jeffallen: Right, but the field is still called GenesisDarc, so the documentation string needs to start with, at least, a capital G...

byzcoin/service.go
jeffallen: ok, makes sense.

byzcoin/proto.go
ineiti: Thanks - now I got it. Yes, it's IntelliJ over-zealously replacing `GenesisDarc` with `AdminDarc`, and me not reading the code on Friday afternoon...

PR #1821, opened 2019-04-26
Add error handling to Calypso, export contracts, and add random-source

ineiti:
Added error handling to the calypso-service to better debug eventual errors.
Export calypso-read and calypso-write contracts for future changes in personhood-spawner.
Added a cost-field to Calypso.ContractWrite, to have paying read-requests.

### DIFF
--- a/calypso/contracts.go
+++ b/calypso/contracts.go
@@ -15,13 +15,14 @@ import (
 // ContractWriteID references a write contract system-wide.
 const ContractWriteID = "calypsoWrite"
 
-type contractWr struct {
+// ContractWrite represents one calypso write instance.
+type ContractWrite struct {
 	byzcoin.BasicContract
 	Write
 }
 
 func contractWriteFromBytes(in []byte) (byzcoin.Contract, error) {
-	c := &contractWr{}
+	c := &ContractWrite{}
 
 	err := protobuf.DecodeWithConstructors(in, &c.Write, network.DefaultConstructors(cothority.Suite))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


calypso/contracts.go
```diff
@@ -51,6 +55,9 @@ func (c *contractWr) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instructi
 			err = errors.New("couldn't unmarshal write: " + err.Error())
 			return
 		}
+		if d := inst.Spawn.Args.Search("darcID"); d != nil {
```
kc1212: Could you explain what this part is used for? Does it means the client who spawns the contract can choose any darc in byzcoin to be the darc that protects the write request? I'm curious whether this approach will have security implications (probably not coz the one that spawns the write request has full responsibility of who gets to read it). Also, if this is the case, can we have a test for it?

calypso/contracts.go
```diff
@@ -68,12 +75,20 @@ func (c *contractWr) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instructi
...
 		}
-		_, _, wc, _, err := rst.GetValues(rd.Write.Slice())
-		if err != nil {
-			return nil, nil, errors.New("referenced write-id is not correct: " + err.Error())
+		if !rd.Write.Equal(inst.InstanceID) {
+			return nil, nil, errors.New("the read request doesn't reference this write-instance")
 		}
-		if wc != ContractWriteID {
-			return nil, nil, errors.New("referenced write-id is not a write instance, got " + wc)
+		if c.Cost.Value > 0 {
```
kc1212: I'd also like to see a test for this feature.

calypso/contracts.go
ineiti: Yes, this means that the client who spawns a write request can define which darc will protect the read requests. It is used when spawning a WriteInstance from a SpawnerInstance, or if you want to create a WriteInstance from a Darc, but want to give another darc for the control.

For security implications, you can 'hijack' a darc and its access patterns, but besides giving other people access to your WriteInstance, there is nothing more I can think of.

With regard to tests, this is currently only tested in TypeScript, but not in go.

calypso/contracts.go
ineiti: There is a test in TypeScript, but not in go... Should I add tests for this and the `darcID` in go, too?

kc1212: Is the TS test going to land in cothority or it's in another repo? If it's only in another repo then I think it's best to have some tests in cothority so that we know we're not breaking anything in the future.

calypso/contracts.go
kc1212: Sounds good. See my response regarding tests in the other comment.

calypso/contracts.go
ineiti: Once my two PRs are in (1820 and 1821), I'll send a PR for the cothority-TS, which is currently in my dynasent-repo.

PR #1822, opened 2019-04-26
adds marshalSize() to the scalar-interface

ineiti:
The scalar interface also should have `marshalSize`, as it's used in some methods, and it's provided at least by the `Ed25519` group.

### DIFF
--- a/external/js/kyber/src/index.ts
+++ b/external/js/kyber/src/index.ts
@@ -165,6 +165,12 @@ export interface Scalar {
   unmarshalBinary(bytes: Buffer): void;
 
   /**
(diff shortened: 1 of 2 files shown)


external/js/kyber/src/index.ts
```diff
@@ -164,6 +164,12 @@ export interface Scalar {
    */
   unmarshalBinary(bytes: Buffer): void;
 
+  /**
+   * Get the length of the buffer after marshaling the scalar
+   * @returns the length as a number
+   */
+  marshalSize(): number;
```
Gilthoniel: You need to implement the function for the BN256Scalar

PR #1826, opened 2019-04-26
BDN signature scheme in JS/Java

Gilthoniel:
This PR adds the support for BDN signatures in the Javascript and the Java libraries. The support for the BLAKE2xs XOF has been copied from the bouncycastle PR to be removed later on when it is merged.


Gilthoniel:
Note that the BLAKE2(x)s files don't need to be reviewed as they will be by the bouncycastle team. I only copied them to move forward on this.

kc1212:
> Note that the BLAKE2(x)s files don't need to be reviewed as they will be by the bouncycastle team. I only copied them to move forward on this.

Nice. Do you know how bouncycastle contributions work? What do we need to do to increase our chances of getting your bouncycastle PR merged? I tried looking on their homepage and repo but didn't find any useful information.

external/java/src/test/java/ch/epfl/dedis/lib/crypto/BdnSigTest.java
```diff
@@ -0,0 +1,80 @@
...
+        Bn256G2Point p1 = new Bn256G2Point(BigInteger.ONE);
+        Bn256G2Point p2 = new Bn256G2Point(BigInteger.valueOf(2));
+        Bn256G2Point p3 = new Bn256G2Point(BigInteger.valueOf(3));
+        List<Point> pubs = Arrays.asList(p1, p2, p3);
+
+        List<Scalar> coefs = BdnSig.hashPointToR(pubs);
+
+        assertEquals("35b5b395f58aba3b192fb7e1e5f2abd3", Hex.printHexBinary(coefs.get(0).toBytes()).toLowerCase());
+        assertEquals("14dcc79d46b09b93075266e47cd4b19e", Hex.printHexBinary(coefs.get(1).toBytes()).toLowerCase());
+        assertEquals("933f6013eb3f654f9489d6d45ad04eaf", Hex.printHexBinary(coefs.get(2).toBytes()).toLowerCase());
```
kc1212: are these from a reference implementation or is it from kyber?

Gilthoniel:
> > Note that the BLAKE2(x)s files don't need to be reviewed as they will be by the bouncycastle team. I only copied them to move forward on this.
> 
> Nice. Do you know how bouncycastle contributions work? What do we need to do to increase our chances of getting your bouncycastle PR merged? I tried looking on their homepage and repo but didn't find any useful information.

Same as you, I didn't find really clear instructions how to contribute so I looked a bit at the previous merged PR and I tried to follow as much as possible their coding style/design.

external/java/src/test/java/ch/epfl/dedis/lib/crypto/BdnSigTest.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Test
+    void testAggregateVerificationVectors() throws Exception {
+        byte[][] vectors = new byte[][]{
+                new byte[]{0b1},
+                new byte[]{0b10},
+                new byte[]{0b101},
+                new byte[]{0b11},
+                new byte[]{0b111}
```
kc1212: if the mask is `000`, does the verification fail or succeed? which is the right behaviour?

kc1212:
if your blake2s goes into bouncycastle, maybe we can also also move the bn256 (not bls) code to bouncycastle as a part of our contribution to the public

external/java/src/test/java/ch/epfl/dedis/lib/crypto/BdnSigTest.java
Gilthoniel: From Kyber, there is a test that generates those values specifically for that purpose.

Gilthoniel: There is no real reference implementation for this, they're quite obscure on the hash function to use.

external/java/src/test/java/ch/epfl/dedis/lib/crypto/BdnSigTest.java
Gilthoniel: In that case the aggregate is null and then you can't create a signature.

Gilthoniel:
@kc1212 The PR has just been accepted so we might talk about your suggestion

PR #1827, opened 2019-04-29
New skipchains use the BDN signature scheme

### LINKED ISSUE #1675: rogue public-key attack on BLS

Gilthoniel:
This changes the default cosi protocol to BDN so that new skipchains will be robust aginst rogue public-key attacks.

Important Notice: any skipchain created after this commit won't be accepted by older versions as they are not aware of the new signature scheme but existing skipchains will continue to operate normally.

Fixes #1675


Gilthoniel:
Waiting on #1826 to make Java and JS tests pass

Gilthoniel:
If you wonder why I moved the the hash function and related from `SkipBlockFix` to `SkipBlock`, I was forced to do so because you can't add anymore fields in `SkipBlockFix` without breaking the backward compatibility because even if you can specify the index, it must be incremental according to the position of the field, then you can't add a new field without changing the indices of the `SkipBlock` fields.

skipchain/skipchain.go
```diff
@@ -211,6 +212,7 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
 		random.Bytes(bl[:], random.New())
 		prop.BackLinkIDs = []SkipBlockID{SkipBlockID(bl[:])}
 		prop.GenesisID = nil
+		prop.SignatureScheme = BdnSignatureSchemeIndex
```
jeffallen: A comment explaining the policy (starting with release v3.1.0, new skipchains default to BDN because...) would be helpful here.

And also perhaps someplace in the top README.md, if there's a place that talks about interoperability guarantees?

skipchain/skipchain.go
```diff
@@ -1348,16 +1318,11 @@ func (s *Service) bftForwardLinkAck(msg, data []byte) bool {
 // order that is more likely to give us non-failing subleaders in the byzcoinx
 // protocol.
 func (s *Service) startBFT(proto string, origRoster, newRoster *onet.Roster, msg, data []byte) (*byzcoinx.FinalSignature, error) {
+	// Before BDN signatures, the new roster was used when it was a rotation so
```
jeffallen: This comment made sense to me until "but the workaround...", and then lost me. Can you improve that last part to be more clear?

skipchain/struct.go
```diff
@@ -477,6 +451,57 @@ func (sb *SkipBlock) GetForwardLen() int {
...
+	hash.Write(sb.Data)
+	if sb.Roster != nil {
+		for _, pub := range sb.Roster.Publics() {
+			_, err := pub.MarshalTo(hash)
+			if err != nil {
+				panic("couldn't marshall point to hash: " + err.Error())
+			}
+		}
+	}
+	if sb.SignatureScheme > 0 {
```
jeffallen: A comment here would be helpful.

skipchain/struct.go
```diff
@@ -553,6 +578,16 @@ func (sbs Proof) verifyChain() error {
 	return nil
 }
 
+// Signature schemes should be ordered by robustness such that for any
```
jeffallen: Why? What if we want to experiment with a new sig scheme that is much cheaper, but also less secure, and so is better for some IoT use-case or whatever?

Gilthoniel: That comment was actually anticipating the moment were we allow signature schemes to evolve for a given skipchain so that you can verify that a request to change only increase the security and it is not a malicious node trying to get a scheme known to be weak.

But yeah your comment makes totally sense. So the problem is that if we give the ordering/something else to prevent downgrade attacks, we might end up in a dead-end later if we want to change the signature schemes. I need to think about it...

byzcoinx/byzcoinx.go
```diff
@@ -132,7 +132,10 @@ func (bft *ByzCoinX) initCosiProtocol(phase phase) (*protocol.BlsCosi, error) {
 	cosiProto.Threshold = bft.Threshold
 	// For each of the prepare and commit phase we get half of the time.
 	cosiProto.Timeout = bft.Timeout / 2
-	cosiProto.SubleaderFailures = bft.SubleaderFailures
+
```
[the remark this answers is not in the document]
Gilthoniel: Gosh, rebasing went wrong.

skipchain/struct.go
Gilthoniel: One solution I think of is to ignore the indices ordering/whatever and use a per index table to assign authorized changes.

README.md
```diff
@@ -56,6 +56,21 @@ Other pieces we hope to achieve in 2018:
...
+### Release v3.1.0
+
+The release introduces the notion of signature scheme for a given skipchain so that
+one can define which co-signing algorithm will be used to sign the forward links. This
+was necessary in the context of weaknesses in the BLS signature algorithm (see the
+[paper](https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html)). New
+skipchains will be created with [BDN](https://github.com/dedis/kyber/tree/master/sign/bdn)
+set as the signature scheme.
+
+Because of a new scheme is default, that means that skipchains created after v3.1.0
```
kc1212: is there an option to create pre-3.1.0 skipchains when running 3.1+?

Gilthoniel: Not without a small change to use the signature scheme provided when creating a genesis block.

kc1212: i guess my question is more like: should we allow such an option? what are the pros/cons?

byzcoinx/byzcoinx.go
```diff
@@ -315,6 +320,18 @@ func InitBFTCoSiProtocol(suite *pairing.SuiteBn256, c *onet.Context, vf, ack pro
 	return nil
 }
 
+// InitBDNCoSiProtocol creates and registers the protocols required to run
+// BFTCoSi to the context c over the BDN signature scheme
+func InitBDNCoSiProtocol(suite *pairing.SuiteBn256, c *onet.Context, vf, ack protocol.VerificationFn, protoName string) error {
```
kc1212: nit: do we have a convention on the capitalisation of acronyms? in this file we have both `BdnCoSi` and `BDNCoSi`

skipchain/struct.go
```diff
@@ -477,6 +451,60 @@ func (sb *SkipBlock) GetForwardLen() int {
 	return len(sb.ForwardLink)
 }
 
+// SignatureProtocol returns the name of the byzcoinx protocols that should
+// be used to sign forward links coming from this block
```
kc1212: can you update the comment here to say the first return value is the name of the new block protocol and the second is the follow block protocol?

external/java/src/main/java/ch/epfl/dedis/skipchain/ByzcoinSig.java
```diff
@@ -17,6 +14,9 @@
  * ByzcoinSig represents a signature from the byzcoin-protocol. It holds both the message and the signature.
  */
 public class ByzcoinSig {
+    public static final int BLS_SIGNATURE_INDEX = 0;
+    public static final int BDN_SIGNATURE_INDEX = 1;
```
kc1212: is it possible to use java enums? like what you did in go

README.md
Gilthoniel: @jeffallen ?

Gilthoniel: Jeff confirmed that we shouldn't allow it.

byzcoinx/byzcoinx.go
Gilthoniel: I have a preference for camel-cased even if it's an acronym. It's slightly more readable.

PR #1830, opened 2019-04-30
adds wss to bcadmin debug

ineiti:
bcadmin now wants a `URL` for `bcadmin debug`, so it's `http://conode.dedis.ch:7771` or `https://conode.dedis.ch`
Added more details in bcadmin debug list for the byzcoins of each node.
Can now give group.toml to bcadmin debug list, to check on a list of servers.


byzcoin/internal/parse.go
```diff
@@ -0,0 +1,139 @@
...
+	"go.dedis.ch/kyber/v3/suites"
+	"go.dedis.ch/kyber/v3/util/encoding"
+	"go.dedis.ch/onet/v3"
+	"go.dedis.ch/onet/v3/app"
+	"go.dedis.ch/onet/v3/log"
+
+	"go.dedis.ch/onet/v3/network"
+)
+
+// StrToSI converts a string to a ServerIdentity. If tls is true, it will se the URL to point
```
jeffallen: This and the next function seem severely misguided; they are all about faking ServerIdentities where you don't have enough information to really know if you'll be able to talk to the server. It seems like they are only used for the Debug methods, which is fine. But then why can't they be private, and next to the debug methods? And have comments saying, "this isn't the right way to make a ServerIdentity, but it's a special case for debugging".

byzcoin/internal/parse.go
```diff
@@ -0,0 +1,139 @@
+package internal
```
jeffallen: What is this package for, and why? bcadmin already has a mysterious dumping ground of random util crap called bcadmin/lib. Not too keen on a second random util/crap package...

byzcoin/internal/parse.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	g, err := app.ReadGroupDescToml(f)
+	if err != nil {
+		return nil, err
+	}
+	return g.Roster, nil
+}
+
+// LoadPrivate loads a private.toml and returns a ServerIdentity.
+func LoadPrivate(file string) (*network.ServerIdentity, error) {
```
jeffallen: I don't understand why this exists, when onet/app/config.go's ParseCothority exists.

ineiti:
OK, the plan is:
- jeff splits the onet/`config.ParseCothority` in two, so that a `LoadCothority` or `LoadPrivate` exists
- Linus removes the `internal` by replacing the `StrToSI` with `ServerIdentity{URL: c.Args().First()}`

jeffallen:
Onet tag v3.0.8 is ready for you to use.

PR #1833, opened 2019-04-30
fixes coin-contract and Dockerfiles

ineiti:
Reverts the change of the coin-contract to be backward-compatible with previous coin-contract.
Fixes Dockerfile(-dev) to use go1.12 (for tls in old linux boxes) and to run ./run_nodes.sh correctly

Closes #1836

PR #1834, opened 2019-04-30
Evoting tools

jeffallen:
These are small fixes to support the evoting cothority bring up.


conode/run_nodes.sh
```diff
@@ -61,10 +61,11 @@ shift $((OPTIND-1))
 [ "${1:-}" = "--" ] && shift
 
 if [ ! -x ./conode ]; then
-	echo "No conode executable found. Use \"go build\" to make it."
+	echo "No conode executable found in the current directory. Use \"go build\" to make it."
 	exit 1
 fi
 
+bindir=`pwd`
```
ineiti: I suppose that the `run_nodes.sh` and `conode` should be in the same directory, that means that the following line is more adapted:

https://github.com/dedis/cothority/pull/1833/files#diff-1829914d2357dde4f1e83f7eb1b60906R63

Which is in fact a revert of the changes from here:

https://github.com/dedis/cothority/commit/678c4cae70f3a24611f57de5f318e31e210e0fae#diff-1829914d2357dde4f1e83f7eb1b60906

jeffallen: I noticed your change and I'll remove this change.

external/js/kyber/package.json
```diff
@@ -49,9 +49,9 @@
     "eslint-config-prettier": "^2.9.0",
     "eslint-plugin-prettier": "^2.5.0",
     "jasmine": "^3.3.1",
-    "jsdoc-to-markdown": "^4.0.1",
+    "jsdoc-to-markdown": "^5.0.0",
```
Gilthoniel: You can actually remove this package. I just checked and it is not used. I would assume a leftover when I switched to typedoc.

evoting/evoting-admin/scripts/create-master.sh
```diff
@@ -5,12 +5,18 @@ if [ -z "$PIN" ]; then
...
+if [ -z "$1" ]; then
+	echo "Need roster.toml as first argument."
+	exit 1
+fi
+
 sc_jallen=289938
 sc_lindo=128871
 sc_giovanni=121769
-admins=$sc_jallen,$sc_lindo,$sc_giovanni
+sc_nkcr=228271
```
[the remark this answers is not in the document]
jeffallen: They are SCIPERS.

PR #1837, opened 2019-05-01
Implementation of a Deferred Contract

### LINKED ISSUE #1758: multi-sig transaction creation
bcadmin (and other clients) needs to support the creation of multi-sig transactions, so that we can implement co-signed transactions that can pass DARCs with AND in them.

We need to think out how this would work from a UI point of view, and if we can use the ledger itself as a rendezvous point to make this easier to do.

At the same time, perhaps this rendezvous point could contribute to solving the problem wherein if I use my signercount 5 for a signature someone else will contribute to, I cannot send any other txns until we send the co-signed txn, because if I send a new signercount of 5, I invalidate my signature on the co-signed txn, and if I send a txn signed with signercount 6 before the co-signed txn is ready and posted, then it will be refused. Perhaps I could post a commit saying, "you haven't seen it yet, but I promise that a txn with signercount X is coming, and my signature will match this hash, so please let me start using signercount x+1 already".

nkcr:
Fixes #1758.

In summary, here is the data stored and the methods this contract implements:

```go

type DeferredData struct {
	// The transaction that signers must sign and can be executed with an
	// "executeProposedTx".
	ProposedTransaction byzcoin.ClientTransaction
	// If the current block index is greater than this value, any Invoke on the
	// deferred contract is rejected. This provides an expiration mechanism.
	ExpireBlockIndex uint64
	// Hashes of each instruction of the proposed transaction. Those hashes are
	// computed using the special "hashDeferred" method.
	InstructionHashes [][]byte
	// The number of time the proposed transaction can be executed. This number
	// decreases for each successful invocation of "executeProposedTx" and its
	// default value is set to 1.
	NumExecution uint64
	// This array is filled with the instruction IDs of each executed
	// instruction when a successful "executeProposedTx" happens.
	ExecResult [][]byte
}

// Functions available translated to pseudo-code
Spawn(transaction ClientTransaction, expireBlockIndex uint64, numExecution=1 uint64) DeferredData
Invoke_addProof(identity Identity, signature []byte, index uint32) DeferredData
Invoke_execProposedTx() DeferredData

```

To agree on an instruction, a client must sign a special hash that is generated for each instruction stored in the given proposed transaction. Those hashes do not take into account the signers nor the signer counters. However, to make each hash uniq, we inclue the instanceID of the deferred contract to the hashing method.

We made the following considerations:

- In order to prevent useless garbage, a call to "addProof" with a wrong signature is rejected
- A call to "addProof" with an identity already stored is rejected (uniqueness of identities stored)
- Each instruction hash is uniq because we inclue the instanceID of the deferred contract
- A proposed transaction has an "expiration date" based on the given `ExpireBlockIndex`
- Each call to this contract use the standard `Verify()` method, which include the signer counters
- The proposed transaction is executed only if every instruction it contains are successful. 
- Instead of having a flag that indicates if the proposed transaction has already been executed, we give the flexibility to specify how many time the proposed transaction can be executed with the `NumExecution` attribute. This attribute is decreased each time a successful "execProposedTx" is launched. When it reaches 0, any "Invoke" is rejected. By default it is set to 1.

### DIFF
--- a/byzcoin/contracts.go
+++ b/byzcoin/contracts.go
@@ -24,6 +24,8 @@ import (
 type Contract interface {
 	// Verify returns nil if the instruction is valid with regard to the signature.
 	VerifyInstruction(ReadOnlyStateTrie, Instruction, []byte) error
+	// Verify returns nil if the instruction is valid with regard to the signature.
+	VerifyDeferedInstruction(ReadOnlyStateTrie, Instruction, []byte) error
 	// Spawn is used to spawn new instances
 	Spawn(ReadOnlyStateTrie, Instruction, []Coin) ([]StateChange, []Coin, error)
 	// Invoke only modifies existing instances
@@ -60,6 +62,13 @@ func (b BasicContract) VerifyInstruction(rst ReadOnlyStateTrie, inst Instruction
 		return err
 	}
 	return nil
+}
+
+// VerifyDeferedInstruction is not implemented in a BasicContract. Types which
+// embed BasicContract must override this method if they want to support
+// deferred executions (using the Deferred contract).
+func (b BasicContract) VerifyDeferedInstruction(rst ReadOnlyStateTrie, inst Instruction, ctxHash []byte) error {
+	return notImpl("VerifyDeferedInstruction")
 }
 
 // Spawn is not implmented in a BasicContract. Types which embed BasicContract
--- a/byzcoin/contracts/deferred.go
+++ b/byzcoin/contracts/deferred.go
@@ -1,0 +1,239 @@
+package contracts
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"strconv"
+	"time"
+
+	"go.dedis.ch/cothority/v3/byzcoin"
+	"go.dedis.ch/cothority/v3/darc"
+	"go.dedis.ch/protobuf"
+)
+
+// The deferred contract allows a group of signers to agree on and sign a
+// proposed transaction, the "proposed transaction".
+
+// ContractDeferredID denotes a contract that can aggregate signatures for a
+// "proposed" instruction
+var ContractDeferredID = "deferred"
+
+// DeferredData contains the specific data of a deferred contract
+type DeferredData struct {
+	ProposedTransaction byzcoin.ClientTransaction
+	Timestamp           uint64
+	ExpireSec           uint64
+	Hash                []byte
+}
+
+type contractDeferred struct {
+	byzcoin.BasicContract
+	DeferredData
+	s *byzcoin.Service
+}
+
+func (s *Service) contractDeferredFromBytes(in []byte) (byzcoin.Contract, error) {
+	c := &contractDeferred{s: s.byzService()}
+
+	err := protobuf.Decode(in, &c.DeferredData)
+	if err != nil {
+		return nil, errors.New("couldn't unmarshal instance data: " + err.Error())
+	}
+	return c, nil
+}
+
+func (c *contractDeferred) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruction, coins []byzcoin.Coin) (sc []byzcoin.StateChange, cout []byzcoin.Coin, err error) {
+	// This method should do the following:
+	//   1. Parse the input buffer
+	//   2. Compute and store the transaction hash
+	//   3. Save the data
+	//
+	// Spawn should have those input arguments:
+	//   - proposedTransaction ClientTransaction
+	//   - expireSec uint64
+	cout = coins
+
+	// Find the darcID for this instance.
+	var darcID darc.ID
+	_, _, _, darcID, err = rst.GetValues(inst.InstanceID.Slice())
+	if err != nil {
+		return
+	}
+
+	// 1. Reads and parses the input
+	proposedTransaction := byzcoin.ClientTransaction{}
+	err = protobuf.Decode(inst.Spawn.Args.Search("proposedTransaction"), &proposedTransaction)
+	timestamp := uint64(time.Now().Unix())
+	expireSec, err := strconv.ParseUint(string(inst.Spawn.Args.Search("expireSec")), 10, 64)
+	if err != nil {
+		return nil, nil, errors.New("couldn't convert expireSec: " + err.Error())
+	}
+
+	// 2. Computes the hash
+	hash := hashDeferred(proposedTransaction.Instructions[0], timestamp)
+
+	// 3. Saves the data
+	data := DeferredData{
+		proposedTransaction,
+		timestamp,
+		expireSec,
+		hash,
+	}
+	var dataBuf []byte
+	dataBuf, err = protobuf.Encode(&data)
+	if err != nil {
+		return nil, nil, errors.New("couldn't encode DeferredData: " + err.Error())
+	}
+
+	sc = append(sc, byzcoin.NewStateChange(byzcoin.Create, inst.DeriveID(""),
+		ContractDeferredID, dataBuf, darcID))
+	return
+}
+
+func (c *contractDeferred) Invoke(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruction, coins []byzcoin.Coin) (sc []byzcoin.StateChange, cout []byzcoin.Coin, err error) {
+	// This method should do the following:
+	//   - Handle the "addProof" invocation
+	//   - Handle the "execProposedTx" invocation
+	//
+	// Invoke:addProof should have the following input argument:
+	//   - identity darc.Identity
+	//   - signature string
+	cout = coins
+
+	// Find the darcID for this instance.
+	var darcID darc.ID
+
+	_, _, _, darcID, err = rst.GetValues(inst.InstanceID.Slice())
+	if err != nil {
+		return
+	}
+
+	switch inst.Invoke.Command {
+	case "addProof":
+		// This invocation appends the identity and the corresponding signature,
+		// which is based on the stored hash of a client. Returns the contract's
+		// data.
+
+		// Get the given Identity
+		identityBuf := inst.Invoke.Args.Search("identity")
+		if identityBuf == nil {
+			return nil, nil, errors.New("Identity args is nil")
+		}
+		identity := darc.Identity{}
+		err = protobuf.Decode(identityBuf, &identity)
+		if err != nil {
+			return nil, nil, errors.New("Couldn't decode Identity")
+		}
+		// Get the given signature
+		signature := inst.Invoke.Args.Search("signature")
+		if signature == nil {
+			return nil, nil, errors.New("Signature args is nil")
+		}
+		// Update the contract's data with the given signature and identity
+		c.DeferredData.ProposedTransaction.Instructions[0].SignerIdentities = append(c.DeferredData.ProposedTransaction.Instructions[0].SignerIdentities, identity)
+		c.DeferredData.ProposedTransaction.Instructions[0].Signatures = append(c.DeferredData.ProposedTransaction.Instructions[0].Signatures, signature)
+		// Save and send the modifications
+		cosiDataBuf, err2 := protobuf.Encode(&c.DeferredData)
+		if err2 != nil {
+			return nil, nil, errors.New("Couldn't encode DeferredData")
+		}
+		sc = append(sc, byzcoin.NewStateChange(byzcoin.Update, inst.InstanceID,
+			ContractDeferredID, cosiDataBuf, darcID))
+		return
+	case "execProposedTx":
+		// This invocation tries to execute the transaction stored with the
+		// "Spawn" invocation. If it is successful, this invocation returns
+		// the InstanceID of the executed proposed transaction.
+
+		instruction := c.DeferredData.ProposedTransaction.Instructions[0]
+
+		// In case it goes well, we want to return the proposed Tx InstanceID
+		rootInstructionID := instruction.DeriveID("").Slice()
+		sc = append(sc, byzcoin.NewStateChange(byzcoin.Update, inst.InstanceID,
+			ContractDeferredID, rootInstructionID, darcID))
+
+		instructionType := instruction.GetType()
+		if instructionType == byzcoin.SpawnType {
+			fn, exists := c.s.GetContractConstructor(instruction.Spawn.ContractID)
+			if !exists {
+				return nil, nil, errors.New("Couldn't get the root function")
+			}
+			rootInstructionBuff, err := protobuf.Encode(&c.DeferredData.ProposedTransaction.Instructions[0])
+			if err != nil {
+				return nil, nil, errors.New("Couldn't encode the root instruction buffer")
+			}
+			contract, err := fn(rootInstructionBuff)
+			if err != nil {
+				return nil, nil, errors.New("Couldn't get the root contract")
+			}
+
+			err = contract.VerifyDeferedInstruction(rst, instruction, c.DeferredData.Hash)
+			if err != nil {
+				return nil, nil, fmt.Errorf("Verifying the root instruction failed: %s", err)
+			}
+
+			rootSc, _, err := contract.Spawn(rst, c.DeferredData.ProposedTransaction.Instructions[0], coins)
+			sc = append(sc, rootSc...)
+		}
+		return
+	default:
+		return nil, nil, errors.New("Cosi contract can only addProof and execRoot")
+	}
+}
+
+func (c *contractDeferred) Delete(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruction, coins []byzcoin.Coin) (sc []byzcoin.StateChange, cout []byzcoin.Coin, err error) {
+	cout = coins
+
+	// Find the darcID for this instance.
+	var darcID darc.ID
+	_, _, _, darcID, err = rst.GetValues(inst.InstanceID.Slice())
+	if err != nil {
+		return
+	}
+
+	sc = byzcoin.StateChanges{
+		byzcoin.NewStateChange(byzcoin.Remove, inst.InstanceID, ContractDeferredID, nil, darcID),
+	}
+	return
+}
+
+// This is a modified version of computing the hash of a transaction. In this
+// version, we do not take into account the signers nor the signers counters. We
+// also add to the hash a timestamp.
+func hashDeferred(instr byzcoin.Instruction, timestamp uint64) []byte {
+	h := sha256.New()
+	h.Write(instr.InstanceID[:])
+	var args []byzcoin.Argument
+	switch instr.GetType() {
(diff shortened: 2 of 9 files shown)


byzcoin/transaction.go
```diff
@@ -291,9 +300,11 @@ func (instr Instruction) Verify(st ReadOnlyStateTrie, msg []byte) error {
 	}
 
 	// check the signature
+	// Save the identities that provide good signatures
+	goodIdentities := make([]string, 1)
```
kc1212: why is the size 1?

byzcoin/contracts/deferred.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+
+	sc = byzcoin.StateChanges{
+		byzcoin.NewStateChange(byzcoin.Remove, inst.InstanceID, ContractDeferredID, nil, darcID),
+	}
+	return
+}
+
+// VerifyInstruction overrides the basic VerifyInstruction
+func (c *contractDeferred) VerifyInstruction(rst byzcoin.ReadOnlyStateTrie, instr byzcoin.Instruction, ctxHash []byte) error {
```
kc1212: i'm not 100% sure whether what's in this function is what we think `VerifyInstruction` should be doing. if you look at the interface documentation, `VerifyInstruction` should only care whether the right signatures are there for the darc, it shouldn't have any other functionality. in my opinion these contract specific verification should go into spawn/invoke. what do you think @jeffallen

byzcoin/contracts.go
```diff
@@ -24,6 +24,8 @@ import (
 type Contract interface {
 	// Verify returns nil if the instruction is valid with regard to the signature.
 	VerifyInstruction(ReadOnlyStateTrie, Instruction, []byte) error
+	// Verify returns nil if the instruction is valid with regard to the signature.
+	VerifyDeferedInstruction(ReadOnlyStateTrie, Instruction, []byte) error
```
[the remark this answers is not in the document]
nkcr: `VerifyDeferredInstruction` is called on each instruction of the proposed transaction, which should be implemented by each contract that wants to support a deferred instruction. We can't make it global and implement it inside the DeferredContract since some contract may have implemented their custom `VerifyInstruction`, thus needing to also implement their custom version of `VerifyDeferredInstruction`.

byzcoin/statetrie.go
```diff
@@ -79,7 +79,8 @@ func (t *stagingStateTrie) Commit() error {
 
 // GetIndex returns the index of the current trie.
 func (t *stagingStateTrie) GetIndex() int {
```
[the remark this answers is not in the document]
kc1212: yes, a staging trie doesn't belong to a block. but we always create staging tries from a real trie, which does belong to a block. so `GetIndex` will return the latest known index. perhaps the comments should clarify that the function returns the index according to what's stored in the trie metadata and not the index which contains the trie.

byzcoin/contracts/deferred.go
```diff
@@ -0,0 +1,379 @@
...
+const defaultNumExecution uint64 = 1
+
+// DeferredData contains the specific data of a deferred contract
+type DeferredData struct {
+	// The transaction that signers must sign and can be executed with an
+	// "executeProposedTx".
+	ProposedTransaction byzcoin.ClientTransaction
+	// If the current block index is greater than this value, any Invoke on the
+	// deferred contract is rejected. This provides an expiration mechanism.
+	ExpireBlockIndex uint64
```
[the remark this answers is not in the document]
kc1212: the problem with a wall-clock time limit is that the consensus result completely relies on network delay and clock synchronisation. a "late" message will cause honest nodes to reject what they would usually accept, causing consensus to fail. using block index won't have this problem but it's true that if there are no transactions, then no blocks are created. i'm not sure what's the best solution here...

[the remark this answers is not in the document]
Gilthoniel: I guess we need to rethink what is behind the expiration, what exactly is the purpose ? Is it like a feature for people to use or is it a cleaning/security feature ?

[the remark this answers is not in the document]
kc1212: i think it's a feature to make sure every deferred contract reaches some "done" state. it's like saying everyone has to sign before the expiry time otherwise this instance become invalid. i don't know how useful it is or the consequences if we don't have this feature.

another way to do something similar is to allow users to remove their proofs. then we don't need to worry about timing 

maybe @nkcr has more ideas?

byzcoin/contracts/deferred.go
```diff
@@ -0,0 +1,379 @@
...
+
+		// Get the given signature
+		signature := inst.Invoke.Args.Search("signature")
+		if signature == nil {
+			return nil, nil, errors.New("Signature args is nil")
+		}
+		// Update the contract's data with the given signature and identity
+		c.DeferredData.ProposedTransaction.Instructions[index].SignerIdentities = append(c.DeferredData.ProposedTransaction.Instructions[index].SignerIdentities, identity)
+		c.DeferredData.ProposedTransaction.Instructions[index].Signatures = append(c.DeferredData.ProposedTransaction.Instructions[index].Signatures, signature)
+		// Save and send the modifications
```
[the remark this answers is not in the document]
nkcr: Each instruction requires a different signature. A transaction is accepted once every instruction has been correctly signed by the required clients. This is why we keep the signatures and identities for each instruction.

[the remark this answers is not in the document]
nkcr: Each instruction requires a different signature. A transaction is accepted once every instruction has been correctly signed by the required clients. This is why we keep the signatures and identities for each instruction.

byzcoin/contracts/deferred.go
nkcr: Makes sense. Moved it to the spawn instruction.

byzcoin/contracts/deferred.go
```diff
@@ -0,0 +1,379 @@
...
+			ContractDeferredID, cosiDataBuf, darcID))
+		return
+	case "execProposedTx":
+		// This invocation tries to execute the transaction stored with the
+		// "Spawn" invocation. If it is successful, this invocation fills the
+		// "ExecResult" field of the "DeferredData" struct.
+
+		instructionIDs := make([][]byte, len(c.DeferredData.ProposedTransaction.Instructions))
+
+		for i, proposedInstr := range c.DeferredData.ProposedTransaction.Instructions {
```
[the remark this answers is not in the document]
nkcr: I tried, but turned out to be quite complex (@kc1212 can confirm) and introduce more complexity compared to the benefit.

byzcoin/contracts/deferred.go
```diff
@@ -0,0 +1,379 @@
...
+			return errors.New("Couldn't decode Identity")
+		}
+		// Get the instruction index
+		indexBuf := instr.Invoke.Args.Search("index")
+		if indexBuf == nil {
+			return errors.New("Index args is nil")
+		}
+		index := binary.LittleEndian.Uint32(indexBuf)
+
+		for _, storedIdentity := range c.DeferredData.ProposedTransaction.Instructions[index].SignerIdentities {
```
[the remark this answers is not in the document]
nkcr: Since this is the only place where we need this functionality, does it make sense to create a specific function for this?

byzcoin/contracts/deferred.go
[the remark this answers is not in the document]
nkcr: I was writing my case to keep the block index but changed my mind..

As @kc1212 said, this is to ensure that a proposed transaction is eventually done and can not stay hidden unfinished and then cause trouble later. This is not meant as a feature that client can use. For this reason, I would rather specify a timestamp that should provide a roughly expire date. Even if a majority of Conodes have like a day of difference (not likely to happen I think), that shouldn't be problem if my timestamps is in weeks, or months, or more.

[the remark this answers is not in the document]
nkcr: So, when spawning the contract, specify an absolute `uint64 timestamp`. Then each Conode check the expiration by checking if the provided timestamps is lower than its current one.

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,404 @@
...
+	proposedTransaction := ClientTransaction{}
+	err = protobuf.Decode(inst.Spawn.Args.Search("proposedTransaction"), &proposedTransaction)
+	if err != nil {
+		return nil, nil, errors.New("couldn't decode proposedTransaction: " + err.Error())
+	}
+	expireBlockIndex, err := strconv.ParseUint(string(inst.Spawn.Args.Search("expireBlockIndex")), 10, 64)
+	if err != nil {
+		return nil, nil, errors.New("couldn't convert expireBlockIndex: " + err.Error())
+	}
+	numExecutionBuff := inst.Spawn.Args.Search("NumExecution")
```
kc1212: @jeffallen suggested that there is no strong need for this feature and i agree. can you remove this parameter?

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,404 @@
...
+		if err2 != nil {
+			return nil, nil, errors.New("couldn't encode DeferredData")
+		}
+		sc = append(sc, NewStateChange(Update, inst.InstanceID,
+			ContractDeferredID, cosiDataBuf, darcID))
+		return
+	case "execProposedTx":
+		// This invocation tries to execute the transaction stored with the
+		// "Spawn" invocation. If it is successful, this invocation fills the
+		// "ExecResult" field of the "deferredData" struct.
```
kc1212: could you add a comment here explaining why something like `processOneTx` cannot be usef?

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,404 @@
...
+		// readonly state try by copying the data.
+		nonce, err2 := rst.GetNonce()
+		if err2 != nil {
+			return nil, nil, errors.New("couldn't get the nonce: " + err2.Error())
+		}
+		sst, err2 := newMemStagingStateTrie(nonce)
+		if err2 != nil {
+			return nil, nil, errors.New("Failed to created stagingStateTrie: " + err2.Error())
+		}
+		rst.ForEach(sst.Set)
```
kc1212: check for error please

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,404 @@
...
+	//   - Handle the "addProof" invocation
+	//   - Handle the "execProposedTx" invocation
+	//
+	// Invoke:addProof should have the following input argument:
+	//   - identity darc.Identity
+	//   - signature []byte
+	//   - index uint32 (index of the instruction wrt the transaction)
+	err2 := c.checkInvoke(rst, inst.Invoke)
+	if err2 != nil {
+		return nil, nil, errors.New("checks of invoke failed: " + err2.Error())
```
kc1212: this function uses named return values, instead of doing what you have on this line,
you can do 
```
err = c.checkInvoke(...
if err != nil {
  return
}
```

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,404 @@
...
+		// "Spawn" invocation. If it is successful, this invocation fills the
+		// "ExecResult" field of the "deferredData" struct.
+
+		// In the following we are creating a new StagingStateTrie from the
+		// readonly state try by copying the data.
+		nonce, err2 := rst.GetNonce()
+		if err2 != nil {
+			return nil, nil, errors.New("couldn't get the nonce: " + err2.Error())
+		}
+		sst, err2 := newMemStagingStateTrie(nonce)
```
kc1212: the places with `err2` can be changed to use named return values

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,404 @@
...
+			case SpawnType:
+				stateChanges, _, err = contract.Spawn(sst, proposedInstr, coins)
+			case InvokeType:
+				stateChanges, _, err = contract.Invoke(sst, proposedInstr, coins)
+			case DeleteType:
+				stateChanges, _, err = contract.Delete(sst, proposedInstr, coins)
+
+			}
+
+			sst.StoreAll(stateChanges)
```
kc1212: do you have a test that checks whether the deferred contract works with instructions that depend on each other (all in the same transaction)? i glanced the tests but didn't see one. maybe use the value contract to set a value and then in a different instruction try to get the same value.

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,404 @@
...
+		sc = append(sc, NewStateChange(Update, inst.InstanceID,
+			ContractDeferredID, resultBuf, darcID))
+
+		return
+	default:
+		return nil, nil, errors.New("deferred contract can only addProof and execProposedTx")
+	}
+}
+
+func (c *contractDeferred) Delete(rst ReadOnlyStateTrie, inst Instruction, coins []Coin) (sc []StateChange, cout []Coin, err error) {
```
kc1212: let me know if i missed it, but i feel there's a missing feature surrounding the delete
during our meeting with @jeffallen we decided to add a feature that allows anybody to delete the instance if has expired, but i didn't find it in this PR. actually, it might be better if it's in another PR. what might be possible is the following: add an optional field in `BasicContract` called `ExpirationBlockIndex`. by default it's 0 and it's not checked. but if it is set, then `VerifyInstruction` will check whether it is expired and if it is, anybody can delete it.

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,404 @@
...
+			}
+			sc = append(sc, stateChanges...)
+
+		}
+
+		c.DeferredData.ExecResult = instructionIDs
+		// At this stage all verification passed. We can then decrease the
+		// NumExecution counter.
+		c.DeferredData.NumExecution = c.DeferredData.NumExecution - 1
+		resultBuf, err2 := protobuf.Encode(&c.DeferredData)
```
kc1212: use named return value instead of `err2`

byzcoin/contracts/value.go
```diff
@@ -84,3 +84,10 @@ func (c *contractValue) Delete(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instr
 	}
 	return
 }
+
+func (c *contractValue) VerifyDeferredInstruction(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruction, ctxHash []byte) error {
+	if err := inst.VerifyWithOption(rst, ctxHash, false); err != nil {
+		return err
+	}
+	return nil
```
kc1212: you can just do `return inst.VerifyWithOption(rst, ctxHash, false)`

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,404 @@
...
+		return
+	}
+
+	// 1. Reads and parses the input
+	proposedTransaction := ClientTransaction{}
+	err = protobuf.Decode(inst.Spawn.Args.Search("proposedTransaction"), &proposedTransaction)
+	if err != nil {
+		return nil, nil, errors.New("couldn't decode proposedTransaction: " + err.Error())
+	}
+	expireBlockIndex, err := strconv.ParseUint(string(inst.Spawn.Args.Search("expireBlockIndex")), 10, 64)
```
kc1212: integers should be serialized as the byte representation of the integer and not the byte representation of the string representation of the integer. in our code we usually use `binary.LittleEndian.PutUint64`

byzcoin/contract_deferred.go
nkcr: By doing so I loose the specific error message added "checks of invoke failed: ...". 

I could do :

```go
if err != nil {
  err = errors.New("checks of invoke failed: " + err2.Error())
  return
} 
```

but that's one line longer and makes the return mechanism a bit more hidden. Is it better?

byzcoin/contract_deferred.go
nkcr: In the case of assigning multiple variable where one of them must be initialized, I can't use `err`. For example, with `sst, err := newMemStagingStateTrie(nonce)` I get "`err` is shadowed during return", because I used the short variable declaration. 

In this case I would need to to the following: 

```go
		var sst *stagingStateTrie
		sst, err = newMemStagingStateTrie(nonce)
		if err != nil {
                        err = errors.New("Failed to created stagingStateTrie: " + err2.Error())
			return
		}
``` 

Instead of

```go
		sst, err2 := newMemStagingStateTrie(nonce)
		if err2 != nil {
			return nil, nil, errors.New("Failed to created stagingStateTrie: " + err2.Error())
		}
```

Which solution should I pick?

byzcoin/contract_deferred.go
nkcr: This is implemented inside the `VerifyInstruction` function, where I allow anyone to delete the contract if it has expired.

byzcoin/contract_deferred.go
nkcr: Added one. The simplest case I could implement is deleting a value contract then trying to call it and see if I get an error.

nkcr:
Thanks a lot the the second review @kc1212 !!
I marked as resolved the requested changes that I applied and made somme comments on others.

byzcoin/contracts/contract_deferred_test.go
```diff
@@ -0,0 +1,1938 @@
...
+	require.Nil(t, err)
+
+	// Need to sleep because we can't predict the output (hence the 'nil')
+	time.Sleep(2 * genesisMsg.BlockInterval)
+	pr, err = cl.WaitProof(byzcoin.NewInstanceID(ctx.Instructions[0].DeriveID("").Slice()), 2*genesisMsg.BlockInterval, nil)
+	require.Error(t, err)
+
+	local.WaitDone(genesisMsg.BlockInterval)
+
+}
```
jeffallen: Correct me if I'm wrong but... I expected to find a test where the proposed instruction depended on having a darc with an AND in it evaluate to true. The test would spawn the deferred, add one proof, try to exec, confirm it failed, add another proof that is not useful to satisfying the DARC, try exec, confirm failed, add the last signature needed to satisfy the darc, and then exec and confirm the proposed txn goes in.

Also, I didn't see a test for deletion?

This test can be added in your next commit, though.

byzcoin/contract_deferred.go
```diff
@@ -0,0 +1,405 @@
...
+	case "addProof":
+		// This invocation appends the identity and the corresponding signature,
+		// which is based on the stored instruction hash (in instructionHashes)
+
+		// Get the given index
+		indexBuf := inst.Invoke.Args.Search("index")
+		if indexBuf == nil {
+			return nil, nil, errors.New("index args is nil")
+		}
+		index := binary.LittleEndian.Uint32(indexBuf)
```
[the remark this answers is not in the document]
nkcr: We are working with a transaction containing multiple instructions.
The index is necessary because each instruction must be signed independently (ie. they each require a different signature and can potentially have a different DARC).

byzcoin/contracts/contract_deferred_test.go
nkcr: That's right. I'm keeping it on the todo list !

PR #1839, opened 2019-05-01
Detect infinite recursive darc delegation

### LINKED ISSUE #1308: Darc delegation cycles
If darc A has "_sign" for darc B (i.e. darc A delegates to darc B; see TestDarc_Delegation) and B delegates to A, that's a loop, and it needs to be detected and/or prevented.

1. Write a test for this.
2. Solve any problems it detects.

kc1212:
In general, there are two solutions to this problem. The first is to add a depth argument in the recursive call and check that it doesn't become too high. The second is to add a list of visited nodes and check that we don't visit a node in that list. I went for the second solution because picking the depth is tricky. My early implementation using a depth of 10,000 takes 0.3 seconds to return error. Making it too small might prevent legitimate uses. 

Fixes #1308

### DIFF
--- a/darc/darc.go
+++ b/darc/darc.go
@@ -53,12 +53,12 @@ import (
 	"go.dedis.ch/kyber/v3/suites"
 	"go.dedis.ch/kyber/v3/util/encoding"
 	"go.dedis.ch/kyber/v3/util/key"
-	"go.dedis.ch/onet/v3/log"
 	"go.dedis.ch/protobuf"
 )
 
 const evolve = "_evolve"
 const sign = "_sign"
+const maxEvalDepth = 10000
 
 // GetDarc is a callback function that we expect the user of this library to
 // supply in some of our methods. The user is free to choose how he/she wants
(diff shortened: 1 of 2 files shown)


darc/darc.go
```diff
@@ -628,42 +627,64 @@ func EvalExprDarc(expr expression.Expr, getDarc GetDarc, acceptDarc bool, ids ..
 			if acceptDarc && found {
 				return true
 			}
+			// prevent cycles by checking the visited map
+			if _, ok := visited[s]; ok {
+				issue = errors.New("cycle detected")
+				return false
+			}
+			newVisited := make(map[string]bool)
```
Gilthoniel: Is it actually necessary to copy the map ? If this is executed synchronously you can reuse the map in the recursions I think.

kc1212: the evaluation will expand every darc identity, suppose we have `darc:a && darc:b`, then the sign expr for both of them are are `darc:c`, then sign expr for `darc:c` is a regular ed25519 public key. in this situation, when evaluating `darc:a && darc:b`, the visited list will contain `darc:a, darc:c` when evaluating the LHS of `&&`. to evaluate the RHS of `&&`, the function will expand `darc:b` and find `darc:c`. if we're not copying the map, the evaluation will fail because `darc:c` is already in the visited list. if we do the copying, the visited list that is passed into `darc:b` expansion will be empty.

in summary, when the darc expansion is on different paths, we don't want to unify the visited list, copying fixes this problem.

I will write a test case for this scenario.

PR #1840, opened 2019-05-01
Changes necessary to use the new Kyber DKG changes

jeffallen:
Also: Upgrade onet to get new message counters in simulations.


calypso/service.go
```diff
@@ -43,7 +43,7 @@ import (
 	"go.dedis.ch/cothority/v3/skipchain"
 	"go.dedis.ch/kyber/v3"
 	"go.dedis.ch/kyber/v3/share"
-	"go.dedis.ch/kyber/v3/share/dkg/pedersen"
+	dkg "go.dedis.ch/kyber/v3/share/dkg/pedersen"
```
ineiti: Makes me smile: between my IntelliJ and your emacs, one deems that the `dkg` is necessary, while the other wants to remove it...

PR #1845, opened 2019-05-02
adds calypso admin binary to set up authorization

ineiti:
- calypso/csadmin that takes the private.toml to sign the byzcoin-id that should be stored
- calypso:Authorize now checks for the signature, except if COTHORITY_ALLOW_INSECURE_ADMIN=true


calypso/proto.go
```diff
@@ -60,14 +60,16 @@ type Read struct {
 // These are the messages used in the API-calls
 // ***
 
-// Authorise is used to add the given ByzCoinID into the list of
-// authorised IDs.
-type Authorise struct {
```
Gilthoniel: The change is right but it breaks backwards compatibility. Should this one be deprecated but still exist ?

calypso/api.go
```diff
@@ -87,16 +89,25 @@ func (c *Client) CreateLTS(ltsRoster *onet.Roster, darcID darc.ID, signers []dar
...
 //
 // It should be called by the administrator at the beginning, before any other
 // API calls are made. A ByzCoinID that is not authorised will not be allowed to
 // call the other APIs.
-func (c *Client) Authorise(who *network.ServerIdentity, what skipchain.SkipBlockID) error {
-	reply := &AuthoriseReply{}
-	err := c.c.SendProtobuf(who, &Authorise{ByzCoinID: what}, reply)
+func (c *Client) Authorize(who *network.ServerIdentity, what skipchain.SkipBlockID) error {
+	reply := &AuthorizeReply{}
+	sig, err := schnorr.Sign(cothority.Suite, who.GetPrivate(), what)
```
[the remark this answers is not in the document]
ineiti: Yes, he could. Should I add a timestamp and only allow authorization requests that are not older than x minutes? Or do we say that an attacker would have to be able to get hold of the signature in the first place, and that this is very unlikely? Given that we have tls-enabled websocket communication, the attacker would have to be _in_ the conode or the client, and then he can do anything he wants anyway.

calypso/proto.go
ineiti: I let this up to DEDIS to decide - I'm not aware of any active project that uses this API-endpoint, so I took the liberty to change the name. But if you want to keep the exact naming, I will change it back to `Authorise`.

calypso/api.go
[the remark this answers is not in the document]
Gilthoniel: I would prefer protect the signature because even if an attacker is inside and can do whatever he wants, that means he can get signatures from other conodes, even after being thrown away by the admin. A timestamp would be perfectly fine here I guess.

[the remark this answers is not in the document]
ineiti: An attacker cannot use a signature from another conode, as it is verified using the public key of the conode. So he can only re-authorize the LTS on the conode he controls. Which he can do anyway...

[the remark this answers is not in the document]
Gilthoniel: Yes you're right. But still, I think it's not totally useless to protect this signature a bit more because if a leak appears for any reason. It's always better to have multiple layers of security and it is not too hard to change I think.

calypso/proto.go
Gilthoniel: @jeffallen ?
You could also mark them as deprecated, and create a deprecated handler that is listening on `Authorise` that can be cleaned up in v4.

calypso/api.go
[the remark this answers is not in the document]
ineiti: OK, added a timestamp to the signature. I'm not totally convinced it is really needed, and just putting one more element in a signature does not necessarily make it better.

calypso/proto.go
jeffallen: Sorry to be absent for so long, please let's respect our backwards compatibility commitment strongly for Go server-side things. I'm ok with a bit more flux on client-side APIs that are very immature.

ineiti: Thumbs-up -> did it...

calypso/api.go
```diff
@@ -87,16 +90,34 @@ func (c *Client) CreateLTS(ltsRoster *onet.Roster, darcID darc.ID, signers []dar
 	return reply, nil
 }
 
-// Authorise adds a ByzCoinID to the list of authorized IDs for each
-// server in the roster. The AuthoriseByzcoinID service refuses requests
-// that do not come from localhost.
+// Authorise - deprecated, please see Authorize
```
Gilthoniel:
```
Authorise blabla...
Deprecated: please see Authorize
```

Gilthoniel: In order to have a tag to find deprecated code later

calypso/proto.go
```diff
@@ -60,16 +60,28 @@ type Read struct {
 // These are the messages used in the API-calls
 // ***
 
-// Authorise is used to add the given ByzCoinID into the list of
-// authorised IDs.
+// Authorise - deprecated, please use Authorize
```
Gilthoniel: Same with the deprecated comment

calypso/proto.go
```diff
@@ -60,16 +60,28 @@ type Read struct {
...
 type Authorise struct {
 	ByzCoinID skipchain.SkipBlockID
 }
 
-// AuthoriseReply is returned upon successful authorisation.
+// AuthoriseReply - deprecated, please use AuthorizeReply
```
Gilthoniel: Same with the deprecated comment

calypso/service.go
```diff
@@ -103,41 +105,51 @@ type vData struct {
...
-		}
-		ip := net.ParseIP(h)
-
-		if !ip.IsLoopback() {
-			return nil, nil, errors.New("authorise is only allowed on loopback")
-		}
-	}
-
-	return s.ServiceProcessor.ProcessClientRequest(req, path, buf)
+// Authorise is deprecated - please use Authorize.
```
Gilthoniel: For the sake of strong backwards compatibility, can we keep this one working with localhost requests ? But still marked as deprecated.

calypso/api.go
ineiti: Thanks for catching up with me...

calypso/service.go
ineiti: Done in latest commit.

PR #1846, opened 2019-05-02
adds fixes to personhood spawner

ineiti:
- correctly spawn coins and credential instances, including controlling darc
- credentialContract: add 'recover' command to allow a set of darcs reset the controlling darc


personhood/contract_credential.go
```diff
@@ -87,6 +116,103 @@ func (c *ContractCredential) Invoke(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.
 
 		sc = append(sc, byzcoin.NewStateChange(byzcoin.Update, inst.InstanceID,
 			ContractCredentialID, credBuf, darcID))
+
+	case "recover":
+		// "recover" checks if enough signatures are present to change the 'evolve' and 'sign' rule
+		// of the darc attached to the credential.
+		sigBuf := inst.Invoke.Args.Search("signatures")
+		if len(sigBuf) == 0 || len(sigBuf)%96 != 0 {
```
Gilthoniel: Do we have a constant for the 96 ?

personhood/contract_credential.go
```diff
@@ -87,6 +116,103 @@ func (c *ContractCredential) Invoke(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.
...
+			return nil, nil, errors.New("wrong signatures argument")
+		}
+		pubBuf := inst.Invoke.Args.Search("public")
+		if len(pubBuf) != 32 {
```
Gilthoniel: Same here, it helps when reading the code to see where this comes from.

personhood/contract_credential.go
```diff
@@ -87,6 +116,103 @@ func (c *ContractCredential) Invoke(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.
...
+		}
+		if threshold == 0 || len(trusteesDarc) == 0 {
+			return nil, nil, errors.New("no threshold or no trustee found")
+		}
+		var valid uint32
+		msg := append(inst.InstanceID.Slice(), pubBuf...)
+		darcVersion := make([]byte, 8)
+		binary.LittleEndian.PutUint64(darcVersion, d.Version)
+		msg = append(msg, darcVersion...)
+		for signer := 0; signer < len(sigBuf); signer += 96 {
```
Gilthoniel: Same about the constants

personhood/contract_credential.go
```diff
@@ -109,3 +235,43 @@ func (c *ContractCredential) Delete(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.
...
+	}
+	if cid != ContractCredentialID {
+		return nil, errors.New("not a credential instance")
+	}
+	return getDarc(rst, darcID)
+}
+
+func checkDarcRule(rst byzcoin.ReadOnlyStateTrie, d *darc.Darc, id string) error {
+	getDarc := func(str string, latest bool) *darc.Darc {
+		if len(str) < 5 || string(str[0:5]) != "darc:" {
```
Gilthoniel: HasPrefix ?

byzcoin/bcadmin/main.go
```diff
@@ -1062,6 +1069,95 @@ func darcShow(c *cli.Context) error {
...
+	servers := local.GenServers(1)
+	s := servers[0].Service(byzcoin.ServiceName).(*byzcoin.Service)
+
+	cl := skipchain.NewClient()
+	var sst *byzcoin.StagingStateTrie
+	sb, err := cl.GetSingleBlock(r, bcID)
+	if err != nil {
+		return err
+	}
+	for {
```
Gilthoniel: Can you move all that logic to the byzcoin Client (api.go) so that we can keep the memory staging trie as private ?

byzcoin/bcadmin/main.go
```diff
@@ -107,6 +108,12 @@ var cmds = cli.Commands{
 		Usage:   "interact with byzcoin for debugging",
 		Aliases: []string{"d"},
 		Subcommands: cli.Commands{
+			{
+				Name:      "follow",
```
Gilthoniel: Can we find another term for this ? I got confused with the "Follow" feature of the skipchain service. I'm not really good at naming so I let you think about it but it feels more like an integrity check.

personhood/contract_credential.go
```diff
@@ -29,24 +36,46 @@ type ContractCredential struct {
 	CredentialStruct
 }
 
+// VerifyInstruction allows for an unsigned "recover" command that will be verified later.
+func (c ContractCredential) VerifyInstruction(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruction, ctxHash []byte) error {
+	if inst.Invoke != nil && inst.Invoke.Command == "recover" {
```
Gilthoniel: Can we have more details here why this particular case needs to skip the verification ?

byzcoin/bcadmin/main.go
```diff
@@ -1062,6 +1069,95 @@ func darcShow(c *cli.Context) error {
...
+		return errors.New("couldn't create roster")
+	}
+	bcID, err := hex.DecodeString(c.Args().Get(1))
+	if err != nil {
+		return err
+	}
+
+	local := onet.NewLocalTest(cothority.Suite)
+	defer local.CloseAll()
+	servers := local.GenServers(1)
```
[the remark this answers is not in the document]
ineiti: I tried first to instantiate the service without the whole context. But `onet.Context` has a lot of private fields, so I had to change all the `log.Lvl1(s.ServerIdentity(), ...`. And finally I thought that using a `LocalTest` was the easiest way to go. After some time it also felt normal, as this functionality _does_ some kind of local test.

> Two requests at the same time will fail because the port won't be open for the second.

I don't understand that. `LocalTest` uses channels, no? So why would it open a port?

byzcoin/service.go
```diff
@@ -1067,7 +1069,12 @@ func (s *Service) catchupAll() error {
 func (s *Service) catchupFromID(r *onet.Roster, scID skipchain.SkipBlockID, sbID skipchain.SkipBlockID) error {
 	// Catch up only friendly skipchains to avoid unnecessary requests
 	if s.db().GetByID(scID) == nil {
-		return fmt.Errorf("%s: Got asked for an unknown skipchain: %x", s.ServerIdentity(), scID)
+		if _, found := s.unknownSkipchains[string(scID)]; found {
+			log.Lvlf3("got asked for unknown skipchain: %x", scID)
+			return nil
+		}
+		s.unknownSkipchains[string(scID)] = true
```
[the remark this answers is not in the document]
ineiti: Yes - so do you prefer adding a `Mutex` here, or just put `log.Lvl3` and be done with it?

byzcoin/statetrie.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"go.dedis.ch/cothority/v3/byzcoin/trie"
 	"go.dedis.ch/cothority/v3/darc"
-	bbolt "go.etcd.io/bbolt"
+	"go.etcd.io/bbolt"
```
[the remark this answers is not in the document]
ineiti: Yes, I don't know how to tell `goimports` to not do this. If it bothers you too much, I'll remove `goimports`, but for the moment it is really useful.

byzcoin/bcadmin/main.go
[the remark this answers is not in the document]
Gilthoniel:
```go
// genLocalHosts returns n servers created with a localRouter
func (l *LocalTest) genLocalHosts(n int) []*Server {
	l.panicClosed()
	servers := make([]*Server, n)
	for i := 0; i < n; i++ {
		port := l.latestPort
		l.latestPort += 10
		servers[i] = l.NewServer(l.Suite, port)
	}
	return servers
}

// NewServer returns a new server which type is determined by the local mode:
// TCP or Local. If it's TCP, then an available port is used, otherwise, the
// port given in argument is used.
func (l *LocalTest) NewServer(s network.Suite, port int) *Server {
	l.panicClosed()
	var server *Server
	switch l.mode {
	case TCP:
		server = l.newTCPServer(s)
		// Set TLS certificate if any configuration available
		if len(l.webSocketTLSCertificate) > 0 && len(l.webSocketTLSCertificateKey) > 0 {
			cert, err := tls.X509KeyPair(l.webSocketTLSCertificate, l.webSocketTLSCertificateKey)
			if err != nil {
				panic(err)
			}
			server.WebSocket.Lock()
			server.WebSocket.TLSConfig = &tls.Config{Certificates: []tls.Certificate{cert}}
			server.WebSocket.Unlock()
		}
		server.StartInBackground()
	default:
		server = l.NewLocalServer(s, port)
	}
	return server
}
```

byzcoin/service.go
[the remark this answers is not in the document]
Gilthoniel: Let's put it as `log.Lvl3`

byzcoin/bcadmin/main.go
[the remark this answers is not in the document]
ineiti: Yes, but:

```
// NewLocalTest creates a new Local handler that can be used to test protocols
// locally
func NewLocalTest(s network.Suite) *LocalTest {
	dir, err := ioutil.TempDir("", "onet")
	if err != nil {
		log.Fatal("could not create temp directory: ", err)
	}

	return &LocalTest{
		Servers:    make(map[network.ServerIdentityID]*Server),
		Overlays:   make(map[network.ServerIdentityID]*Overlay),
		Services:   make(map[network.ServerIdentityID]map[ServiceID]Service),
		Trees:      make(map[TreeID]*Tree),
		Nodes:      make([]*TreeNodeInstance, 0, 1),
		Check:      CheckAll,
		mode:       Local,
		ctx:        network.NewLocalManager(),
		Suite:      s,
		path:       dir,
		latestPort: 2000,
	}
}
```

So it's in `Local` by default, and doing `NewLocalServer` - no?

byzcoin/service.go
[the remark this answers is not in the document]
ineiti: Oh, but then what about the caller of `catchupFromID`? If I return an error, he will print the error...

byzcoin/bcadmin/main.go
ineiti: Hmm - difficult. If I move it all to `api.go`, I don't see a nice way to handle the `LocalTest` I'm using. Also, I think being able to replay things outside of ByzCoin makes also sense to other projects, e.g., if somebody wants to re-create the state of a given block. So I'd rather vote to have this as a useful export of the memory staging trie.

byzcoin/bcadmin/main.go
ineiti: s/follow/replay/

byzcoin/bcadmin/main.go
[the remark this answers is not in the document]
Gilthoniel: Oh god yes I missed that, nevermind then.

byzcoin/service.go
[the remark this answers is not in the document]
ineiti: Added a lock.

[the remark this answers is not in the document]
Gilthoniel: `log.Lvl3` and return nil. We can consider that as not an actual error.

byzcoin/bcadmin/main.go
Gilthoniel: What's wrong with using `LocalTest` there ?
Well it's funny because I had the same thinking so this piece of code could be available with the Client (instead of being private in bcadmin) so other projects don't have to deal with the logic of the trie.

Gilthoniel: After talking with Jeff, we agreed that the logic inside the for loop should be a function provided by the byzcoin service. The function would take a callback that returns a block like the line 1096. Does it make sense ?

ineiti: No, I don't see how and why you want to add a callback in the byzcoin service. Also, should this be authenticated like I proposed?

ineiti:
moved the `bcadmin debug replay` part to https://github.com/dedis/cothority/pull/1858

byzcoin/bcadmin/main.go
ineiti: Closing this conversation - please have a look at #1858

jeffallen:
Somehow the merge left things not working, so I've rebased onto master instead. I needed to update the exact same things, and "make proto" still shows no changes, so... no idea if this build will work.

PR #1852, opened 2019-05-03
bcadmin: Adds a new option to edit a Darc's description

### LINKED ISSUE #1811: `bcadmin darc` cannot change the description of an existing darc
Add that feature. Add an integration test.

(Probably a startup task for Noémien.)

nkcr:
Runing `bcadmin darc cdesc --did <darc id> --desc <descritpion>` will update the description of the provided darc id. 
If `--did` is not provided, it uses the config darc admin.

Fixes #1811


ineiti:
Why is it `darc cdesc` and not `darc desc`?

nkcr:
> Why is it `darc cdesc` and not `darc desc`?

`cdesc` stands for "Change Description".

ineiti:
Given that the other commands are:
- `bcadmin darc add`
- `bcadmin darc rule`
- `bcadmin darc show`
I think `bcadmin darc desc` is more in line that `bcadmin darc cdesc`...

nkcr:
They are all verbs and specify actions, except `rule`, where an additional verb is used, like 
* `bcadmin darc rule --replace`. 
or
* `bcadmin darc rule --delete`

This makes sense since there are multiple actions based on `rule`. But this is not the case with `desc`, where the only action is to update the description.

So, to keep this logic, it would be either:

* `bcadmin darc desc --update`
or
* `bcadmin darc cdesc`

I prefer the later one since I only provide one action, and adding a "second level" of parameter is unnecessary. (And not using a verb to describe an action makes things unclear.)

byzcoin/bcadmin/main.go
```diff
@@ -1062,6 +1082,96 @@ func darcShow(c *cli.Context) error {
...
+	}
+	if err != nil {
+		return err
+	}
+
+	counters, err := cl.GetSignerCounters(signer.Identity().String())
+
+	invoke := byzcoin.Invoke{
+		ContractID: byzcoin.ContractDarcID,
+		Command:    "evolve_unrestricted",
```
jeffallen: This should be "evolve", because we are not changing the rules. See the comments on the top of contract_darc.go.

byzcoin/bcadmin/test.sh
```diff
@@ -236,4 +237,16 @@ testQR() {
...
 
+testUpdateDarcDesc() {
+  # We update the description of the latest darc, then we get the latest darc
+  # and check if the description changed.
+  runCoBG 1 2 3
+  runGrepSed "export BC=" "" runBA create --roster public.toml --interval .5s
+  eval $SED
+  [ -z "$BC" ] && exit 1
+
+  testOK runBA darc cdesc --desc "New description"
```
jeffallen: Please make a new darc without evolve_unrestricted rights, and see that it can update it's own description.

nkcr:
Thank you for the review @jeffallen. I implemented the suggested changes.

PR #1853, opened 2019-05-04
checks that proto-files are updated

ineiti:
Lately I find myself often with stalled proto-files. This checks that all proto-
files are correctly updated. It does not run
'protoc', as this is very nitpicky in creating files and changes from one version
to another.


.travis.yml
```diff
@@ -14,9 +14,11 @@ matrix:
         - go get github.com/dedis/Coding || true
 
       script:
+        - set -e
+        - make test_proto
         - make -C conode verify
-        - env GO111MODULE=on make test
```
Gilthoniel: why this change ?

Makefile
```diff
@@ -19,6 +19,11 @@ test_playground:
 	done;
 
 proto:
+	@pv=`protoc --version`; \
```
Gilthoniel: This is already done in the proto.sh file, no ?

Gilthoniel: Ok I reached the change in proto.sh. Is it because Travis uses a different version and you can't install the 3.6.1 one ?

.travis.yml
ineiti: To put all tests that take little time at the beginning. So you don't have to wait for the `make test` that takes a long time, before finding out that your PR fails because you forgot to update the proto-files.

Makefile
ineiti: `protoc` is not installed at all in travis, so I would've to make sure that it is installed in the correct version. And in fact `protoc` is never used in `proto.sh` at all, only in `external/Makefile`. So it should go in there.

ineiti: Put check of protoc-version in `external/Makefile`.

PR #1856, opened 2019-05-07
Implements the spawn and invoke update of the value contract

nkcr:
This is the first step of using deferred contracts with bcadmin. The plan is to implement the manipulation of contracts in the `bcadmin/clicontracts`* package.
 
Each contract should be usable on its own. For example, creating a new value contract with `bcadmin contract value spawn --value "myValue"`), but also on its deferred version, where a pipe and a special param (like `--redirect`) is used to redirect the transaction to a deferred contract:

- `bcadmin contract value spawn --value "myValue" --redirect | bcadmin contract deferred spawn`

with this, the value contract is not directly spawned, but the transaction is passed as argument to the spawn of a deferred contract.

**The changes of this PR**:
- Moves utility functions to lib/utils.go
- Adds a new package "clicontracts", which will hold the handling of contracts for bcadmin
- Implements the value contract in the "clicontracts" package

We use the following logic to manipulate contracts:
- `bcadmin contract <contract name> <action> [command] [--<arg1> <value arg1>, ... ] [--instID <instID>]`

For example (this is what is implemented with this PR):
- `bcadmin contract value spawn --value "myValue"`
- `bcadmin contract value invoke update --value "newValue" --instID deadbeef1234`

 (* "cli" stands for "Command Line Interface". This idea is to provide the "cli" version of a contract. Let me know if there is a better name.)

### DIFF
--- a/byzcoin/bcadmin/clicontracts/value.go
+++ b/byzcoin/bcadmin/clicontracts/value.go
@@ -1,0 +1,177 @@
+package clicontracts
+
+import (
+	"encoding/hex"
+	"errors"
+	"fmt"
+
+	"go.dedis.ch/cothority/v3/byzcoin"
+	"go.dedis.ch/cothority/v3/byzcoin/bcadmin/lib"
+	"go.dedis.ch/cothority/v3/byzcoin/contracts"
+	"go.dedis.ch/cothority/v3/darc"
+	"gopkg.in/urfave/cli.v1"
+)
+
+// ValueSpawn is used to spawn a new contract.
+func ValueSpawn(c *cli.Context) error {
+	bcArg := c.String("bc")
+	if bcArg == "" {
+		return errors.New("--bc flag is required")
+	}
+
+	value := c.String("value")
+	if value == "" {
+		return errors.New("--value flag is required")
+	}
+	valueBuf := []byte(value)
+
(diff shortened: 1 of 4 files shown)


byzcoin/bcadmin/clicontracts/value.go
```diff
@@ -0,0 +1,177 @@
+package clicontracts
```
jeffallen: How about if this directory is simply called `contracts`. Then there would be byzcoin/contracts/*.go and byzcoin/bcadmin/contracts/*.go for the server and client sides.

nkcr: That's what I originally did but it seems that this is not allowed. I get the following error:
```
contracts redeclared as imported package name previous declaration at...
```

jeffallen: The fix for this, if you wanted, would have been to rename the import of the byzcoin/contracts package, like this:

```
package conrtacts
import (
  bcContracts "go.dedis.ch/cothority/v3/byzcoin/contracts"
)
```

But calling this directory clicontracts is fine too.

jeffallen:
Please resolve the conflict and then I'll merge it.

nkcr:
@jeffallen, up to date with the master branch now!

PR #1858, opened 2019-05-07
adds bcadmin debug replay command  …

ineiti:
Before a new byzcoin version is released, it must be verified that the new contracts
correctly create the same StateChanges given the same inputs. The `bcadmin debug replay`
fetches all blocks and applies them one by one and verifies the merkle tree root of
the global state for each block.

### DIFF
--- a/byzcoin/bcadmin/main.go
+++ b/byzcoin/bcadmin/main.go
@@ -1191,7 +1191,7 @@ func debugReplay(c *cli.Context) error {
 
 	cl := skipchain.NewClient()
 	cb := func(ro *onet.Roster, sib skipchain.SkipBlockID) (*skipchain.SkipBlock, error) {
-		return cl.GetSingleBlock(r, sib)
+		return cl.GetSingleBlock(ro, sib)
 	}
 
 	_, err = s.ReplayState(bcID, r, cb)
--- a/byzcoin/bcadmin/main.go
+++ b/byzcoin/bcadmin/main.go
@@ -27,6 +27,7 @@ import (
 	"go.dedis.ch/cothority/v3/byzcoin/contracts"
 	"go.dedis.ch/cothority/v3/darc"
 	"go.dedis.ch/cothority/v3/darc/expression"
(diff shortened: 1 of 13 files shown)


ineiti:
@Gilthoniel
Can you move all that logic to the byzcoin Client (api.go) so that we can keep the memory staging trie as private ?

@ineiti
Hmm - difficult. If I move it all to api.go, I don't see a nice way to handle the LocalTest I'm using. Also, I think being able to replay things outside of ByzCoin makes also sense to other projects, e.g., if somebody wants to re-create the state of a given block. So I'd rather vote to have this as a useful export of the memory staging trie.

@Gilthoniel
What's wrong with using LocalTest there ?
Well it's funny because I had the same thinking so this piece of code could be available with the Client (instead of being private in bcadmin) so other projects don't have to deal with the logic of the trie.

@Gilthoniel
After talking with Jeff, we agreed that the logic inside the for loop should be a function provided by the byzcoin service. The function would take a callback that returns a block like the line 1096. Does it make sense ?

@ineiti
No, I don't see how and why you want to add a callback in the byzcoin service. Also, should this be authenticated like I proposed?

ineiti:
Note for later: if this goes in, it should not go in as a service endpoint, but only as a 'library' in `byzcoin`. The main reason I want to include this is to check that the chain will be OK _before_ it is deployed on all nodes. And this is of course not possible if this code goes in as a service endpoint...

ineiti:
Discussing this with @jeffallen, we found one very strong argument to make the `StateTrie` public: the data in the blocks is public, and as such it should be possible for anybody interested to work on that data to use it. If `StateTrie` is private to `byzcoin`, people would have to implement their own `StateTrie` if they want to do things like snapshotting or other. From that point of view, there is no reason to keep `StateTrie` private.

Compare this to Ethereum, where their libraries allow you to work on all the data without having to re-implement any of the data structures.

Gilthoniel:
That's why the function I'm writing returns a `ReadOnlyStateTrie` which is public so you can ask for the global state of a block and then read the values.

PR #1863, opened 2019-05-09
KyberJS: add pairing module to the bundle

Gilthoniel:
This updates the index to also export the pairing module so that
it is available in the bundle.


Gilthoniel:
@PizzaWhisperer

PR #1864, opened 2019-05-09
 Adds a string representation of Deferred data

nkcr:
This is a small step toward using deferred contract from bcadmin.


byzcoin/contract_deferred.go
```diff
@@ -43,6 +43,29 @@ type DeferredData struct {
 	ExecResult [][]byte
 }
 
+// String returns a human readable string representation of the deferred data
+func (dd DeferredData) String() string {
```
jeffallen: Performance for this particular use case does not matter, but string concatenation like this in Go causes lots of copying and garbage creation. It would be better to use https://godoc.org/strings#Builder.

nkcr: Thanks for the tip! 👍   (I based it on what I found in `byzcoin/transaction.go:222`)

PR #1867, opened 2019-05-09
bcadmin: adds --redirect for value contract

nkcr:
When using

```
$ bcadmin contract value spawn --value myValue
```

with the `--redirect` option, it does not spawn the value contract but redirects the transaction to stdout, later allowing this output to be used as the input of a spawn:deferred with pipe. (like `bcadmin contract value spawn --value myValue --redirect | bcadmin contract deferred spawn`)

This is a small step towards using bcadmin with deferred contract.


byzcoin/bcadmin/clicontracts/value.go
```diff
@@ -63,6 +66,30 @@ func ValueSpawn(c *cli.Context) error {
 		},
 	}
 
+	redirect := c.Bool("redirect")
```
jeffallen: Shell redirection is what you do when you have some output and want to send it someplace else. To me, it doesn't have a meaning in CLI apps.

I propose `--export` instead. Then you can export to a file, or redirect stdout to a pipe and export directly to `bcadmin deferred` to post the tx and give the first sig.

Also, I expected the --export flag to be be global: `bcadmin --export contract value add`. So that it can take effect on all kinds of bcadmin operations later, like `bcadmin --export roster del XXX`.

That would mean that you'd wrap AddTransactionAndWait with something that either exports to stdout, or does the real transaction, if --export is not present.

In CLI apps, orthogonality is your friend.

PR #1868, opened 2019-05-10
Bcadmin: deferred clicontract

nkcr:
Finally implements the cli version of a deferred contract. I added all the functions except delete of the deferred contract: 

* Spawn:deferred
* Invoke:deferred.addProof
* Invoke:deferred.execProposedTx

Combined with the value contract, we can successfully defer the invocation of a transaction (at least with the value contract yet). Here is a scenario that works:

```bash
# Run the nodes, create roster and set up the config
➜ ./run_nodes.sh -n 5 -c -t -v 2
➜ export BC_CONFIG=~/GitHub/cothority/conode
➜ bcadmin create -roster ~/GitHub/cothority/conode/public.toml 

# Add the rules specific to the value and deferred contracts.
# We use the admin identity.
➜ bcadmin darc rule -rule spawn:value --identity ed25519:... 
➜ bcadmin darc rule -rule spawn:deferred --identity ed25519:...                                                         
➜ bcadmin darc rule -rule invoke:deferred.addProof --identity ed25519:...        
➜ bcadmin darc rule -rule invoke:deferred.execProposedTx --identity ed25519:...                                                                                                                                                     

# Spawn a value contract, but redirect the transaction to the spawn of a deferred contract
➜ bcadmin contract value spawn --value myValue --redirect | bcadmin contract deferred spawn

# Add the proof on the single instruction of the deferred transaction 
# (the --hash and --instID values are given when we spawn the deferred contract)
➜ bcadmin contract deferred invoke addProof --hash ... --instID ... --iid 0

# Finally execute the deferred transaction.
# This will call the Spawn:value(myValue) transaction.
# If we hadn't called the addProof before, it wouldn't have worked.
➜ bcadmin contract deferred invoke execProposedTx --instID ...
```

Moreover, instead of using pipe, we can also save the proposed transaction to a file, and then pass it to the spawn of a deferred contract:

```bash
➜ bcadmin contract value spawn --value myValue --redirect > export.bin
➜ bcadmin contract deferred spawn < export.bin
```

### DIFF
--- a/byzcoin/bcadmin/clicontracts/deferred.go
+++ b/byzcoin/bcadmin/clicontracts/deferred.go
@@ -28,13 +28,13 @@ func DeferredSpawn(c *cli.Context) error {
 	// ---
 	proposedTransactionBuf, err := ioutil.ReadAll(os.Stdin)
 	if err != nil {
-		return errors.New("Failed to read from stding: " + err.Error())
+		return errors.New("failed to read from stding: " + err.Error())
 	}
 
 	proposedTransaction := byzcoin.ClientTransaction{}
 	err = protobuf.Decode(proposedTransactionBuf, &proposedTransaction)
 	if err != nil {
-		return errors.New("Failed to decode transaction: " + err.Error())
+		return errors.New("failed to decode transaction: " + err.Error())
 	}
 
 	// ---
@@ -184,7 +184,7 @@ func DeferredInvokeAddProof(c *cli.Context) error {
 	}
 	hash, err := hex.DecodeString(hashStr)
 	if err != nil {
-		return errors.New("Coulndn't decode the hash string: " + err.Error())
+		return errors.New("coulndn't decode the hash string: " + err.Error())
 	}
 
 	instID := c.String("instID")
@@ -207,12 +207,12 @@ func DeferredInvokeAddProof(c *cli.Context) error {
 	identity := signer.Identity()
 	identityBuf, err := protobuf.Encode(&identity)
 	if err != nil {
-		return errors.New("Coulndn't encode the identity: " + err.Error())
+		return errors.New("coulndn't encode the identity: " + err.Error())
 	}
 
 	signature, err := signer.Sign(hash)
 	if err != nil {
-		return errors.New("Couldn't sign the hash: " + err.Error())
+		return errors.New("couldn't sign the hash: " + err.Error())
 	}
 
 	// ---
--- a/byzcoin/bcadmin/clicontracts/deferred.go
+++ b/byzcoin/bcadmin/clicontracts/deferred.go
@@ -1,0 +1,374 @@
+package clicontracts
+
+import (
+	"encoding/binary"
+	"encoding/hex"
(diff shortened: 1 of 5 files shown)


byzcoin/bcadmin/clicontracts/deferred.go
```diff
@@ -0,0 +1,374 @@
...
+	if sstr == "" {
+		signer, err = lib.LoadKey(cfg.AdminIdentity)
+	} else {
+		signer, err = lib.LoadKeyFromString(sstr)
+	}
+	if err != nil {
+		return err
+	}
+
+	expireBlockIndexInt := uint64(6000)
```
Gilthoniel: What's the reasoning behind 6000 ?

byzcoin/bcadmin/main.go
```diff
@@ -422,6 +426,94 @@ var cmds = cli.Commands{
 					},
 				},
 			},
+			{
+				Name:  "deferred",
+				Usage: "Manipulate a vadeferred contract",
```
Gilthoniel: vadeferred ?

byzcoin/bcadmin/main.go
```diff
@@ -422,6 +426,94 @@ var cmds = cli.Commands{
...
+							},
+							cli.StringFlag{
+								Name:  "sign",
+								Usage: "public key of the signing entity (default is the admin public key)",
+							},
+						},
+					},
+					{
+						Name:  "invoke",
+						Usage: "invokw on a deferred contract ",
```
Gilthoniel: invoke

byzcoin/bcadmin/main.go
```diff
@@ -422,6 +426,94 @@ var cmds = cli.Commands{
...
+								Action: clicontracts.DeferredInvokeAddProof,
+								Flags: []cli.Flag{
+									cli.StringFlag{
+										Name:   "bc",
+										EnvVar: "BC",
+										Usage:  "the ByzCoin config to use (required)",
+									},
+									cli.UintFlag{
+										Name:  "iid",
+										Usage: "the instruction index of the transaction (starts from 0) (fefault is 0)",
```
Gilthoniel: default

byzcoin/bcadmin/clicontracts/deferred.go
nkcr: For the moment it is an arbitrary high value because I have to specify an expireBlockIndex to the deferred contract, but don't want the client to specify it from the command line (the client should only have to add `--redirect` and nothing else to specify a proposed transaction). 
The plan is to remove this parameter here and make it an optional argument for the deferred contract, so that the deferred contract can use a default value when `expireBlockIndex` is not given (like `current_blockIdx + 100`).

nkcr: If #1875 is accepted before this one I can refactor this part.

Gilthoniel: It's ok, you can do it in another PR

PR #1876, opened 2019-05-14
Adds get and delete methods to clicontracts

nkcr:
Until now only the value contract had the "get" method. 

In this PR, I add "get" and "delete" methods to the value and deferred clicontracts.

The get method is particularly convenient for testing. I could one more time verify that a deferred transaction is successfully created by performing a "get" of a proposed transaction after executing `invoke:deferred.execProposedTx`.

I also changed the "--iid" flag to "--instID" for consistency.

### DIFF
--- a/byzcoin/bcadmin/clicontracts/README.md
+++ b/byzcoin/bcadmin/clicontracts/README.md
@@ -67,9 +67,11 @@ bcadmin contract deferred invoke addProof --hash ... --instID ... --iid 0
 
 ```bash
 # Run the nodes, create roster and set up the config
-./run_nodes.sh -n 5 -c -t -v 2
-export BC_CONFIG=~/GitHub/cothority/conode
+~/GitHub/cothority/conode/run_nodes.sh -n 5 -c -t -v 2
 bcadmin create -roster ~/GitHub/cothority/conode/public.toml 
+
+# Copy/Paste from the output of the previous command
+export BC="..."
 
 # Add the rules specific to the value and deferred contracts.
 # We use the admin identity.
--- a/byzcoin/bcadmin/clicontracts/README.md
+++ b/byzcoin/bcadmin/clicontracts/README.md
@@ -62,3 +62,31 @@ Invoke an addProof on a deferred contract:
 # The --hash and --instID values are given when we spawn the deferred contract
 bcadmin contract deferred invoke addProof --hash ... --instID ... --iid 0
 ```
+
+**Working scenario**:
+
+```bash
+# Run the nodes, create roster and set up the config
+./run_nodes.sh -n 5 -c -t -v 2
+export BC_CONFIG=~/GitHub/cothority/conode
+bcadmin create -roster ~/GitHub/cothority/conode/public.toml 
+
+# Add the rules specific to the value and deferred contracts.
+# We use the admin identity.
+bcadmin darc rule -rule spawn:value --identity ed25519:... 
+bcadmin darc rule -rule spawn:deferred --identity ed25519:...                                                         
+bcadmin darc rule -rule invoke:deferred.addProof --identity ed25519:...        
+bcadmin darc rule -rule invoke:deferred.execProposedTx --identity ed25519:...                                                                                                                                                     
+
+# Spawn a value contract, but redirect the transaction to the spawn of a deferred contract
+bcadmin contract value spawn --value myValue --redirect | bcadmin contract deferred spawn
+
+# Add the proof on the single instruction of the deferred transaction 
+# (the --hash and --instID values are given when we spawn the deferred contract)
+bcadmin contract deferred invoke addProof --hash ... --instID ... --iid 0
+
+# Finally execute the deferred transaction.
+# This will call the Spawn:value(myValue) transaction.
+# If we hadn't called the addProof before, it wouldn't have worked.
+bcadmin contract deferred invoke execProposedTx --instID ...
+```
--- a/byzcoin/bcadmin/clicontracts/deferred_test.sh
+++ b/byzcoin/bcadmin/clicontracts/deferred_test.sh
@@ -117,7 +117,7 @@ testGet() {
         }'`
     echo -e "Here is the instance ID:\t$DEFERRED_INSTANCE_ID"
 
-    # We know the array conaining the hash to sign is the second line after
+    # We know the array containing the hash to sign is the second line after
     # "- Instruction hashes:" and we remove the "--- " prefix.
     HASH=`echo "$OUTRES" | sed -n ' 
         /- Instruction hashes:/ {
@@ -128,7 +128,7 @@ testGet() {
         }'`
     echo -e "Here is the hash:\t\t$HASH"
 
-    # We know use the get function to check if have the right informations:
+    # We now use the get function to check if we have the right informations:
     OUTRES=`runBA contract deferred get --instID $DEFERRED_INSTANCE_ID`
     testGrep "action: spawn:value" echo "$OUTRES"
     testGrep "identities: \[\]" echo "$OUTRES"
@@ -139,7 +139,7 @@ testGet() {
     
     testOK runBA contract deferred invoke addProof --instID "$DEFERRED_INSTANCE_ID" --hash "$HASH" --instrIdx 0 --sign "$KEY" --darc "$ID"
 
-    # Since we prformed an addProof, the result should now contrain a new
+    # Since we performed an addProof, the result should now contrain a new
     # identity and the field signature set to 1.
     OUTRES=`runBA contract deferred get --instID $DEFERRED_INSTANCE_ID`
     testGrep "action: spawn:value" echo "$OUTRES"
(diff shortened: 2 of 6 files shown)


byzcoin/bcadmin/clicontracts/README.md
```diff
@@ -61,4 +61,32 @@ Invoke an addProof on a deferred contract:
...
+bcadmin darc rule -rule spawn:deferred --identity ed25519:...                                                         
+bcadmin darc rule -rule invoke:deferred.addProof --identity ed25519:...        
+bcadmin darc rule -rule invoke:deferred.execProposedTx --identity ed25519:...                                                                                                                                                     
+
+# Spawn a value contract, but redirect the transaction to the spawn of a deferred contract
+bcadmin contract value spawn --value myValue --redirect | bcadmin contract deferred spawn
+
+# Add the proof on the single instruction of the deferred transaction 
+# (the --hash and --instID values are given when we spawn the deferred contract)
+bcadmin contract deferred invoke addProof --hash ... --instID ... --iid 0
```
jeffallen: I understand how you came up with this `bcadmin contract deferred invoke addProof` command, but it's just way too big. The cli tools are meant to be used by humans, and need to be short and possible to remember. For instance, `bcadmin -x ... | bcadmin deferred` to start one, and `bcadmin deferred sign $id` to sign one.

byzcoin/bcadmin/clicontracts/deferred_test.sh
```diff
@@ -83,4 +85,120 @@ testContractDeferredInvoke() {
...
+
+    # We know the instance ID is the next line after "Spawned new deferred contract..."
+    DEFERRED_INSTANCE_ID=`echo "$OUTRES" | sed -n ' 
+        /Spawned new deferred contract/ {
+            n
+            p
+        }'`
+    echo -e "Here is the instance ID:\t$DEFERRED_INSTANCE_ID"
+
+    # We know the array conaining the hash to sign is the second line after
```
jeffallen:
```suggestion
    # We know the array containing the hash to sign is the second line after
```

byzcoin/bcadmin/clicontracts/deferred_test.sh
```diff
@@ -83,4 +85,120 @@ testContractDeferredInvoke() {
...
+    HASH=`echo "$OUTRES" | sed -n ' 
+        /- Instruction hashes:/ {
+            n
+            n
+            s/--- //
+            p
+        }'`
+    echo -e "Here is the hash:\t\t$HASH"
+
+    # We know use the get function to check if have the right informations:
```
jeffallen:
```suggestion
    # We now use the get function to check if we have the right information.
```

byzcoin/bcadmin/clicontracts/deferred_test.sh
```diff
@@ -83,4 +85,120 @@ testContractDeferredInvoke() {
...
+    testGrep "action: spawn:value" echo "$OUTRES"
+    testGrep "identities: \[\]" echo "$OUTRES"
+    testGrep "counters: \[\]" echo "$OUTRES"
+    testGrep "signatures: 0" echo "$OUTRES"
+    testGrep "Spawn:	value" echo "$OUTRES"
+    testGrep "Args:value" echo "$OUTRES"
+    
+    testOK runBA contract deferred invoke addProof --instID "$DEFERRED_INSTANCE_ID" --hash "$HASH" --instrIdx 0 --sign "$KEY" --darc "$ID"
+
+    # Since we prformed an addProof, the result should now contrain a new
```
jeffallen:
```suggestion
    # Since we performed an addProof, the result should now contain a new
```

byzcoin/bcadmin/clicontracts/deferred_test.sh
```diff
@@ -83,4 +85,120 @@ testContractDeferredInvoke() {
...
+    # Note on the regex used in grep. We want to be sure an identity of form
+    # [ed25519:aef123] is added.
+    #
+    # \[            An opening angle bracket
+    # [             A group a chars that appears 1..*
+    #   :a-f0-9     Any hexadecimal chars and ":"
+    # ]+
+    # \]            A closing angle bracket
+    #
+    testGrep "identities: \[[:a-f0-9]+\]" echo "$OUTRES"
```
jeffallen: It seems like you should know the exact identity you expect here?

byzcoin/bcadmin/clicontracts/README.md
nkcr: Before anything, here is the convention I am using for the clicontrant commands:

```bash
   bcadmin contract CONTRACT { spawn  --bc <byzcoin config> 
                                      [--<arg name> <arg value>, ...]
                                      [--darc <darc id>] 
                                      [--sign <pub key>] 
                                      [--redirect],
                               invoke <command>
                                      --bc <byzcoin config>
                                      --instID <instance ID>
                                      [--<arg name> <arg value>, ...]
                                      [--darc <darc id>] 
                                      [--sign <pub key>],
                               get    --bc <byzcoin config>
                                      --instID <instance ID>,
                               delete --bc <byzcoin config>
                                      --instID <instance ID>
                                      [--darc <darc id>] 
                                      [--sign <pub key>]     
                             }
```


I knew I would need to fight for this case 😁, let me explain :

- **Short command line doesn't necessarily reduce the effort to remember**. I personally will have less of a tendency to forget if something is logical and follows a convention than if it's short. Having a convention also allows me to "rebuild" the command in case I forget it because it follows a standard, which brings the next point
- **Everyone one eventually forgets**, which means that, even if the command is shot, it is likely that you will have to look at the doc to re-learn the specific command you need to use. If a convention is used, one has only to remember the logic and convention used in order to fully appreciate the handling of clicontracts.
- **clicontract is a cli interface for a contract**, which is why I am not re-inventing nor simplifying commands, but rather **mapping** the contract implementation to a command line usage. This allows anyone that understands how contract are implemented to use the clicontract and vice-versa, without any cognitive effort requested to map special command to the corresponding contract function (in your proposal, you require users to understand that `bcadmin deferred` == `Spawn:deferred` and `bcadmin deferred sign` == `Invoke:deferred.addProof`. Establishing conventions and using them has the huge benefit of improving implementation, usage, maintenance, and learning of a system. In this case it also clearly describe what the system is doing.

Those points stand because there are more than one clicontracts. I would agree to use "special short" command it they were only one case of the deferred contract with - let's say - the value contract. But here I think it's very interesting to have a standard way to implement clicontracts, which is a very nice interface for handling them. 
(also I did not forget that somme of the command will move at the "contract" level, and --redirect will be renamed to --export).

byzcoin/bcadmin/clicontracts/README.md
```diff
@@ -61,4 +61,32 @@ Invoke an addProof on a deferred contract:
...
 # The --hash and --instID values are given when we spawn the deferred contract
 bcadmin contract deferred invoke addProof --hash ... --instID ... --iid 0
+```
+
+**Working scenario**:
+
+```bash
+# Run the nodes, create roster and set up the config
+./run_nodes.sh -n 5 -c -t -v 2
+export BC_CONFIG=~/GitHub/cothority/conode
```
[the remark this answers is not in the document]
nkcr: If it is not specified, the key and the config files are saved in the current location. I like to specify it in order to have all the mess in one place.

byzcoin/bcadmin/clicontracts/deferred_test.sh
nkcr: Nice observation, that's totally right!

byzcoin/bcadmin/clicontracts/README.md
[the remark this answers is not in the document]
Gilthoniel: Hm, if `BC_CONFIG` is not set it will go in a specific configuration folder (e.g. with Linux it goes in `.config`). Usual people shouldn't have to worry about this. If it really goes to your current location, there's a bug to fix.

byzcoin/bcadmin/clicontracts/README.md
kc1212: Maybe a way to satisfy the two conflicting points (easy to use vs following a logical standard) is to use aliases. Perhaps there's a `bcadmin deferred` that only contains aliases of the frequently used commands. Like `bcadmin deferred sign $inst --bc $config` = `bcadmin contract deferred invoke "sign" --bc $bc --instID $inst`. It could just be a shell script that uses `bcadmin` (not saying that's a good idea). Just like `git pull` is like an alias for fetching and then merging.

I think the current approach is fine. But when many more people start using deferred, we may need to provide an alias.

byzcoin/bcadmin/clicontracts/README.md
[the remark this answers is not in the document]
kc1212: I would also move this out of the README because people reading it might be confused and it might create two places where the configs are stored if they're not sure what they're doing.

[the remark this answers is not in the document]
nkcr: README updated! I also changed to an absolute path when running `./run_nodes.sh` since it might not be obvious where the script is.
@Gilthoniel yes you're right, when I checked I made the mistake to reset the BC_CONFIG with `export BC_CONFIG=""`, which is wrong.

byzcoin/bcadmin/clicontracts/README.md
nkcr: I'm totally for it! Let's add an alias when/if the need arises.

PR #1879, opened 2019-05-15
Add an option to use a different suite for DKG

### LINKED ISSUE #1878: DKG protocol is fixed to cothority.Suite
Our DKG protocol uses `cothority.Suite`, but it should be able to use other suites, e.g., a suite with G2 as its curve like in drand.

kc1212:
Fixes #1878

### DIFF
--- a/dkg/pedersen/dkg.go
+++ b/dkg/pedersen/dkg.go
@@ -183,7 +183,6 @@ func (o *Setup) rootStartDeal(replies []structInitReply) error {
 func (o *Setup) allStartDeal(ssd structStartDeal) error {
 	var err error
 	if o.NewDKG == nil {
-		// TODO: Should be newdistkeyhandler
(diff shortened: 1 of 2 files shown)


dkg/pedersen/dkg.go
```diff
@@ -174,7 +184,7 @@ func (o *Setup) allStartDeal(ssd structStartDeal) error {
 	var err error
 	if o.NewDKG == nil {
 		// TODO: Should be newdistkeyhandler
```
Gilthoniel: What about this todo ?

kc1212: Not too sure, `NewDistKeyGenerator` appears to be just a convenience method for `NewDistKeyHandler`. There no sign that `NewDistKeyGenerator` is going away so I'll remove the TODO.

PR #1885, opened 2019-05-16
bcadmin debug replay checks links

ineiti:
For investigating a cothority-ts misbehaviour I wanted to make sure that the forward-links
are correct. As this might also serve other verifications, I propose it as a PR.

### DIFF
--- a/byzcoin/statereplay.go
+++ b/byzcoin/statereplay.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 
+	"go.dedis.ch/kyber/v3/pairing"
+
 	"go.dedis.ch/cothority/v3/skipchain"
 	"go.dedis.ch/onet/v3"
 	"go.dedis.ch/onet/v3/log"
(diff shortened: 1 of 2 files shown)


byzcoin/statereplay.go
```diff
@@ -95,8 +97,36 @@ func (s *Service) ReplayState(id skipchain.SkipBlockID, ro *onet.Roster, cb Bloc
...
+					}
+				}
 				return nil, replayError(sb, "merkle tree root doesn't match with trie root")
 			}
+
+			log.Lvl2("Checking links for block", sb.Index)
+			pubs := sb.Roster.ServicePublics(skipchain.ServiceName)
+			for j, fl := range sb.ForwardLink {
+				if fl.From == nil || fl.To == nil ||
+					len(fl.From) == 0 || len(fl.To) == 0 {
```
jeffallen: FYI, the length of a nil slice is 0, not a crash. So some of these checks are not needed. But NBD (no big deal).

PR #1887, opened 2019-05-16
Improvements and fix for deferred

nkcr:
This PR contains the side stuff that had to be implemented/refactored in order to use the deferred update_config from bcadmin.

This PR does the following:

- Adds a `VerifyDeferredInstruction(...)` for the Darc and Config contracts
- Fixes a wrong mechanism in the deferred contract when getting contract instances of proposed transactions
- Adds a String representation of a ChainConfig struct

I have another PR ready that depends on this one, which implements the deferred update_config of from bcadmin.


kc1212:
Are there tests for this PR? or it's in another one?

nkcr:
You are right 👍  I'm adding a scenario that uses "Invoke:config.update_config" as the proposed transaction.

PR #1892, opened 2019-05-17
Adds config clicontract with invoke:update_config

nkcr:
It is now possible to update the contract config with a deferred transaction. There is a complete scenario in the readme, but here is the main command:

```bash
bcadmin contract config invoke updateConfig --blockInterval 7s \
                                            --maxBlockSize 5000000 \
                                            --darcContractIDs darc,darc2 \
                                            --redirect | bcadmin contract deferred spawn
```

UPDATE:

I added the get method, which displays the latest chain config:

```bash
bcadmin contract config get
```

### DIFF
--- a/byzcoin/bcadmin/clicontracts/config_test.sh
+++ b/byzcoin/bcadmin/clicontracts/config_test.sh
@@ -2,6 +2,7 @@
 
 testContractConfig() {
     run testContractConfigInvoke
+    testContractConfigGet
 }
 
 # In this test we check the behavior of the invoke:update_config function. We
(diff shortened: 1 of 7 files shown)


kc1212:
@nkcr can you fix the conflict please?

byzcoin/bcadmin/clicontracts/config_test.sh
```diff
@@ -0,0 +1,66 @@
+# This method should be called from the byzcoin/bcadmin/test.sh script
+
+testContractConfig() {
+    run testContractConfigInvoke
+    testContractConfigGet
```
kc1212: does this one need a `run`?

nkcr:
Good point for the missing `proof.Verify(...)`. I added them to the other clicontracts too.

PR #1893, opened 2019-05-17
Prevent block with infinite loop

Gilthoniel:
This cleans the forward links before proposing a block to prevent
old valid forward links to be stored in new blocks.


Gilthoniel:
This is a partial fix of #1894 but it doesn't close the issue.

PR #1895, opened 2019-05-17
Improvements to E-voting

jeffallen:
Previously, e-voting authentication was designed to be
decentralised, with the signature being posted in the blocks
and the signature being checked during the verification of the
blocks by the conodes.

But because the signature was essentially a bearer token,
it must remain confidential, or else it can be used for
replay attacks.

With this change, the signature is now private data, which is
only known by the client and by the lead conode. The lead conode
becomes a centralised point of authentication, which is no worse
than our current system, which already has a centralised
authentication system (Tequila). In production, the same
organisation operates Tequila as the lead conode.

The process of shuffling and decrypting the votes is still
executed in a decentralised manner, with the private key
for an election still sharded across the cothority.


ineiti:
Why hasn't this been solved with the proxy-authentication? That seemed to be the correct way forward... What am I missing?

If you want to move to a decentralized authentication system now (like in the `credential` contract of ByzCoin), it will be difficult...

jeffallen:
@ineiti proxy auth requires cooperation from the auth provider that we were not willing/interested to require from Tequila, as well as heavy front-end changes I was not willing to get involved in making to this app. However, with time short before the election, I realised that posting the signature into the blocks is (and was last year also) risky, because an attacker could replay the admin's signature from the election open, and open other elections, or close the current election.

@Gilthoniel I agree that it's not 100% clear how a follower decides if a skipblock comes from the leader or not, so I will add a leader signature, taking the place of the previous user signature. That way there will be a chain of (centralised) trust from Tequila thru the lead conode and on to the followers.

evoting/lib/transaction.go
```diff
@@ -79,31 +75,24 @@ func NewTransaction(data interface{}, user uint32, signature []byte) *Transactio
...
-		election, _ := GetElection(s, genesis, false, t.User)
-		if election == nil {
-			return nil
-		}
-		message = election.Master
+// Sign generates a signature on the marshaled form of the txn,
+// without a Signature, then puts that signature into the txn.
+func (t *Transaction) Sign(priv kyber.Scalar) error {
+	t.Signature = nil
+	data, err := protobuf.Encode(t)
```
Gilthoniel: I would somehow add something more to the signature because if someone votes twice, an attacker could potentially replay the first vote to erase the last one, right ?

jeffallen: hmm, how about the block index? seems doable, but will cause me to re-refactor the code.

PR #1896, opened 2019-05-17
Deferred contract: Adds test scenarios

nkcr:
- execution of a proposed transaction where two signers are required
- delete of a deferred instance

### DIFF
--- a/byzcoin/contracts/contract_deferred_test.go
+++ b/byzcoin/contracts/contract_deferred_test.go
@@ -99,7 +99,7 @@ func TestDeferred_ScenarioSingleInstruction(t *testing.T) {
 	}
 	require.Nil(t, ctx.FillSignersAndSignWith(signer))
 
-	_, err = cl.AddTransaction(ctx)
+	_, err = cl.AddTransactionAndWait(ctx, 10)
 	require.Nil(t, err)
 
 	myID := ctx.Instructions[0].DeriveID("")
@@ -165,7 +165,7 @@ func TestDeferred_ScenarioSingleInstruction(t *testing.T) {
 	}
 	require.Nil(t, ctx.FillSignersAndSignWith(signer))
 
-	_, err = cl.AddTransaction(ctx)
+	_, err = cl.AddTransactionAndWait(ctx, 10)
 	require.Nil(t, err)
 
 	proposedTransaction.Instructions[0].SignerIdentities = append(proposedTransaction.Instructions[0].SignerIdentities, identity)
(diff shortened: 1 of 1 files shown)


nkcr:
This was discussed in https://github.com/dedis/cothority/pull/1837#discussion_r281644554

byzcoin/contracts/contract_deferred_test.go
```diff
@@ -2276,3 +2277,666 @@ func TestDeferred_ScenarioUpdateConfig(t *testing.T) {
...
+	dataBuf, _, _, err = pr.Get(myID.Slice())
+	require.Nil(t, err)
+
+	result = byzcoin.DeferredData{}
+	err = protobuf.Decode(dataBuf, &result)
+	require.Nil(t, err)
+
+	// We can not do this test because the identities have to be compared using
+	// the Equal() method.
+	//require.Equal(t, result.ProposedTransaction, proposedTransaction)
```
kc1212: this won't work because `proposedTransaction` doesn't have the signer identities? you could add the expected signer identities to `proposedTransaction` and then compare the hash of `proposedTransaction` with the hash of `result.ProposedTransaction`, then u won't need to do any of the comparison below because the signer identity is a part of the hash.

byzcoin/contracts/contract_deferred_test.go
```diff
@@ -2276,3 +2277,666 @@ func TestDeferred_ScenarioUpdateConfig(t *testing.T) {
...
+			Invoke: &byzcoin.Invoke{
+				ContractID: byzcoin.ContractDeferredID,
+				Command:    "execProposedTx",
+			},
+			SignerCounter: []uint64{4},
+		}},
+	}
+	require.Nil(t, ctx.FillSignersAndSignWith(signer))
+
+	_, err = cl.AddTransaction(ctx)
```
kc1212: you can change a lot of `AddTransaction` to `AddTransactionAndWait(ctx, 10)` then it will only return when the transaction is included in the block. that way you don't need to do `time.Sleep(2 * genesisMsg.BlockInterval)` which may cause the test to run slower.

nkcr: Nice observation! Tests are now 2 times faster 🏆

byzcoin/contracts/contract_deferred_test.go
nkcr: True! Although we can only compute the hash of an instruction (right?)

kc1212: there should be a function that computes the hash for a list of instructions

PR #1900, opened 2019-05-21
extends Rule and Credential

ineiti:
This PR adds some useful methods to two clases:

- CredentialStruct: setCredential, delAttribute
- Rule: append, remove, which work directly on the expression

It also moves the OR and AND from Rules to Rule.

### DIFF
--- a/external/js/cothority/src/darc/rules.ts
+++ b/external/js/cothority/src/darc/rules.ts
@@ -7,6 +7,9 @@ import { IIdentity } from "./identity-wrapper";
  */
 export class Rule extends Message<Rule> {
 
+    static OR = "|";
+    static AND = "&";
+
     /**
      * @see README#Message classes
      */
@@ -15,7 +18,7 @@ export class Rule extends Message<Rule> {
     }
 
     readonly action: string;
-    readonly expr: Buffer;
+    expr: Buffer;
 
     constructor(props?: Properties<Rule>) {
         super(props);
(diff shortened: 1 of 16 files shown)


external/js/cothority/src/darc/rules.ts
```diff
@@ -49,9 +91,6 @@ export class Rule extends Message<Rule> {
  */
 export default class Rules extends Message<Rules> {
 
-    static OR = "|";
```
Gilthoniel: I put those here because I'm expecting people to update/add rules from this class and not rule by rule. Is there a good reason to move it ?

external/js/cothority/src/personhood/credentials-instance.ts
```diff
@@ -280,6 +297,26 @@ export class CredentialStruct extends Message<CredentialStruct> {
...
 
+    /**
+     * Removes the attribute from the given credential. If the credential or the
+     * attribute doesn't exist, it returns 'undefined', else it returns the
+     * content of the deleted attribute.
+     *
+     * @param credential the name of the credential
+     * @param attribute the attribute to be deleted
+     */
+    delAttribute(credential: string, attribute: string): Buffer {
```
Gilthoniel: `deleteAttribute`. For three letters, it's nicer.

external/js/cothority/src/darc/rules.ts
ineiti: Well, the `OR` and `AND` are used in the `Rule` class, in the expression, not in `Rules`. Also, this PR proposes a more fine-grained approach to the expressions, not setting whole expressions at once, but adding the possibility to add or remove an identity.

external/js/cothority/src/darc/rules.ts
```diff
@@ -15,7 +18,7 @@ export class Rule extends Message<Rule> {
     }
 
     readonly action: string;
-    readonly expr: Buffer;
```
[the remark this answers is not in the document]
ineiti: No, because you have to be able to read the action of the rule...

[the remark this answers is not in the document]
Gilthoniel: I was only talking about the `expr` as it should have a function to test if a list of identities pass the expression but it doesn't need to be publicly available. But yes, `action` is readonly and should stay that way.

PR #1901, opened 2019-05-21
View change improvement

### LINKED ISSUE #1835: Leader gets stuck in view change
**To be investigated further**

1. The leader is listening for its heartbeat and it reaches timeout when trying to add a block which then triggers a view change request from the **current** leader to elect another node.

2. View change requests are triggered when a conode restarts after while. Basically it should wait for the catch up before starting the heartbeat monitor.

Gilthoniel:
This PR updates the view change to have an upper bound to the timeout value so that situation where the view change is triggered but the signature fails doesn't create a huge timeout.

Also: more verbose log for future debugging
Also: use two locks for catching and trie update

Fixes #1835

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -2201,78 +2201,95 @@ func (s *Service) startAllChains() error {
 	s.pollChan = make(map[string]chan bool)
 	s.pollChanMut.Unlock()
 
-	// Catch up is done before starting the chains to prevent undesired events
-	err = s.catchupAll()
-	if err != nil {
-		return fmt.Errorf("%v couldn't sync: %s", s.ServerIdentity(), err.Error())
+	// All the logic necessary to start the chains is delayed to a goroutine so that
+	// the other services can start immediately and are not blocked by Byzcoin.
+	go func() {
+		s.working.Add(1)
+		defer s.working.Done()
+
+		// Catch up is done before starting the chains to prevent undesired events
+		err = s.catchupAll()
+		if err != nil {
+			log.Errorf("%v couldn't sync: %s", s.ServerIdentity(), err.Error())
+			return
+		}
+
+		gas := &skipchain.GetAllSkipChainIDs{}
+		gasr, err := s.skService().GetAllSkipChainIDs(gas)
+		if err != nil {
+			log.Errorf("%v couldn't get the skipchains: %s", s.ServerIdentity(), err.Error())
+			return
+		}
+
+		for _, gen := range gasr.IDs {
+			err := s.startChain(gen)
+			if err != nil {
+				log.Error("catch up error: ", err)
+			}
+		}
+
+		s.monitorLeaderFailure()
+	}()
+
+	return nil
+}
+
+func (s *Service) startChain(genesisID skipchain.SkipBlockID) error {
+	if !s.hasByzCoinVerification(genesisID) {
+		return nil
 	}
 
-	gas := &skipchain.GetAllSkipChainIDs{}
-	gasr, err := s.skService().GetAllSkipChainIDs(gas)
+	interval, _, err := s.LoadBlockInfo(genesisID)
+	if err != nil {
+		return fmt.Errorf("%s Ignoring chain %x because we can't load blockInterval: %s",
+			s.ServerIdentity(), genesisID, err)
+	}
+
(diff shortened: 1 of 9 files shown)


byzcoin/service.go
```diff
@@ -1340,7 +1355,7 @@ func (s *Service) updateTrieCallback(sbID skipchain.SkipBlockID) error {
 	// Check if the polling needs to be updated.
 	s.pollChanMut.Lock()
 	scIDstr := string(sb.SkipChainID())
-	if nodeIsLeader {
+	if nodeIsLeader && !s.catchingUp {
```
kc1212: if not mistaken, the `s.catchingUp` flag is locked/unlocked by `s.updateTrieLock` instead of `s.catchingLock`, which seems a bit confusing. i'm wondering whether we can refactor the catchup logic into its own go-routine and the byzcoin service will interact with it only using channels. i'm not saying we should change anything now, just throwing the idea out there.

byzcoin/service.go
```diff
@@ -1261,13 +1274,15 @@ func (s *Service) updateTrieCallback(sbID skipchain.SkipBlockID) error {
...
-		if s.catchingUp {
-			log.Warn(s.ServerIdentity(), "Got new block while catching up - ignoring block for now")
-			return nil
-		}
+		log.Warn(s.ServerIdentity(), "Got new block while catching up - ignoring block for now")
+		go func() {
+			// This new block will catch up at the end of the current catch up (if any)
+			// and be ignored if the block is already known.
+			s.catchingLock.Lock()
+			s.catchUp(sb)
```
kc1212: does this need a `s.working.Add(1)` to make sure there are no leaky go-routines?

byzcoin/service.go
Gilthoniel: The problem here is that the callback after a block is stored is called during a catch up (it has to) but some actions like starting the heartbeat/monitor should not be trigger until the end of the catch up. So you need a way to tell if the callback is called in a catch up context, and I'm not sure that channels could solve this.

byzcoin/viewchange/viewchange.go
```diff
@@ -159,11 +158,11 @@ func (c *Controller) Start(myID network.ServerIdentityID, genesis skipchain.Skip
...
 				}
 			}
-			if meta.countOf(ctr) > 2*f && meta.stateOf(ctr) < startedTimerState && meta.acceptOf(ctr) {
+			if meta.countOf(ctr) >= 2*f && meta.stateOf(ctr) < startedTimerState && meta.acceptOf(ctr) {
 				// To avoid starting the next view-change too
 				// soon, start view-change timer after
-				// receiving 2*f+1 view-change messages.
-				timer.Reset(time.Duration(math.Pow(2, float64(ctr))) * initialDuration)
+				// receiving 2*f view-change messages.
+				timer.Reset(initialDuration * 2)
```
kc1212: i'm not too sure about this change. in section 4.5.2 of the PBFT paper, the time need to increase exponentially. this is because we're working in a partially synchronous network where we don't know the real network latency. the paper also says we need to wait for 2f+1 messages before starting the timer.

in my opinion, if the problem we're trying to solve is allow a 5-node cluster to operate when 2 nodes are down (at least try to make sure that the 3 nodes are not stuck), then I don't think this is the right fix. because this scenario violates the PBFT assumption and it's expected that the remaining nodes will be suck or get into an inconsistent state.

kc1212: The best approach I can think of is to upper bound the expiration timer. Because it's unrealistic for computer networks to have a delay of 5+ minutes. Alternatively you can write a bcadmin command to reset the viewchange expiration timer.

Gilthoniel: That's right for 2f+1, I misread a paragraph. The upper bound solution is fine I guess.

ineiti: I always thought it's `2f+2` for ByzCoin, no?

Gilthoniel: the view change is following this paper: http://pmg.csail.mit.edu/papers/osdi99.pdf which states `2f+1`

ineiti: Thanks - I was not aware that the view-change has another threshold than ByzCoin...

PR #1903, opened 2019-05-21
adds NoVerify option to GetProof

### LINKED ISSUE #1902: GetProof is very slow
On the current DEDIS-byzcoin `9cc3`, a `GetProof` takes about 5 seconds! So if there are multiple requests, it takes a long time before the proofs arrive.

ineiti:
The current GetProof is very slow, as it verifies the whole proof at the end of the call.
This takes about 500ms on the test-system for 50 blocks, or 5s on the DEDIS-ByzCoin.

This PR adds a NoVerify field to the GetProof call, so that the verification is
skipped. Of course the verification still needs to be done at the client.

Closes #1902


Gilthoniel:
I'm wondering if we should remove the verification from the handler as it doesn't make any sense to do it there because the client will have to verify anyway. What do you think of keeping this handler and removing the verification, and if we have actual calls from the service itself, adding a API function that will fetch proof and verify it ?

ineiti:
> I'm wondering if we should remove the verification from the handler as it doesn't make any sense to do it there because the client will have to verify anyway. What do you think of keeping this handler and removing the verification, and if we have actual calls from the service itself, adding a API function that will fetch proof and verify it ?

IIRC, I added the code because for some time the proofs were not valid (race-conditions). But I think we don't have that case anymore. And, as you write, the client has to verify that the proof is valid, anyway.

Proposing alternative here: #1905

PR #1906, opened 2019-05-22
Eventlog streaming

kc1212:
Implement event streaming in go and java. Under the hood, we still use the old skipblock streaming feature of byzcoin but parse the blocks into events using a handler wrapper. The streaming feature allows the user to stream from blocks in the past. In such cases, we read all the blocks from the one given by the client and then call the handler on those, then we call the handler on the new blocks as before (duplicates are removed in the process).

### DIFF
--- a/eventlog/api.go
+++ b/eventlog/api.go
@@ -3,6 +3,8 @@ package eventlog
 import (
 	"bytes"
 	"errors"
+
+	"go.dedis.ch/cothority/v3/skipchain"
 	"go.dedis.ch/onet/v3/network"
 
 	"go.dedis.ch/cothority/v3"
@@ -22,6 +24,7 @@ type Client struct {
 	Signers    []darc.Signer
 	Instance   byzcoin.InstanceID
 	c          *onet.Client
+	sc         *skipchain.Client
 	signerCtrs []uint64
 }
 
@@ -31,6 +34,7 @@ func NewClient(ol *byzcoin.Client) *Client {
 	return &Client{
 		ByzCoin:    ol,
 		c:          onet.NewClient(cothority.Suite, ServiceName),
+		sc:         skipchain.NewClient(),
 		signerCtrs: nil,
 	}
 }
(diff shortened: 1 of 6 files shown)


eventlog/api.go
```diff
@@ -202,3 +206,136 @@ func (c *Client) Search(req *SearchRequest) (*SearchResponse, error) {
...
 	return reply, nil
 }
+
+// StreamHandler is the signature of the handler used when streaming events.
+type StreamHandler func(event Event, blockID []byte, err error)
+
+// Close closes all the websocket connections.
+func (c *Client) Close() error {
+	err := c.ByzCoin.Close()
+	if err2 := c.sc.Close(); err != nil {
```
Gilthoniel: `err2 != nil` and 4 lines below also

eventlog/api.go
```diff
@@ -202,3 +206,136 @@ func (c *Client) Search(req *SearchRequest) (*SearchResponse, error) {
...
+		err = err2
+	}
+	if err2 := c.c.Close(); err != nil {
+		err = err2
+	}
+	return err
+}
+
+// StreamEvents is a blocking call where it calls the handler on even new event
```
Gilthoniel: -event- new event

eventlog/api.go
```diff
@@ -202,3 +206,136 @@ func (c *Client) Search(req *SearchRequest) (*SearchResponse, error) {
...
+		_ = handleBlocks(handler, resp.Block)
+	}
+	// the following blocks
+	return c.ByzCoin.StreamTransactions(h)
+}
+
+// StreamEventsFrom is a blocking call where it calls the handler on even new
+// event from (inclusive) the given block ID until the connection is closed or
+// the server stops.
+func (c *Client) StreamEventsFrom(handler StreamHandler, id []byte) error {
```
Gilthoniel: Should `id` be of type skipchain.SkipBlockID ?

external/java/src/main/java/ch/epfl/dedis/eventlog/EventLogInstance.java
```diff
@@ -172,6 +177,194 @@ public SearchResponse search(String topic, long from, long to) throws CothorityE
...
+                current = b;
+                super.receive(b);
+            }
+
+            for (Pair p : buffer) {
+                if (p.block == null) {
+                    super.error(p.error);
+                }
+                else if (p.block.getBackLinks().size() == 0) {
+                    super.error("no black links");
```
Gilthoniel: I'm not sure the color matters ;)

external/java/src/main/java/ch/epfl/dedis/eventlog/EventLogInstance.java
```diff
@@ -172,6 +177,194 @@ public SearchResponse search(String topic, long from, long to) throws CothorityE
...
+     * The function blocks while it reads old events from the given block.
+     * If an error is thrown, the caller should unsubscribe the tag.
+     *
+     * @param handler is the event handler, it will be called in a different thread.
+     * @param from is the block which marks the beginning of the subscription.
+     * @return an integer tag that can be used to unregister the handler.
+     * @throws CothorityCommunicationException
+     * @throws CothorityCryptoException
+     */
+    public int subscribeEvents(EventHandler handler, byte[] from) throws CothorityCommunicationException, CothorityCryptoException {
```
Gilthoniel: `from` should be of type SkipblockId or is this for convenience ?

eventlog/api.go
kc1212: It is for convenience, I didn't want to expose the user to the skipchain package. It's the same reason why the handler which you implement uses a byte slice `type StreamHandler func(event Event, blockID []byte, err error)`.

external/java/src/main/java/ch/epfl/dedis/eventlog/EventLogInstance.java
kc1212: same reason as https://github.com/dedis/cothority/pull/1906#discussion_r286958784

PR #1908, opened 2019-05-22
Optimize block proof

### LINKED ISSUE #1907: Tool to optimize forward links
We need a way to create missing forward links because it may happen that level 1 or more fail due to bad network or anything that can prevent the signature to be created. The first step would be to add a command in the `scmgr` CLI to trigger a request to fix a given chain. We can then think about triggering this automatically when necessary.

Gilthoniel:
In other words, this PR implements a tool to create missing forward-links by checking that the proof of a given block is the shortest possible and then propagate it to the given roster. If the block ID is a genesis, it will fetch the latest block and use it as the target.

Fixes #1907

### DIFF
--- a/blscosi/protocol/protocol.go
+++ b/blscosi/protocol/protocol.go
@@ -83,11 +83,15 @@ func GlobalRegisterDefaultProtocols() {
 	onet.GlobalProtocolRegister(DefaultSubProtocolName, NewDefaultSubProtocol)
 }
 
+// DefaultFaultyThreshold computes the maximum number of faulty nodes
+func DefaultFaultyThreshold(n int) int {
+	return (n - 1) / 3
+}
+
 // DefaultThreshold computes the minimal threshold authorized using
 // the formula 3f+1
 func DefaultThreshold(n int) int {
-	f := (n - 1) / 3
-	return n - f
+	return n - DefaultFaultyThreshold(n)
 }
 
 // NewBlsCosi method is used to define the blscosi protocol.
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -1832,7 +1832,7 @@ func newSkipchainService(c *onet.Context) (onet.Service, error) {
 	}
 	log.ErrFatal(s.RegisterHandlers(s.StoreSkipBlock, s.GetUpdateChain,
 		s.GetSingleBlock, s.GetSingleBlockByIndex, s.GetAllSkipchains,
-		s.GetAllSkipChainIDs,
+		s.GetAllSkipChainIDs, s.OptimizeProof,
 		s.CreateLinkPrivate, s.Unlink, s.AddFollow, s.ListFollow,
 		s.DelFollow, s.Listlink, s.ForwardLinkHandler))
 	s.ServiceProcessor.RegisterStatusReporter("Skipblock", s.db)
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -422,7 +422,7 @@ func (s *Service) OptimizeProof(req *OptimizeProofRequest) (*OptimizeProofReply,
 	index := 0
 	newProof := Proof{}
 
-	for _, sb := range pr {
+	for _, sb := range pr[:len(pr)-1] {
 		if sb.Index < index {
 			// Skip blocks thanks to the new forward-link.
 			continue
@@ -431,16 +431,16 @@ func (s *Service) OptimizeProof(req *OptimizeProofRequest) (*OptimizeProofReply,
 		diff := math.Log(float64(target.Index - sb.Index))
 		base := math.Log(float64(target.BaseHeight))
 		// Target the maximum height authorized for this block.
-		h := int(math.Min(diff/base, float64(sb.Height)))
+		h := int(math.Min(diff/base, float64(sb.Height-1)))
 
-		if len(sb.ForwardLink) < h {
+		if h > 0 && len(sb.ForwardLink) <= h {
 			// We need to round the e^x operation because of the floating-point
 			// precision but unlike the previous division, this will always
 			// produce the index minus ε < 10^-9.
 			index = sb.Index + int(math.Round(math.Exp(float64(h)*base)))
 			to := pr.Search(index)
 			if to == nil {
-				return nil, errors.New("chain is inconsistent")
+				return nil, fmt.Errorf("chain is inconsistent: block at index %d not found", index)
 			}
 
 			req := &ForwardSignature{
@@ -450,7 +450,7 @@ func (s *Service) OptimizeProof(req *OptimizeProofRequest) (*OptimizeProofReply,
 			}
 			reply := &ForwardSignatureReply{}
 
-			log.Lvlf2("Request to create forward-link at index %d with height %d", sb.Index, h)
+			log.Lvlf2("Request to create forward-link at index %d with height %d / %d", sb.Index, h, index)
 			// The signature must be asked to the roster of the block
 			err := sendForwardLinkRequest(sb.Roster, req, reply)
 
@@ -470,6 +470,8 @@ func (s *Service) OptimizeProof(req *OptimizeProofRequest) (*OptimizeProofReply,
 
 		newProof = append(newProof, sb)
 	}
+
+	newProof = append(newProof, target)
 
 	// Propagate the optimized proof to the given roster
 	err = s.startPropagation(s.propagateProof, req.Roster, &PropagateProof{newProof})
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -390,7 +390,7 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
 	return reply, nil
 }
 
-// sendForwardLinkRequest sends a request to the given until until either the forward-link is
+// sendForwardLinkRequest sends requests to conodes in the given roster until either the forward-link is
 // created or there's not enough online nodes to get a valid signature.
 func sendForwardLinkRequest(ro *onet.Roster, req *ForwardSignature, reply *ForwardSignatureReply) (err error) {
 	cl := NewClient()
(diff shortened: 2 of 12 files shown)


skipchain/skipchain.go
```diff
@@ -381,6 +390,95 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
 	return reply, nil
 }
 
+// sendForwardLinkRequest sends a request to the given until until either the forward-link is
```
jeffallen:
```suggestion
// sendForwardLinkRequest sends requests to conodes in the given roster until either the forward-link is
```

skipchain/skipchain.go
```diff
@@ -381,6 +390,95 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
...
+		err = cl.SendProtobuf(ro.List[i], req, reply)
+		if err == nil {
+			return nil
+		}
+	}
+
+	return err
+}
+
+// OptimizeProof creates missing forward links to optimize the proof of the block
```
jeffallen: I'm concerned about who is allowed to call this. It seems like anyone can call it, not just admins? Is that potentially expensive? Dangerous?

skipchain/skipchain.go
```diff
@@ -381,6 +390,95 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
...
+			}
+
+			req := &ForwardSignature{
+				TargetHeight: h,
+				Previous:     sb.Hash,
+				Newest:       to,
+			}
+			reply := &ForwardSignatureReply{}
+
+			log.Lvlf2("Request to create forward-link at index %d with height %d / %d", sb.Index, h, index)
```
jeffallen:
```suggestion
			log.Lvlf2("requesting missing forward-link at index %d with height %d / %d", sb.Index, h, index)
```

skipchain/skipchain.go
```diff
@@ -381,6 +390,95 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
...
+			// The signature must be asked to the roster of the block
+			err := sendForwardLinkRequest(sb.Roster, req, reply)
+
+			if err != nil {
+				log.Error("couldn't create a forward link:", err)
```
jeffallen:
```suggestion
				log.Error("could not create a missing forward link:", err)
```

(reasoning: the error messages in this function should be distinct from the ones in normal skipblock creation)

skipchain/skipchain.go
```diff
@@ -381,6 +390,95 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
...
+				// reset the index to try to create lower levels
+				index = sb.Index
+			} else {
+				// save the new forward link
+				err = sb.AddForwardLink(reply.Link, h)
+				if err != nil {
+					log.Error("couldn't store the forward-link:", err)
```
jeffallen:
```suggestion
					log.Error("could not store the missing forward-link:", err)
```

skipchain/struct.go
```diff
@@ -1024,6 +1035,56 @@ func (db *SkipBlockDB) GetProof(sid SkipBlockID) (sbs []*SkipBlock, err error) {
...
+		if target == nil {
+			return errors.New("couldn't find the block")
+		}
+
+		sb, err := db.getFromTx(tx, target.SkipChainID())
+		if err != nil {
+			return err
+		}
+		if sb == nil {
+			// It should never happen if the previous is found..
```
jeffallen:
```suggestion
			// It should never happen if the previous is found.
```

skipchain/skipchain.go
```diff
@@ -381,6 +390,95 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
...
+		diff := math.Log(float64(target.Index - sb.Index))
+		base := math.Log(float64(target.BaseHeight))
+		// Target the maximum height authorized for this block.
+		h := int(math.Min(diff/base, float64(sb.Height-1)))
+
+		if h > 0 && len(sb.ForwardLink) <= h {
+			// We need to round the e^x operation because of the floating-point
+			// precision but unlike the previous division, this will always
+			// produce the index minus ε < 10^-9.
+			index = sb.Index + int(math.Round(math.Exp(float64(h)*base)))
```
ineiti: See above about making this a method on the `SkipBlock` structure.

skipchain/skipchain.go
```diff
@@ -381,6 +390,95 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
...
+			to := pr.Search(index)
+			if to == nil {
```
ineiti: I'm quite convinced that this is correct, but I have a little doubt. Would it be possible that a node doesn't have this `to` block here? I think not, but am only 99% sure...

skipchain/skipchain.go
Gilthoniel: That was also a concern for me but we have a handler opened for `ForwardSignature` requests anyway and this will only try to optimize what is necessary. Yes you could ask to optimize any block in the chain but eventually all forward links will be created and it won't start anything. In the worst case one or more can't be created, well an attacker could already send a `ForwardSignature` message to start the creation process. If that makes sense.

skipchain/skipchain.go
```diff
@@ -381,6 +390,95 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
...
+
+			req := &ForwardSignature{
+				TargetHeight: h,
+				Previous:     sb.Hash,
+				Newest:       to,
+			}
+			reply := &ForwardSignatureReply{}
+
+			log.Lvlf2("Request to create forward-link at index %d with height %d / %d", sb.Index, h, index)
+			// The signature must be asked to the roster of the block
```
ineiti:
```suggestion
			// The signature must be created by the roster of the block
```

skipchain/skipchain.go
ineiti: Why would it allow lower-level forward links if it failed on higher level ones? I think you can simply return an error here and abort.

skipchain/skipchain.go
```diff
@@ -1168,41 +1268,56 @@ func (s *Service) bftForwardLinkLevel0Ack(msg []byte, data []byte) bool {
...
+		return errors.New("didn't get ForwardSignature message")
+	}
+
+	_, err := s.ForwardLinkHandler(fsOrig)
+	return err
+}
+
+// ForwardLinkHandler receives a forward-link signature request for a block already
+// appended to the chain (height >= 1).
+func (s *Service) ForwardLinkHandler(req *ForwardSignature) (*ForwardSignatureReply, error) {
```
ineiti: I'm lazy: can you tell me whether this method makes sure that the proposed link is actually valid, as in: the `from` and `to` are a valid pair within the restrictions given by the `BaseHeight` and `MaximumHeight`, respective to the indexes of the blocks?

skipchain/skipchain.go
```diff
@@ -381,6 +390,95 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
...
+			// Skip blocks thanks to the new forward-link.
+			continue
+		}
+
+		diff := math.Log(float64(target.Index - sb.Index))
+		base := math.Log(float64(target.BaseHeight))
+		// Target the maximum height authorized for this block.
+		h := int(math.Min(diff/base, float64(sb.Height-1)))
+
+		if h > 0 && len(sb.ForwardLink) <= h {
```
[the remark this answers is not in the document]
Gilthoniel: Wait how can `h` be bigger than `MaximumHeight` if it's bounded by `Height` ? Isn't it the sole purpose of `MaximumHeight` to limit the `Height` ?

skipchain/skipchain.go
Gilthoniel: If the chain is consistent, a forward-link will only be a shortcut between x and z by skipping y so if you create a forward-link to it, z is supposed to be in the chain.

skipchain/skipchain.go
Gilthoniel: My thought was that a conode might be preventing the highest link to be created so you fallback to lower ones to still optimize a bit.

skipchain/skipchain.go
Gilthoniel: Looking at it, the BFT callback checks only against the size of the back links but the forward-link won't be updated in the db if the target height is wrong. But the BFT callback could be more clever.

blscosi/protocol/protocol.go
```diff
@@ -83,11 +83,15 @@ func GlobalRegisterDefaultProtocols() {
 	onet.GlobalProtocolRegister(DefaultSubProtocolName, NewDefaultSubProtocol)
 }
 
+// DefaultFaultyThreshold computes the maximum number of faulty nodes
+func DefaultFaultyThreshold(n int) int {
```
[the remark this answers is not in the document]
Gilthoniel: https://www.usenix.org/system/files/conference/usenixsecurity16/sec16_paper_kokoris-kogias.pdf

`3f+2` is due to the microblock/keyblock structure of the paper where the consensus group is not the same. But we only have one consensus group that can evolve block after block.

skipchain/skipchain.go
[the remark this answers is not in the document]
ineiti: You're absolutely correct - I simply didn't see line 434 :(

skipchain/skipchain.go
ineiti: OK, now I got the restriction: you're only adding higher-than-existing forward-links, this is why it works. Because if you have the following `BaseHeight=2` skipchain (block-# and links):

- 0 :: 1 - x - 4
- 1 :: 2
- 2 :: 3 - 4
- 3 :: 4
- 4 :: 5 - 6 - 8

and a node is missing block `2`, the chain the node stores is still consistent, but the node couldn't add a forward-link from block `0` to block `2`.

skipchain/skipchain.go
ineiti: why would he reject the highest link but accept a lower link?

skipchain/skipchain.go
Gilthoniel: Yes that's correct for the higher-than-existing.

skipchain/skipchain.go
ineiti: I opened an issue here:

https://github.com/dedis/cothority/issues/1912

does it make sense?

Gilthoniel: Yes. Closing here then.

skipchain/skipchain.go
Gilthoniel: I guess the comment might be more explicit but the loop is for the overall proof so for a chain:
1 <- 2 <- 3 <- 4 <- 5

If the FL between 1 and 5 fails, you continue by following the level 0 until 3 where you can ry to create the link between 3 and 5.

Gilthoniel: That's what I mean by lower levels.

blscosi/protocol/protocol.go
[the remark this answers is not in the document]
Gilthoniel: Concerning the API, I won't remove then because of backwards compatibility but I will replace the inner logic to use those functions.

PR #1910, opened 2019-05-23
Prevent infinite looping when following forward-links

### LINKED ISSUE #1894: Prevent infinite loop when following forward links
Take the `GetProof` method of the skipblock DB. It assumes forward links have been correctly added in the blocks but a recent bug was storing new blocks with previous forward links if they were present, which was creating infinite loop in the path as we don't verify the from and to when reading them.

Those infinite loops should be prevented.

Gilthoniel:
This adds a check in functions that are following forward-links
to insure that it moves to a more recent block every time and
return an error otherwise.

Also: check forward-links when storing a unknown block.
    
Fixes #1894

### DIFF
--- a/skipchain/struct.go
+++ b/skipchain/struct.go
@@ -773,11 +773,9 @@ func (db *SkipBlockDB) StoreBlocks(blocks []*SkipBlock) ([]SkipBlockID, error) {
 				}
 
 				// As we don't store the target height in the forward-link, there is no
-				// way to insure that it will be stored at the right height if the target
+				// way to ensure that it will be stored at the right height if the target
 				// block is not yet discovered like in a catch up scenario that will end
-				// up in this path. But we can at least check that the forward-link belongs
-				// to the block and is not a leftover when creating a new block and that
-				// we don't have more than expected.
+				// up in this path.
 				// Deprecated: This is a notice for v4 to add the target height in the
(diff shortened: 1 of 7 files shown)


skipchain/struct.go
```diff
@@ -767,6 +771,35 @@ func (db *SkipBlockDB) StoreBlocks(blocks []*SkipBlock) ([]SkipBlockID, error) {
 						return fmt.Errorf("Tried to store unlinkable block: %+v", sb.SkipBlockFix)
 					}
 				}
+
+				// As we don't store the target height in the forward-link, there is no
+				// way to insure that it will be stored at the right height if the target
```
kc1212: insure -> ensure

skipchain/struct.go
```diff
@@ -767,6 +771,35 @@ func (db *SkipBlockDB) StoreBlocks(blocks []*SkipBlock) ([]SkipBlockID, error) {
...
+				// block is not yet discovered like in a catch up scenario that will end
+				// up in this path. But we can at least check that the forward-link belongs
```
kc1212: I understand the first sentence until "that will end up in this path". Can you rephrase it? Or remove the final part it if it's not super important.

Gilthoniel: I removed it, I guess it's more confusing than informative.

PR #1911, opened 2019-05-23
Make el consistent with bcadmin

kc1212:
- use the bcadmin library for operations in el
- fix integration test (flakiness on jenkins was observed locally too, it should be fixed on this PR)
- the OpenID authentication are untested and left as before, they will be fixed in a future PR

Discussion point:
I copy pasted the key function from bcadmin to el because the keys for logging events need to stay in the el config directory. Alternatively, we tell el the bcadmin config directory by adding a flag (e.g., `--bcconfig`) and remove the `key` function from el. The first approach has nicer UI, but the second approach removes repeated functionality.

### DIFF
--- a/eventlog/README.md
+++ b/eventlog/README.md
@@ -36,7 +36,7 @@ below) *must* be authorised to use these rules.
 An example transcript of correctly setting up and using an Eventlog is:
 
 ```
-# make the ByzCoin instance
+# Make the ByzCoin instance
 
 $ bcadmin c -roster ../../conode/public.toml
 Created ByzCoin with ID 7ad741d44e216fc4475da60b8656b904937639415ec27f7003e13408d6e0510c.
--- a/eventlog/README.md
+++ b/eventlog/README.md
@@ -15,7 +15,7 @@ auditability. Below are some of the main features that sets us apart.
   think it is valid, e.g., the timestamp is reasonable, the client is
   authorised and so on.
 - Distributed access control - fine-grained client access control with
-  delegation support is configured using [DARC](../byzcoin/README.md#darc).
+  delegation support is configured using [DARC](../darc/README.md#darc).
 - Configurable acceptance criteria - we execute a smart-contract on all nodes,
   nodes only accept the event if the smart-contract returns a positive result.
 - Existance proof - once an event is logged, an authorised client can request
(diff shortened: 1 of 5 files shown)


nkcr:
> we tell el the bcadmin config directory

Isn't it the purpose of the global `config` param / `BC_CONFIG` env var ?

nkcr:
> we tell el the bcadmin config directory by adding a flag (e.g., `--bcconfig`) and remove the `key` function from el

Can't we use bcadmin and specify the el directory with the `--config` param ?

kc1212:
> > we tell el the bcadmin config directory by adding a flag (e.g., `--bcconfig`) and remove the `key` function from el
> 
> Can't we use bcadmin and specify the el directory with the `--config` param ?

You could, but the user need to be cautious because he/she shouldn't use `--config` for non-eventlog keys. Actually, now that I think about it, when using `el` in production, the user is probably on a different machine than the one running `bcadmin`. The `el` user might not even have the `bcadmin` tool. So it needs to create the key using `el key`, send the public part to the administrator and then the administrator will send back the new DARC. This way, the `el` user doesn't need to switch between `el` and `bcadmin`. In my opinion we should keep the `key` subcommand in `el` for better user experience.

eventlog/el/test.sh
```diff
@@ -18,52 +18,42 @@ main(){
...
-	eval $SED
-	[ -z "$PRIVATE_KEY" ] && exit 1
-	ID=`awk '/^Identity: / { print $2}' < $RUNOUT`
-	[ -z "$ID" ] && exit 1
-
+testEventLog(){
+	##### setup phase
 	runCoBG 1 2 3
 	# block interval of 2 seconds to this particular test to fail because of small intervals
 	# (Note: might go back to 0.5 after #1813)
```
jeffallen: #1813 is fixed, remove comment? reduce interval?

eventlog/README.md
```diff
@@ -39,67 +38,63 @@ An example transcript of correctly setting up and using an Eventlog is:
 ```
 # make the ByzCoin instance
```
jeffallen:
```suggestion
# Make the ByzCoin instance
```

eventlog/README.md
```diff
@@ -112,9 +107,5 @@ long now = System.currentTimeMillis() * 1000 * 1000;
 SearchResponse resp = el.search("", now - 1000, now + 1000);
 ```
 
-Please refer to the javadocs for more information. The javadocs are not hosted
```
jeffallen: Want to give a link to javadoc.io?

kc1212: the java doc is given above https://github.com/dedis/cothority/pull/1911/files#diff-3ff608f4768b8ab0247453c05725047fR76

PR #1913, opened 2019-05-24
Skipchain doc update

nkcr:
I found some dead links in the doc and took this opportunity to update its content. 
The concept of height can take a bit of time to understand for newcomers (like me), so I also added some explanations about it.


skipchain/README.md
```diff
@@ -5,41 +5,138 @@ Skipchain
...
-  passwordless logins
-  - Web-pages - store your webpage on your personal blockchain and only allow
-  changes if a threshold of administrators sign off
-  - Free values - add any key/value pair to your personal skipchain
-- [RandHound](../randhound/README.md) - produce verifiable random numbers
-and store them on a blockchain for later proofs
-- [OCS](../ocs/README.md) - a distributed, secure document sharing service that
-features access-control and auditability. The documents and access-control
-lists are stored on the skipchain.
+users to use a personal blockchain where they can store data. Skipchain is used
```
kc1212: I know this is not new, but I'm wondering what it means by "personal blockchain". Every conode stores the same chain as every other conode so the blockchain is not personal in that sense. Does "personal blockchain" have another meaning?

nkcr: @ineiti ?

ineiti: Yes, that's probably not correct anymore. That was an old idea to have everybody using his own skipchain, and then putting them together somehow. So this can probably just be removed.

PR #1915, opened 2019-05-24
Skip to next block if link creation fails

ineiti:
PR to make it clearer what is happening. The previous code was supposed to try lower-level links, but actually did the same as here: just skipping to the next skipblock and trying to get a new forward-link there.

### DIFF
--- a/skipchain/skipchain.go
+++ b/skipchain/skipchain.go
@@ -448,9 +448,7 @@ func (s *Service) OptimizeProof(req *OptimizeProofRequest) (*OptimizeProofReply,
 			err := sendForwardLinkRequest(sb.Roster, req, reply)
 
 			if err != nil {
-				log.Error("could not create a missing forward link:", err)
-				// reset the index to try to create lower levels
-				index = sb.Index
+				log.Warn("could not create missing forward link, trying next block:", err)
 			} else {
 				// save the new forward link
 				err = sb.AddForwardLink(reply.Link, h)


skipchain/skipchain.go
```diff
@@ -448,9 +448,7 @@ func (s *Service) OptimizeProof(req *OptimizeProofRequest) (*OptimizeProofReply,
 			err := sendForwardLinkRequest(sb.Roster, req, reply)
 
 			if err != nil {
-				log.Error("could not create a missing forward link:", err)
-				// reset the index to try to create lower levels
-				index = sb.Index
+				log.Warn("could not create missing forward link, trying next block:", err)
```
Gilthoniel: Like I said in the PR, if you don't reset the index, you will miss potential forward-links because it will assume the forward-link has been created and the returned proof will be wrong.

ineiti: hmm, if I brain-trace the code, it will simply add the current block as is, so without the new forward-link, and then go on to the next block. But perhaps my brain-trace-capability is in Friday mode. I just checked it again, and I'm still sure it should work.

Gilthoniel:
```go
if sb.Index < index {
    // Skip blocks thanks to the new forward-link.
    continue
}
```

If the forward-link fails and index keep the target value, it will skip the following blocks of the previous proof until the target index. **But** the forward-link doesn't exist as it failed.

ineiti: Friday brain...

PR #1919, opened 2019-05-27
Release v3.1.0

Gilthoniel:
This updates the version of the different Cothority libraries to reflect the major changes.


kc1212:
should we sync the versions? i.e., move onet to v3.1.0 too

Gilthoniel:
I thought about it but Onet itself doesn't have major changes, it's only Cothority. Maybe @jeffallen has an opinion about it ?

kc1212:
Indeed, I'm ok with it either way.

PR #1927, opened 2019-05-28
Cosmetic updates for bcadmin deferred

nkcr:
- Updated `--instID` to `--instid`
- Fixes a wrong description of `-darc` parameter for `bcadmin darc rule`
- Removes unnecessary printing of a proposed Tx

### DIFF
--- a/byzcoin/bcadmin/main.go
+++ b/byzcoin/bcadmin/main.go
@@ -365,14 +365,14 @@ var cmds = cli.Commands{
                                       [--redirect],
                                invoke <command>
                                       --bc <byzcoin config>
-                                      --instid <instance ID>
+                                      --instid, i <instance ID>
                                       [--<arg name> <arg value>, ...]
                                       [--darc <darc id>] 
                                       [--sign <pub key>],
                                get    --bc <byzcoin config>
-                                      --instid <instance ID>,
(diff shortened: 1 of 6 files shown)


kc1212:
Can you add an abbreviation of `-i` for `--instid`? if we don't have a conflict. I think @jeffallen wanted to see it. Also, there are a lot of 
```go
cli.StringFlag{
    Name:  "instid",
    Usage: "the instance ID of the value contract",
}
```
is it possible to use a variable for for all these?

like
```go
[]cli.Flag{
    cli.StringFlag{...},
    instFlag,
    cli.StringFlag{...},
    // ...
}
```

byzcoin/bcadmin/main.go
```diff
@@ -472,7 +472,7 @@ var cmds = cli.Commands{
 								Usage:  "the ByzCoin config to use (required)",
 							},
 							cli.StringFlag{
-								Name:  "instID",
+								Name:  "instid",
```
jeffallen: Please use `"instid, i"` in order to allow the short argument -i for the frequently used instance id. See https://github.com/urfave/cli#alternate-names

nkcr: ok, done

PR #1936, opened 2019-05-29
 Java: use the JSR 356 spec implemented by Tomcat/Apache

### LINKED ISSUE #1825: Java-websockets not closed?
ByzGen said that the java-websockets are not closed when there is an error - is it true?

Gilthoniel:
This changes the implementation of the websocket requests and close the connection when required.

Fixes #1825

### DIFF
--- a/external/java/pom.xml
+++ b/external/java/pom.xml
@@ -123,11 +123,27 @@
             <artifactId>eddsa</artifactId>
             <version>0.2.0</version>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/javax.websocket/javax.websocket-api -->
         <dependency>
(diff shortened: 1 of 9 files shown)


Gilthoniel:
@kc1212 Can you confirm the `StreamHandler` interface can be moved or if it absolutely needs to stay for backwards compatibility ?

kc1212:
It can be moved

jeffallen:
To best check that this works with Android, I want to also add respect for the Url in a Roster, which the Java client does not have yet, so I will add that into this PR, if it's OK with you.

Gilthoniel:
Sure, go for it.

Gilthoniel:
Keeping that around if one day someone needs it

Issue #1938, opened 2019-05-31
Deferred contract: index out of range with rst.GetIndex()


nkcr:
I found an edge case where getting the current index with `rst.GetIndex()` returns `runtime error: index out of range`. This happens on the deferred contract there : https://github.com/dedis/cothority/blob/master/byzcoin/contract_deferred.go#L343

It happens when invoking a deferred instruction with a deferred contract, for example performing a deferred `invoke:deferred.addProof` with a deferred contract... still there? Here is an example:

Code to reproduce the error:

```bash
# Add the rules
bcadmin darc rule -rule spawn:value --identity <admin id>
bcadmin darc rule -rule spawn:deferred --identity <admin id>
bcadmin darc rule -rule invoke:deferred.addProof --identity <admin id>
bcadmin darc rule -rule invoke:deferred.execProposedTx --identity <admin id>
# Spaw a deferred value contract
bcadmin --export contract value spawn --value "Hello world" | bcadmin contract deferred spawn
# Perform a deferred addProof on the previous deferred value contract
bcadmin --export contract deferred invoke addProof --hash <hash> --instid <inst id> | bcadmin contract deferred spawn
# Add proof on the second deferred contract
bcadmin contract deferred invoke addProof --instid <inst id 2> --hash <hash 2>
# Exec the deferred addProof
bcadmin contract deferred invoke execProposedTx --instid <inst id 2>
> runtime error: index out of range
```
.
  
@kc1212 you helped me implementing this `rst.GetIndex()`, do you have an idea how to fix this?

kc1212:
is there a stack trace for this problem? or it's just an error message

nkcr:
Apparently, this is related to `dfs` not saving metadata: https://github.com/dedis/cothority/blob/master/byzcoin/trie/trie.go#L579

kc1212:
Thanks, I'll take care of this issue.

kc1212:
I'm revisiting this issue and I'm not sure what can be done about the `ForEach`. It is not designed to iterate through the metadata and I'm not sure if it should. If it does then the caller may be confused and interpret metadata as the real key/value pairs in the trie. As such, `ForEach` shouldn't be used to copy the whole trie so maybe `err = rst.ForEach(sst.Set)` in the deferred contract should be something else.

[...]

PR #1939, opened 2019-05-31
bcadmin: Refactors --redirect to a global --export (-x)

nkcr:
Now any transaction can be exported using the `--export` (`--x`) option and the `lib.AddTransactionAndWaitWithOption(..)` method, which automaticaly detects if the global `--export` argument is used.

Note: Test with --export and deferred contract is written but commented because of a bug: see #1938

### DIFF
--- a/byzcoin/bcadmin/lib/utils.go
+++ b/byzcoin/bcadmin/lib/utils.go
@@ -1,13 +1,18 @@
 package lib
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
+	"os"
 	"strings"
 
 	"go.dedis.ch/cothority/v3/byzcoin"
 	"go.dedis.ch/cothority/v3/darc"
+	"go.dedis.ch/protobuf"
+	"gopkg.in/urfave/cli.v1"
 )
 
 // StringToDarcID converts a string representation of a DARC to a byte array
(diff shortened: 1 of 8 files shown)


byzcoin/bcadmin/main.go
```diff
@@ -703,6 +695,10 @@ func init() {
 			Value:  getDataPath(cliApp.Name),
 			Usage:  "path to configuration-directory",
 		},
+		cli.BoolFlag{
+			Name:  "export, x",
+			Usage: "redirects the transaction to stdout",
+		},
```
kc1212: is it not possible to move the export flag under contracts? is it because you're using `c.GlobalBool("export")`?

I prefer to minimise the use of globals in functions, it's perfectly fine if `AddTransactionAndWaitWithOption` has an extra argument.

byzcoin/bcadmin/lib/utils.go
```diff
@@ -70,3 +75,43 @@ func GetDarcByID(cl *byzcoin.Client, id []byte) (*darc.Darc, error) {
 
 	return d, nil
 }
+
+// AddTransactionWithOption add an option to export the transaction instread of
+// sending it. An export will redirect the transaction to stdout.
+func AddTransactionWithOption(c *cli.Context, cl *byzcoin.Client, tx byzcoin.ClientTransaction) (*byzcoin.AddTxResponse, error) {
```
kc1212: is this function used?

byzcoin/bcadmin/lib/utils.go
```diff
@@ -70,3 +75,43 @@ func GetDarcByID(cl *byzcoin.Client, id []byte) (*darc.Darc, error) {
...
+			return nil, errors.New("failed to encode tx: " + err.Error())
+		}
+		reader := bytes.NewReader(buf)
+		_, err = io.Copy(os.Stdout, reader)
+		if err != nil {
+			return nil, errors.New("failed to copy to stdout: " + err.Error())
+		}
+		os.Exit(0)
+	}
+	return cl.AddTransactionAndWait(tx, wait)
```
kc1212: the way the functions are arranged in this file is a bit strange.

I think it'll be cleaner if there are two distinct functions. one for adding a real transaction, the other for printing the transaction. 

Something like this:
```
TransactionToString(cl *byzcoin.Client, tx byzcoin.ClientTransaction) (string, error)
AddTransaction(...)
```

actually you don't need the second one because you can just do `cl.AddTransactionAndWait` or `cl.AddTransaction`.

The caller needs to decide which one to use.

kc1212: the way I see it is `AddTransactionAndWaitWithOption` does two very different things depending on the global flag. there is no overlap between the two tasks (the `if` statement covers most of the function) except the parameters. so there's no need to put these two functionalities into one function.

kc1212:
I like the new name for the flag. I think it's nicer to move the `export` flag under `contract` and avoid using globals, what do you think?

byzcoin/bcadmin/main.go
nkcr: Using --export at the contract level was indeed the plan, but I am not able to access flags declared before the final subcommand.

byzcoin/bcadmin/lib/utils.go
nkcr: The downside of it is that we must then implement the if/else block everywhere we commit a transaction (because we want it to be widely available): 

```go
if c.GlobalBool("export") {
  TransactionToString(...)
} else {
  AddTransaction(...)
}
```

...or I could implement this if/else block in a new function function here? `AddWaitOrPrintTransaction(...)`?

byzcoin/bcadmin/main.go
kc1212: can you try using the parent context? https://godoc.org/github.com/urfave/cli#Context.Parent

byzcoin/bcadmin/lib/utils.go
kc1212: personally i prefer seeing the explicit `if/else`, it's less "magical"

nkcr: I agree. @jeffallen is it okay like that?

nkcr:
@kc1212 thanks for the review, the last commit matches your comments:
- `--export` at contract level
- refactors the export function with less magic

PR #1941, opened 2019-06-03
Java client: GFp gc optimization

### LINKED ISSUE #1877: Java bn256 makes too much garbage
@yanngabbud noticed frequent GCs when using the Cothority java client inside of an Android app. Memory profiling indicates that this is due to thousands of BigNum, which come from GFp[2,6,12]'s, which come from the loop in OptAte.miller.

(Yann will add additional info to this issue when he can export it.)

The challenge is to try to find a way to make less garbage, perhaps by calling new GFpX less often in the miller test's loop, by reusing existing points on each time through the loop.

nkcr:
This PR implements and uses a pool of GFp objects, which reduces the number of garbage collection operations needed.
The class that implements the pool is generic and work with GFp2, GFp6, and GFp12.

Fixes #1877


external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/GFpItf.java
```diff
@@ -0,0 +1,5 @@
+package ch.epfl.dedis.lib.crypto.bn256;
+
+public interface GFpItf {
```
kc1212: Please add documentation for public interfaces.

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/GFpPool.java
```diff
@@ -0,0 +1,72 @@
+package ch.epfl.dedis.lib.crypto.bn256;
+
+import java.util.LinkedList;
+
+/**
+ * This singleton class keeps a pool of different GFp. Instead of creating manually new GFp objects that will then
+ * be eventually deleted by the GC, one can use this class which will keep no longer used objects and re-use them.
+ * If the pool is empty a new object is created. We initialize GFp instances to ZERO.
+ */
+public class GFpPool {
```
kc1212: does the pool need to be public? i think the classes in the same package can access it if it's just `class GFpPool`

external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/GFpPool.java
```diff
@@ -0,0 +1,72 @@
...
+    public int getC2()  { return gfp2Pool.count;  }
+    public int getC6()  { return gfp6Pool.count;  }
+    public int getC12() { return gfp12Pool.count; }
+
+    private GFpItf get(Pool pool) {
+        pool.count++;
+        if(pool.cache.isEmpty())
+            return pool.create();
+
+        return pool.cache.pop().setZero();
```
kc1212: I have a feeling that `setZero` is not needed, could you try to remove it and see if the tests still pass?

If the `setZero` is not essential they maybe you don't need the `GFpItf` interface. You can just put `Object` in the pool (`private void put(Pool pool, Object... gfps)`) and then cast them back to the right type. You can also keep an empty private interface.

kc1212: In my opinion removing the interface and using `private void put(Pool pool, Object... gfps)` is probably better, because the the public methods have the correct typed anyway.

Issue #1924, opened 2019-05-28
bcadmin deferred usage improvements


jeffallen:
Tasks:

- [x] Update `bcadmin Darc rule` that says -darc has no default
- [x] Discuss upper cased commands like `instID` 
- [x] ~~Create restricted darc with some rules~~
- [x] Hide darc IDs
- [x] Implement a `bcadmin login`
- [x] Deferred spawn: print only once the proposed Tx
- [x] Show args of the instructions in a proposed transaction
- [x] Refactor `--redirect` to `--export` at the contract level
- [x] Make easy to create darc with "N of M" rules
- [x] Make a wrapper to create a deferred Darc

==== Jeff comments ====

1. bcadmin darc rule says -darc has no default, but the default is actually the gnesis darc.
2. --instID is the only flag that's got upper case, can we make flags case insensitive?
3. `bcadmin darc add` cannot set rules at the same time it creates a darc, but this is required when an admin wants to create a darc that will be restricted, since the final users of the darc won't be able to evolve it to add new rules.
4. Never show darc ID, because only the base id is useful in the UI
5. it should be possible to do something like `bcadmin login` that says "from now on, my darc is X and use key Y to sign my requests"

## Deferred usability

1. in spawn: print the instance id of the deferred last, and the proposed txn once.
2. to see the proposed argument, use `bcadmin contract deferred get`, but it needs to show the args, using a heuristic to try to show ASCII if there is enough, see strconv.Quote. Or maybe BasicContract should implement String() and contract writers should be encouraged to override it?

Also: -redirect does not reflect what's happening, and should not be on individual functions. It needs to be on the root of the args tree (`bcadmin -export ...`) so that every call to AddTransactionAndWait in the whole bcadmin can be converted to "deferred capable". In theory, you could make an entire ByzCoin, from the genesis DARC through all the other DARCs and contracts, require a 2-man rule for decentralised authentication. Deferred is not for special contracts, it needs to be possible on all the main contracts, i.e. darc, insecure darc, config, etc.

Consider: easy to use wrappers that help setup a deferred-enabled DARC. Input, list of signers (https://godoc.org/github.com/urfave/cli#StringSliceFlag) and the output is the new DARC id that they will put into the id in the rule of the DARC controlling access to the thing. And then make this thing able to calculate N of M rules (i.e. 2 out of the 4 admins must agree to evolve the roster).

ineiti:
[...]

PR #1946, opened 2019-06-04
 bcadmin: updates the link command

### LINKED ISSUE #1929: Adds a `bcadmin login`
#1924 :
> it should be possible to do something like bcadmin login that says "from now on, my darc is X and use key Y to sign my requests"

@jeffallen can you imagine using env var like DARC and SIGNER that are set when running `bcadmin login` ? Like this we only have to update the current parameter to be also settable via EnvVar, and we keep the previous behavior.

nkcr:
This update refactors the argument name of `bcadmin link` and allows anyone to link to a roster given a darc (--darc) and an identity (--identity). This will create or update a config file that will be set as the BC env var, except if `--silent` is given.
It also add a new `bcadmin info` that prints information about the current BC config used.

Related to #1924 
Fixes #1929

### DIFF
--- a/byzcoin/bcadmin/main.go
+++ b/byzcoin/bcadmin/main.go
@@ -13,7 +13,6 @@ import (
 	"math/big"
 	"math/rand"
 	"os"
-	"path"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


byzcoin/bcadmin/main.go
```diff
@@ -847,54 +863,84 @@ func link(c *cli.Context) error {
...
 			}
-			ad, err = darc.NewFromProtobuf(adBuf)
+
+			newDarc, err = darc.NewFromProtobuf(darcBuf)
 			if err != nil {
 				return errors.New("invalid darc stored in byzcoin: " + err.Error())
 			}
 		}
+
+		identity := cothority.Suite.Point()
```
Gilthoniel: So the default identity will always be the null point ? Does it make sense ? And even if it was randomly generated, you wouldn't have the private key associated with it meaning you cannot sign anything. Is there an interest in generating such config ? Same question for the empty darc.

nkcr: Good questions ! Note that this is the original behaviour. 
I found it strange too, but after all I decided to trust the original author...

Gilthoniel: @ineiti Do you remember the reason of this ?

ineiti: yes, the naming is not good (problem #1 in programming). It's not the `default`, but probably `empty` would be better.

It's useful if you want to interact with a byzcoin but don't have a private key associated with a darc. This way you still can `link` with a chain and get the latest blocks or other statistics.

Gilthoniel: I wonder if we should enforce the identity and darc parameters and have a `--passive` or a better name that indicates you only want to interact with the chain. In any case it would nice to have better comments on this as we were two to be confused about it.

nkcr: I updated the comments.

PR #1947, opened 2019-06-04
conode setup: improve non-interactive mode

tharvik:
Currently, `conode setup --non-interactive` doesn't respect given config path (always saved to `$HOME/.config` on Linux) and only allow the creation of nodes with loopback address.

* added `--host` to allow to select host
* use `--config` directly (the default is the same anyway)
* remove a bit of dead code
  * "config" and "debug" should be read via `cli.GlobalString`, path was never triggered
  * btw, to actually have the wanted check in place, it would be best to have non-interactive as a subcmd

### DIFF
--- a/conode/conode.go
+++ b/conode/conode.go
@@ -75,7 +75,7 @@ func main() {
 				cli.StringFlag{
 					Name:  "host",
 					Usage: "which host to listen on",
-					Value: "localhost",
+					Value: "",
 				},
 				cli.IntFlag{
 					Name:  "port",
--- a/conode/conode.go
+++ b/conode/conode.go
@@ -218,13 +218,6 @@ func checkConfig(c *cli.Context) error {
 }
 
 func setup(c *cli.Context) error {
-	if c.String("config") != "" {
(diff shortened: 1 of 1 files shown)


conode/conode.go
```diff
@@ -213,18 +218,12 @@ func checkConfig(c *cli.Context) error {
 }
 
 func setup(c *cli.Context) error {
-	if c.String("config") != "" {
-		log.Fatal("[-] Configuration file option cannot be used for the 'setup' command")
-	}
-	if c.String("debug") != "" {
```
Gilthoniel: Please keep this one as it is still available as a global flag.

Gilthoniel: (the debug one I mean)

tharvik: That'd be a breaking change, as `c.String("debug")` is not a defined flag thus currently always returns the default value. Changing it to `GlobalString` will make it work but might break some build script depending on it. Is it intented?

Gilthoniel: Oh right, no that's correct. Ok let's keep it that way.

PR #1949, opened 2019-06-05
bcadmin: deferred darc creation and minimum ids rule

nkcr:
**New commands**:

(still not convinced by the names: ~~"adeferred" for *add deferred*~~, and "prule" for *print rule*)

- ~~`bcadmin darc adeferred -id A -id C -id D`...~~
- `bcadmin darc add -deferred -id A -id C -id D`...

...that creates a deferred enabled Darc with the needed rules

- `bcadmin darc prule -id A -id C -id "C & D" --minimum 2`...

...prints `((A) & (B)) | ((A) & (C & D)) | ((B) & (C & D))`

**Updated command**:

- `bcadmin darc rule -rule xx:xx -id A -id B -id C --minimum 2`

Can now support multiple identities and compute minimum rule like "M out of N"

Related to #1924


byzcoin/bcadmin/main.go
```diff
@@ -297,6 +297,61 @@ var cmds = cli.Commands{
 					},
 				},
 			},
+			{
+				Name:   "adeferred",
```
Gilthoniel: Why do we need another command instead of extending the existing one ?

byzcoin/bcadmin/main.go
```diff
@@ -297,6 +297,61 @@ var cmds = cli.Commands{
...
+						Usage: "output file for the darc key (optional)",
+					},
+					cli.StringFlag{
+						Name:  "desc",
+						Usage: "the description for the new DARC (default: random)",
+					},
+				},
+			},
+			{
+				Name:   "prule",
```
Gilthoniel: What about using `--dry-run` on the darc creation that will print the darc instead of creating it ?

nkcr: My aim was not to simulate a Darc, but just have a handy way to play with those minimum rules groups, so it can be either directly copy/pasted or further modified.

PR #1950, opened 2019-06-07
Allow for different calypso roster

ineiti:
Before: Calypso roster was always the same as the byzcoin roster
Now: Allow for different roster

### DIFF
--- a/external/js/cothority/src/calypso/calypso-rpc.ts
+++ b/external/js/cothority/src/calypso/calypso-rpc.ts
@@ -17,11 +17,15 @@ import { DecodeKey, OnChainSecretInstance } from "./calypso-instance";
 export class OnChainSecretRPC {
     static serviceID = "Calypso";
     private socket: IConnection;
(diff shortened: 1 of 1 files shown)


external/js/cothority/src/calypso/calypso-rpc.ts
```diff
@@ -17,11 +17,15 @@ import { DecodeKey, OnChainSecretInstance } from "./calypso-instance";
 export class OnChainSecretRPC {
     static serviceID = "Calypso";
     private socket: IConnection;
-    private list: ServerIdentity[];
+    private readonly list: ServerIdentity[];
```
Gilthoniel: Why is this necessary ?

ineiti: you mean it should be only `private`, but not `private readonly`? I think it was an IntelliJ suggestion, because the variable is not changed outside of the constructor. So I accepted the suggestion.

PR #1951, opened 2019-06-07
Try other nodes if command fails

ineiti:
This PR switches to a next node if the current node returns an error. It
tries to keep the same 'working' node as long as possible.

### DIFF
--- a/external/js/cothority/src/network/connection.ts
+++ b/external/js/cothority/src/network/connection.ts
@@ -113,15 +113,23 @@ export class WebSocketConnection implements IConnection {
             });
 
             ws.onClose((code: number, reason: string) => {
-                if (code !== 1000) {
-                    Log.error("Got close:", code, reason);
-                    reject(new Error(reason));
+                switch (code) {
+                    case 1000:
+                        Log.lvl3("Normal closing of connection");
+                        break;
+                    case 1006:
+                        Log.lvl2("abnormal close of connection");
+                        reject(new Error(reason));
+                        break;
+                    default:
+                        Log.error("Got error close:", code, reason);
+                        reject(new Error(reason));
+                        break;
                 }
             });
 
             ws.onError((err: Error) => {
                 clearTimeout(timer);
-
                 reject(new Error("error in websocket " + path + ": " + err));
             });
         });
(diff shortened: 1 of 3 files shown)


external/js/cothority/src/network/connection.ts
```diff
@@ -141,25 +149,28 @@ export class RosterWSConnection extends WebSocketConnection {
...
             try {
                 // we need to await here to catch and try another conode
                 return await super.send(message, reply);
             } catch (e) {
-                Log.lvl3(`fail to send on ${addr} with error:`, e);
+                Log.lvl3(`failed to send on ${this.url} with error:`, e);
                 errors.push(e.message);
-                // TODO: send again over all nodes
-                return Promise.reject(e.message);
+                if (i > 2) {
```
Gilthoniel: Can we have that as a modifiable parameter of the class ?

external/js/cothority/src/network/connection.ts
```diff
@@ -113,15 +113,23 @@ export class WebSocketConnection implements IConnection {
             });
 
             ws.onClose((code: number, reason: string) => {
-                if (code !== 1000) {
-                    Log.error("Got close:", code, reason);
-                    reject(new Error(reason));
+                switch (code) {
```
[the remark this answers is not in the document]
ineiti: the 'normal' close returns error-code `1000`, as in the end of `onMessage`, so I check that, and simply return if it's a 'normal' close.

I changed it now to a simple `if`, as I don't remember why the code `1006` was in there specifically...

external/js/cothority/src/network/connection.ts
```diff
@@ -141,25 +149,28 @@ export class RosterWSConnection extends WebSocketConnection {
...
-                Log.lvl3(`fail to send on ${addr} with error:`, e);
+                Log.lvl3(`failed to send on ${this.url} with error:`, e);
                 errors.push(e.message);
-                // TODO: send again over all nodes
-                return Promise.reject(e.message);
+                if (i > 2) {
+                    return Promise.reject(errors.join(" :: "));
+                }
+                Log.error("Error while sending - trying with next node");
+                this.addresses = this.addresses.slice(1);
```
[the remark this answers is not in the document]
ineiti: Even though I have to make a comment about `javascript` and `readable` not going well together ;)

external/js/cothority/src/network/connection.ts
```diff
@@ -113,15 +113,15 @@ export class WebSocketConnection implements IConnection {
             });
 
             ws.onClose((code: number, reason: string) => {
-                if (code !== 1000) {
-                    Log.error("Got close:", code, reason);
-                    reject(new Error(reason));
+                if (code === 1000) {
+                    Log.lvl3("Normal closing of connection");
```
Gilthoniel: There's a `return` missing here

ineiti: oups - nice catch. Wonder why it passed the tests?

Gilthoniel: onClose so I guess the resolve is called before

ineiti: so if the `resolve` is called, a later `reject` doesn't work anymore? Makes sense.

PR #1952, opened 2019-06-07
Speed up proofs

ineiti:
Before: every proof was always verified starting at the genesis block
Now: it keeps track of the latest known block and only verifies the proof
starting on this latest block, even if more blocks are available.

### DIFF
--- a/external/js/cothority/src/byzcoin/proof.ts
+++ b/external/js/cothority/src/byzcoin/proof.ts
@@ -1,3 +1,4 @@
+import { Point } from "@dedis/kyber";
 import { createHash } from "crypto";
 import _ from "lodash";
 import Long from "long";
@@ -72,17 +73,24 @@ export default class Proof extends Message<Proof> {
     get key(): Buffer {
         return this.inclusionproof.key;
     }
+
+    /**
+     * Serves as a cache of the known blocks in different skipchains. The key is the
+     * skipchain-id (byzcoin-id), while the value is an array of known block-ids. The
+     * known block-ids are in order, but because of the skipchain nature, they will
+     * have 'holes' in it.
+     */
+    static knownBlocks: Map<Buffer, Buffer[]> = new Map<Buffer, Buffer[]>();
+
     /**
      * @see README#Message classes
      */
     static register() {
         registerMessage("byzcoin.Proof", Proof, InclusionProof, SkipBlock, ForwardLink);
     }
-
     readonly inclusionproof: InclusionProof;
     readonly latest: SkipBlock;
     readonly links: ForwardLink[];
-
     protected _state: StateChangeBody;
 
     constructor(props: Properties<Proof>) {
@@ -91,6 +99,13 @@ export default class Proof extends Message<Proof> {
         this.links = this.links || [];
     }
 
+    /**
+     * Verifies that the proof is of the given type and starts at the given byzcoin-id.
+     * @param genesisID the hash of the first block, the byzcoin-id
+     * @param contractID what contract the instance is supposed to be of
+     * @throws an error if the proof is not based on genesisID, if it is a proof of absence,
+     * or if the instance is not of type contractID.
+     */
     async getVerifiedInstance(genesisID: InstanceID, contractID: string): Promise<Instance> {
         const err = this.verify(genesisID);
         if (err != null) {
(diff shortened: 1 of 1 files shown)


external/js/cothority/src/byzcoin/proof.ts
```diff
@@ -121,15 +123,36 @@ export default class Proof extends Message<Proof> {
...
+        if (!links[0].to.equals(genesisID)) {
             return new Error("first link must come from the genesis block");
         }
 
-        const links = this.links;
-        for (let i = 1; i < links.length; i++) {
+        let latestKnown = 0;
+        const latestBlocks = Proof.knownBlocks.get(genesisID);
+        if (latestBlocks) {
+            links.map((l) => l).reverse().find((link, index) => {
```
Gilthoniel: `links.slice().reverse()`

external/js/cothority/src/byzcoin/proof.ts
```diff
@@ -121,15 +123,36 @@ export default class Proof extends Message<Proof> {
...
+                return false;
+            });
+        }
+
+        if (latestKnown === links.length - 1) {
+            return null;
+        }
+
+        let publics: Point[];
+        links.slice(0, latestKnown + 1).forEach((link) => {
```
Gilthoniel: This could go reverse and exit at the first defined `newRoster`.

ineiti: Dang - I'm not even sure this works in all cases - will have to look at it later with a fresh head...

ineiti: It was wrong - with multiple roster changes this would've failed...

ineiti:
> Looks good to me but I'm concerned of one thing: this _knownBlocks_ map is open to anyone with access to the front-end including malicious scripts wherever they are (extensions, console access, ...). And then an attacker in this context could populate with any hash what so ever.

If you have an attacker _in_ your computer, he can just overwrite the `proof` method - no need to populate the cache with wrong values, that's too complicated ;)

external/js/cothority/src/byzcoin/proof.ts
```diff
@@ -148,6 +189,10 @@ export default class Proof extends Message<Proof> {
             }
         }
 
+        // Update the known list of links. If we had access to the link-height, we could
+        // very much optimize this. But as it is, we can only store the current link list.
+        Proof.knownBlocks.set(genesisID, links.map((link) => link.to));
```
Gilthoniel: Last point I'm just thinking of. If you ask for multiple proofs of the same byzcoin chain but very different blocks, you will kinda loose the benefit of what you did as you override the previous links. Like if a proof has a few links and the next one as a direct link from the genesis.

Also the array search is slow in Javascript compared to an object (map) access. Line 150.

All of that to propose that you convert the IDs into string and use it as a key:
```javascript
const register = knownBlocks.get(genesisID);
const id = buf2hex(link.to);
register[id] = null;

_.has(register, id) // lodash - true
_.has(register, "abc") // false
```

And you can update the register with new verified IDs. Just a thought, feel free to discard.

ineiti: > Last point I'm just thinking of. If you ask for multiple proofs of the same byzcoin chain but very different blocks, you will kinda loose the benefit of what you did as you override the previous links. Like if a proof has a few links and the next one as a direct link from the genesis.

If I suppose that the proofs I get will be either from the same block or from a later block (which sounds as a very reasonable assumption to me), there is not much I can do against this inefficiency.

In a next step, one could implement https://github.com/dedis/cothority/issues/1960 to make it better.

> Also the array search is slow in Javascript compared to an object (map) access. Line 150.
> 
> All of that to propose that you convert the IDs into string and use it as a key:
> 
> ```js
> const register = knownBlocks.get(genesisID);
> const id = buf2hex(link.to);
> register[id] = null;
> 
> _.has(register, id) // lodash - true
> _.has(register, "abc") // false
> ```
> 
> And you can update the register with new verified IDs. Just a thought, feel free to discard.

Hmm - two questions:
- is it slow even compared to the signature-verification?
- how would the map-definition look like with this new definition?

Gilthoniel: It would have been a concern if we keep the previous block IDs but I guess your assumption is right so it doesn't matter.

PR #1843, opened 2019-05-02
Updates cothority-ts

ineiti:
Depends on #1853 and #1845 - will rebase against master once those are in. For testing, please rebase on those.

Making cothority-ts more consistent within itself:
- worked on *-instance.ts to always have same signature for `constructor`, `fromByzCoin`, `spawn`, and `create`
- removed getters/setters that only pass through a private variable
- updated `spawner-instance` and `credentials-instance` to fit latest version of personhood project
- allow correct signing for multi-instruction transactions with different signers for each transaction
- removed all circular dependencies
- made logger output variable-type
- added constants for commands and arguments to instances
- added recovery mode for credentials, so that it can be recovered by a threshold of other identities
- added calypso-instance for creating write- and read-instances
- added personhood-rpc for interaction with the personhood service
- added ring signatures for use in personhood-parties

PR #1953, opened 2019-06-07
Correct the ed25519_signer representation

ineiti:
Before: ed25519_signer hex values were represented as protobuf-encoded values
Now: ed25519_signer hex values are the raw MarshalBinary values

Why: this is how it is handled in the go-code. It also makes more sense, as the
hex-values are always preceded by 'ed25519:', so it is clear what kind of point
it is.

### DIFF
--- a/external/js/cothority/src/darc/identity-ed25519.ts
+++ b/external/js/cothority/src/darc/identity-ed25519.ts
@@ -1,7 +1,8 @@
 import { curve, Point, PointFactory, sign } from "@dedis/kyber";
 import { Message, Properties } from "protobufjs/light";
-import { registerMessage } from "../protobuf";
-import { IIdentity } from "./identity-wrapper";
+import Log from "../log";
+import { EMPTY_BUFFER, registerMessage } from "../protobuf";
+import IdentityWrapper, { IIdentity } from "./identity-wrapper";
 
 const {schnorr} = sign;
 const ed25519 = curve.newCurve("edwards25519");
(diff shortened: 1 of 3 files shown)


Gilthoniel:
This is a protobuf message class, the field must be filled as a it would be in protobuf. I made a PR no later than last week to fix this.

ineiti:
@Gilthoniel can you elaborate on https://github.com/dedis/cothority/pull/1953 ? As far as I see, this is used in the rules of the darcs. At least in version 3.0, the darc-rules for ed25519-keys were simply of the form `ed25519:` + the marshalled point. This gave a total length of 40 bytes. Are you proposing to change this to `ed25519:` + protobuffed marshalled point, for a total length of 48 bytes?
Also, can you please point me to the PR where you fixed this?

Gilthoniel:
https://github.com/dedis/cothority/pull/1928/files

Gilthoniel:
I think I need the context where this is an issue for you because from what I have in front of my eyes is that the string representation is `ed22519.${this.public.toString()}` which doesn't render the _protobuf_ shape but the _marshal_ shape.

ineiti:
> I think I need the context where this is an issue for you because from what I have in front of my eyes, the string representation is `ed22519.${this.public.toString()}` which doesn't render the _protobuf_ shape but the _marshal_ shape.

OK, now it makes sense. I think what happened was:
- broken implementation pre- #1843 because the string returned was not a correct darc-string
- halfway fixed implementation in #1843 , except when unmarshalling the protobuf-structure of `IdentityEd25519`
- fixed and well working in #1928 
- I mistakenly thought that this is pre-#1843 and created this PR

PR #1961, opened 2019-06-13
Improve Jenkins stability

### LINKED ISSUE #1956: [View Change] Heartbeats and transaction collection
I recently added a check inside `byzcoin/tx_pipeline.go` to prevent the actual leader to collect transaction when a block is processing so that if a change of leader will be executed, we don't loose transactions in the process.

```go
sb, doCatchUp := s.skService().WaitBlock(s.scID, nil)
if sb == nil && !doCatchUp {
	log.Lvl2("Skip tx collection during block processing")
	return nil, nil
}
```

But this causes troubles when the block processing is slow (typically like in Jenkins with a small window rotation value and a small block interval) because the heartbeat timeout will be reached before the block is actually created. And then a new view change process is triggered even if the leader is responsive.

I see a few different ways to fix this:

Gilthoniel:
Also: improve collection tx to skip collection when block is processing
but preserve the heartbeat

Fixes #1956

### DIFF
--- a/byzcoin/tx_pipeline.go
+++ b/byzcoin/tx_pipeline.go
@@ -95,11 +95,7 @@ func (s *defaultTxProcessor) CollectTx() ([]ClientTransaction, error) {
 		return nil, err
 	}
 
-	sb, doCatchUp := s.skService().WaitBlock(s.scID, nil)
-	if sb == nil && !doCatchUp {
-		log.Lvl2("Skip tx collection during block processing")
-		return nil, nil
-	}
+	_, isNotProcessingBlock := s.skService().WaitBlock(s.scID, nil)
 
 	latest, err := s.db().GetLatestByID(s.scID)
 	if err != nil {
@@ -124,6 +120,12 @@ func (s *defaultTxProcessor) CollectTx() ([]ClientTransaction, error) {
 	root := proto.(*CollectTxProtocol)
 	root.SkipchainID = s.scID
 	root.LatestID = latest.Hash
+	// When a block is processed, we prevent conodes to send us back transactions
+	// until the next collection.
+	if !isNotProcessingBlock {
+		root.MaxNumTxs = 0
+	}
+
 	if err := root.Start(); err != nil {
 		log.Error(s.ServerIdentity(), "Failed to start the protocol with error: "+err.Error()+
 			" Start() only returns an error when the protocol is not initialised correctly,"+
--- a/eventlog/api_test.go
+++ b/eventlog/api_test.go
@@ -355,10 +355,15 @@ func TestClient_StreamEventsFrom(t *testing.T) {
 	// log the first batch
 	ids, err := c.Log(batch1...)
 	require.NoError(t, err)
-	waitForKey(t, leader.omni, c.ByzCoin.ID, ids[len(batch1)-1], testBlockInterval)
+	// Makes sure the block is processed on each conode.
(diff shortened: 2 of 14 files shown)


Gilthoniel:
https://jenkins.cothority.net/blue/organizations/jenkins/gaylor_long/activity

byzcoin/tx_pipeline.go
```diff
@@ -124,6 +120,12 @@ func (s *defaultTxProcessor) CollectTx() ([]ClientTransaction, error) {
 	root := proto.(*CollectTxProtocol)
 	root.SkipchainID = s.scID
 	root.LatestID = latest.Hash
+	// When a block is processed, we prevent conodes to send us back transactions
+	// until the next collection.
+	if !isNotProcessingBlock {
```
jeffallen: So, if isNotProcessingBlock ever gets stuck to be false, then (1) we'll never pick up any more txns from followers, and (2) they will never require a view-change, because they continue to hear from the leader.

Also: the double negative here implies that maybe this flag should be flipped and called isProcessingBlock. Would at least make this check easier to reason about.

eventlog/api_test.go
```diff
@@ -355,11 +355,16 @@ func TestClient_StreamEventsFrom(t *testing.T) {
...
-	waitForKey(t, leader.omni, c.ByzCoin.ID, ids[len(batch1)-1], testBlockInterval)
+	// Makes sure the block is processed on each conode.
+	for _, srv := range s.services {
+		waitForKey(t, srv.omni, c.ByzCoin.ID, ids[len(batch1)-1], testBlockInterval)
+	}
 	for _, id := range ids {
 		checkProof(t, leader.omni, id, c.ByzCoin.ID)
 	}
 
+	time.Sleep(10 * time.Second)
```
jeffallen: Is this still needed, or was a debugging aid?

skipchain/struct.go
```diff
@@ -1120,15 +1122,31 @@ func (db *SkipBlockDB) GetProof(sid SkipBlockID) (sbs []*SkipBlock, err error) {
 		sbs = append(sbs, sb)
 
 		for sb != nil && len(sb.ForwardLink) > 0 {
-			id := sb.ForwardLink[len(sb.ForwardLink)-1].To
-			sb, err = db.getFromTx(tx, id)
+			// In some cases, a forward-link could be stored before the
```
jeffallen: Is this an unrelated change? (no big deal, just trying to understand the motivation here)

byzcoin/tx_pipeline.go
Gilthoniel: It's actually called `doCatchUp` in other places as the purpose of the function is first to check if a block is stored or currently in the block buffer, but this naming here was unclear so I changed it to that to make it a bit more obvious on the purpose.

It won't be stuck because the block buffer is cleaned if the block fails to be created. It's up to the leader to be consistent on that. Now if you're afraid that a leader could send heartbeats without creating blocks, it was already the case. Not saying we shouldn't do anything but looks like another issue to open.

skipchain/struct.go
Gilthoniel: Related to Jenkins stability and it also makes sense. Returning an error if the very latest block is not yet fully stored is wrong because we can send the previous state as the next one is not yet ready.

Gilthoniel:
You mean to request a leader with a recent version ?

PR #1962, opened 2019-06-14
Offers better printing of instructions

### LINKED ISSUE #1937: Show args of the instructions in a proposed transaction
#1924 
> to see the proposed argument, use bcadmin contract deferred get, but it needs to show the args, using a heuristic to try to show ASCII if there is enough, see strconv.Quote. Or maybe BasicContract should implement String() and contract writers should be encouraged to override it?

This happens to be actually much more tricky. Using `strconv.Quote` or other heuristic methods is not sufficient to provide meaningful information about an instruction, especially for more complex instruction like `invoke:config.update_config` that has a `Config` as argument.

**Problem statement**:

Given an `Instruction` only, print this instruction along with the potential arguments contained in the `spawn` or `invoke` methods, which depend on the contracts.

**The problem**:

Only contracts know how to print their arguments, but we are left with only an `Instruction` that can provide an `action` string of form `invoke:config.update_config`.

**One solution**:

One dirty solution would be to manually check the `action` string, extract the contract name, and call a new `PrintArgs` function on that contract that can be overrided by contracts or has a default behavior (like using `strconv.Quote`) implemented on `BasicContract`. It would be like:

```go
name := extractName(action)
switch(name) {
  case "value":
    return PrintValueArgs(action, args)
  case "config":
    return PrintConfigArgs(action, args)
  ... 
}
```

nkcr:
Related to #1924 
Fixes #1937 

Implements a mechanism so that it is now possible to print instructions with respect to the kind of contract that uses it. Here are the changes:

- Moves the `contract` map of `byzcoin.Service` to a public variable that is statically filled with `init()`. This variable is called `ContractsFn`.
- Implements a default `PrintMethod(instr Instruction)` for the `BasicContract`, which can be overrided by contracts. This method is used to print an instruction.
- Contract have no longer access to a `Service` instance, which is fine since the only use case for this was to get the constructor of another contract.
- Rename the `NumExecution` attribute in deferred contract to `MaxNumExecution` to be more explicit.
- Remove the service created in the "Contract" package. It's not needed anymore because of  the new public `ContractsFn` that can be directly filled.


byzcoin/contracts.go
```diff
@@ -36,6 +39,40 @@ type Contract interface {
 	Invoke(ReadOnlyStateTrie, Instruction, []Coin) ([]StateChange, []Coin, error)
 	// Delete removes the current instance
 	Delete(ReadOnlyStateTrie, Instruction, []Coin) ([]StateChange, []Coin, error)
+	// Print ...
```
jeffallen: Please finish this comment.

Naming is hard:

This is a function which formats an Instruction into a string. It works on all types of instructions, not just Invokes (which are the methods of contracts). It does no I/O, so it doesn't "print".

So it should be named "FormatInstruction".

(This is bike shedding, but I hope you can understand my reasoning. Naming is hard, but worth doing right.)

byzcoin/service.go
```diff
@@ -2163,7 +2173,7 @@ func (s *Service) monitorLeaderFailure() {
 // registerContract stores the contract in a map and will
```
jeffallen: I think this is not called anymore?

byzcoin/struct.go
```diff
@@ -674,24 +674,24 @@ func (c ChainConfig) checkNewRoster(newRoster onet.Roster) error {
...
+// -- BlockInterval: 7s
+// -- Roster: {1e89775c-636a-536a-bc39-1ec951c86dc9 [tls://localhost:2002 tls://localhost:2004 tls://localhost:2006] 467abd382f78222e898d323194c0fb30f7096bb6bb885ea31284979f794e558a}
+// -- MaxBlockSize: 5000000
+// -- DarcContractIDs:
+// --- darc contract ID 0: darc
+// --- darc contract ID 1: darc2
+// --- darc contract ID 2: darc3'
 // ```
 func (c ChainConfig) String() string {
 	var res strings.Builder
```
jeffallen:
```suggestion
	res := new(strings.Builder)
```

And then all the other references to res do not need an &.

byzcoin/transaction.go
```diff
@@ -206,38 +207,60 @@ func (instr Instruction) DeriveID(what string) InstanceID {
...
+	var out strings.Builder
+	out.WriteString("- instruction:\n")
+	fmt.Fprintf(&out, "-- hash: %x\n", instr.Hash())
+	fmt.Fprintf(&out, "-- instID: %v\n", instr.InstanceID)
+	fmt.Fprintf(&out, "-- action: %s\n", instr.Action())
+	fmt.Fprintf(&out, "-- identities: %v\n", instr.SignerIdentities)
+	fmt.Fprintf(&out, "-- counters: %v\n", instr.SignerCounter)
+	fmt.Fprintf(&out, "-- signatures: %d\n", len(instr.Signatures))
+	out.WriteString(
+		regexp.MustCompile(`(?m)^(.+)$`).ReplaceAllString(methodStr, "-$1"))
```
jeffallen: Regexp compilation should happen in the package scope, so that they happen once during init, instead of on every execution of String(). (Also the other call to MustCompile above.)

A comment on this regexp would be helpful too.

byzcoin/contracts/service.go
```diff
@@ -2,37 +2,13 @@ package contracts
...
-
-// interface to byzcoin.Service
-func (s *Service) byzService() *byzcoin.Service {
-	return s.Service(byzcoin.ServiceName).(*byzcoin.Service)
-}
-
-func newService(c *onet.Context) (onet.Service, error) {
-	s := &Service{
-		ServiceProcessor: onet.NewServiceProcessor(c),
+	if byzcoin.ContractsFn == nil {
```
jeffallen: This is strange, it seems like Go init semantics guarantee that package byzcoin's init was already run.

byzcoin/service.go
```diff
@@ -78,6 +81,13 @@ func init() {
 	log.ErrFatal(err)
 	network.RegisterMessages(&bcStorage{}, &DataHeader{}, &DataBody{})
 	viewChangeMsgID = network.RegisterMessage(&viewchange.InitReq{})
+
+	if ContractsFn == nil {
```
jeffallen: Remove this, and put the `make()` at declaration time, so that it happens one single time, no if checks.

byzcoin/service.go
```diff
@@ -34,6 +34,9 @@ import (
 	uuid "gopkg.in/satori/go.uuid.v1"
 )
 
+// ContractsFn ...
+var ContractsFn map[string]ContractFn
```
jeffallen:
```suggestion
var ContractsFn = make(map[string]ContractFn)
```

byzcoin/service.go
```diff
@@ -34,6 +34,9 @@ import (
 	uuid "gopkg.in/satori/go.uuid.v1"
 )
 
+// ContractsFn ...
```
jeffallen: Finish this comment please. In particular mention: `Access to this map is not threadsafe, so it must only be written to from a package's init() function.`

The Go runtime is single-threaded while running all of the init() functions.

byzcoin/contracts.go
nkcr: I agree on the "Print" side, however this function only formats the method of an instruction, not the instruction itself. I'll go then for `FormatMethod`.

byzcoin/service.go
nkcr: I kept it for backward compatible reasons, but I am happy to remove it.

nkcr: Actually we can't just removing it because it is used in other places.

Gilthoniel:
Moving to WIP until @nkcr is done with the contract registry.

Issue #1959, opened 2019-06-11
Attack on ByzCoin proof


ineiti:
# Problem

The current byzcoin proof is supposed to proof the existence or absence of an instance, given only the byzcoin-id. It does this by providing the following items:

1. forward-links that link the byzcoin-id to the latest block
2. latest block including the hash of the global state merkle-tree-root
3. inclusion proof from merkle-tree-root to the instance leaf
4. the instance leaf itself

Unfortunately, `1.` is flawed. It is supposed to link the byzcoin-id, which is the hash of the genesis-block, to the latest block. It uses the forward-links, which is a signature on:
- hash of the source block
- hash of the destination block
- the roster of the destination block, if the roster changed

This signature can be checked against the aggregate public key of the source block. Now the client only has the byzcoin-id, so he cannot check the link of the genesis-block to the next block. This is why the first forward-link in the proof points from a `nil` source block to the genesis block, and includes the roster. But there is a problem:
- this first, initial forward link is not signed
- the client cannot verify if the roster is correct

So an attacker can simply create a new roster, use it to sign an erroneous forward-link to a new block, and then prove anything he wants.

# Proposed solution

The following solution makes sure that
1. during some phase-out, old clients can still rely on the wrong proof
2. new clients can correctly check the proof

I propose to add an optional `Genesis` field to https://github.com/dedis/cothority/blob/master/byzcoin/proto.go#L158 :
```go
type Proof struct {
	// InclusionProof is the deserialized InclusionProof
	InclusionProof trie.Proof
	// Providing the latest skipblock to retrieve the Merkle tree root.
	Latest skipchain.SkipBlock
	// Proving the path to the latest skipblock. The first ForwardLink has an
	// empty-sliced `From` and the genesis-block in `To`, together with the
	// roster of the genesis-block in the `NewRoster`.
	Links []skipchain.ForwardLink
	// The header of the genesis-block to prove the initial roster of nodes
	Genesis *skipchain.SkipBlockFix `protobuf:"opt"`
}
```

As long as old clients are around, the `Links` field needs to keep the first link pointing from `nil` to the genesis block. Old clients will simply ignore the `Genesis` field, and rely on this first link. New clients will ignore the first link, and use the `Genesis` field to verify the byzcoin-id and get the initial roster.

# Further thoughts

Once we allow for proofs that don't start at the genesis-block, the `Genesis` field will be empty, and the links start at the latest known block to the client.

Gilthoniel:
[...]

PR #1965, opened 2019-06-18
Byzcoin proof attack

Gilthoniel:
This fixes the Byzcoin proof attack by making sure the roster provided inside the first link is correct and not tampered. It also adds an API to ask for shorter proofs but the default request keeps asking for the proof from the genesis block so that this proof can be shared among peers.

Also: remove duplication of the proof verification already done by the request API

Part of #1959

### DIFF
--- a/byzcoin/api.go
+++ b/byzcoin/api.go
@@ -501,7 +501,7 @@ func (c *Client) getServer() *network.ServerIdentity {
 func (c *Client) fetchGenesis() error {
 	skClient := skipchain.NewClient()
 
-	// Integrity is done by the request function.
+	// Integrity check is done by the request function.
 	sb, err := skClient.GetSingleBlock(&c.Roster, c.ID)
 	if err != nil {
 		return err
--- a/byzcoin/api.go
+++ b/byzcoin/api.go
@@ -134,9 +134,10 @@ func (c *Client) GetProof(key []byte) (*GetProofResponse, error) {
 // starting from the latest known block by this client. The proof
 // can prove the existence or the absence of the key. Note that the integrity
 // of the proof is verified.
-// Caution: the proof will be verifiable only by client/service that know the
-// state of the chain up to the block. If you want to share the proof, you may
-// prefer to use GetProof.
+// Caution: the proof will be verifiable only by client/service that knows the
+// state of the chain up to the block. If you need to pass the Proof onwards to
+// another server, you must use GetProof in order to create a complete standalone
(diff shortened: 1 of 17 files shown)


Gilthoniel:
Note that I removed the TODO comment about verifying the roster in calypso because the roster is coming from the proof that has been signed by the Cothority.

Gilthoniel:
This doesn't fix the issue as the next step will be to fix the Java and Javascript libraries.

ineiti:
1. I still think that this is a band-aid rather than a fix: the proof in itself is still not secure. If you want to have it secure, you need additional data. So the 'nice' thing about saying that our base-truth is simply the 32-bytes of the byzcoin-ID is gone, as now it also needs the genesis-block itself to prove that it's correct.
2. To optimize it, I would not store the `SkipBlock`, but only the `SkipBlockFix`, as the latter is enough to calculate the byzcoin-ID, so you save some bytes and need to store a little bit less data

Gilthoniel:
>     1. I still think that this is a band-aid rather than a fix: the proof in itself is still not secure. If you want to have it secure, you need additional data. So the 'nice' thing about saying that our base-truth is simply the 32-bytes of the byzcoin-ID is gone, as now it also needs the genesis-block itself to prove that it's correct.

In the case of a short proof (one that doesn't start from the genesis), having the block in the proof is not sufficient because you would need the proof of this block. It assumes that you already know the block beforehand and that you verified it.
Yes, adding the genesis block for full proofs would allow them to be verifiable offline but only for those specific proofs. But this genesis block is needed for most the verification anyway so why would you want to provide it all the time ? An example is the Java client that is fetching the genesis block at the beginning all the time...
> 
>     2. To optimize it, I would not store the `SkipBlock`, but only the `SkipBlockFix`, as the latter is enough to calculate the byzcoin-ID, so you save some bytes and need to store a little bit less data

As I already mentioned, the `SignatureScheme` is used in the hash but because of the protobuf indexing issue, the field has to be in `SkipBlock`.

ineiti:
> As I already mentioned, the SignatureScheme is used in the hash but because of the protobuf indexing issue, the field has to be in SkipBlock.

True - forgot about that :( We're still experimenting in what happens if we have a _real_ protobuf workflow with the `.proto` files that generate also the `.go` files - if you're interested, you can contact Valerian.

Gilthoniel:
That would definitely be an interesting thing to investigate for v4 !

PR #1970, opened 2019-06-20
Proof verification for external libraries

### LINKED ISSUE #1959: Attack on ByzCoin proof

### LINKED ISSUE #1960: Shorter proofs
Now: all proofs start the forward-links at the genesis-block

Wanted: the client can give the latest know block-id, and the node returns a proof starting from this block-id, supposing the client has the block and the roster at this time.

Gilthoniel:
Follow up on #1959 to implement the fix for Java and Javascript.

Fixes #1959 
Fixes #1960

### DIFF
--- a/external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
+++ b/external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
@@ -35,6 +35,12 @@ public class Proof {
     private final byte[] key;
     private final StateChangeBody body;
 
+    /**
+     * Creates a proof given a protobuf representation.
+     *
+     * @param p     The proof encoded as a protobuf representation
+     * @param iid   The key of the instance
+     */
     public Proof(ByzCoinProto.Proof p, InstanceId iid) {
         this.proof = p.getInclusionproof();
         this.latest = new SkipBlock(p.getLatest());
--- a/external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
+++ b/external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
@@ -36,7 +36,8 @@ public class Proof {
     private final StateChangeBody body;
 
     /**
-     * Creates a proof given a protobuf representation.
+     * Creates a proof given a protobuf representation. It is the caller's responsibility to call
+     * verify() on the resulting Proof.
      *
      * @param p     The proof encoded as a protobuf representation
      * @param iid   The key of the instance
@@ -58,7 +59,7 @@ public class Proof {
 
     /**
      * Creates a new proof given a protobuf-representation and a trusted skipchain ID.
-     * Deprecated: the verification should be performed against a known block
+     * Deprecated: the verification should be performed against a known block.
      *
      * @param p    is the protobuf representation
      * @param scID is the skipchain ID
--- a/external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
+++ b/external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
@@ -35,19 +35,11 @@ public class Proof {
     private final byte[] key;
     private final StateChangeBody body;
 
-    /**
-     * Creates a new proof given a protobuf-representation and a trusted skipchain ID.
-     *
-     * @param p    is the protobuf representation
-     * @param scID is the skipchain ID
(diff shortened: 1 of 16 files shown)


external/java/src/main/java/ch/epfl/dedis/byzcoin/Proof.java
```diff
@@ -36,18 +36,16 @@
     private final StateChangeBody body;
 
     /**
-     * Creates a new proof given a protobuf-representation and a trusted skipchain ID.
+     * Creates a proof given a protobuf representation.
```
jeffallen:
```suggestion
     * Creates a proof given a protobuf representation. It is the caller's responsibility to call
     * verify() on the resulting Proof.
```

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -179,16 +181,56 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
+        }
+
+        return this.getProofFrom(this.genesis, id, waitMatch, interval);
+    }
+
+    /**
+     * Gets a proof from byzcoin to show that a given instance is in the
+     * global state. The proof starts from the latest known block.
+     * Caution: If you need to pass the Proof onwards to another server,
+     * you must use GetProof in order to create a complete standalone
```
jeffallen:
```suggestion
     * you must use getProof in order to create a complete standalone
```

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -179,16 +181,56 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
+        }
+
+        return this.getProofFrom(this.latest, id, waitMatch, interval);
+    }
+
+    /**
+     * Gets a proof from byzcoin to show that a given instance is in the
+     * global state. The proof starts from the block given in parameter.
+     * Caution: If you need to pass the Proof onwards to another server,
+     * you must use GetProof in order to create a complete standalone
```
jeffallen:
```suggestion
     * you must use getProof in order to create a complete standalone
```

PR #1971, opened 2019-06-21
Parallel sending of messages

ineiti:
When using the cothority-ts in the omniledger-demonstrator I see sometimes a big
lag in the messages, specifically if dedis.nella.org is replying (I love that node
exactly for this reason!)

This PR solves the problem by sending all messages in parallel and returning the first
successful reply. It still tries to fetch all nodes, in case there are multiple
errors.

### DIFF
--- a/external/js/cothority/src/network/connection.ts
+++ b/external/js/cothority/src/network/connection.ts
@@ -44,7 +44,7 @@ export interface IConnection {
  * Single peer connection
  */
 export class WebSocketConnection implements IConnection {
-    url: string;
+    private url: string;
     private service: string;
     private timeout: number;
 
@@ -129,28 +129,31 @@ export class WebSocketConnection implements IConnection {
 }
 
 /**
- * Multi peer connection that tries all nodes one after another
+ * Multi peer connection that tries all nodes one after another. It can send the command to more
+ * than one node in parallel and return the first success if 'parallel' i > 1.
  */
 export class RosterWSConnection {
-    static parallel: number = 2;
-    addresses: string[];
-    addressNext: number;
-    connections: WebSocketConnection[] = [];
+    private addresses: string[];
+    private connectionsActive: WebSocketConnection[];
+    private connectionsPool: WebSocketConnection[];
 
     /**
      * @param r         The roster to use
      * @param service   The name of the service to reach
+     * @param parallel how many nodes to contact in parallel
      */
-    constructor(r: Roster, service: string) {
+    constructor(r: Roster, private service: string, parallel: number = 2) {
+        if (parallel < 1) {
+            throw new Error("parallel must be >= 1");
+        }
         this.addresses = r.list.map((conode) => conode.getWebSocketAddress());
         shuffle(this.addresses);
-        this.addressNext = RosterWSConnection.parallel;
-        if (this.addressNext > this.addresses.length) {
-            this.addressNext = this.addresses.length;
-        }
-        for (let i = 0; i < this.addressNext; i++) {
-            this.connections.push(new WebSocketConnection(this.addresses[i], service));
-        }
+        // Initialize the pool of connections
+        this.connectionsPool = this.addresses.map((addr) => new WebSocketConnection(addr, service));
+        // And take the first 'parallel' connections
+        this.connectionsActive = this.connectionsPool.splice(0, parallel);
+        // Upon failure of a connection, it is pushed to the end of the connectionsPool, and a
+        // new connection is taken from the beginning of the connectionsPool.
     }
 
     /**
@@ -164,15 +167,17 @@ export class RosterWSConnection {
      */
     async send<T extends Message>(message: Message, reply: typeof Message): Promise<T> {
         const errors: string[] = [];
-        let rotate = this.addresses.length - this.connections.length;
+        let rotate = this.addresses.length - this.connectionsActive.length;
 
         // Get the first reply - need to take care not to return a reject too soon, else
         // all other promises will be ignored.
-        return Promise.race(this.connections.map((connection) => {
+        // The promises that never 'resolve' or 'reject' will later be collected by GC:
+        // https://stackoverflow.com/questions/36734900/what-happens-if-we-dont-resolve-or-reject-the-promise
+        return Promise.race(this.connectionsActive.map((_, i) => {
             return new Promise<T>(async (resolve, reject) => {
                 do {
                     try {
-                        const sub = await connection.send(message, reply);
+                        const sub = await this.connectionsActive[i].send(message, reply);
                         // Signal to other connections that have an error that they don't need
                         // to retry.
                         rotate = -1;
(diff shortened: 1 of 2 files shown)


external/js/cothority/src/network/connection.ts
```diff
@@ -132,42 +131,85 @@ export class WebSocketConnection implements IConnection {
...
      * @param r         The roster to use
      * @param service   The name of the service to reach
-     * @param retry = 2 How many times to retry a failing connection with another node
      */
-    constructor(r: Roster, service: string, private retry: number = 2) {
-        super("", service);
+    constructor(r: Roster, service: string) {
         this.addresses = r.list.map((conode) => conode.getWebSocketAddress());
         shuffle(this.addresses);
+        this.addressNext = RosterWSConnection.parallel;
```
Gilthoniel: Please let that be parameterized so we can disable it if necessary.

external/js/cothority/src/network/connection.ts
```diff
@@ -132,42 +131,85 @@ export class WebSocketConnection implements IConnection {
...
+        if (this.addressNext > this.addresses.length) {
+            this.addressNext = this.addresses.length;
+        }
+        for (let i = 0; i < this.addressNext; i++) {
+            this.connections.push(new WebSocketConnection(this.addresses[i], service));
```
Gilthoniel: Please create an array of addresses instead so you don't have to have the URL modifiable. A new ws is created anyway when sending so creating the connection object is cheap.

external/js/cothority/src/network/connection.ts
```diff
@@ -44,7 +44,7 @@ export interface IConnection {
  * Single peer connection
  */
 export class WebSocketConnection implements IConnection {
-    protected url: string;
+    url: string;
```
Gilthoniel: See other comment

external/js/cothority/src/network/connection.ts
ineiti: If you put `parallel = 1`, it will only do one connection at the time, and still rotate through the nodes if that connection fails or times out.

Another solution would be to add a new `RosterWSParallelConnection` and then use that if the user wants it.

Gilthoniel: No no that's fine, that's what I expect. You may want to have only one connection at a time so even if most of the time it will use the default value, it's always nice to have the choice.

external/js/cothority/src/network/connection.ts
```diff
@@ -132,42 +131,85 @@ export class WebSocketConnection implements IConnection {
...
+        return Promise.race(this.connections.map((connection) => {
+            return new Promise<T>(async (resolve, reject) => {
+                do {
+                    try {
+                        const sub = await connection.send(message, reply);
+                        // Signal to other connections that have an error that they don't need
+                        // to retry.
+                        rotate = -1;
+                        resolve(sub as T);
+                    } catch (e) {
```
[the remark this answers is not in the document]
ineiti: >  If a request fails and there's no more rotation, it won't reject if another promise resolved as they won't have enough errors to reject.

That's by design and exactly the point: I don't want to know if any connection failed, I want to know if any connection succeeds. I only want to know about failure if _all_ connections fail. Else I only want success.

Also, I don't understand how the accepted solution in the SO link is acceptable: to me it looks like it would wait for all connections to be either resolved or rejected, while the OP (and this case) wants to have the first successful connection.

And, yes, there will be open connections when the `Promise.race` returns. This seems to be by design of `Promise.race`, and I don't see how I can change that.

external/js/cothority/src/network/connection.ts
ineiti: So instead of `static parallel: number = 2;` you want to have `setParallelCalls` as a method?

external/js/cothority/src/network/connection.ts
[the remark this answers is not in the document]
Gilthoniel: > 
> That's by design and exactly the point: I don't want to know if any connection failed, I want to know if any connection succeeds. I only want to know about failure if _all_ connections fail. Else I only want success.
That's the problem, I got that you want to achieve that but the fact is that some promise could not be fulfilled at all, meaning they will be leaking. It should always either resolve or reject.

> 
> Also, I don't understand how the accepted solution in the SO link is acceptable: to me it looks like it would wait for all connections to be either resolved or rejected, while the OP (and this case) wants to have the first successful connection.
Well that allows you to resolve or reject any promises by using the property of `Promise.all`. `Promise.race` returns either with rejection or resolution **but** `Promise.all` returns only if all have resolve **or** on the first failure. So you inverse that you can get the first promise that resolves or all the rejections.

> 
> And, yes, there will be open connections when the `Promise.race` returns. This seems to be by design of `Promise.race`, and I don't see how I can change that.
I know, it's not the problem.

external/js/cothority/src/network/connection.ts
ineiti: That complicates things a lot if I want to keep some memory of the nodes that are accessible and the ones that are not accessible. With the current solution, if a node fails, it won't be used until it rotates back in. If I put all nodes in a list, then I need to re-arrange the list to make sure the failing node is at the end. Possible but less elegant, IMHO.

Also, if there are many nodes (100s), we'll keep them all around for no apparent reason. Is that not a problem?

external/js/cothority/src/network/connection.ts
Gilthoniel: It can be a parameter in the constructor.

external/js/cothority/src/network/connection.ts
Gilthoniel: No you don't need to change a lot. Just remove the `connections` array and create them in the `do ... while`, I'm not asking more. You already have the array of addresses (sorry that wasn't clear).

external/js/cothority/src/network/connection.ts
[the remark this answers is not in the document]
ineiti: Thanks for clarifying the SO solution - I didn't understand that they reject on success... Between rejecting on success and ignoring the promise, I'd chose the latter, which seems to still get garbage collected:

https://stackoverflow.com/questions/36734900/what-happens-if-we-dont-resolve-or-reject-the-promise

But if you think this is not OK, I'll change it around and reject on success.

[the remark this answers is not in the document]
Gilthoniel: Hm no you're right. I missed the point that it will be GCed because of the race. I got stuck by the `all` logic.

external/js/cothority/src/network/connection.ts
ineiti: Hmm - I don't see how that would work. The current `connections` is kind of a pool of workers. Every time I want to send a command to the nodes, I start all workers until one succeeds. While waiting for the first success, every failing worker will change its address and try again.

This means that if there are failing nodes, the workers in `connections` have their addresses changed so as not to include those failing nodes. A subsequent call to `send` then only tries nodes that didn't fail before.

I'm feeling I'm missing again something, sorry ;)

Gilthoniel: You can hardly talk about workers because on the `send` it will create a new websocket every time. That's why I was saying it doesn't change much if you create a connection for each address, the end result will be the same in terms of memory footprint (almost). The small overhead allows you to keep the URL unique per connection which was my first goal with the object (allowing later optimization to reuse connections). Thus I was proposing something like:

```javascript
let promises = [];
for (let i = 0; i < 2; i++) {
  promises.push(new Promise((r, rr) => {
    let connection = new WebSocketConnection(this.addresses[i], this.service);
    // do ... while
  });
}
return Promise.race(promises);
```

ineiti: But like this you always send the requests to the same conodes. So suppose `parallel = 2` and the first two conodes in the list are down. This means that every time you send a request, you have to wait for those conodes to time out before you can try the next ones...

PR #1976, opened 2019-06-25
conode/Dockerfile*: use CMD exec form

tharvik:
Forward of dedis/cothority_template#54 (base image issue already fixed).

### DIFF
--- a/conode/Dockerfile-dev
+++ b/conode/Dockerfile-dev
@@ -10,4 +10,4 @@ COPY exe/conode.Linux.x86_64 ./conode
 
 EXPOSE 7770 7771
 
-CMD [ "./run_nodes.sh", "-n 1" ]
+CMD ["./run_nodes.sh", "-n", "1"]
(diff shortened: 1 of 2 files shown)


conode/Dockerfile-dev
```diff
@@ -10,4 +10,4 @@ COPY exe/conode.Linux.x86_64 ./conode
 
 EXPOSE 7770 7771
 
-CMD [ "./run_nodes.sh", "-n 1" ]
+CMD ["./run_nodes.sh", "-n", "1"]
```
nkcr: Will it work if it's not the full path? 
> If you want to run your command without a shell then you must express the command as a JSON array and give the full path to the executable. [1](https://docs.docker.com/v17.09/engine/reference/builder/#cmd)

tharvik: Yes, it takes the `WORKDIR` into account; and that makes a single place to change when one wants to run it as non-root.

tharvik: Fix for docker documentation in https://github.com/docker/docker.github.io/issues/9025

PR #1977, opened 2019-06-25
conode/Dockerfile*: avoid code duplicates

tharvik:
Based on #1976.

* use a `Dockerfile-base` containing the common code of `Dockerfile{,-dev}`
* `Dockerfile{,-dev}` depends on a new build arg, `FROM`, describing where to get the base image
* add phony Makefile target `docker-base` which ensure that the base image is build

### DIFF
--- a/conode/Dockerfile
+++ b/conode/Dockerfile
@@ -1,19 +1,14 @@
+ARG FROM
+
 FROM golang:1.12 as builder
 ARG BUILD_TAG=none
 ARG ldflags="-s -w -X main.gitTag=unknown"
 RUN go get go.dedis.ch/cothority
 RUN cd /go/src/go.dedis.ch/cothority && git checkout $BUILD_TAG && GO111MODULE=on go install -ldflags="$ldflags" ./conode ./byzcoin/bcadmin ./eventlog/el ./scmgr 
 
-FROM debian:stretch-slim
-RUN apt update; apt install -y procps ca-certificates; apt clean
-WORKDIR /root/
-RUN mkdir /conode_data
-RUN mkdir -p .local/share .config
-RUN ln -s /conode_data .local/share/conode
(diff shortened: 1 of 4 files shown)


ineiti:
I'm not sure this is wanted / possible. The difference between those two is the following:
- `Dockerfile` is used to create a docker-image from the sources as found in github.com
- `Dockerfile.dev` is used to create a docker-image from the local sources, so running `go build` in the local directory, which picks up any `replace` in the `go.mod`

tharvik:
> I'm not sure this is wanted / possible. The difference between those two is the following:

First, let state that there is multi stage docker build now, so, we can have a builder image providing the `conode` executable from any source.

> * `Dockerfile` is used to create a docker-image from the sources as found in github.com

So it's using a builder which is fetching the sources on github and building them.

> * `Dockerfile.dev` is used to create a docker-image from the local sources, so running `go build` in the local directory, which picks up any `replace` in the `go.mod`

Using no builder, simply copy the locally generated `conode`.

NB: the core of the conode docker doesn't change, same env, same need, only the way to provide things around it.

nkcr:
I am not convinced this solves an actual problem that justifies the extra amount of complexity. Waiting for other opinions.

ineiti:
It removes some redundancy that bit us last time when we didn't have the ca-certificates in the docker image. So I +1 the PR - it's just a little step towards nicer Dockerfiles...

nkcr:
I have another proposal in #2006: use a common script. What do you think?

ineiti:
> I have another proposal in #2006: use a common script. What do you think?

I like the script for running the conodes, but for setting up the docker-images, I prefer a common, third, `Dockerfile` - else you have two dockerfiles and a script to understand...

Anyway - that's just my opinion. You'll have to maintain it ;)

PR #1978, opened 2019-06-25
Global contract registration

Gilthoniel:
This extracts the contract registry from the byzcoin service to make it available to CLIs. Because of backwards compatibility, it is still possible to register the contract only for the service but it should be moved to the module initialization if it needs to be globally available.

Related to #1962


byzcoin/service.go
```diff
@@ -2348,7 +2370,7 @@ var existingDB = regexp.MustCompile(`^ByzCoin_[0-9a-f]+$`)
 func newService(c *onet.Context) (onet.Service, error) {
 	s := &Service{
 		ServiceProcessor:       onet.NewServiceProcessor(c),
-		contracts:              make(map[string]ContractFn),
+		contracts:              globalContractRegistry.Clone(),
```
nkcr:
```suggestion
		contracts:              GetContractRegistry(),
```

Gilthoniel:
Yes the `contracts` is a copy of the static registry that can be changed per server in the tests for instance.

byzcoin/service.go
Gilthoniel: The public returns a read-only registry but the service should have the writable type.

Gilthoniel:
@nkcr Revised with backwards compatibility. Can you please have another look ?

byzcoin/contract_darc.go
```diff
@@ -83,7 +87,7 @@ func (c *contractSecureDarc) Spawn(rst ReadOnlyStateTrie, inst Instruction, coin
 	// If we got here this is a spawn:xxx in order to spawn
 	// a new instance of contract xxx, so do that.
 
-	cfact, found := c.s.GetContractConstructor(inst.Spawn.ContractID)
+	cfact, found := c.contracts.Search(inst.Spawn.ContractID)
```
nkcr: What if `SetRegistry` has not been called? Shouldn't you first check if `c.contracts == nil` ? (nothing prevents you from not using `GetContract()`)

byzcoin/contracts.go
```diff
@@ -38,19 +39,111 @@ type Contract interface {
...
 }
 
+// ReadOnlyContractRegistry is the read-only interface for the contract registry.
+type ReadOnlyContractRegistry interface {
+	Search(contractID string) (ContractFn, bool)
+}
+
+// SpawnerContract is an interface to detect contracts that need a reference
+// to the registry.
+type SpawnerContract interface {
```
nkcr: Not convinced by the name. Spawning is the only use case when a contract needs the registry?

byzcoin/contracts.go
```diff
@@ -38,19 +39,111 @@ type Contract interface {
...
+// contractRegistry maps a contract ID with its constructor function. As soon
+// as the first cloning happens, the registry will be locked and no new contract
+// can be added for the global call.
+type contractRegistry struct {
+	registry map[string]ContractFn
+	locked   bool
+	sync.Mutex
+}
+
+func (cr *contractRegistry) register(contractID string, f ContractFn, ignoreLock bool) error {
```
nkcr: I would comment why the ignoreLock is needed

byzcoin/contracts.go
```diff
@@ -38,19 +39,111 @@ type Contract interface {
...
+		return errors.New("contract already registered")
+	}
+
+	cr.registry[contractID] = f
+	cr.Unlock()
+	return nil
+}
+
+// Search looks up the contract ID and returns the constructor function
+// it it exists and nil otherwise.
```
nkcr:
```suggestion
// if it exists and nil otherwise.
```

byzcoin/service.go
```diff
@@ -1951,10 +1964,33 @@ func (s *Service) processOneTx(sst *stagingStateTrie, tx ClientTransaction) (Sta
 // GetContractConstructor gets the contract constructor of the contract
 // contractName.
 func (s *Service) GetContractConstructor(contractName string) (ContractFn, bool) {
-	fn, exists := s.contracts[contractName]
+	fn, exists := s.contracts.Search(contractName)
 	return fn, exists
 }
 
+// GetContract creates a contract given the ID and the bytes input if the contract
+// exists or nil otherwise.
```
nkcr:
```suggestion
// exists or nil otherwise. It also sets the contract registry if needed.
```

byzcoin/service.go
```diff
@@ -1951,10 +1964,33 @@ func (s *Service) processOneTx(sst *stagingStateTrie, tx ClientTransaction) (Sta
...
+func (s *Service) GetContract(contractName string, in []byte) (Contract, error) {
```
nkcr: In fact a contract is an abstraction, what we return here is an instance of a contract: maybe `getInstance()`, or `getContractInstance()` ?

byzcoin/contracts.go
Gilthoniel: Ahah me neither. I tried to find a name for a contract that needs to look up other contracts but miserably failed. Something like `ContractWithRegistry` sounds weird but at least it's clear, I don't know.

nkcr: I didn't find any better name either, I would go for `ContractWithRegistry` as it's at least clear.

PR #1984, opened 2019-06-28
Testing and simplifying catchup

### LINKED ISSUE #1980: syncChain can go into an infinite loop
Ineiti writes: This happened when I deleted the `.db` on `gasser.blue` and then tried to get the chain loaded from the roster again.

Proposed changes:
- [x] verify why it passes the test, and get the test to fail, then fix it
- [x] work on a better way to have a node update / download a given chain

ineiti:
The previous tests only unit-tested the catchup and downloadDB methods. This test
runs it in a more production-like environment with transactions being processed
as the node catches up. A couple of problems got fixed:

- downloadDB now stores the new blocks
- byzcoin.verifySkipblock checks if it needs to catchup

Depends on dedis/onet#559
Closes #1980


byzcoin/collect_tx.go
```diff
@@ -112,6 +112,12 @@ func (p *CollectTxProtocol) Start() error {
 
 // Dispatch runs the protocol.
 func (p *CollectTxProtocol) Dispatch() error {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Print(r)
+			log.Print(log.Stack())
```
kc1212: use `log.Error`
also: under what condition does `Dispatch` panic? is it not possible to fix the panic?

byzcoin/service.go
```diff
@@ -1001,106 +1001,115 @@ func (s *Service) downloadDB(sb *skipchain.SkipBlock) error {
...
-					return nil
-				})
-				if err != nil {
-					return fmt.Errorf("Couldn't store entries: %s", err.Error())
-				}
-				if len(resp.KeyValues) < catchupFetchDBEntries {
-					break
-				}
+		// Create a roster with all nodes other than us
+		// TODO: remove this and use the future parallel client
```
kc1212: are you using the parallel client? the PR says it depends on it but I don't see where the parallel client is used

ineiti: This is in #2001

byzcoin/collect_tx.go
ineiti: Oups - debugging stuff. There was an error coming from `dedis/protobuf`, where the defer/recover in there tried to cast a value to an `error`, but it failed to do so. Unfortunately I cannot reproduce it anymore...

byzcoin/service.go
ineiti: The adding of `SendProtobufParallel` was much bigger than thought, as I had to adjust a couple of tests to the new parallelized calls. Now I don't have any `SendProtobufParallel` in it anymore, so I removed it from the description...

byzcoin/service.go
```diff
@@ -1001,106 +1001,115 @@ func (s *Service) downloadDB(sb *skipchain.SkipBlock) error {
...
+		s.stateTries[idStr] = st
+		s.stateTriesLock.Unlock()
+		chain, err := skCl.GetUpdateChain(newRoster, sb.SkipChainID())
+		if err != nil {
+			return err
 		}
-		log.Errorf("Couldn't load database from %s - got error %s", roster.List[0], err)
+		for _, sb := range chain.Update {
+			log.Lvlf2("Storing block %d: %x", sb.Index, sb.CalculateHash())
+			s.db().Store(sb)
```
kc1212: when you're storing the skipblock, can you also check that you're only storing the skipblocks that are at the same index or below the stateTrie? otherwise the blocks and the trie might become out of sync

ineiti: for every block that is stored in the db, `updateTrieCallback` will be called. If the block is newer than the trie, which might happen if another block gets created while the db is downloaded, `updateTrieCallback` will make sure that the block will be added to the trie:

- If it's just one new block, all is well, as this is the normal path in `updateTrieCallback`
- if it's two new blocks, the go-routine with `s.catchUp` in it will be called, and all is well
- if it's three or more new blocks, multiple go-routines will be called, each with a `s.catchUp` call. The first `s.catchUp` will update to the latest block, and all further `s.catchUp` will pass through as there is nothing to update to

And no, I didn't think this through while writing the code ;)

PR #1985, opened 2019-07-01
bcadmin link displays only byzcoin chains

### LINKED ISSUE #1935: bcadmin link roster.toml shows non-byzcoin IDs
The CLI just runs `GetAllSkipChainIDs`, but some of the IDs might not be running byzcoin, see https://github.com/dedis/cothority/blob/f38ffc4f7caced9ff9eaa2eccfc86d8cfb63aee8/byzcoin/bcadmin/main.go#L797

Gilthoniel:
Fixes #1935


byzcoin/api.go
```diff
@@ -89,6 +89,17 @@ func newLedgerWithClient(msg *CreateGenesisBlock, c *Client) (*CreateGenesisBloc
 	return reply, nil
 }
 
+// GetAllChainIDs returns the list of Byzcoin chains known by the server given in
+// parameter.
+func (c *Client) GetAllChainIDs(si *network.ServerIdentity) (*GetAllChainIDsResponse, error) {
```
kc1212: The name might cause a confusing because it's quite similar to `GetAllSkipChainIDs`, "chain" in our system is still skipchain. Why not call it `GetAllByzCoinChainIDs`?

ineiti: `GetAllByzCoinIDs`?

PR #1988, opened 2019-07-02
Verify contract ID in state changes

### LINKED ISSUE #1966: byzcoin should check that contracts of StateChanges are valid
If a contract emits StateChanges with a contract name that is unknown, then the execution of the contract should fail.

Gilthoniel:
Fixes #1966

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -2051,6 +2051,12 @@ func (s *Service) executeInstruction(st ReadOnlyStateTrie, cin []Coin, instr Ins
 	// instruction, we need to get the version from the trie
 	vv := make(map[string]uint64)
 	for i, sc := range scs {
+		// Make sure that the contract either exists or is empty.
+		if _, ok := s.contracts.Search(sc.ContractID); !ok && sc.ContractID != "" {
+			log.Errorf("Found unknown contract ID \"%s\"", sc.ContractID)
+			return nil, nil, errors.New("unknown contract ID")
+		}
+
 		ver, ok := vv[hex.EncodeToString(sc.InstanceID)]
 		if !ok {
 			_, ver, _, _, err = st.GetValues(sc.InstanceID)


Gilthoniel:
@jeffallen Is this what you expected ? Or a more strict policy where the contract can only produce state changes with its contract ID ?

ineiti:
I'm very surprised that this passes all tests - can you try a `bcadmin replay` with these changes?

Gilthoniel:
I'm not sure what is surprising as it's only looking up existing contracts so if the value is a known contract or empty it passes. I don't think we use other values in the contract ID field.

ineiti:
Heh - got caught again reading only the comments but not the code ;) The comment reads:

```
In other words, we don't allow cross-contract state changes.
```

And the `tight` in the previous phrase should be `tied`...

which is not what the code does... So I was under the impression that you will reject cross-contract changes, which surprised me that it passes.

Anyway, our current EVM implementation still creates `StateChange`s that are not tied to a contract. Would the correct fix on our end be to create a dummy contract for those statechanges?

byzcoin/service.go
```diff
@@ -2012,6 +2012,14 @@ func (s *Service) executeInstruction(st ReadOnlyStateTrie, cin []Coin, instr Ins
 	// instruction, we need to get the version from the trie
 	vv := make(map[string]uint64)
 	for i, sc := range scs {
+		// Security check that the state changes produced by the contract
+		// are either tight to its ID or empty. In other words, we don't
```
ineiti:
```suggestion
		// are either tied to its ID or empty. In other words, we don't
```

Gilthoniel:
Oh gosh I changed the code but not the comment. Now I got why you asked that..

So can you be a bit more precise how you're creating state changes for your EVM implementation ? Is it an empty contract ID or do you need different values for whatever reason ?

Gilthoniel:
@jeffallen The empty string is related to the discussion about eventlog (and potentially bevm) where the contract can spawn multiple instances but only a subset should be accessible to updates then if the contract ID is empty, the instance is stored and no invoke instruction can accidentally (or on purpose) update it. The contract can still access it with internal references.

c/c @kc1212

kc1212:
> @jeffallen The empty string is related to the discussion about eventlog (and potentially bevm) where the contract can spawn multiple instances but only a subset should be accessible to updates then if the contract ID is empty, the instance is stored and no invoke instruction can accidentally (or on purpose) update it. The contract can still access it with internal references.
> 
> c/c @kc1212

Yep, the empty string instruction comparison should be kept there.

Gilthoniel:
After talking with @cgrigis it seems that the bevm implementation can use empty contract ID for their state changes. The implementation is only using the state trie as a secured key/value storage but it is not tied to a contract.

We can merge the PR imho.

PR #1989, opened 2019-07-02
Remove duplicate transactions

### LINKED ISSUE #1987: Leader does not filter identical transactions
When a client sends a transaction to ByzCoin, it will send it to multiple members of the roster for redundancy. The leader may consequently receive multiple times the same transaction, and does not currently seem to filter them. All transaction copies after the first one will then be rejected with an error about invalid counters, such as the following:
```
(...) and returned error: instruction verification failed: for pk ed25519:7072c949808ca44df779d79bf03a8e0c12b3d03713b442876022cef44442b26a, got counter=12, but need 13
```

It has no consequence on the correctness of ByzCoin, but is a waste of time and resources.

ineiti:
With the new use of the Client-calls, where the same
instruction is sent in parallel to more than one node,
it can happen that the same transaction is proposed
twice in the same block. This removes duplicate transactions
in same and directly following blocks.

Closes #1987

### DIFF
--- a/byzcoin/service_test.go
+++ b/byzcoin/service_test.go
@@ -76,7 +76,7 @@ func TestService_CreateGenesisBlock(t *testing.T) {
 		Version: 0,
 		Roster:  *s.roster,
 	})
-	require.NotNil(t, err)
+	require.Error(t, err)
 
 	// invalid: max block too small, big
 	resp, err = service.CreateGenesisBlock(&CreateGenesisBlock{
@@ -84,13 +84,13 @@ func TestService_CreateGenesisBlock(t *testing.T) {
 		Roster:       *s.roster,
 		MaxBlockSize: 3000,
 	})
-	require.NotNil(t, err)
+	require.Error(t, err)
 	resp, err = service.CreateGenesisBlock(&CreateGenesisBlock{
 		Version:      0,
 		Roster:       *s.roster,
 		MaxBlockSize: 30 * 1e6,
 	})
-	require.NotNil(t, err)
+	require.Error(t, err)
 
 	// invalid darc
 	resp, err = service.CreateGenesisBlock(&CreateGenesisBlock{
@@ -98,18 +98,18 @@ func TestService_CreateGenesisBlock(t *testing.T) {
 		Roster:      *s.roster,
 		GenesisDarc: darc.Darc{},
 	})
-	require.NotNil(t, err)
+	require.Error(t, err)
 
 	// create valid darc
 	signer := darc.NewSignerEd25519(nil, nil)
 	genesisMsg, err := DefaultGenesisMsg(CurrentVersion, s.roster, []string{"spawn:dummy"}, signer.Identity())
-	require.Nil(t, err)
+	require.NoError(t, err)
 	genesisMsg.BlockInterval = 100 * time.Millisecond
 	genesisMsg.MaxBlockSize = 1 * 1e6
 
 	// finally passing
 	resp, err = service.CreateGenesisBlock(genesisMsg)
-	require.Nil(t, err)
+	require.NoError(t, err)
 	assert.Equal(t, CurrentVersion, resp.Version)
 	assert.NotNil(t, resp.Skipblock)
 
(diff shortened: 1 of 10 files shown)


byzcoin/tx_pipeline.go
```diff
@@ -327,13 +330,26 @@ func (p *txPipeline) processTxs(txChan <-chan ClientTransaction, initialState *t
...
+	leaderLoop:
 		for {
 			select {
 			case tx, ok := <-txChan:
 				if !ok {
 					log.Lvl3("stopping txs processor")
 					return
 				}
+				for _, txHash := range txHashes {
+					if bytes.Compare(txHash, tx.Instructions.Hash()) == 0 {
```
Gilthoniel: This is hashing the instructions for each loop I think. Can we get that out of it ?

byzcoin/tx_pipeline.go
```diff
@@ -327,13 +330,26 @@ func (p *txPipeline) processTxs(txChan <-chan ClientTransaction, initialState *t
...
+		var txHashes [][]byte
+	leaderLoop:
 		for {
 			select {
 			case tx, ok := <-txChan:
 				if !ok {
 					log.Lvl3("stopping txs processor")
 					return
 				}
+				for _, txHash := range txHashes {
```
Gilthoniel: Using a map would be more efficient probably.

byzcoin/tx_pipeline.go
```diff
@@ -327,13 +330,26 @@ func (p *txPipeline) processTxs(txChan <-chan ClientTransaction, initialState *t
...
+					if bytes.Compare(txHash, tx.Instructions.Hash()) == 0 {
+						log.Lvl2("Got a duplicate transaction, ignoring it")
+						continue leaderLoop
```
Gilthoniel: You don't need a label, you can just break the select.

byzcoin/tx_pipeline.go
```diff
@@ -327,13 +330,26 @@ func (p *txPipeline) processTxs(txChan <-chan ClientTransaction, initialState *t
...
+					}
+				}
+				txHashes = append(txHashes, tx.Instructions.Hash())
+				if len(txHashes) > maxTxHashes {
```
Gilthoniel: I think you're right we need limit this but I would prefer seeing that in the collection step so that we don't have multiple truncation in different places. In fact the collection already has the option to limit the number of transactions sent by the conodes so if a block is slow to be created we could ask for 0. What do you think ?

byzcoin/tx_pipeline.go
ineiti: there is a `for` between here and the `select`. But I'll have a look at it if I can take it out with your previous comment.

byzcoin/tx_pipeline.go
ineiti: Can you elaborate a bit more? I don't see what you're describing here. Where would you ask for `0`?

byzcoin/tx_pipeline.go
ineiti: but then it's more difficult to remove the oldest elements. In this case, I can simply drop the front of the slice. With a map, I would have to go through the whole map and decide for each element if it is too old.

byzcoin/tx_pipeline.go
Gilthoniel: My point is that instead of taking of the size of the array/map you're filling with hashes, the number of txs collected should be limited in the collection process so that you insure that no more than x transactions will be processed and then you don't need to clean the array/map (which make the use of the map interesting). You ask for `0` when you received too many transactions until the next block is created (when it's actually created, you start again to ask for >0 transactions).

ineiti: Hmm, I think I got most of it. The reason I'm using the `txHashes` is because transactions from previous blocks might also be double transactions. I had a first version that only made sure that all transactions from the current block are unique. But transactions might repeat also across blocks, so I added the `maxTxHashes`. But instead of counting the # of blocks where I make sure to have only unique hashes, I take the number of transactions that I want to be sure there are only unique hashes.

Does that make sense?

byzcoin/service_test.go
```diff
@@ -394,6 +394,69 @@ func TestService_AddTransaction_ValidInvalid(t *testing.T) {
...
+	defer log.SetShowTime(log.ShowTime())
+	log.SetShowTime(true)
+	s := newSerN(t, 1, testInterval, 4, disableViewChange)
+	defer s.local.CloseAll()
+
+	// add the first tx to create the instance
+	log.Lvl1("Adding a tx twice")
+	dcID := random.Bits(256, false, random.New())
+	tx1, err := createOneClientTxWithCounter(s.darc.GetBaseID(), dummyContract, dcID, s.signer, 1)
+	require.Nil(t, err)
```
Gilthoniel: Could you change those occurrences to be `require.NoError(t, err)` ?

ineiti: Here and forever after...

PR #1993, opened 2019-07-05
Version in block to fix the hash of an instruction

### LINKED ISSUE #1854: Add version information to blocks
As we are using pre-compiled contracts, one of the problems we'll be facing a lot is the change of functionality in a contract. This is important to handle correctly, specifically if we want to keep the possibility of recreating the global state of the blockchain starting from the genesis-block.

I propose we add a field to the `byzcoin.DataHeader`, `Version *int`, that is also available in the `ReadOnlyStateTrie`. So a contract that wants to change its behaviour can check the version it should apply, and so behave correctly for different blocks.

Some details:
- `byzcoin.verifySkipBlock` has to check that the version of the proposed new block is `>=` the version of the previous block
- the `Version` needs to be a pointer to int - a `protobuf:"opt"` will still create the field when using `protobuf.Encode`, and thus create incompatible blocks
- to allow correct updating of the byzcoin instance: all the nodes return their version in `getTx`, and the leader choses the highest version with at least `2n/3+2` representatives

### LINKED ISSUE #1990: byzcoin.Instruction.Hash doesn't hash Invoke.Command
The method `Instruction.Hash` should also take into account the `Instruction.Invoke.Command` field in the hash, but it doesn't. As this is a completely backwards-incompatible change, here some propositions:

1. implement #1854 and then pass this information somehow to `Instruction`
2. Change the `Version` of the byzcoin-protocol https://github.com/dedis/cothority/blob/master/byzcoin/messages.go#L27 and:
  - allow both instructions for a number of blocks, check the signatures against both hashes
  - allow only signatures against new hashes in later blocks
3. Add a version to the `ClientTransaction` that is 0 by default, and change the hash according to this version, doing the same as above: allowing both during a phase-out time, then only allowing the new

As far as I can see, this should include #1854 in any case.

Gilthoniel:
Add a version field to the block header in ByzCoin and provide new APIs to create transactions that will use the correct instruction hash function.

Closes #1854
Closes #1990


Gilthoniel:
1. Version of the ByzCoin protocol in the block

The DataHeader structure has been updated to include a version field. Older blocks will simply have this field filled with 0 thanks to the backwards compatibility property of Protobuf and it doesn't break the hash neither. The verification insures that the value is either above or equal but never lower.

New proposed block will reuse the previous value so that current chain will keep on with the version 0 and newly created chain will use the latest version. After this PR, we can implement a command to switch the chain to a different version above the current one signed by the main DARC of the chain.

2. Hash of an instruction

This is the big part of the PR, specially for the Go side. As we're mostly using structures and no interfaces, there are only two ways (that I can see) to solve this. Either we add a different hash function to the instruction that will hash correctly and you fill the code with if to use the correct one at the correct moment, or you use a private field that will indicate the current version so that you take action to hash correctly. I went for the second solution as it allows to evolve the instruction and then pass it around "safely" so clients only have to use the correct API where the second solution forces to either have more APIs or the client to know about the hash fix.

From there, two functions to notice:
- Client.CreateTransaction that will use the latest block to look up the version and then create transactions compatible.
- ClientTransaction.instructions.Upgrade(v Version) that will update the private field of the instructions

Gilthoniel:
**Backwards compatibility**

Tested locally through 3 conodes running on 3.1.x and 3 conodes on 3.2.x. I created a chain with the 3.1.x ones and added the 3.2.x incrementally. The test included block creation, leader switch, etc ... (with transaction implying an invoke of course). Everything went right.

byzcoin/proto.go
```diff
@@ -196,6 +198,9 @@ type Instruction struct {
 	// Signatures that are verified using the Darc controlling access to
 	// the instance.
 	Signatures [][]byte
+	// version is a private field that can allow a instruction to be passed
```
ineiti:
```suggestion
	// version is a private field that can allow an instruction to be passed
```

ineiti:
I don't see how this closes #1854 - `ReadOnlyStageTrie` does not have a `GetVersion` method, or even better `GetHeader` or something similar. And this was one of the main reasons I wrote 1854, so that the _contracts_ also know which version they should run.

Gilthoniel:
Thanks for pointing that out, the latest commit makes it available to contracts.

byzcoin/statetrie.go
```diff
@@ -227,3 +238,18 @@ func newMemStateTrie(nonce []byte) (*stateTrie, error) {
 	}
 	return &st, nil
 }
+
+type metadataReader interface {
+	GetMetadata([]byte) []byte
+}
+
+func readVersion(t metadataReader) Version {
```
ineiti: I'm not sure this is correct. At what point is a new version stored in the block/stateTrie? Given only version-0 blocks, what happens if the following transactions get processed in block n:

1. BEVM-update with version 0
2. Config-update to version 1
3. BEVM-update with version **x**

What is **x** here? Would it be version 0 or version 1? Can you make a test that has an instruction that behaves differently for version 0 and version 1, and then create a block with the above clienttransaction list?

Also, did you go through what happens if a node wants to catch up?

What happens if a node wants to re-create a statechange-cache?

Gilthoniel: I updated the test to illustrate how I was thinking the version would switch to a new one. Basically a block will be used to mark the beginning of a new version. Note that the function to create this block is not the definitive one but we still need to discuss with the others how we will make the switch so I just wrote it to show you.

byzcoin/service.go
```diff
@@ -1719,6 +1812,19 @@ func (s *Service) verifySkipBlock(newID []byte, newSB *skipchain.SkipBlock) bool
...
+
+		prevBlock := s.skService().GetDB().GetByID(newSB.BackLinkIDs[0])
+		if prevBlock == nil {
+			return errors.New("missing previous block")
+		}
+		prevHeader, err := decodeBlockHeader(prevBlock)
+		if err != nil {
+			return err
+		}
+		if header.Version < prevHeader.Version {
```
kc1212: is there a "limit" for `header.Version`? for instance, whoever calls `createUpgradeVersionBlock` can create their own version. is there a way to stop a malicious node from proposing a block with version 99999 when the current version is 1? will it be accepted? do we need to prevent such a scenario?

byzcoin/service.go
```diff
@@ -1929,6 +2038,8 @@ func (s *Service) createStateChanges(sst *stagingStateTrie, scID skipchain.SkipB
 		}
 	}
 
+	txOut.Upgrade(version)
```
kc1212: is `Upgrade` the right name? what the function does it more like `SetVersion`

byzcoin/service.go
ineiti: We discussed that with @Gilthoniel over luch, and it should indeed make sure that the following should hold:

```
prevVersion <= currentVersion <= conodeVersion
```

kc1212: is `conodeVersion` a hard-coded thing in byzcoin? there might be consensus issues if the conodes use different `conodeVersion`s. suppose this expression `prevVersion <= currentVersion <= conodeVersion` fails on half of the conodes but passes on the other half, then we will not have consensus. depending on the deployment, you could say that if consensus fails for this reason, then the conode admins need to make sure they update their conodes.

ineiti: We talked about this with Noemien, and the conclusion was that for starters, it's simplest if the owner of the leader can call a method to send an empty block with the new version. If 2/3+1 of the nodes accept this block by signing the forward-link, all future blocks will be of the new version.

A more elaborate version of this would be to extend the `CollectTxResponse` structure with a field like `ConodeVersion`. Every node would always fill this field, whenever the leader asks to send the latest transactions. Then the leader could decide by himself if he can get a consensus of nodes to sign a new version or not.

Gilthoniel:
I'll implementation the version upgrade mechanism in a second PR as this one is already big enough.

README.md
```diff
@@ -71,6 +71,40 @@ existing skipchains will continue to operate normally. In summary, if you need t
 create skipchains after updating to v3.1.0, make sure every conode is at least using
 v3.0.1 aswell.
 
+### Release v3.2.0
+
+A new field has been added to the *DataHeader*, *Version*, so that new features or
+upgrades can be coordinated between the conodes to only start using it when enough
+of them are up to date. The leader will propose an change of version when it detects
```
ineiti:
```suggestion
of them are up to date. The leader will propose a change of version when it detects
```

README.md
```diff
@@ -71,6 +71,40 @@ existing skipchains will continue to operate normally. In summary, if you need t
...
+that enough of the participants can reach a consensus. A successful increase of
+version is announced by an empty block that will have act as a barrier between the
```
ineiti:
```suggestion
version is announced by an empty block that will act as a barrier between the
```

README.md
```diff
@@ -71,6 +71,40 @@ existing skipchains will continue to operate normally. In summary, if you need t
...
+previous and the new version. Its *DataHeader* data will contain the new version.
+
+When creating a ledger, the default version is the most recent one and block are
```
ineiti:
```suggestion
When creating a ledger, the default version is the most recent one and blocks are
```

byzcoin/api.go
```diff
@@ -212,6 +229,37 @@ func (c *Client) GetProofFrom(key []byte, from *skipchain.SkipBlock) (*GetProofR
 	return reply, nil
 }
 
+// GetDeferredData makes a request to retrieve the deferred instruction data
+// and return the reply if the proof can be verified.
+func (c *Client) GetDeferredData(instrID InstanceID) (*DeferredData, error) {
```
ineiti: Is this correctly included here, or should this be a separate PR?

byzcoin/bcadmin/clicontracts/deferred.go
```diff
@@ -85,14 +85,13 @@ func DeferredSpawn(c *cli.Context) error {
...
 
-	ctx := byzcoin.ClientTransaction{
-		Instructions: []byzcoin.Instruction{
-			{
-				InstanceID:    byzcoin.NewInstanceID(d.GetBaseID()),
-				Spawn:         &spawn,
-				SignerCounter: []uint64{counters.Counters[0] + 1},
-			},
-		},
+	ctx, err := cl.CreateTransaction(byzcoin.Instruction{
```
ineiti: Do I understand correctly that you're calling out to a websocket within a contract? If yes, then this should be removed and replaced with `NewClientTransaction`, as you should never depend on external states within a contract!

byzcoin/contract_deferred.go
```diff
@@ -419,33 +419,7 @@ func (c *contractDeferred) VerifyDeferredInstruction(rst ReadOnlyStateTrie, inst
...
-		binary.LittleEndian.PutUint64(nameLenBuf, uint64(len(nameBuf)))
-		h.Write(nameLenBuf)
-		h.Write(nameBuf)
-
-		valueLenBuf := make([]byte, 8)
-		binary.LittleEndian.PutUint64(valueLenBuf, uint64(len(a.Value)))
-		h.Write(valueLenBuf)
-		h.Write(a.Value)
-	}
+	instr.hashType(h)
```
ineiti: I'm lazy: is this backward-compatible, so that instances from old `spawn`s can still be correctly verified?

byzcoin/proto.go
```diff
@@ -198,6 +200,9 @@ type Instruction struct {
 	// Signatures that are verified using the Darc controlling access to
 	// the instance.
 	Signatures [][]byte
+	// version is a private field that can allow an instruction to be passed
```
ineiti: I'm not a fan of passing the version as a private field in this struct. Is it to prevent adding new methods that take the version-number as argument?

If you keep it, I would add an additional comment that states something like:
```
This field must be the last field of the struct, so that the protobuf-library enumerates the fields correctly.
```

byzcoin/service.go
```diff
@@ -2582,10 +2673,9 @@ func (s *Service) fixInconsistencyIfAny(genesisID skipchain.SkipBlockID, st *sta
 		return err
 	}
 
-	var header DataHeader
-	err = protobuf.Decode(currSB.Data, &header)
+	header, err := decodeBlockHeader(currSB)
 	if err != nil {
-		return errors.New("couldn't unmarshal header: " + err.Error())
+		return err
```
ineiti: Please be careful when removing these "stacked errors", as it can be difficult to track down the error when you only have the error from `decodeBlockHeader`. As the `decodeBlockHeader` is called multiple times, if you find the error in the logs, you don't know where it failed. This is why these `return errors.New("more error details" + err.Error())` are useful.

byzcoin/service.go
```diff
@@ -1012,6 +1037,55 @@ func (s *Service) createNewBlock(scID skipchain.SkipBlockID, r *onet.Roster, tx
 	return ssbReply.Latest, nil
 }
 
+// createUpgradeVersionBlock has the sole purpose of proposing an empty block with the
+// version field of the DataHeader updated so that new blocks will use the new version,
+func (s *Service) createUpgradeVersionBlock(scID skipchain.SkipBlockID, version Version) (*skipchain.SkipBlock, error) {
```
kc1212: when is this function called in practice? i only see it in the test.

in a later PR are you planning to add perhaps a bcadmin feature that will allow a client to ask the leader to send a version upgrade transaction? is this the version upgrade mechanism?

byzcoin/api.go
Gilthoniel: I was changing the client transaction when I saw a lot of code repetition so I made this function. There's no new feature from it. If that was the question.

byzcoin/contract_deferred.go
Gilthoniel: Yes the only difference I saw was the instance ID included in the hash at the end. But having the copy/pasted logic here is annoying when fixing stuff like the command name.

byzcoin/proto.go
Gilthoniel: Yes, that means a lot of function prototype changes and some part of the client code that would need to take this into account so for me it's better to do it like that. It can still be improved for v4 with better abstractions.

Noted for the comment.

byzcoin/service.go
Gilthoniel: Yes for the moment only in the test, that was for demonstration. A PR is on the way to let the leader detect when other conodes are up-to-date by using the collectTx protocol.

Issue #1982, opened 2019-06-27
Update to trie and the skipblock does not happen atomically


kc1212:
In the `StoreBlocks` function, the skipblock update happens in one boltdb transaction and then it makes a callback to `updateTrieCallback` which makes another transaction to update the trie. These should happen atomically otherwise we may get into an inconsistent state when the conode crashes in between these two transactions.

kc1212:
Making the trie storage and block storage is tricky with the current API because the primary way we access the trie does not allow us to specify a boltdb tx. Each function call to the trie APIs, for example `stateTrie.GetValues` or `stateTrie.VerifiedStoreAll` create new transactions. There are three ways that I can think of to resolve this issue:

1. Create a new set of `stateTrie` APIs that operate inside a transaction by making use of `trie.*WithBucket`. The `ReadOnlyStateTrie` may need to be updated too because it also creates new transactions, if it's used in `updateTrieCallback`. This would be the correct way to fix this issue but it will involve a lot of work, not only on the `stateTrie` side but also the `SkipBlockDB` side because we also use it to get information on skipblocks in the callback.

2. Instead of modifying the trie directly inside `updateTrieCallback`, have the callback return a set of instructions that will be executed by `SkipBlockDB`. This approach might involve less work but it will change the order of operations which might make cothority unstable.

3. Keep the `updateTrieCallback` the same but add a mechanism that detects the inconsistency and tries to recover from it on startup. This approach does not address the root cause but it is probably the simplest solution.

ineiti:
When starting to read I thought directly of `3.`, without reading it ;) So I vote +1 for `3.`...

Gilthoniel:
I see there an occasion to fix the problem we have when a forward link level 0 is stored but the block is still processing. Now it could end up in a crazy design idea but I think it might be worth investigate.

It would great to have a transaction running from the beginning of the block creation to the point it's actually stored. For the moment it's only a bunch of DB calls there and there.

That means the transaction would need to be transmitted from the `forwardLinkLevel0` function through the callback and plus the hard work you said about the trie but the good side is that everything is private so there should not be backwards compatibility troubles.

So I kinda vote for +1

kc1212:
I analyzed the complexity of doing option 1. It's big but not unmanageable. I've grouped the API changes into 5 sections.

1. We need to create new versions of `byzcoin.{LoadBlockInfo, LoadConfig, computeInitialDuration}`, which are basically wrappers around `stateTrie`.
[...]

PR #1996, opened 2019-07-05
Fix consistency issue if any

kc1212:
This commit implements a workaround for #1982. Instead of doing
transactions atomically, we detect consistency issues on startup and fix
it if possible.

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -30,7 +30,7 @@ import (
 	"go.dedis.ch/onet/v3/log"
 	"go.dedis.ch/onet/v3/network"
 	"go.dedis.ch/protobuf"
-	bbolt "go.etcd.io/bbolt"
+	"go.etcd.io/bbolt"
 	uuid "gopkg.in/satori/go.uuid.v1"
 )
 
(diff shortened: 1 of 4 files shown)


byzcoin/service.go
```diff
@@ -2305,6 +2311,16 @@ func (s *Service) startChain(genesisID skipchain.SkipBlockID) error {
 		return nil
 	}
 
+	// before doing anything, verify that byzcoin is consistent
+	st, err := s.getStateTrie(genesisID)
+	if err != nil {
+		return err
+	}
+	if err := s.fixInconsistencyIfAny(genesisID, st); err != nil {
```
Gilthoniel: I might be missing something but if you only repair the inconsistencies when booting the conode, that means that until you restart the conode, one inconsistency can be propagated to client requests / catch up / whatever ?

kc1212: Yes, this is the intended behaviour. The issue that this PR tries to solve is: how to make sure the trie and skipblocks are consistent after a crash or power failure. Since we store the skipblock and then the trie one of the other, the only time that the database becomes inconsistent internally is when a crash happens in between these two operations. The inconsistency is only visible after a restart, that's why we do the repair at startup. https://github.com/dedis/cothority/issues/1982#issuecomment-507201045 explains it in another way

PR #2001, opened 2019-07-09
Use parallel

ineiti:
Uses the latest `SendProtobufParallel` from `onet` to simplify and speed up real-life usage of byzcoin. A lot of tests fail when sending commands to multiple nodes, mainly for racy-reasons. All failing tests got a `cl.UseNode(0)` to force the `byzcoin.Client` to send only to the leader.

As the main goal of this PR is to have workable CLI binaries when a node is down, I made sure that all integration tests are passing. For this to work, I added the following elements:
- `--wait` flag in `bcadmin` that polls all nodes at the end of a transaction to make sure that all nodes have the latest block. In the integration-tests, an `export BC_WAIT=true` is used.
- replaced some `GetProof` in `bcadmin` with `WaitProof`, whenever the `GetProof` was just after an `AddTransactionAndWait` with an instruction-type of `Spawn`. When the instruction-type is `Invoke` or `Delete`, a `WaitPropagation` is needed.
- rewrote some of the integration-test scripts to correctly call all tests (`bcadmin/clicontracts` / `calypso/csadmin/clicontracts`)
- for testing, added a `bcadmin debug counters` that shows the counters from all nodes - handy ;)

So as I ran the tests a lot I found a lot of edge-cases that sometimes trigger and sometimes not. This is why there are a lot of small changes that make sure the system is always in a coherent state before the next command is launched.

See #2002 for a list of tests that fail and a first idea what might be wrong with the tests.


byzcoin/bcadmin/main.go
```diff
@@ -99,7 +99,7 @@ var cmds = cli.Commands{
 			cli.IntFlag{
 				Name:  "server",
 				Usage: "which server number from the roster to contact (default: 0)",
-				Value: 0,
+				Value: -1,
```
kc1212: the doc is out of date

eventlog/api.go
```diff
@@ -31,8 +31,11 @@ type Client struct {
 // NewClient creates a new client to talk to the eventlog service.
 // Fields DarcID, Instance, and Signers must be filled in before use.
 func NewClient(ol *byzcoin.Client) *Client {
+	var bc byzcoin.Client
+	bc = *ol
+	bc.UseNode(0)
```
kc1212: this is interesting, why can't you just write `ol.UseNode`?

skipchain/api.go
```diff
@@ -21,13 +18,34 @@ import (
 // service from the outside
 type Client struct {
 	*onet.Client
+	// Used for SendProtobufParallel. If it is nil, default values will be used.
+	Options *onet.ParallelOptions
```
kc1212: can this be private?

eventlog/api.go
ineiti: Because `ol` is passed as a pointer, so if I write `ol.UseNode`, the passed value gets changed, which is not really intended.

The correct way would be to pass `ol` as value, but this would break the API... Perhaps a good issue for `v4`?

ineiti:
> can you change if the integration tests are still passing?

`s/change/check/` ?

Good point - in fact the main reason I wanted to have `use_parallel` in, is to have the CLIs working OK again. I got stuck on multiple occasions when I wanted to use them and one node was down... So I started with `integration_test.sh` and will try to get them all to pass... Wish me good luck ;)

kc1212:
yep, it should be "check". did the tests pass?

ineiti:
Of course not ;) but I'm working on them...

byzcoin/bcadmin/lib/utils.go
```diff
@@ -142,6 +143,50 @@ func CombinationAnds(list []string, m int) []string {
...
+				log.Lvl2("Found new block:", sb.Index)
+				cc, err := cl.GetChainConfig()
+				if err != nil {
+					return err
+				}
+				cl.Roster = cc.Roster
+				continue searchLatest
+			} else if cl.Latest.Index < sb.Index {
+				log.Lvlf2("Node %d returned earlier block: %d", node, cl.Latest.Index)
+				continue searchLatest
```
kc1212: what would happen if one of the node always returns an earlier block? will this function wait infinitely? i think the outer for-loop should have some kind of limit

ineiti: OK, added an outer for-loop and a wait for every subsequent query, to avoid pinging the nodes too frequently. With javascript I would've added default parameters, but that doesn't exist in go, so I hard-coded them ;)

calypso/csadmin/clicontracts/lts_test.sh
```diff
@@ -14,9 +14,8 @@ testContractLTSInvoke() {
     eval $SED
     [ -z "$BC" ] && exit 1
 
-    OUTRES=`runCA contract lts spawn`
-    matchOK "$OUTRES" "^Spawned a new LTS contract. Its instance id is:
-[0-9a-f]{64}$"
+    testGrep "Spawned a new LTS contract. Its instance id is:
```
nkcr: @ineiti  By changing `matchOK` to `testGrep` you are actually changing the test. `matchOK` looks for an exact match, while `textGrep` only ensures that it finds AT LEAST the content....

PR #2003, opened 2019-07-10
Implements CLI for Calypso service

nkcr:
~~We should discuss whether or not this work should be included - and thus supported - in cothority. This is needed for one of our project with an external partner.~~

~~This PR does not add any new features to Calypso, but only provides a simple way to use Calypso with the command line.~~

~~I see one major disadvantage of including it versus 2 advantages:~~

~~Con:~~

~~- Increases the load of stuff to further maintain in cothority~~

~~Pros:~~

~~- Offers a new, straightforward, and (in my opinion) easy interface to interact with Calypso~~
~~- Adds more tests that should consolidate the stability of the system~~

---

The team is ok to merge this PR.

The following commands are added (see the [csadmin/README.md](https://github.com/dedis/cothority/tree/csadmin-new-cli/calypso/csadmin)):

- csadmin contract lts spawn
- csadmin dkg start
- csadmin contract write spawn
- csadmin contract read spawn
- csadmin decrypt
- csadmin recover


ineiti:
I vote +1 for including this. Having more CLIs is always nice. Not having looked at the code, but only at the text for the PR, I propose to simplify the commands. Instead of
`csadmin contract lts spawn`
I propose
`csadmin spawn lts`
It's less to type ;) Also, I'm usually trying to keep a hierarchy in the commands by putting the most generic parts to the left. So for the four-word command, it would be:
`csadmin contract spawn lts`

nkcr:
> I propose to simplify the commands. Instead of
> `csadmin contract lts spawn`
> I propose
> `csadmin spawn lts`

It follows the convention used in `bcadmin`. I [love conventions](https://github.com/dedis/cothority/pull/1876#discussion_r284719510).. ;) 

This is an interesting idea, however I see situations where it has some limits. It works well if we only keep `spawn` and `invoke` as exclusive actions for contract (I know that a `csadmin invoke` only concerns a contract). But with `delete` and `get` (that could be added later) there might be situations where those action are not exclusively related to a contract, for example a potential `csadmin get status`. 

And then users are not able anymore to understand what's going on: does `csadmin delete lts` deletes the key? In fact no, it only destroys the instance of the lts contract, which is explicitly described with an `csadmin contract lts delete`. And that goes towards an interesting idea from one of SURI's talk:

"Make design that facilitates mental model of software developper and end-users"

kc1212:
I also support including this CLI in cothority, it's not really specific to any particular external partner

ineiti:
> "Make design that facilitates mental model of software developper and end-users"

+1 - but please use _my_ mental model (that was a joke of course ;)

My mental model tries to keep the scope from the most general to the most detailed. This is why the 

`csadmin contract lts delete` 'jumps' around in my head: the `lts` is more detailed than the `delete`... So when I will be using this command, the most natural for _me_ is to do:
- `csadmin` - well, I want to admin calypso
- `contract` - I want to use the contract (for the `csadmin` I could even do without...)
- `delete` - I want to delete something (which is more general than `lts`)
- `lts` - this is the most detailed information in the command, so it should come last

Anyway, this is _my_ mental model. So if you want to make it simple for _me_ to use, then you'll have to use that one ;) Now if you want to make a big study of which one is the most common mental model, that might be in fact very interesting. But you'll find that there are a lot of them, and everybody thinks his is the best...

And you get the additional benefit that when sorting alphabetically, common things stay together. This is why I like the `YYYY-MM-DD` format, and none of `MM-DD-YYYY` or, even worse, `DD-MM-YYYY`...

nkcr:
Ok thank you for your feedback: its **ready for review**. 

I am sorry for giving a 17 files changed and +1,392 lines added single PR. 😧 

Here is an exhaustive list of changes. 

- `calypso/csadmin/commands.go`: CLI commands are removed from the `main.go` and stored on a separate file (we should probably do the same for bcadmin)
- `calypso/csadmin/main.go`:
-- Use the same global `-c` flag as bcadmin (the config path)
-- Implement the functions not related to a contract:
--- `csadmin dkg start` (dkgStart)
--- `csadmin decrypt` (decrypt)
--- `csadmin recover` (recover)

- `calypso/csadmin/clicontracts/`
-- This folder follows the same logic as in bcadmin. Each contract has its file and its test file.

- `calypso/csadmin/test.sh`
-- uses the same logic as in bcadmin. Tests for function not related to a contract are there and each clicontract has its own test file next to it. This file will then call each of those files.

- `libtest.sh`
-- Implements new utility functions `matchOK` and `matchNOK`. Those allows to check the exact matching of a string with a pattern.

- "Side features" (a lot of files but only weak changes):
 -- `byzcoin/bcadmin/lib/config.go`: Comments formatting and refactors an error print. I also put the name of bcadmin, which is used for the `cliApp.Name` and the config folder, in a constant.
-- `calypso/api.go`: Code formatting
-- `calypso/contracts.go`Comments formatting
-- `byzcoin/bcadmin/main.go` Uses the new constant for the name of bcadmin
-- `calypso/service.go` Refactors error printing
-- `calypso/struct.go` Only formatting
-- `unit_test.sh` Only formatting

- `calypso/csadmin/README.md`
-- Describes a scenario using csadmin

calypso/csadmin/clicontracts/lts_test.sh
```diff
@@ -0,0 +1,50 @@
...
+    ID=`cat ./darc_id.txt`
+    KEY=`cat ./darc_key.txt`
+
+    # Fail because it does not have the "spawn:longTermSecret"
+    testFail runCA contract lts spawn --darc "$ID"
+
+    # Fail because this identity is not allowed in the default used admin darc
+    #
+    # This test should be done, but we end up with a "Got a duplicate
+    # transaction, ignoring it" later it we do.
```
kc1212: could you prepend the comment with a TODO? so it's easy to grep these kind of things later

calypso/csadmin/clicontracts/read.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	if !exist {
+		return errors.New("proof not found")
+	}
+
+	match := proof.InclusionProof.Match(instidbuf)
+	if !match {
+		return errors.New("proof does not match")
+	}
+
```
kc1212: since the proof has the value inside, you can also check whether the instance is a write instance

calypso/csadmin/clicontracts/lts_test.sh
```diff
@@ -0,0 +1,50 @@
+# This method should be called from the calypso/csadmin/test.sh script
+
+testContractLTS() {
+    run testContractLTSInvoke
```
kc1212: why do we need an extra wrapper? can we not just do `run testContractLTSInvoke` inside `test.sh`?

calypso/csadmin/clicontracts/read_test.sh
```diff
@@ -0,0 +1,58 @@
+# This method should be called from the calypso/csadmin/test.sh script
+
+testContractRead() {
+    run testContractReadInvoke
```
kc1212: why do we need an extra wrapper? can we not just do run testContractLTSInvoke inside test.sh?

calypso/csadmin/clicontracts/read_test.sh
```diff
@@ -0,0 +1,58 @@
...
+    bcID=$( ls config/bc-* | sed -e "s/.*bc-\(.*\).cfg/\1/" )
+    testOK runCA authorize co1/private.toml $bcID
+    testOK runCA authorize co2/private.toml $bcID
+    testOK runCA authorize co3/private.toml $bcID
+    # Creat LTS and save the public key
+    runCA dkg start --instid "$LTS_ID" -x > key.pub
+    PUB_KEY=`cat key.pub`
+    matchOK $PUB_KEY ^[0-9a-f]{64}$
+
+    # Add the Calypso rule "pawn:calypsoWrite"
```
kc1212: spawn

calypso/csadmin/clicontracts/write.go
```diff
@@ -0,0 +1,140 @@
...
+		dstr = cfg.AdminDarc.GetIdentityString()
+	}
+	d, err := lib.GetDarcByString(cl, dstr)
+	if err != nil {
+		return err
+	}
+
+	data := c.String("data")
+	if data == "" {
+		return errors.New("please provide data with --data")
```
kc1212: i think it'll be quite convenient to allow reading data from stdin if the `--data` flag is missing, e.g., when you want to put a file on calypso

calypso/csadmin/clicontracts/write_test.sh
```diff
@@ -0,0 +1,52 @@
+# This method should be called from the calypso/csadmin/test.sh script
+
+testContractWrite() {
+    run testContractWriteInvoke
```
kc1212: same question as the other ones

calypso/csadmin/clicontracts/write_test.sh
```diff
@@ -0,0 +1,52 @@
...
+    testOK runCA authorize co1/private.toml $bcID
+    testOK runCA authorize co2/private.toml $bcID
+    testOK runCA authorize co3/private.toml $bcID
+    # Creat LTS and save the public key
+    runCA dkg start --instid "$LTS_ID" -x > key.pub
+    
+    PUB_KEY=`cat key.pub`
+    matchOK $PUB_KEY ^[0-9a-f]{64}$
+
+    # Fail because the Calypso rule "pawn:calypsoWrite" has not been added
```
kc1212: spawn

calypso/csadmin/commands.go
```diff
@@ -0,0 +1,198 @@
...
+			},
+			cli.BoolFlag{
+				Name:  "export, x",
+				Usage: "the decrypted data is exported to stdout",
+			},
+		},
+	},
+	{
+		Name:  "contract",
+		Usage: "use contracts",
```
kc1212: this usage information is not too useful, maybe briefly describe which contracts are available?

calypso/csadmin/commands.go
```diff
@@ -0,0 +1,198 @@
...
+						Action: clicontracts.WriteSpawn,
+						Flags: []cli.Flag{
+							cli.StringFlag{
+								Name:   "bc",
+								EnvVar: "BC",
+								Usage:  "the ByzCoin config to use (required)",
+							},
+							cli.StringFlag{
+								Name:  "darc",
+								Usage: "DARC with the right to create an LTS (default is the admin DARC)",
```
kc1212: right to create a write instance

calypso/csadmin/commands.go
```diff
@@ -0,0 +1,198 @@
...
+								EnvVar: "BC",
+								Usage:  "the ByzCoin config to use (required)",
+							},
+							cli.StringFlag{
+								Name:  "sign, s",
+								Usage: "public key of the signing entity (default is the admin)",
+							},
+							cli.StringFlag{
+								Name:  "instid, i",
+								Usage: "the instance id of the spawned LTS contract",
```
kc1212: isn't it the write contract?

calypso/csadmin/README.md
```diff
@@ -0,0 +1,123 @@
...
+$ csadmin contract read spawn --instid <write instance id> --key <hex pub key>
+```
+
+**6) Send a decrypt request**
+
+With the read and write instances, it is now possible to request the encrypted
+data. The response, `DecryptKeyReply`, is exported and saved to a file:
+
+```bash
+$ csadmin decrypt --writeid <write instance id> --readid <read instance id> -x > reply.bin
```
kc1212: decrypt and recover might be confusing. especially the decrypt command doesn't do any decryption. maybe decrypt can be called "reencrypt" and recover can be called "decrypt"?

calypso/csadmin/main.go
```diff
@@ -83,3 +91,234 @@ func authorize(c *cli.Context) error {
...
+		"- xHat: %s\n"+
+		"- X: %s\n", reply.C, reply.XhatEnc, reply.X)
+
+	return nil
+}
+
+// recover decrypts an encrypted key stored in a DecryptKeyReply. It expects the
+// DecryptKeyReply to be protobuf encoded and passed in STDIN.
+// With the --export option, the recovered data is sent to STDOUT.
+func recover(c *cli.Context) error {
```
kc1212: see my comment in the README regarding using a different name

calypso/csadmin/main.go
```diff
@@ -83,3 +91,234 @@ func authorize(c *cli.Context) error {
...
+
+	xc, err := signer.GetPrivate()
+	if err != nil {
+		return errors.New("failed to get private key: " + err.Error())
+	}
+
+	xcInv := xc.Clone().Neg(xc)
+	XhatDec := dkr.X.Clone().Mul(xcInv, dkr.X)
+	Xhat := XhatDec.Clone().Add(dkr.XhatEnc, XhatDec)
+	XhatInv := Xhat.Clone().Neg(Xhat)
```
kc1212: is it not possible to use the `RecoverKey` function?

calypso/csadmin/commands.go
nkcr: absolutely!

calypso/csadmin/main.go
nkcr: absolutely!

calypso/csadmin/clicontracts/lts_test.sh
nkcr: Yes we could, this is in anticipation if more functions are added to this clicontract (for example `get` or `delete`). 

It comes from what has been done in `byzcoin/bcadmin/clicontracts`.

calypso/csadmin/clicontracts/write.go
nkcr: Actually I'm realising this should not be called "data" and it may confuse the reader. The "data" I mention here is actually the symmetric key that is embedded into a `kyber.Point`. This has a limited space of [29 bit](https://github.com/dedis/kyber/blob/master/group/edwards25519/point.go#L107) and the user should not use it to store a file.

Instead, there is actually a `data` field that is used to stored the encrypted data. So I should rename "--data" to "--key" and add a "--data" that can be read from STDIN or given in argument and that is stored on the actual `data` field of the Writer contract. But this content has to be encrypted by the user.

calypso/csadmin/commands.go
nkcr: Text is updated. I prefer not mentioning the list of contracts since it is likely to evolve and I will forget to update it. This information is dynamically available with `-h`.

calypso/csadmin/README.md
nkcr: I agree and will make the change!

Interestingly, this naming only reflects the implementation of the Calypso service (`DecryptKey(..)` and `RecoverKey(...)`) . Having the end-user point of view sometimes raises interesting observations, as we should now ideally update the service too (but it's not in the scope of this PR).

calypso/csadmin/README.md
```diff
@@ -0,0 +1,123 @@
...
+It is also possible to directly export the pub key to a file with "-export",
+which send the hex string representation to STDOUT.
+
+**4) Spawn a write instance**
+
+With the instance id of the previously spawned LTS contract and the public key,
+spawn the write instance:
+
+```bash
+$ csadmin contract write spawn --instid <lts instance id> --data "Hello, world." -key <lts public key>
```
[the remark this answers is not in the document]
nkcr: It is now possible to provide data either with `--data` or from STDIN with `--readin` set. 

The old `--data` has been changed to `--secret`.

nkcr:
@kc1212 thank you for the valuable review ! Your comments have been addressed.

PR #2004, opened 2019-07-10
Get ByzCoinIDs in Java

kc1212:
Because it was missing...

### DIFF
--- a/external/java/src/main/java/ch/epfl/dedis/byzcoin/ByzCoinRPC.java
+++ b/external/java/src/main/java/ch/epfl/dedis/byzcoin/ByzCoinRPC.java
@@ -687,6 +687,24 @@ public class ByzCoinRPC {
     }
 
     /**
+     * Asks the roster for all the existing ByzCoin IDs.
+     *
+     * @param roster list of all cothority servers with public keys
+     * @return a list of SkipblockIds which are have a ByzCoin service running
+     * @throws CothorityCommunicationException if the communication fails or the message is corrupted
+     */
+    public static List<SkipblockId> getAllByzCoinIDs(Roster roster) throws CothorityCommunicationException {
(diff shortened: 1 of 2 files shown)


external/java/src/main/java/ch/epfl/dedis/byzcoin/ByzCoinRPC.java
```diff
@@ -686,6 +686,24 @@ public static Darc makeGenesisDarc(Signer admin, Roster roster) throws Cothority
...
 
+    /**
+     * Asks the roster for all the existing ByzCoin IDs.
+     *
+     * @param roster list of all cothority servers with public keys
+     * @return a list of SkipblockIds which are have a ByzCoin service running
+     * @throws CothorityCommunicationException if the communication fails or the message is corrupted
+     */
+    public static List<SkipblockId> getAllByzCoinIDs(Roster roster) throws CothorityCommunicationException {
+        ByzCoinProto.GetAllByzCoinIDsRequest.Builder b = ByzCoinProto.GetAllByzCoinIDsRequest.newBuilder();
```
Gilthoniel: I'm not sure how much this must be backward compatible but calling conode running 3.1.1 will fail as the message doesn't exist.

kc1212: Well, the go client API also has this problem if you put it that ways: https://github.com/dedis/cothority/blob/master/byzcoin/api.go#L94

I think it's ok because old clients will continue to work when using the old APIs on new conodes.

Gilthoniel: Oh right the fix was at the CLI level. Ok that's fair.

PR #2006, opened 2019-07-11
Uses a common script for Dockerfile and Dockerfile.dev

nkcr:
Proposal instead of #1977

Instead of having a third docker image that runs bash commands, I suggest using a common script file that each of the `Dockerfile` and `Dockerfile.dev` copy and run.

This has the advantage of not bringing additional complexity to the makefile and not having to specify a third docker image. 

The ideal would be to have a single docker image with multiple stages that could be skipped, but [this is not read yet](https://github.com/docker/cli/issues/1134).


tharvik:
> Instead of having a third docker image that runs bash commands, I suggest using a common script file that each of the `Dockerfile` and `Dockerfile.dev` copy and run.

That won't fix my initial (small) issue; when I used alternatively `Dockerfile{,dev}`, I found myself using two very different system, because one was based on `golang:1.12` and the other on `debian:stretch-slim`. To avoid this (and the related issues that might arise from that, such as missing ca-certificates, sh/bash mismatch, some system tools not available, ...), I tried to commonize the code as much as possible.

Now, the PR you propose doesn't fix that, because you have to express `Dockerfile` semantic in the common part.

> This has the advantage of not bringing additional complexity to the makefile and not having to specify a third docker image.

I don't see the added complexity in the Makefile, its supposed to be transparent to the user: I'll be build anyway as a deps when building any ~top image.

> The ideal would be to have a single docker image with multiple stages that could be skipped, but [this is not read yet](https://github.com/docker/cli/issues/1134).

One could use the `FROM ${FROM}` trick I used, and then, via Makefile, choose the dependency logic the situation demands.

nkcr:
Ok, I feel ready to go for your proposal !

Closing this PR in favour of #1977

Issue #1792, opened 2019-04-03
Clean logs to better reflect the levels


Gilthoniel:
My conode is the leader of the current cothority and I can see a lot of unnecessary logs that are growing the log files. It's also not very useful for debugging.

I actually don't know if there is a definition of the levels like:
1. If the event must be see by an admin because it requires further actions
2. If the event must be see by an admin but it is only informative
3. Important event happening from time to time in the normal path of execution
4. Event happening in the normal path of execution and helpful for debugging
5. Total crap useful for debugging

Take the tx collect protocol that is writing a lot of logs in level 3 but it's doesn't really bring much except really hard log file to decrypt.

Can I have your input on that @jeffallen @ineiti ? Thanks

ineiti:
There are in fact six more levels:
- `Fail`: die instantly with `os.Exit(1)` - it's in fact never used in the code...
- `Panic`: the internal state of the conode is inconsistent, and it's not possible for the node itself to get into a consistent state again, human intervention necessary
- `Error`: something bad happened that _really_ should never happen, but the internal state is consistent, and the node can go on without putting data into danger
- `Warn`: something non-critical that should not happen, but that the admin should know of
- `Info`: will always be printed, general info about the run. With `DEBUG_LVL==0` it doesn't show the line-number. Useful for CLIs
- `Print`: should probably be renamed to `Debug`, and I use it only to print variables during a test-run. Whenever I have `log.Print` in a PR, Jeff tells me to remove it, and I comply happily ;)
- `Lvl1`: Most important verbose informations - should very rarely happen, like at startup, shutdown, and a couple of times a day
- `Lvl2`: Unflooding verbose information - around 1/minute, showing the good execution of the conode and giving some details
- `Lvl3`: More verbose information that might be interesting to a sysadmin - around 1/second
- `Lvl4`: More detailed debugging-logs of what is happening, x/second, only useful in `go test`
- `Lvl5`: **FLOOD** - print every byte of the network, low-level debugging, `go test` only

I think the current debug-levels are a bit more verbose, meaning that this list's `Lvl3` is merged into the current `Lvl2`, and this lists `Lvl4` is more like the current `Lvl3`. So if you like my list, this is something that could be changed.

Whenever I had to chose the debug-level, I was trying to satisfy (more or less) the msg/sec description. Which means that over time what was `Lvl2` became `Lvl3` to avoid spamming of the output.

[...]

PR #2007, opened 2019-07-11
Changes a log.Lvl2 to log.Lvl3

nkcr:
This log prints a lot of information while the client does nothing.
In my opinion log =< 2 should not print anything if the client doesn't interact with the system.

If this log is valuable but there is already too much things printed at log level 3, then we should move some stuff from log level 3 to log level 4.


ineiti:
Moving things from lvl3 to lvl4 is actually a good idea... See #1792

PR #2010, opened 2019-07-12
Adding ClientTransaction.Hash

ineiti:
The detection of duplicate transactions makes it difficult to test
in the case where you want to prove that
1. the transaction doesn't go through
2. something changes (darc gets updated)
3. the same transaction as in 1 goes through

Using ClientTransaction.Hash, the same transaction can be re-submitted
after a new signature.

Also reverts patches to tests that didn't work with the old duplicate transaction detection.

This _should_ be backwards-compatible, as the test is only done in the leader. So if different version of conodes are available, it might be that sometimes given transactions are rejected as being doubles, and sometimes not...

### DIFF
--- a/byzcoin/transaction.go
+++ b/byzcoin/transaction.go
@@ -90,6 +90,20 @@ func (args Arguments) Names() []string {
 	return names
 }
 
+// Hash calculates the hash over all instructions and their signatures.
+// This creates a unique hash with regard to two sets of instructions
+// that only differ with their signature.
+func (ctx *ClientTransaction) Hash() []byte {
+	h := sha256.New()
+	for _, inst := range ctx.Instructions {
+		h.Write(inst.Hash())
+		for _, sig := range inst.Signatures {
+			h.Write(sig)
+		}
+	}
+	return h.Sum(nil)
+}
+
 // FillSignersAndSignWith fills the SignerIdentities field with the identities of the signers and then signs all the
 // instructions using the same set of  signers. If some instructions need to be signed by different sets of signers,
 // then use the SignWith method of Instruction.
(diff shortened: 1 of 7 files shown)


Gilthoniel:
Again this fix is dangerous, you rely on the fact that the signature is nondeterministic which might not be always the case. Can we take a moment to think about that ?

Maybe it's the time to think about the signer counters w.r.t. to that situation. I guess any change wouldn't be backward compatible so this may just be a temporary fix but the issue is more about the lack of randomness / something in the transaction hash.

nkcr:
For me this PR only solves one particular problem: compute the hash of a transaction wrt to all of its elements. 
Then the way we use this hash and what's inside a transaction is another story...  I'm glad we were able to spend at least one month without discussing about the counters... :Griesemer

ineiti:
@nkcr - take care - Bryan's pet peeve for some time has been to correct people who said "discuss about". It's just "discuss the counters" ;)

And even if the signature is deterministic, it's not worse than before...

byzcoin/transaction.go
```diff
@@ -90,6 +90,20 @@ func (args Arguments) Names() []string {
 	return names
 }
 
+// Hash calculates the hash over all instructions and their signatures.
+// This creates a unique hash with regard to two sets of instructions
+// that only differ with their signature.
+func (ctx *ClientTransaction) Hash() []byte {
```
kc1212: this function seems to be similar to `DeriveID`, except `DeriveID` does domain separation. can we use that instead?

kc1212: oh nevermind, `DeriveID` is only for one `Instruction`.

my general feeling is that future developers will be confused which `Hash` to use. they might not be aware that there's one under `Instructions` if they see the `ClientTransaction` one first. my suggestion is to move this one to `Instructions` but rename is to `HashWithSignatures`, for example. so that future developers see two versions of it at the same place and then needs to make a conscious decision on which one to use

PR #2011, opened 2019-07-12
Progress download

### LINKED ISSUE #1792: Clean logs to better reflect the levels

ineiti:
Show a progress indicator when the node downloads the DB of the
global state.

Closes #1792


byzcoin/service.go
```diff
@@ -554,11 +555,13 @@ func (s *Service) DownloadState(req *DownloadState) (resp *DownloadStateResponse
 		s.downloadState.stop = make(chan bool)
 		nonce := binary.LittleEndian.Uint64(random.Bits(64, true, random.New()))
 		s.downloadState.nonce = nonce
+		total := make(chan int)
 		go func(ds downloadState) {
 			idStr := fmt.Sprintf("%x", ds.id)
 			db, bucketName := s.GetAdditionalBucket([]byte(idStr))
 			err := db.View(func(tx *bbolt.Tx) error {
 				bucket := tx.Bucket(bucketName)
+				total <- bucket.Stats().KeyN
```
kc1212: i think `bucket.Stats().KeyN` will give you the total number of key-value pairs in the boltdb bucket which is always higher than the total number of states in byzcoin because the intermediate nodes in the trie are also stored as key-value pairs in boltdb. so i'm not sure whether this is what you want to show

byzcoin/proto.go
```diff
@@ -311,6 +311,8 @@ type DownloadStateResponse struct {
 	// is generated by the server, and will be set
 	// for every subsequent reply, too.
 	Nonce uint64
+	// Total key/value pairs.
+	Total int
```
kc1212: could you add `protobuf:"opt"` for backward-compatibility?

byzcoin/service.go
kc1212: nevermind, it should be fine because `len(KeyValues)` is actually the boltdb key-values

ineiti: Yes, and we need to transmit the whole state, because the merkle-tree retains some history of the adding and deleting of instances. So if we just transmit the instances, the created merkle-tree will have a different root!

byzcoin/proto.go
ineiti: Done.

PR #2014, opened 2019-07-16
Return better errors in AddTransactionAndWait

### LINKED ISSUE #2013: Return errro messages for transaction failures
We need a better to present the errors during a transaction failure. Right now the error simply gets logged by `log.Error`. The return value of a `AddTransactionAndWait` is always something like "transaction got refused" but it never says why it got refused.

We need to return a better error so that the caller and the test can check exactly what happened.

kc1212:
Closes #2013


byzcoin/service.go
```diff
@@ -391,8 +393,16 @@ func (s *Service) AddTransaction(req *AddTxRequest) (*AddTxResponse, error) {
...
 				if !success {
-					return nil, errors.New("transaction is in block, but got refused")
+					if !exists {
+						return nil, errors.New("transaction is in block, but got refused for unknown error")
+					}
+					return nil, errors.New(trimErrorMsg(errMsg))
+				}
+
+				if exists {
+					log.Warn(s.ServerIdentity(), "transaction is accepted but there are errors: ", errMsg)
```
ineiti:
```suggestion
					log.Warn(s.ServerIdentity(), "transaction is in block but got refused for the following reason: ", errMsg)
```

byzcoin/service.go
```diff
@@ -1996,9 +2036,16 @@ func (s *Service) createStateChanges(sst *stagingStateTrie, scID skipchain.SkipB
...
-// processOneTx takes one transaction and creates a set of StateChanges. It also returns the temporary StateTrie
-// with the StateChanges applied. Any data from the trie should be read from sst and not the service.
-func (s *Service) processOneTx(sst *stagingStateTrie, tx ClientTransaction) (StateChanges, *stagingStateTrie, error) {
+// processOneTx takes one transaction and creates a set of StateChanges. It
+// also returns the temporary StateTrie with the StateChanges applied. Any data
+// from the trie should be read from sst and not the service.
+func (s *Service) processOneTx(sst *stagingStateTrie, tx ClientTransaction) (newStateChanges StateChanges, newStateTrie *stagingStateTrie, err error) {
+	defer func() {
+		if err != nil {
+			s.txErrorBuf.add(string(tx.Instructions.Hash()), err.Error())
```
ineiti:
```suggestion
			s.txErrorBuf.add(string(tx.Instructions.HashWithSignatures()), err.Error())
```

byzcoin/service.go
```diff
@@ -2656,6 +2718,7 @@ func newService(c *onet.Context) (onet.Service, error) {
 		closed:                 true,
 		catchingUpHistory:      make(map[string]time.Time),
 		rotationWindow:         defaultRotationWindow,
+		txErrorBuf:             newRingBuf(10),
```
ineiti: This should be big enough to keep at least all errors from all transactions from two blocks. If we suppose that all transactions create errors, for a 1MB block with 1kB transactions, that is 1024 transactions. For two blocks (just to be sure in case `AddTransaction` is behind one block), this makes 2048 errors.

```suggestion
		txErrorBuf:             newRingBuf(2048),
```

byzcoin/ringbuf.go
```diff
@@ -0,0 +1,44 @@
...
+func newRingBuf(size int) ringBuf {
+	return ringBuf{
+		current: 0,
+		size:    size,
+		items:   make([]ringBufElem, size),
+	}
+}
+
+type ringBufElem struct {
+	key   string
```
ineiti: why is this `string` and not `[]byte`?

byzcoin/service.go
```diff
@@ -391,8 +393,16 @@ func (s *Service) AddTransaction(req *AddTxRequest) (*AddTxResponse, error) {
 		for found := false; !found; {
 			select {
 			case success := <-ch:
+				errMsg, exists := s.txErrorBuf.get(string(ctxHash))
```
ineiti:
```suggestion
				errMsg, exists := s.txErrorBuf.get(string(req.Transaction.Instructions.HashWithSignatures()))
```

ineiti:
Nice ringbuffer - the only change I would make is not to use `Hash`, but rather `HashWithSignatures`. Else you might return wrong errors...

ineiti:
Also, looking at the failed test, I see that the error-message is:

`" this invalid contract always returns an error"`

So your `trimError` method leaves a leading space, which is really not nice and might confuse people when they try to match the error...
Also, `trimError` looks a bit dangerous to me: imagine that somebody matches an error, and then we change the `fmt.Errorf` in `processOneTx`, and all of a sudden the error of the contract is 'different'... Personally I would remove `trimError` and just return the whole string.

ineiti:
Commented on the error-size in the issue...

byzcoin/service.go
kc1212: i wanted to say the transaction is accepted because the `success` flag is true, but for some reason there is a an error message (which should never happen) so we print a warning

byzcoin/ringbuf.go
kc1212: because history... it started as a map then changed to a slice but the string type stayed

now i've changed it to `[]byte`

kc1212:
Due to the issue regarding error message size and the fact that we cannot send both a reply and an error at the same time. I've decided to keep this PR more or less as is (except the issues raised by @ineiti are addressed). The new `trimErrorMsg` doesn't try to do anything fancy with the colons anymore, it just checks if length is too long and truncates it.

byzcoin/ringbuf.go
```diff
@@ -0,0 +1,45 @@
+package byzcoin
+
```
nkcr: I know it's not a common practice in this repo, but can you add few sentences describing why this file is needed? (I have a bad memory, and I know that I will soon wonder why this file is here...)

byzcoin/service.go
```diff
@@ -415,6 +425,14 @@ func (s *Service) AddTransaction(req *AddTxRequest) (*AddTxResponse, error) {
 	}, nil
 }
 
+func trimErrorMsg(errMsg string) string {
+	const limit = 105
```
nkcr: Can you comment why this value?

kc1212:
Thanks @nkcr, documentation added

byzcoin/service.go
```diff
@@ -316,7 +318,10 @@ func (s *Service) CreateGenesisBlock(req *CreateGenesisBlock) (
 	}, nil
 }
 
-// AddTransaction requests to apply a new transaction to the ledger.
+// AddTransaction requests to apply a new transaction to the ledger. Note is
```
jeffallen:
```suggestion
// AddTransaction requests to apply a new transaction to the ledger. Note
```

PR #2018, opened 2019-07-23
external/js/kyber: add @types/bn.js for users

tharvik:
Currently, using `@dedis/kyber` with typescript requires to do `npm i -D @types/bn.js` before building. Moving it from `devDependencies` to `dependencies` fixes that. (it was also bumped during that).

Can also someone make a NPM release after merging please?

### DIFF
--- a/external/js/kyber/package-lock.json
+++ b/external/js/kyber/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@dedis/kyber",
-  "version": "3.0.6",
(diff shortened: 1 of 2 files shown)


kc1212:
In the code, `@dedis/kyber` is still in `devDependencies`

tharvik:
> In the code, `@types/bn.js` is still in `devDependencies`

You're right, found out that `npm i smth` and `npm uninstall && npm i smth` doesn't yield the same result.

PR #2019, opened 2019-07-23
Implement StoreAllToReplica

### LINKED ISSUE #1938: Deferred contract: index out of range with rst.GetIndex()

kc1212:
Doing ForEach in the deferred contract might become expensive when the
state grows big. So we introduce StoreAllToReplica that makes a
"lightweight" copy of the trie and then applies the state changes. It
still makes copies of key/values but only those that are waiting to be
committed to a new block.

Closes #1938

### DIFF
--- a/byzcoin/statetrie.go
+++ b/byzcoin/statetrie.go
@@ -95,10 +95,12 @@ func (t *stagingStateTrie) GetIndex() int {
 	return int(index)
 }
 
+// StoreAllToReplica creates a copy of the read-only trie and applies the state
+// changes to the copy.
 func (t *stagingStateTrie) StoreAllToReplica(scs StateChanges) (ReadOnlyStateTrie, error) {
 	newTrie := t.Clone()
 	if err := newTrie.StoreAll(scs); err != nil {
-		return nil, err
+		return nil, errors.New("replica failed to store state changes: " + err.Error())
 	}
 	return newTrie, nil
 }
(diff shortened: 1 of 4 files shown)


nkcr:
Do not forget to uncomment https://github.com/dedis/cothority/blob/master/byzcoin/bcadmin/clicontracts/deferred_test.sh#L9.

kc1212:
I'm keeping the integration test commended because it wasn't trivial to fix. Here's the issue to track is https://github.com/dedis/cothority/issues/2021

byzcoin/statetrie.go
```diff
@@ -90,6 +95,14 @@ func (t *stagingStateTrie) GetIndex() int {
 	return int(index)
 }
 
+func (t *stagingStateTrie) StoreAllToReplica(scs StateChanges) (ReadOnlyStateTrie, error) {
```
nkcr: Can you add a comment on this one? Like" it copies the stagingStateTrie but keeps the disk-backed tries".

Also I know its not common but wrapped errors are much easier to debug. Can you format the error with something like `"replica failed to store state changes: " + err.Error()`?

PR #2026, opened 2019-07-26
Merges the /conode doc

nkcr:
The documentation for setting up a conode was shared between multiple .md files that were sometimes containing the same duplicated information.
I think that having multiple distinct .md files linked together makes the job of reading and maintaining the documentation harder than sticking to a single README.md. This is why I merged them and refactored the doc to a single one.

I packed up a script that automatically creates/update a table of content for a given markdown file. This makes the reading of long md files easier.


conode/README.md
```diff
@@ -1,56 +1,612 @@
...
+
+Once the setup is done with one of the two options, you can finally run your
+conode depending on the previously chosen option.
+
+### Option 1: :computer: Run with the command line
+
+Simply launch the following:
+
+```bash
+$ conode server
```
kc1212: we often run the command like `conode server | tee logfile.txt` so maybe it's a good idea to mention that the log files are not written anything so the user should use `tee` if they want to keep old logs

conode/README.md
```diff
@@ -1,56 +1,612 @@
...
 
-For most of the apps you need at least 3 running nodes. Once you have them up
+```
+$ wget https://raw.githubusercontent.com/dedis/cothority/conode/conode.service
+$ systemctl --user enable conode.service
+$ systemctl --user start conode
+```
+
+Unfortunately, systemd doesn't allow a user to run a service at system startup,
+and all user services get stopped once the user logs out!
```
kc1212: I think it is possible to run user systemd services at startup, see https://wiki.archlinux.org/index.php/Systemd/User#Automatic_start-up_of_systemd_user_instances

conode/README.md
```diff
@@ -1,56 +1,612 @@
...
+2. The file containing all the public configurations will be in
+"tmp/public.toml".
+
+# Operating a conode
+
+## Backups
+
+On Linux, the following files need to be backed up:
+1. `$HOME/.config/conode/private.toml`
+2. `$HOME/.local/share/conode/$PUBLIC_KEY.db`
```
kc1212: the db files are not hex-encoded public keys, they are the sha256 of the public keys (see the `dbFileName()` function in onet)

kc1212: probably just ask the user to backup the whole conode folder

conode/README.md
```diff
@@ -1,56 +1,612 @@
...
+
+If using the command line, or the `$HOME/conode_data` folder if using docker.
```
kc1212: I didn't understand this sentence

conode/README.md
```diff
@@ -1,56 +1,612 @@
...
+  - [Environment setup](#environment-setup)
+  - [Configuration setup](#configuration-setup)
+    - [Option 1: :computer: Configuration setup with the command line](#option-1-computer-configuration-setup-with-the-command-line)
+    - [Option 2: :whale: Configuration setup with docker](#option-2-whale-configuration-setup-with-docker)
+    - [Post setup instructions](#post-setup-instructions)
+  - [Run your conode](#run-your-conode)
+    - [Option 1: :computer: Run with the command line](#option-1-computer-run-with-the-command-line)
+    - [Option 2: :whale: Run with docker](#option-2-whale-run-with-docker)
+    - [Option 3: `run_nodes.sh`](#option-3-run_nodessh)
+- [Operating a conode](#operating-a-conode)
```
kc1212: "Operating a conode" is very similar to "Running a conode". Maybe change it to "Maintaining a conode"?

conode/README.md
```diff
@@ -1,56 +1,612 @@
...
+All the configuration is stored in `conode_data`
+
+## Development version with docker
+
+For the latest and greatest version of the conode, you can replace
+`conode:latest` with `conode:dev` and you should get a stable, but changing
+conode. This means, that to use all the functionalities you need to update the
+apps and follow the latest `conode:dev` container regularly.
+
+## Creating Your Cothority
```
kc1212: this section seems to be out-of-place, it's in between sub-sections about docker. maybe combine it with "Setting up more than one node" or "Verifying your server"? you can tell the user to create the combined roster.toml and then the `conode check` command should check all the conodes.

doctoc.sh
```diff
@@ -0,0 +1,180 @@
...
+#
+# Attributions: Updated from https://gitlab.com/pedrolab/doctoc.sh/tree/master,
+# which comes from
+# https://gist.github.com/meleu/57867f4a01ede1bd730f14b2f018ae89.
+#
+# The list of invalid chars come from https://github.com/thlorenz/anchor-\
+# markdown-header/blob/56f77a232ab1915106ad1746b99333bf83ee32a2/anchor-\
+# markdown-header.js#L25
+#
+
```
kc1212: it's a good practice to start the shell script with
```
set -e
set -u
```
so that it exists early when an error occurs

doctoc.sh
```diff
@@ -0,0 +1,180 @@
...
+INVALID_CHARS="'[]/?!:\`.,()*\";{}+=<>~$|#@&–—"
+
+# For Mac user, we need to use 'gsed'
+SED=sed
+
+appname='doctoc.sh'
```
kc1212: why not just use `basename $0`?

conode/README.md
nkcr: I think this section comes from @ineiti, I left it "as is"

nkcr:
Thanks for the review @kc1212, comments have been addressed 💯

PR #2031, opened 2019-07-30
Use getProofFromLatest

ineiti:
To speed up usage in a bigger ByzCoin chain, use
getProofFromLatest by default.

### DIFF
--- a/external/js/cothority/src/byzcoin/byzcoin-rpc.ts
+++ b/external/js/cothority/src/byzcoin/byzcoin-rpc.ts
@@ -75,7 +75,7 @@ export default class ByzCoinRPC implements ICounterUpdater {
         rpc.genesis = await skipchain.getSkipBlock(skipchainID);
         rpc.latest = rpc.genesis;
 
-        const ccProof = await rpc.getProofFromLatest(CONFIG_INSTANCE_ID, waitMatch, interval);
+        const ccProof = await rpc.getProof(CONFIG_INSTANCE_ID, waitMatch, interval);
         rpc.config = ChainConfig.fromProof(ccProof);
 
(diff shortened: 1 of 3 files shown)


external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -75,7 +75,7 @@ export default class ByzCoinRPC implements ICounterUpdater {
         rpc.genesis = await skipchain.getSkipBlock(skipchainID);
         rpc.latest = rpc.genesis;
 
-        const ccProof = await rpc.getProof(CONFIG_INSTANCE_ID, waitMatch, interval);
+        const ccProof = await rpc.getProofFromLatest(CONFIG_INSTANCE_ID, waitMatch, interval);
```
Gilthoniel: I didn't change this one because the RPC is created a few lines above and the latest block is populated with the genesis. That means this call of `getProofFromLatest` won't get any benefit as it will be the first time we check the full chain (and it is required as you only know the genesis).

external/js/cothority/src/calypso/calypso-rpc.ts
```diff
@@ -43,7 +43,7 @@ export class OnChainSecretRPC {
         await ctx.updateCountersAndSign(this.bc, [signers]);
         await this.bc.sendTransactionAndWait(ctx);
         // Ask for the full proof which is easier to verify.
-        const p = await this.bc.getProof(ctx.instructions[0].deriveId());
+        const p = await this.bc.getProofFromLatest(ctx.instructions[0].deriveId());
```
Gilthoniel: The proof is sent to create the LTS. In that situation, it's dangerous to send a short proof as you don't know the state of the server that will receive the message.

ineiti:
Closed as all calls to `getProof` are needed to interact correctly with LTS.

PR #2033, opened 2019-08-06
DARC attributes

### LINKED ISSUE #2012: Verify additional data in DARCs
Some of our partners may need to do additional verification using DARCs other than relying on only public keys. For example, our DARCs cannot verify a statement like "requests for medical records must be made between 9 AM and 5 PM".

The type of additional data is not clearly specified at the point so our implementation needs to be extensible. Nevertheless, from looking at a few examples in the [XACML spec](http://docs.oasis-open.org/xacml/2.0/access_control-xacml-2.0-core-spec-os.pdf), we are able to classify the types of additional data into the following categories.

1. Additional data based on global/system data in ByzCoin, e.g., time and block index).
2. Additional data based on instance. Suppose we are keeping medical records, we may have a rule that says the queriers are able to access his own records, which means the signer ID must match one of the records in the instance.
3. Additional data based on external data, e.g., weather of the day or whether an event happened.
4. Additional data based on "temporary" or "local" data, e.g., the IP address of the client who sent the transaction.

For the first two data-types, we can implement it in the following way. Add `AdditionalRules map[Action]Rule` to the DARC struct. Where the key is an `Action` which must be an existing rule in `Rules`. We also add a `func (instr Instruction) VerifyAdditionalRules(st ReadOnlyStateTrie) error` which has the ability to verify some default `AdditionalRules`. Suppose a doctor is only allowed to modify patient data between 9 AM and 5 PM. The standard DARC rule would look like `invoke:modify_record => ed25519:doctor1 | ed25519:doctor1`. Then the additional rule would look like `time_interval:9AM;5PM` which is keyed on `invoke:view_record`. If `VerifyAdditionalRules` implements `time_interval` verification and is used by default in `BasicContract`, then the rule will be enforced automatically.

For the third data-type, we need an oracle to do the verification. I believe the rule can still be encoded the same way as above in `AdditionalRules`. But the contract author needs to write his/her custom verification function that contracts the oracle.

Finally, the fourth data-type cannot be easily verified collectively because the information needed for verification only exists on one node. If it must be done, we need to offload the information to the oracle. For example, the oracle independently checks the IP addresses of clients and the conodes will trust the oracle to correctly record the IP address. 

Here's the proposed change and it should be backwards compatible (only the new types and functions no implementation).

```diff
diff --git a/byzcoin/contracts.go b/byzcoin/contracts.go
index 992a15563..854e1b94a 100644
--- a/byzcoin/contracts.go
+++ b/byzcoin/contracts.go
@@ -196,7 +196,10 @@ func notImpl(what string) error { return fmt.Errorf("this contract does not impl
 // VerifyInstruction offers the default implementation of verifying an instruction. Types
 // which embed BasicContract may choose to override this implementation.
 func (b BasicContract) VerifyInstruction(rst ReadOnlyStateTrie, inst Instruction, ctxHash []byte) error {
-       return inst.Verify(rst, ctxHash)
+       if err := inst.Verify(rst, ctxHash); err != nil {
+               return err
+       }
+       return inst.VerifyAdditionalRules(st)
 }

 // VerifyDeferredInstruction is not implemented in a BasicContract. Types which
diff --git a/byzcoin/transaction.go b/byzcoin/transaction.go
index 5dcb98285..4526f3fdd 100644
--- a/byzcoin/transaction.go
+++ b/byzcoin/transaction.go
@@ -371,6 +371,12 @@ func (instr Instruction) VerifyWithOption(st ReadOnlyStateTrie, msg []byte, chec
        return darc.EvalExpr(d.Rules.Get(darc.Action(instr.Action())), getDarc, goodIdentities...)
 }

+// VerifyAdditionalRules verifies the metadata rules.
+func (instr Instruction) VerifyAdditionalRules(st ReadOnlyStateTrie) error {
+       // TODO add some default verification here like date/time
+       return nil
+}
+
 // InstrType is the instruction type, which can be spawn, invoke or delete.
 type InstrType int

diff --git a/darc/proto.go b/darc/proto.go
index 282e7f29e..517c48ce8 100644
--- a/darc/proto.go
+++ b/darc/proto.go
@@ -51,6 +51,15 @@ type Darc struct {
        // verify this darc. It is not needed in online verification where the
        // verifier stores all darcs.
        VerificationDarcs []*Darc
+       // AdditionalRules is used for adding extra properties/rules for the
+       // existing rules. The rules that should go in this field are ones that
+       // are not related to signatures but are nevertheless globally
+       // verifiable. For example, time or block index is a global property
+       // that any conode can verify (if we assume perfect clocks). Access
+       // control policy of specific records is also a global property. But
+       // the IP address of the sender is not a global property because it can
+       // only be seen by the initial receiver of the transaction.
+       AdditionalRules map[Action]Rule `protobuf:"opt"`
 }

 // Identity is a generic structure can be either an Ed25519 public key, a Darc
```

kc1212:
Closes #2012


darc/darc.go
```diff
@@ -619,9 +630,35 @@ func EvalExpr(expr expression.Expr, getDarc GetDarc, ids ...string) error {
...
+			if attrFunc, ok = attrFuncs[tokens[1]]; !ok {
+				issue = errors.New("no such attr interpreter: " + tokens[1])
+				return false
+			}
+			if err := attrFunc(tokens[2]); err != nil {
+				issue = err
+				return false
+			}
+			return true
+		}
```
cgrigis: Suggestion: for clarity, the `if` branch could have its contents isolated into its own function :
```
func evalAttr(s string, attrFuncs AttrInterpreters) error {
    tokens := strings.SplitN(s, ":", 3)
    if len(tokens) != 3 {
        return errors.New("attr has an invalid format")
    }
    if tokens[0] != "attr" {
        return errors.New("first token should be attr")
    }

    var ok bool
    var attrFunc func(string) error
    if attrFunc, ok = attrFuncs[tokens[1]]; !ok {
        return errors.New("no such attr interpreter: " + tokens[1])
    }
    return attrFunc(tokens[2])
}
```
and reduced to:
```
if strings.HasPrefix(s, "attr") {
    if err := evalAttr(s, attrFuncs); err != nil {
        issue = err
        return false
    }
   return true
}
```
But that's just because I like short functions. :-)

darc/expression/expression_test.go
```diff
@@ -188,6 +188,50 @@ func TestParsing_RealIDs(t *testing.T) {
 	}
 }
 
+func TestParsing_Attr(t *testing.T) {
+	expr := []byte("attr:xyz_-:zys=123&sdy=234")
```
cgrigis: Maybe add a test to ensure JSON-encoded attribute arguments are supported, as it might be a practical way to pass more complex data?

byzcoin/contracts.go
```diff
@@ -206,6 +207,52 @@ func (b BasicContract) VerifyDeferredInstruction(rst ReadOnlyStateTrie, inst Ins
 	return notImpl("VerifyDeferredInstruction")
 }
 
+// AttrInterpreters provides one default attribute verification which check
+// whether the transaction is sent after a certain block index and before
+// another block index.
+func (b BasicContract) AttrInterpreters(rst ReadOnlyStateTrie, inst Instruction) map[string]func(string) error {
```
cgrigis:
```suggestion
func (b BasicContract) AttrInterpreters(rst ReadOnlyStateTrie, inst Instruction) darc.AttrInterpreters {
```

cgrigis:
```suggestion
func (b BasicContract) GetAttrInterpreters(rst ReadOnlyStateTrie, inst Instruction) map[string]func(string) error {
```
The function rename would be to avoid confusion with the type itself.

byzcoin/transaction.go
```diff
@@ -368,6 +379,10 @@ func (instr Instruction) VerifyWithOption(st ReadOnlyStateTrie, msg []byte, chec
 		}
 		return d
 	}
+
+	if ops.EvalAttr != nil {
+		return darc.EvalExprAttr(d.Rules.Get(darc.Action(instr.Action())), getDarc, ops.EvalAttr, goodIdentities...)
```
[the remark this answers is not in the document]
kc1212: yep

byzcoin/contracts.go
kc1212: I'm changing it to `MakeAttrInterpreters` because it's dynamically creating those interpreters from the given `rst` and `inst`.

PR #2034, opened 2019-08-08
Leader detects when to upgrade the byzcoin protocol version

Gilthoniel:
This updates the transaction collection to include the version of the
Byzcoin protocol of the conode so that the leader can synchronize an
upgrade.

### DIFF
--- a/byzcoin/collect_tx.go
+++ b/byzcoin/collect_tx.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"time"
 
+	"go.dedis.ch/cothority/v3/byzcoinx"
 	"go.dedis.ch/cothority/v3/skipchain"
 	"go.dedis.ch/onet/v3"
 	"go.dedis.ch/onet/v3/log"
@@ -21,16 +22,17 @@ func init() {
 // CollectTxProtocol is a protocol for collecting pending transactions.
 type CollectTxProtocol struct {
 	*onet.TreeNodeInstance
-	TxsChan      chan []ClientTransaction
-	SkipchainID  skipchain.SkipBlockID
-	LatestID     skipchain.SkipBlockID
-	MaxNumTxs    int
-	requestChan  chan structCollectTxRequest
-	responseChan chan structCollectTxResponse
-	getTxs       getTxsCallback
-	Finish       chan bool
-	closing      chan bool
-	version      int
+	TxsChan           chan []ClientTransaction
+	CommonVersionChan chan Version
+	SkipchainID       skipchain.SkipBlockID
+	LatestID          skipchain.SkipBlockID
+	MaxNumTxs         int
+	requestChan       chan structCollectTxRequest
+	responseChan      chan structCollectTxResponse
+	getTxs            getTxsCallback
+	Finish            chan bool
+	closing           chan bool
+	version           int
 }
 
 // CollectTxRequest is the request message that asks the receiver to send their
@@ -45,7 +47,8 @@ type CollectTxRequest struct {
 // CollectTxResponse is the response message that contains all the pending
 // transactions on the node.
 type CollectTxResponse struct {
-	Txs []ClientTransaction
+	Txs            []ClientTransaction
+	ByzcoinVersion Version
 }
 
 type structCollectTxRequest struct {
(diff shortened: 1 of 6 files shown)


byzcoin/service_test.go
```diff
@@ -513,6 +514,76 @@ func TestService_AddTransactionVersion(t *testing.T) {
...
+		wg.Add(1)
+		defer wg.Done()
+
+		c := uint64(1)
+		shutdown := false
+		wait := 0
+		for !shutdown {
+			select {
+			case <-closing:
+				shutdown = true
```
ineiti: Just out of curiosity: why doesn't this just `return`?

byzcoin/service_test.go
```diff
@@ -513,6 +514,76 @@ func TestService_AddTransactionVersion(t *testing.T) {
...
+			require.NoError(t, err)
+
+			c++
+			time.Sleep(50 * time.Millisecond)
+		}
+	}(closing)
+
+	time.Sleep(testInterval)
+
+	// Simulate a upgrade of the conodes.
```
ineiti:
```suggestion
	// Simulate an upgrade of the conodes.
```

byzcoin/tx_pipeline.go
```diff
@@ -30,6 +37,9 @@ type txProcessor interface {
 	// function should only return when a decision has been made regarding
 	// the proposal.
 	ProposeBlock(*txProcessorState) error
+	// ProposeUpgradeBlock should create a barrier block between two Byzcoin
+	// version so that next blocks will use the version in parameter.
```
ineiti:
```suggestion
	// version so that future blocks will use the new version.
```

byzcoin/service_test.go
Gilthoniel: because I wanted the last transaction to take the `wait = 10` so everything closes correctly.

byzcoin/tx_pipeline.go
```diff
@@ -150,9 +160,14 @@ func (s *defaultTxProcessor) CollectTx() ([]ClientTransaction, error) {
 	protocolTimeout := time.After(bcConfig.BlockInterval / 2)
 
 	var txs []ClientTransaction
+	commonVersion := Version(0)
+
 collectTxLoop:
 	for {
 		select {
+		case commonVersion = <-root.CommonVersionChan:
+			// At this point, we only know if other conodes are running the same
```
ineiti: I don't understand the comment, here...

Gilthoniel: Just to say that the value gives a version that is the same for a threshold of conodes but it can be the latest version available so it needs to check that to not create a block to upgrade from version x to x (which is not an upgrade per se)

ineiti: OK, now I understand. I'd put your answer in the comment then ;)

byzcoin/tx_pipeline.go
```diff
@@ -12,12 +12,19 @@ import (
 	"go.dedis.ch/protobuf"
 )
 
+// CollectTxResult contains the aggregated response of the conodes to the
+// collectTx protocol.
+type CollectTxResult struct {
```
kc1212: does it need to be public?

byzcoin/collect_tx.go
```diff
@@ -165,6 +177,10 @@ func (p *CollectTxProtocol) Dispatch() error {
 			case <-p.closing:
 				return nil
 			}
+
+			if vb.hasThresholdFor(leaderVersion) {
+				p.CommonVersionChan <- leaderVersion
+			}
```
kc1212: why is this part inside the for loop? shouldn't it be outside because it's waiting for all the responses and then computing the common version from those responses?

kc1212: if it's outside the channel size can be 1

byzcoin/tx_pipeline_test.go
```diff
@@ -178,6 +201,20 @@ func newSlowCollectMockTxProc(t *testing.T, batch int, txs []ClientTransaction,
 	return proc
 }
 
+func newDefaultMockTxVersion(t *testing.T, batch int, txs []ClientTransaction, failAt int) mockTxProc {
```
kc1212: is this used?

PR #2036, opened 2019-08-14
Mobile update

ineiti:
Adding changes from the mobile-branch of omniledger. Mostly cosmetic changes and copy-methods to access private fields.

### DIFF
--- a/external/js/cothority/src/calypso/calypso-rpc.ts
+++ b/external/js/cothority/src/calypso/calypso-rpc.ts
@@ -105,6 +105,13 @@ export class LongTermSecret extends OnChainSecretRPC {
             roster = bc.getConfig().roster;
         }
         const ocs = new OnChainSecretRPC(bc);
+        try {
+            await ocs.authorizeRoster(roster);
+        } catch (e) {
+            if (e.toString().indexOf("already authorised") < 0) {
+                throw new Error(e);
+            }
+        }
         const lr = await ocs.createLTS(roster, darcID, signers);
         return new LongTermSecret(bc, lr.instanceid, lr.X, roster);
     }
--- a/external/js/cothority/src/log.ts
+++ b/external/js/cothority/src/log.ts
@@ -15,6 +15,7 @@ export class Logger {
         return this._lvl;
     }
     _lvl: number;
+    stackFrameOffset: number = 0;
 
     constructor(lvl: number) {
         this._lvl = lvl === undefined ? defaultLvl : lvl;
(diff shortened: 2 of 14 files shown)


external/js/cothority/src/calypso/calypso-rpc.ts
```diff
@@ -105,6 +105,13 @@ export class LongTermSecret extends OnChainSecretRPC {
             roster = bc.getConfig().roster;
         }
         const ocs = new OnChainSecretRPC(bc);
+        try {
+            await ocs.authorizeRoster(roster);
+        } catch (e) {
+            if (e.toString().indexOf("already authorised") < 0) {
```
nkcr: If this is the only way of checking it, I would at least comment the code where the error is raised to indicate this dangerous coupling.

external/js/cothority/src/log.ts
```diff
@@ -75,7 +73,7 @@ export class Logger {
             // @ts-ignore
             return (file).padEnd(20);
         } catch (e) {
-            return this.printCaller(new Error("Couldn't get stack - " + e), i + 2);
+            return this.out("Couldn't get stack - " + e.toString(), (i + 2).toString());
```
nkcr: Where does this magic value of 2 come from? 
There seems to be an unnecessary extra space.

external/js/cothority/src/personhood/spawner-instance.ts
```diff
@@ -38,6 +38,10 @@ export default class SpawnerInstance extends Instance {
             .add(this.struct.costCredential.value);
     }
```
nkcr: To be consistent with the rest I would comment this function

external/js/cothority/src/log.ts
ineiti: It's because the stack includes the `printCaller` and `{lvl*,print,...}` method.

external/js/cothority/src/calypso/calypso-rpc.ts
ineiti: This is raised in `dedis/cothority/byzcoin`. Any better idea how the coupling can be removed?

external/js/cothority/src/personhood/spawner-instance.ts
ineiti: Removed `signupCost`, as it is too specific and not used at all.

external/js/cothority/src/calypso/calypso-rpc.ts
nkcr: I don't have a better idea, I would simply add a comment where you mentioned, and update the test to capture this new behaviour.

ineiti: Done.

PR #2039, opened 2019-08-16
revamp travis.yml

tharvik:
Nearly full rewrite of `travis.yml`. With magic YAML feature such as anchors to reduce code duplication.

* separate the build by stages, so that we avoid running the tests if the build fails for example ie faster feedback
* ensure that we really support the wanted version of go/js: every job using go is duplicated into 1.11 and 1.12
* dropping support for go 1.10, because it doesn't support modules; for eg, it can't lint the project
  * it's easy to add back if wanted, but there isn't currently full support for it
* add deployment to NPM of external/js/{kyber,cothority} on a `dev` tag (the feature I initially wanted when starting this PR)
* IMHO, with this PR, it's easier to read/code/manage `travis.yml`

NB: if you want to actually use the deploy stage, one would need to **set `DEPLOY_NPM_TOKEN` into cothority's travis' settings before merging**; also, to test it before pusing it to master, one need to explicitly set `jobs.include.deploy.on.branch` to match the branch one's want.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,10 @@ _: &stage_test_go
   script: GO111MODULE=on make test_goveralls
 _: &stage_deploy_npm
   <<: *language_js
-  script: skip
+  script: 
+    - CHANGED_FILES=$(git diff --name-only $TRAVIS_COMMIT_RANGE)
+    - JS_COTHORITY_N=`echo "$CHANGED_FILES" | grep -E -c "^external/js/cothority/" || true`
+    - JS_KYBER_N=`echo "$CHANGED_FILES" | grep -E -c "^external/js/kyber/" || true`
   before_deploy: echo "//registry.npmjs.org/:_authToken=${DEPLOY_NPM_TOKEN}" > $HOME/.npmrc
 
 
@@ -43,8 +46,7 @@ stages:
   - lint
   - build
   - test
-  - name: deploy
-    if: branch = master
+  - deploy
 
 jobs:
   include:
@@ -112,22 +114,28 @@ jobs:
       name: "NPM: js > kyber"
       <<: *stage_deploy_npm
       deploy:
+        on:
+          branch: master
+          condition: $JS_KYBER_N != 0
         provider: script
         script: >-
           cd external/js/kyber &&
           npm ci &&
-          npm version prerelease --preid=p`date +%Y%m%d%H%M%S` &&
+          npm version prerelease --preid=p`date +%y%m.%d%H.%M%S` &&
           ./publish.sh --tag dev
     - name: "NPM: js > cothority"
       <<: *stage_deploy_npm
       deploy:
+        on:
+          branch: master
+          condition: $JS_COTHORITY_N != 0
         provider: script
         script: >-
           cd external/js/kyber && npm ci && npm run link && cd ../../.. &&
           cd external/js/cothority &&
           npm ci &&
           npm link @dedis/kyber &&
-          npm version prerelease --preid=p`date +%Y%m%d%H%M%S` &&
+          npm version prerelease --preid=p`date +%y%m.%d%H.%M%S` &&
           ./publish.sh --tag dev
 
 notifications:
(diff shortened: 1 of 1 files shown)


nkcr:
I am trying to check if the build works but even if I remove the condition, I still get the message "Skipping a deployment with the script provider because this branch is not permitted: travis-test": https://travis-ci.org/dedis/cothority/builds/577699483

tharvik:
> I am trying to check if the build works but even if I remove the condition, I still get the message "Skipping a deployment with the script provider because this branch is not permitted: travis-test": https://travis-ci.org/dedis/cothority/builds/577699483

Hum, I was wrong, it's not working as wanted: the working fix is to add
```yaml
  deploy:
    ...
    on:
      branch: my-custom-branch
```

.travis.yml
```diff
@@ -1,104 +1,135 @@
...
 
+    - stage: deploy
+      name: "NPM: js > kyber"
+      <<: *stage_deploy_npm
+      deploy:
+        provider: script
+        script: >-
+          cd external/js/kyber &&
+          npm ci &&
+          npm version prerelease --preid=p`date +%Y%m%d%H%M%S` &&
```
nkcr:
```suggestion
          npm version prerelease --preid=pre &&
```

.travis.yml
```diff
@@ -1,104 +1,135 @@
...
+    - name: "NPM: js > cothority"
+      <<: *stage_deploy_npm
+      deploy:
+        provider: script
+        script: >-
+          cd external/js/kyber && npm ci && npm run link && cd ../../.. &&
+          cd external/js/cothority &&
+          npm ci &&
+          npm link @dedis/kyber &&
+          npm version prerelease --preid=p`date +%Y%m%d%H%M%S` &&
```
nkcr:
```suggestion
          npm version prerelease --preid=pre &&
```

tharvik:
> Got it worked with your suggested use of `deploy.on.branch`, we have a much nicer travis now, thank you

Happy to help :)

> I suggest we drop the date timestamp and only add a "pre" prefix for the build versions.

I don't think it'll work, as it will be publishing to the same version each time it runs; from the [npm-publish doc](https://docs.npmjs.com/cli/publish)
> Fails if the package name and version combination already exists in the specified registry.
>
> Once a package is published with a given name and version, that specific name and version combination can never be used again, even if it is removed with npm-unpublish.

nkcr:
> I don't think it'll work, as it will be publishing to the same version each time it runs; from the [npm-publish doc](https://docs.npmjs.com/cli/publish)

`npm version prerelease` automatically adds a version number at the end and increments it. So we would end up with something like:

v3.1.5-pre-1
v3.1.5-pre.0

In the meantime, I am wondering if there is way to generate such build only if the sources of the package has changed. We are constantly merging to master for things not concerned by the js packages and that might generate a lot of similar prerelease.

tharvik:
> > I don't think it'll work, as it will be publishing to the same version each time it runs; from the [npm-publish doc](https://docs.npmjs.com/cli/publish)
> 
> `npm version prerelease` automatically adds a version number at the end and increments it. So we would end up with something like:
> 
> v3.1.5-pre-1
> v3.1.5-pre.0

Correct, but it's only local (npm registry is not check for the latest version if I'm correct), so for a given version (`3.1.5`), the travis build will always generate the same prerelease version (`3.1.5-pre-0`) thus always pushing to the same version.

> In the meantime, I am wondering if there is way to generate such build only if the sources of the package have changed. We are constantly merging to master for things not concerned by the js packages and that might generate a lot of similar prereleases.

Yep, it usually pretty similar; one could add a check using `TRAVIS_COMMIT_RANGE` to see the list of changed files and if any is in `js/external/*`, publish it

nkcr:
> Correct, but it's only local (npm registry is not check for the latest version if I'm correct), so for a given version (`3.1.5`), the travis build will always generate the same prerelease version (`3.1.5-pre-0`) thus always pushing to the same version.

Sadly true ... : /   I then updated a bit the prerelease semantic to be more readable and I added the conditions.
I let you check if that works for you.

tharvik:
LGTM

PR #2045, opened 2019-08-30
Fix calypsoRead spawn

ineiti:
There is a test-case that is only used in the dynacred, where a calypsoRead is spawned
using a spawnerInstance and a coinInstance that has to pay for it. For this use-case,
CalypsoReadInstance.spawn didn't use ClientTransaction.make


nkcr:
Time to see if the automatic build works :D

PR #2049, opened 2019-09-05
travis: deploy without checking for changes

tharvik:
Currently, there is a hook to avoid pushing new NPM packages with the same content. But there isn't really a stable/safe way to do it. By discussing with @nkcr, we decided to drop this feature because it was already failing for some use cases (push force, new branch, ...).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,11 +33,8 @@ _: &stage_test_go
   script: GO111MODULE=on make test_goveralls
 _: &stage_deploy_npm
   <<: *language_js
-  script: 
-    - CHANGED_FILES=$(git diff --name-only $TRAVIS_COMMIT_RANGE)
-    - JS_COTHORITY_N=`echo "$CHANGED_FILES" | grep -E -c "^external/js/cothority/" || true`
-    - JS_KYBER_N=`echo "$CHANGED_FILES" | grep -E -c "^external/js/kyber/" || true`
   before_deploy: echo "//registry.npmjs.org/:_authToken=${DEPLOY_NPM_TOKEN}" > $HOME/.npmrc
+  script: skip  # default to `make test`
 
 
 dist: trusty
@@ -116,7 +113,6 @@ jobs:
       deploy:
         on:
           branch: master
-          condition: $JS_KYBER_N != 0
         provider: script
         script: >-
           cd external/js/kyber &&
@@ -128,7 +124,6 @@ jobs:
       deploy:
         on:
           branch: master
-          condition: $JS_COTHORITY_N != 0
         provider: script
         script: >-
           cd external/js/kyber && npm ci && npm run link && cd ../../.. &&


ineiti:
You could do the following:
```
FIRST_COMMIT_IN_PR=`git log --pretty="%H" -1`
if [ "$TRAVIS_COMMIT_RANGE" ]; then
  FIRST_COMMIT_IN_PR=${TRAVIS_COMMIT_RANGE/%..*/}
fi
PREVIOUS_COMMIT=`git log --pretty="%H" $FIRST_COMMIT_IN_PR -2 | tail -n 1`
if [ "$( git diff $PREVIOUS_COMMIT external/js/cothority )" ]; then
  ./publish
fi
```

Ugly, but should work...

tharvik:
I don't think that works.

>  `FIRST_COMMIT_IN_PR=${TRAVIS_COMMIT_RANGE/%..*/}`

You're assuming that the first commit is an existing one, which isn't the case when push forcing.

I understood `TRAVIS_COMMIT_RANGE ` as the set of commits last pushed when using a branch, or for a PR, the range of commit from the merging point to the tip of the branch.

ineiti:
The documentation reads:

> TRAVIS_COMMIT_RANGE: The range of commits that were included in the push or pull request. (Note that this is empty for builds triggered by the initial commit of a new branch.)

So if it is `from..to`, then both `from` and `to` should be in the push/PR, no? The problem is that the `git diff from..to` doesn't include the `from^1..from`, and we're also interested in that to check if something changed.

tharvik:
> The documentation reads:
> 
> > TRAVIS_COMMIT_RANGE: The range of commits that were included in the **push** or pull request. (Note that this is empty for builds triggered by the initial commit of a new branch.)
> 
> So if it is `from..to`, then both `from` and `to` should be in the push/PR, no?

Yes and no, it depends how you pushed the commits (or understand "push"). I've made a small transcript of the simple process of creating a new branch (a commit away from master) and push, creating a new commit and push, and amending the second commit.

```
1er push
        git
                aa3e417..ea84749
        travis
                empty
2nd push
        git
                ea84749..d37d211
        travis
                ea84749..d37d211
2nd push force
        git
                d37d211..be66de0
        travis
                d37d211..be66de0
```

In this case, there is some issues, first, a new commit on a new branch gives us no usable information, second, in the "2nd push force", the commit range contains an obsolete first commit. Also, the commit range is not really helpful, as not for the whole range (ie, the branch we want to merge into another).

To get a better idea of why I think this feature it's not possible with travis (and is impossible to fix for branches in any general CI/CD without explicitling merge point), we first have a small lemma: you can't assume that someone would want to merge to master, as it's just of some tip of path in your graph. Now, you find yourself in a git with HEAD to the tip of branch you want to merge: what do you do? Nothing as you don't know which of the several branch, encountered during traversal to the root commit, you want to merge into.

> The problem is that the `git diff from..to` doesn't include the `from^1..from`, and we're also interested in that to check if something changed.

If you want to find the previous commit, you can sed it: `sed s,\.,^1.,`

PR #2050, opened 2019-09-10
RoPaSci calypso contract

ineiti:
Implements an extension to the Rock-Paper-Scissors contract by allowing the player 1
to store his move (the pre-hash) encrypted in a CalypsoWrite structure. Player 2 will
create a CalypsoRead instance that can be used to recover the pre-hash, even if
Player 1 is absent.

If Player 2 is absent, Player 1 can still confirm the game and collect the wins.

There is a documentation at the beginning of the personhood/contract_ropasci.go file.


jeffallen:
I am not very clear where the LTS comes from in production, but the test code makes sense to me.

PR #2057, opened 2019-09-21
Adding connectivity command to status

### LINKED ISSUE #2053: Connection-test service
Add a service with a CLI that allows to send a `ping` to all other nodes given in a roster:
- the CLI should use the `private.toml` of the conode to sign the start of the request
- to avoid abuse, a rate limitation of 1/sec should be given
The connection-test can be done using https://github.com/dedis/cothority/blob/master/messaging/broadcast.go:
1. the requester uses the CLI tool to send a ping-request over the websocket
2. the service of the requester node sends out a broadcast-protocol `ContactNodes` using `CreateProtocol(name, tree, serviceID)`
3. all receiver nodes (excluding the requester node) catch this protocol-message in their `NewProtocol` and send out another broadcast-protocol `ContactNodes`, but this time using `CreateProtocol(name, tree, NilServiceID)`
4. upon reception of the reply of 3., the nodes reply back to 2.
5. the service replies back whether all could talk to all other nodes

If it's possible to have some stats, e.g., how long it took to contact whom, and which connections failed, that would be awesome.

ineiti:
**What this PR does**
In order to verify that all nodes can reach each other, the status
service gets an additional command connectivity that tests an
all nodes to all nodes connectivity.

The test can also try multiple list of nodes until all nodes can reach
each other.

To prevent abuse, the request must be signed by the node that
launches the request.

Closes #2053 

---

🙅‍ Friendly checklist:

- [x] 5. Hard limit of 80 chars per line is always respected in the comments, 100 chars per line for code.


status/service/api.go
```diff
@@ -25,3 +31,49 @@ func (c *Client) Request(dst *network.ServerIdentity) (*Response, error) {
...
+	}
+	resp := &ConnectivityReply{}
+	err = c.SendProtobuf(list[0], conn, resp)
+	if err != nil {
+		return nil, err
+	}
+	return resp.Nodes, nil
+}
+
+func (c *Connectivity) hash() ([]byte, error) {
```
Gilthoniel: This does not include the `Time` value so one could change the value and send the message around. Same for `FindFaulty` but it's less of a big deal.

external/proto/status.proto
```diff
@@ -16,3 +16,21 @@ message Response {
    map<string, onet.Status> status = 1;
   optional network.ServerIdentity serveridentity = 2;
 }
+
+// Connectivity is sent by a client to check the connectivity of a given roster. The Time must
+// be within 2 minutes of the server's time. The signature must be a schnorr-signature using the
+// private conode-key on the following message:
+//   sha256( bytes.LittleEndian.PutUInt64(Time) | protobuf.Encode(List[0]) |
+//   protobuf.Encode(List[1])... )
+message Connectivity {
```
Gilthoniel: Can we name this `ConnectivityRequest` ? So we can see in the code that's a message right away.

status/service/status_test.go
```diff
@@ -19,7 +21,36 @@ func NewTestClient(l *onet.LocalTest) *Client {
...
+	for i := range ro.List {
+		require.True(t, ro.List[i].Equal(repl[i]))
+	}
+	require.NoError(t, local.WaitDone(time.Second))
+
+	servers[2].Pause()
+	repl, err = cl.Connectivity(priv, ro.List, time.Second, false)
+	require.Error(t, err)
+
+	repl, err = cl.Connectivity(priv, ro.List, time.Second, true)
```
Gilthoniel: Can you check you get the right set ?

status/service/api.go
ineiti: Thanks, missed that. Updated to hash:

```
Time | Timeout | FindFaulty ? byte(1) : byte(0) | List
```

external/proto/status.proto
ineiti: Most of the other API endpoints (excluded the status-request itself) have the following naming convention:

- service-method `Name`
- api-method `Client.Name`
- request-structure `Name`
- reply-structure `NameReply` or `NameResponse`

That should probably be part of some _Cleanup-task_ for v4... So whatever you chose, all the other APIs should follow. The above naming-scheme is, as far as I know, used in the majority of API calls over the websockets.

ineiti:
One important thing: we discussed this on Friday, and I went ahead and did this proof-of-concept work. So if you think that this should be done in another way, please speak up.

To put it in the comments here: the goal is to make sure that the conodes can communicate with each other and are not blocked.

One proposition was to have the status-service do this automatically and log which connections are OK and which are bad. But I think this is difficult as you need to decide:
- when to update the good/bad queue (I love caching, but...)
- which roster to take (all known connections? all rosters from all byzcoins? past/present blocks?)
So I went ahead and did this simple "test this set of nodes" PR.

The only downside is that now anybody can set up a node and then ask the other nodes to check the connections, thus DOSsing the system. But it's even easier to just open websockets to the nodes to bring 'em down. So I don't see why somebody should go through the pain of setting up a node to DOS the system...

external/proto/status.proto
```diff
@@ -16,3 +16,22 @@ message Response {
    map<string, onet.Status> status = 1;
   optional network.ServerIdentity serveridentity = 2;
 }
+
+// Connectivity is sent by a client to check the connectivity of a given roster. The Time must
```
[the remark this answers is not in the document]
ineiti: I tried setting the wrap at 80 characters - but it's _really_ narrow. The line

```
func (st *Stat) CheckConnectivity(req *CheckConnectivity) (*CheckConnectivityReply, error) {
```

would be needed to split like this:
```
func (st *Stat) CheckConnectivity(req *CheckConnectivity)(
	*CheckConnectivityReply, error) {
```

Wrt @Gilthoniel comment in #2058 - can we go with 120 characters? 100?

ineiti:
> * some errors not wrapped

Question/discussion about error-wrapping:

https://github.com/dedis/cothority/issues/2061

external/proto/status.proto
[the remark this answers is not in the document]
jeffallen: I would just like to register my vote here as "abstain". I've no opinion about formatting other than "use gofmt".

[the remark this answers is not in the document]
nkcr: > code at 100

.. when a wrapping at 80 is not doable

Gilthoniel:
> One important thing...

I don't have any problem with that implementation. I was hoping to have something that would allow us to include the connectivity status in the skipchain explorer so we don't manually need to check but as you said, this might be complicated so for the small benefit, let's do it your way.

status/status.go
```diff
@@ -46,6 +47,29 @@ func main() {
 			Value: 0,
 			Usage: "debug-level: `integer`: 1 for terse, 5 for maximal",
 		},
+		cli.BoolFlag{
+			Name: "connectivity, c",
+		},
+	}
+	app.Commands = cli.Commands{
+		{
+			Name:      "connectivity",
```
nkcr: This reminds me https://youtu.be/eMz0vni6PAw, where it encourages making your command "read like sentence (9:53)", and "avoid positional arguments where the order matters (10:35)".
For me it would make more sense having `status checkConnectivity --key KEY --roster ROSTER`. Flag arguments also forces you to describe what they are (not sure what is `private.toml` actually..).

nkcr: This was just a thought for future implementation of CLIs (starts to be a lot of them in cothority).

status/service/status.go
```diff
@@ -46,14 +51,94 @@ func (st *Stat) Request(req *Request) (network.Message, error) {
...
+	// Check signature
+	hash, err := req.hash()
+	if err != nil {
+		return nil, errors.New("couldn't hash message: " + err.Error())
+	}
+	err = schnorr.Verify(cothority.Suite, st.ServerIdentity().Public, hash, req.Signature)
+	if err != nil {
+		return nil, errors.New("signature verification failed: " + err.Error())
+	}
+	if math.Abs(time.Now().Sub(time.Unix(req.Time, 0)).Seconds()) > 120 {
```
nkcr: where does this magic value of 120 comes from?

status/service/status.go
```diff
@@ -46,14 +51,94 @@ func (st *Stat) Request(req *Request) (network.Message, error) {
...
+}
+
+func (st *Stat) testNodes(nodes []*network.ServerIdentity, to time.Duration) error {
+	r := onet.NewRoster(nodes)
+	tree := r.GenerateBinaryTree()
+	p, err := st.CreateProtocol(messaging.BroadcastName, tree)
+	if err != nil {
+		return errors.New("protocol creation failed: " + err.Error())
+	}
+	bc := p.(*messaging.Broadcast)
```
nkcr: hope this will never panic, isn't it better checking if the assertion is ok  and return a nice error if not?

status/status.go
ineiti: Hmm - I'll follow along whatever you decide, but not before I gave my own opinion ;)

For me, flags are optional. Things that are needed for the command should be part of the arguments. So I always cringe if I have to type

```
bcadmin create --roster public.toml
```

Because the `create` command needs the roster in all cases. Where a flag like `--interval .5s` makes sense to me, as this is not absolutely needed.

status/service/status.go
ineiti: Yes, generally it's good to test if assertions are OK. But here it's just 4 lines after the `CreateProtocol`, which should definitively give back a `messaging.Broadcast` structure. Else a panic is a good idea anyway, because something is really broken and we shouldn't continue.

status/status.go
nkcr: > Things that are needed for the command should be part of the arguments

I totally agree on this clever principle, which is nice and actually work for commands that have one needed argument, or two if there is a logic (like `cp SOURCE DESTINATION`). 

Otherwise I won't remember the order: is it the roster then the key? or is it the key then the roster? If you don't know it by hearth then you must use `-h`. In my case I don't mind, I know I must provide a key and a roster so, as long as I provide `--roster` and `--key`, my cli is happy. But I am not 100% convinced by myself, so don't change it.

It is actually not easy to find an example of a cli with multiple positional arguments, apart from `cp, mv, ...`.

ineiti: Well you have the firewall `ufw` which takes sometimes more than 2 arguments. But, yes, I have to do `ufw -h` every time I want to use those. But I'm not sure I wouldn't need to if there were flags.

What I try to do in general is to start with the most global, long-term arguments first, then the more specific ones. So for a `roster`, a `config`, and a `key`, this would be the order, as the `roster` is the most global one, then comes my `config`, and finally there is a `key`.

But in general having the CLIs being something like spoken language is nice. Then you just need to avoid the French-based sentence order, else all is off again ;)

PR #2065, opened 2019-09-26
Make tests working without talking exclusively to the leader

### LINKED ISSUE #2002: Tests only work with leader
While working on #2001, some tests showed up that only work if the leader is contacted. This seems bad in most cases, and should be extended to also work for non-leaders.

As far as I can tell, often something like this happens:
1. `AddTransaction` is sent to a node, it returns as soon as this node processed the transaction
2. `GetProof` is sent to two nodes, the node who doesn't have the transaction is faster and returns an empty proof

All tests that use `byzcoin.Client.UseNode` are suspicious:

- [x] `byzcoin.TestClient_GetProof`
- [x] `byzcoin.TestClient_GetProofCorrupted`
- [x] `byzcoin.TestSecureDarc`
- [x] `byzcoin.TestService_SetConfigRosterReplace`
- [x] `byzcoin.TestService_SetConfigRosterReplace`
- [x] `byzcoin.Client.DownloadState`
- [x] `byzcoin.contracts.TestDeferred_ScenarioMultiInstructions_`
- [x] `byzcoin.contracts.TestDeferred_ScenarioMultipleSigners`
- [x] `byzcoin.contracts.TestDeferred_ScenarioUpdateConfig`
- [x] `byzcoin.contracts.TestDeferred_SimpleDelete`
- [x] `byzcoin.contracts.TestDeferred_PublicDelete`
- [x] `byzcoin.contracts.TestInsecureDarc`
- [x] `calypso.TestService_ReshareLTS_Different`
- [x] `calypso.TestClient_CreateLTS`
- [x] `calypso.TestClient_Calypso`
- [x] `calypso.TestService_ReshareLTS_OneMore`
- [x] `eventlog.Client`
- [x] `personhood.Tests`

The following is good:
- `byzcoin.simulation.SimulationService.Run`, because all transactions need to be sent to the same node, else the counters are not correct
- `TestContract_Write_Benchmark` - same reason: multiple transactions with same signer

Gilthoniel:
**What this PR does**

It makes the tests working without having to contact the leader exclusively.

Fixes #2002 

---

🙅‍ Friendly checklist:


byzcoin/api.go
```diff
@@ -199,7 +199,43 @@ func (c *Client) GetProofFromLatest(key []byte) (*GetProofResponse, error) {
 		return c.GetProof(key)
 	}
 
-	return c.GetProofFrom(key, c.Latest)
+	return c.GetProofFrom(key, c.Latest, protobuf.Decode)
+}
+
+// GetProofAfter returns a proof for the key stored in the skipchain
+// starting from the latest known block by this client. The proof will always
+// be older than the barrier or it will return an error.
```
ineiti:
```suggestion
// be newer than the barrier or it will return an error.
```

byzcoin/api.go
```diff
@@ -209,13 +245,13 @@ func (c *Client) GetProofFromLatest(key []byte) (*GetProofResponse, error) {
 // state of the chain up to the block. If you need to pass the Proof onwards to
 // another server, you must use GetProof in order to create a complete standalone
 // proof starting from the genesis block.
-func (c *Client) GetProofFrom(key []byte, from *skipchain.SkipBlock) (*GetProofResponse, error) {
+func (c *Client) GetProofFrom(key []byte, from *skipchain.SkipBlock, decoder onet.Decoder) (*GetProofResponse, error) {
```
ineiti: Isn't this a backward-incompatible change in an API that is in use?

byzcoin/api.go
```diff
@@ -232,7 +268,14 @@ func (c *Client) GetProofFrom(key []byte, from *skipchain.SkipBlock) (*GetProofR
 // GetDeferredData makes a request to retrieve the deferred instruction data
 // and return the reply if the proof can be verified.
 func (c *Client) GetDeferredData(instrID InstanceID) (*DeferredData, error) {
-	pr, err := c.GetProofFromLatest(instrID.Slice())
+	return c.GetDeferredDataAfter(instrID, time.Unix(0, 0))
+}
+
+// GetDeferredDataAfter makes a request to retrieve the deferred instruction data
+// and return the reply if the proof can be verified and the block is not
```
ineiti:
```suggestion
// and returns the reply if the proof can be verified and the block is not
```

byzcoin/api.go
ineiti: I'm not a big fan of putting all these checks in the receiving method to decide if the received proof is acceptable or not.

We should better define what `SendProtobuf` actually guarantees - triaging the results from different nodes with the `decoder` method looks very flaky to me - sometimes you might get a good reply, sometimes you might not.

E.g., after the call to `getFinalizedParty` you're not really sure if the error is due to the call being too fast, or if the error is due to the fact that the finalized party is not there.

The culprit, IMHO, lies in the `AddTransactionAndWait`. Because, what happens, and what is taken care of not in an ideal way, IMO, in the `getFinalizedParty` case, is that:

1. Transaction is added to a node
2. the node receives a block containing the transaction and returns to the caller
3. the caller asks another node for a proof
4. other nodes still catch up with the latest block

So you get an error in 3., because you ask the proof of a node that did not store the transaction yet (which is normal in a distributed setting).

I think that the `AddTransaction` api needs some flags:
- `WaitForAllNodes` - which makes sure that the block has been processed by all nodes - probably difficult to implement and also flaky
- `ReturnProof` - returns the proof of the instruction - needs to take care which proof should actually be returned
- `ReturnStateChange` - return the statechanges done by the transaction

Gilthoniel: That would be nice to get the proof from the `AddTransactionAndWait` call but you still scenarios where it wouldn't be enough. Like calling a CLI multiple times where we don't share states between the calls so you would still need something. I first thought about using a request to specifically request a proof of existence but it can exist and still be out-of-date.. so that's the best solution I found.

ineiti: CLI: yes, if one command sends the transaction and a next command gets the proof, you're in trouble. But with your solution you'd need to pass the timestamp from the 1st to the 2nd command, so that the second command can correctly call `GetProofAfter`.

Tests/js-frontend/mobile app: here the problem is mostly the `AddTransaction` / `GetProof` combo. In the latest mobile app, I solve the problem by sending the `AddTransaction` and `GetProof` to the same node. So I'm sure that if the `AddTransaction` is done, the `GetProof` is correct, too.

What happens probably most of the time is that `GetProofAfter` accepts the proof from the node that got the `AddTransaction` in the first place (or another node closer to the leader). So it should not be flakey in fact.

ineiti: In fact, `SendProtobufParallel` returns the `network.ServerIdentity` that replied first. So you could use that to send the `GetProof` to, and you'd be sure that the proof includes the latest `ClientTransaction`.

Gilthoniel: It's Friday so let's be crazy:
1. `AddTransaction` request to Bob's node works, everything looks good so far
2. Bob is having a bad day so decides to take his node down
3. `GetProof` to Bob's node fails as it's offline

One of the reason  of `SendProtobufParallel` is to prevent that kind of situation. But yes, you could optimize by asking Bob's node first as you would expect a correct reply from it.

ineiti: OK, looking at `GetProof`, I see that there is `ID`, which can be any block know to the client. So:
1. `AddTransaction` has to return the block where the transaction is actually stored (optional field to be backwards-compatible)
2. `GetProof` gives this block - nodes that are not yet up-to-date will return an error, nodes that do have this block will return the correct proof

ineiti: This should survive even a Friday-crazyness ;)

Gilthoniel: Hm, if you ask for a proof from block `ID`, it will return a proof starting from `ID` which is not always verifiable.

ineiti: True - so we're down to adding a `ReturnProofFrom` flag to the `AddTransaction` message, so that you can get the proof directly if your transaction has been valid. The value of this flag would be used in the same way as the `ID` in `GetProof`.

A more complicated setup would be to add a `ThisBlockNeedsToBeInTheChain` field in the `GetProof` message, but that looks overly complicated and still misses the point if the node did not catch up yet.

Gilthoniel: I was investigating on the last one. We could get the ID from the transaction, and send it to the GetProof so the server is responsible to send back an error if it doesn't know the block. But it can be problematic if a malicious node doesn't stick to the behavior so the client needs to check that the proof indeed contains the block.

Overall that sounds feasible.

ineiti: I think it's OK to say that `AddTransaction` needs to return the proof in such a way that the block containing the transaction needs to be in `Proof.Latest`. Then the client can look at this block and verify whether his transaction is actually in this latest block. And if the rest of the proof checks out, the client can be sure that his transaction has been included in the chain.

PR #2068, opened 2019-09-27
Spawn value

ineiti:
personhood/spawner-contract:
- Adds the possibility to spawn a value instance
- Checks that a spawned darc doesn't have other rules than `_sign` and `invoke:darc.evolve`

external/js/cothority:
- adds the value instance
- adds the value instance to the spawner instance

---

🙅‍ Friendly checklist:


personhood/contract_spawner.go
```diff
@@ -80,25 +82,46 @@ func (c *ContractSpawner) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Inst
...
 		if err = c.getCoins(cout, c.CostDarc); err != nil {
 			return
 		}
 		instBuf = inst.Spawn.Args.Search("darc")
 		d, err := darc.NewFromProtobuf(instBuf)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, errors.New("couldn't decode darc: " + err.Error())
+		}
+		// Whitelist allowed darc rules - there is no spawn allowed, and all
```
jeffallen: I really don't understand this. Why is this needed now? Was it wrong before? Is this something about the value contract?

ineiti: The spawner-contract allows any user with coins to create new instances. In order not to have a user create a darc with rules like `spawn:darc` or `invoke:coin.mint`, the allowed rules are restricted.

Chosing between a blacklist and whitelist approach I chose the latter, to avoid that somebody adds an `invoke` to any contract and then opens a security hole in the spawner-contract.

If you prefer, I can create a separate PR for that 'fix' of the spawner-contract.

PR #2071, opened 2019-09-27
Updated the roadmap.

jeffallen:
Updated the roadmap to give clear end of life dates.


ineiti:
as soon as it passes travis, this should be ok ;)

Gilthoniel:
One thing that could be added actually is the Go version support, or is it somewhere already ?

Issue #2069, opened 2019-09-27
Travis takes too long


ineiti:
The test for Travis to go over the cothority-repo is taking anywhere from 30' to 1h, depending on the load of Travis.

Some ideas to speed it up:
- reduce the number of tests
  `+` easy to do
  `-` test-coverage decreases
- replace travis with circleCI and run it on an EPFL machine with more oomph
  `+` keep all tests
  `-` need to rewrite the .travis.yml
  `-` need to setup circleCI
- replace travis with Jenkins, as Jenkins is already setup - https://resources.github.com/whitepapers/practical-guide-to-CI-with-Jenkins-and-GitHub/
  `+` keep all tests
  `-` rewrite .travis.yml
- rewrite tests to be faster and more unit-test based
  `+` keep travis
  `-` lots of work

SuddenGunter:
hi
also recommend to look at drone CI, as it's easy to setup + you may cache builds on some external storage for faster execution

tharvik:
5. retry for/rewrite tests to be parallel; currently goveralls does each dir individually, and sequentially in each
`+` harder, better, faster, stronger
`-` might be a lot of work

nkcr:
With travis each job in a stage is run in parallel, so we don't gain much by removing go 1.12.x

ineiti:
@nkcr travis only runs a limited number of jobs in parallel, to avoid overloading of the machines. There are two ways to get around this:
- pay more money to get more parallel builds
- install your own CI infrastructure, e.g. in Kubernetes, then you pay as you go...

PR #2072, opened 2019-09-30
Move to Go 1.13

jeffallen:
Change to using Go 1.13, improve documentation about our policy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
-_: &language_go_1_11
-  name: "go 1.11"
-  language: go
-  go: "1.11.x"
 _: &language_go_1_12
   name: "go 1.12"
   language: go
   go: "1.12.x"
+_: &language_go_1_13
+  name: "go 1.13"
+  language: go
+  go: "1.13.x"
 _: &language_js
   language: node_js
   node_js: "lts/*"
@@ -26,7 +26,6 @@ _: &stage_lint_go
 _: &stage_build_go
   script:
     - make -C conode bindist tooldist
-    - GO111MODULE=on go get ./...
     - GO111MODULE=on go build ./...
 _: &stage_test_go
   before_script: *get_coding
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,11 +1,11 @@
...
-  language: go
-  go: "1.11.x"
 _: &language_go_1_12
   name: "go 1.12"
   language: go
   go: "1.12.x"
+_: &language_go_1_13
+  name: "go 1.12"
+  language: go
+  go: "1.12.x"
```
nkcr: go 1.**13**.x

.travis.yml
```diff
@@ -80,9 +79,9 @@ jobs:
 
     - stage: test
       <<: *stage_test_go
-      <<: *language_go_1_11
-    - <<: *stage_test_go
       <<: *language_go_1_12
```
ineiti: Would this be a good moment to remove the 1_12 testing here, as proposed in #2069 ?

.travis.yml
jeffallen: oh god, what an idiot.

.travis.yml
```diff
@@ -79,10 +72,9 @@ jobs:
         - npm run build
 
     - stage: test
-      <<: *stage_test_go
```
ineiti: Travis tries to run some ruby - I wonder if this line should go back in? But I got to admit that @tharvik script-fooed the hell out of .travis.yml and that I never know where a `-` is needed and where not.

tharvik: `<<:` is the merge operator, taking its definition and merging it with the current dict.
```yaml
- stage: test
<<: *stage_test_go
---
- stage: test
  before_script: *get_coding
  script: GO111MODULE=on make test_goveralls
```
```yaml
- stage: test
- <<: *stage_test_go
---
- stage: test
- before_script: *get_coding
  script: GO111MODULE=on make test_goveralls
```

.travis.yml
jeffallen: done.

.travis.yml
```diff
@@ -79,10 +75,8 @@ jobs:
         - npm run build
 
     - stage: test
+      <<: *language_go_1_13
```
tharvik: I find it nicer if is `stage` first then `language` (to stay coherent with the others)

jeffallen: But then language gets set to Ruby. Travis is very fragile and mysterious when using the merge operator.

PR #2073, opened 2019-09-30
Allow darcs with spawn:calypsoRead

ineiti:
**What this PR does**

When using the spawner-instance, darcs with spawn:calypsoRead should
be allowed, as they can only be used as instructions to a calypsoWrite
instance that will check if coins are needed or not.

---

🙅‍ Friendly checklist:

### DIFF
--- a/personhood/contract_spawner.go
+++ b/personhood/contract_spawner.go
@@ -101,10 +101,18 @@ func (c *ContractSpawner) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Inst
 		if err != nil {
 			return nil, nil, errors.New("couldn't decode darc: " + err.Error())
 		}
-		// Whitelist allowed darc rules - there is no spawn allowed, and all
+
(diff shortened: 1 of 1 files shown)


personhood/contract_spawner.go
```diff
@@ -111,8 +111,9 @@ func (c *ContractSpawner) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Inst
 			"coin\\.(fetch|store|transfer)|" +
 			"credential\\.(update|recover)|" +
 			"popParty\\.(barrier|finalize|mine|addParty)|" +
-			"ropasci\\.(second|confirm)" +
-			"value\\.update))$")
+			"ropasci\\.(second|confirm)|" +
```
nkcr: > When using the spawner-instance, darcs with spawn:calypsoRead should
be allowed, as they can only be used as instructions to a calypsoWrite
instance that will check if coins are needed or not.

This explanation should be added in the comment above.

PR #2076, opened 2019-10-02
adds bcadmin debug block

ineiti:
**What this PR does**

This command fetches a block from one or more nodes and displays information
about the header and the body. It can be used to display parts of the chain
as seen from different nodes.

---

🙅‍ Friendly checklist:


jeffallen:
In the process of reviewing this, I made some changes to it to check my assumptions, and so I'm going to add a commit here with my minor changes, and then merge it.

PR #2077, opened 2019-10-02
Proposal for error-handling

ineiti:
Is this OK like that, or should we keep the "shouldn't|couldn't|wouldn't|hasn't"-format?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,7 +11,7 @@ This PR
 - [ ] 1. Design choices, user documentation and any additional doc are added (or updated) in READMEs.
 - [ ] 2. Any new behaviour is tested and small units of code that can be are unit tested.
 - [ ] 3. Code comments are added on tests to explain what they do.
-- [ ] 4. Errors are systematically wrapped, following the `[failed to | couldn't do] ACTION THAT FAILED: " + err.Error()` format.
+- [ ] 4. Errors are systematically wrapped using `xerrors`.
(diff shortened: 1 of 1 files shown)


nkcr:
> Is this OK like that, or should we keep the "shouldn't|couldn't|wouldn't|hasn't"-format?

IMHO errors should never be thrown without a message: it helps reading the errors (obviously) and the source code. So we could add `Errors are systematically wrapped using xerrors AND a meaningful message`.

tharvik:
Might want to also update `.github/PULL_REQUEST_TEMPLATE.md`

ineiti:
> Might want to also update `.github/PULL_REQUEST_TEMPLATE.md`

That's the change I'm proposing ;)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,7 +11,7 @@ This PR
 - [ ] 1. Design choices, user documentation and any additional doc are added (or updated) in READMEs.
 - [ ] 2. Any new behaviour is tested and small units of code that can be are unit tested.
 - [ ] 3. Code comments are added on tests to explain what they do.
-- [ ] 4. Errors are systematically wrapped, following the `[failed to | couldn't do] ACTION THAT FAILED: " + err.Error()` format.
+- [ ] 4. Errors are systematically wrapped using `xerrors` AND a meaningful message is added.
```
nkcr:
```suggestion
- [ ] 4. Errors are systematically wrapped with a meaningful message using `xerrors.Errorf` and the `%+v` verb.
```

PR #2078, opened 2019-10-02
external/cothority: use wss when got via https

tharvik:
**Simplify `Address || Url -> websocket URL` translation.**

Currently, when using a Roster with servers without `Url`, when using `ServerIdentity.getWebSocketAddress`, the system default to building one from `Address`.
The default protocol used is `ws:` which, when providing js/cothority via `https:`, is killed with some browser (see https://github.com/c4dt/omniledger/issues/143#issuecomment-536920787). But the Roster didn't always provide the `Url` field, so, when following the chain, it happens that some are contacted via unsecure WebSocket.

This PR fixes that by defaulting to `wss:` when missing `Url`, and some other small changes.
* rm unused `static ServerIdentity.{urlToWebsocket,addressToWebsocket}`
* enforce more structure on `Address` and `Url`: they should both be valid URL when calling `getWebSocketAddress` (not checked within constructor)
* I dropped the specific handling of the trailing slash
  * previsouly, it removed the last slash if existing and I don't understand why this behavior

---

🙅‍ Friendly checklist:

### DIFF
--- a/external/js/cothority/src/network/connection.ts
+++ b/external/js/cothority/src/network/connection.ts
@@ -1,5 +1,6 @@
 import { Message, util } from "protobufjs/light";
 import shuffle from "shuffle-array";
+import URL from "url-parse";
 import Log from "../log";
 import { Roster } from "./proto";
 import { BrowserWebSocketAdapter, WebSocketAdapter } from "./websocket-adapter";
@@ -53,7 +54,12 @@ export class WebSocketConnection implements IConnection {
      * @param service   Name of the service to reach
      */
     constructor(addr: string, service: string) {
-        this.url = addr;
+        const url = new URL(addr, {});
+        if (typeof window !== "undefined" && window.isSecureContext === true) {
+            url.set("protocol", "wss");
+        }
+        this.url = url.href;
+
         this.service = service;
         this.timeout = 30 * 1000; // 30s by default
     }
--- a/external/js/cothority/package-lock.json
+++ b/external/js/cothority/package-lock.json
@@ -1115,6 +1115,11 @@
       "resolved": "https://registry.npmjs.org/@types/sprintf-js/-/sprintf-js-1.1.1.tgz",
       "integrity": "sha512-h8jF5yPUoHH1QUIDfHgqFs7Y0UykKB5CJ1Z32PHGmHRLmW0pI+kYKAEnXVqeUZ3ygFmDkyvhD4vUZN+RZyTd1w=="
     },
+    "@types/url-parse": {
+      "version": "1.4.3",
+      "resolved": "https://registry.npmjs.org/@types/url-parse/-/url-parse-1.4.3.tgz",
+      "integrity": "sha512-4kHAkbV/OfW2kb5BLVUuUMoumB3CP8rHqlw48aHvFy5tf9ER0AfOonBlX29l/DD68G70DmyhRlSYfQPSYpC5Vw=="
+    },
     "@types/ws": {
       "version": "6.0.1",
       "resolved": "https://registry.npmjs.org/@types/ws/-/ws-6.0.1.tgz",
@@ -3067,8 +3072,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -3089,14 +3093,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
@@ -3111,20 +3113,17 @@
         "code-point-at": {
           "version": "1.1.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "concat-map": {
           "version": "0.0.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "console-control-strings": {
           "version": "1.1.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "core-util-is": {
           "version": "1.0.2",
@@ -3241,8 +3240,7 @@
         "inherits": {
           "version": "2.0.3",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "ini": {
           "version": "1.3.5",
@@ -3254,7 +3252,6 @@
           "version": "1.0.0",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "number-is-nan": "^1.0.0"
           }
@@ -3269,7 +3266,6 @@
           "version": "3.0.4",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "brace-expansion": "^1.1.7"
           }
@@ -3277,14 +3273,12 @@
         "minimist": {
           "version": "0.0.8",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "minipass": {
           "version": "2.3.5",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "safe-buffer": "^5.1.2",
             "yallist": "^3.0.0"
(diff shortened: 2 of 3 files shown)


external/js/cothority/src/network/proto.ts
```diff
@@ -251,15 +218,28 @@ export class ServerIdentity extends Message<ServerIdentity> {
...
     /**
-     * Returns websocket version of this.url if set, otherwise converts the server
-     * address to match the websocket format.
+     * Returns websocket URL to connect to this ServerIdentity.
      * @returns the websocket address
      */
     getWebSocketAddress(): string {
-        if (this.url) {
-            return ServerIdentity.urlToWebsocket(this.url);
+        if (this.url !== undefined && this.url !== "") {
```
Gilthoniel:
```javascript
if (this.url) {
```

external/js/cothority/src/network/proto.ts
```diff
@@ -251,15 +218,28 @@ export class ServerIdentity extends Message<ServerIdentity> {
...
+            const url = new URL(this.url, {});
+            // TODO why not supporting proto ws: & wss: ?
```
Gilthoniel: I guess the todo can be removed

external/js/cothority/src/network/proto.ts
```diff
@@ -251,15 +218,28 @@ export class ServerIdentity extends Message<ServerIdentity> {
...
-     * Returns websocket version of this.url if set, otherwise converts the server
-     * address to match the websocket format.
+     * Returns websocket URL to connect to this ServerIdentity.
      * @returns the websocket address
      */
     getWebSocketAddress(): string {
-        if (this.url) {
-            return ServerIdentity.urlToWebsocket(this.url);
+        if (this.url !== undefined && this.url !== "") {
+            const url = new URL(this.url, {});
```
Gilthoniel:
```javascript
new URL(this.url)
```
the 2nd is optional

external/js/cothority/src/network/proto.ts
```diff
@@ -251,15 +218,28 @@ export class ServerIdentity extends Message<ServerIdentity> {
...
+            // TODO why not supporting proto ws: & wss: ?
+            switch (url.protocol) {
+                case "http:": url.set("protocol", "ws:"); break;
+                case "https:": url.set("protocol", "wss:"); break;
+                default: throw new Error(`unknown url protocol: ${url.protocol}`);
+            }
+            return url.href;
         } else {
-            return ServerIdentity.addressToWebsocket(this.address);
+            const addr = new URL(this.address, {});
```
Gilthoniel: same

external/js/cothority/src/network/proto.ts
```diff
@@ -251,15 +218,28 @@ export class ServerIdentity extends Message<ServerIdentity> {
...
+            const port = parseInt(addr.port, 10);
+            if (isNaN(port)) {
```
Gilthoniel:
```javascript
Number.isNaN(port)
```
The global tries to cast the value to a number but it will always be a number/NaN from parseInt function

Gilthoniel:
I'm not totally convinced by that PR for backwards compatibility reasons. The default behavior is to return a non-secure address when URL is not set. It would be interesting to deprecate this one and introduce a new one that returns a URL, and return a secure address by default if URL is not set (or if address is defined as tls::// and non-secure otherwise).

external/js/cothority/src/network/proto.ts
tharvik: Why so? It's less precise and wrong if `this.url == null` (as `tsconfig.strictNullChecks` is not set).

external/js/cothority/src/network/proto.ts
tharvik: If `this.url` is relative (it might as there isn't checks for it), it will use `windows.location` to resolve it. Specifing an empty second argument forces the URL to be absolute.

external/js/cothority/src/network/proto.ts
tharvik: Oups, I should have put it in the description.
Still, wouldn't an URL with proto `ws` be more correct than `http`? The first one is clearly the protocol used.

tharvik:
> I'm not totally convinced by that PR for backwards compatibility reasons. The default behavior is to return a non-secure address when URL is not set.

There is clearly a backward compatibility issue. IMO, now that all the servers reachable from the chain of Roster are available via `wss`, we can pretty safely default to a secure connection.

> It would be interesting to deprecate this one and introduce a new one that returns a URL, and return a secure address by default if URL is not set (or if address is defined as tls::// and non-secure otherwise).

Sure, I can do that, I've got some questions then
  * deprecation: how to do it? a `Log.warning` with a deprecated message?
  * got a naming preference? `urlClient`?
    * while adding a new name, I can move validation into the constructor and having it as a field directly, to trigger potential errors earlier

external/js/cothority/src/network/proto.ts
Gilthoniel: because if `this.url` is equal to null, the if clause will be true

tharvik: Yup, because it shouldn't be `null` and thus indicating a bug. (it's also because denoting missing value by empty string wrong IMO, so expliciting the fact that we support empty string makes it more readable)

Gilthoniel: It's not a bug to have either `undefined`, `null` or an empty string. The type allows it so it needs to be handled.

Gilthoniel: It's using raw types as the class is a message coming from the network.

external/js/cothority/src/network/proto.ts
Gilthoniel: Historically it has been `http` so let's keep it that way.

Gilthoniel:
You can use `@deprecated` in the comment, it should be globally supported soon.

For the name `getWebsocketURL()`. You can't change the validation of the constructor again for backward compatibility concerns.

tharvik:
> I'm not totally convinced by that PR for backwards compatibility reasons. The default behavior is to return a non-secure address when URL is not set. It would be interesting to deprecate this one and introduce a new one that returns a URL, and return a secure address by default if URL is not set (or if address is defined as tls::// and non-secure otherwise).

Hum, I'm rethinking about it: that won't fix my issue, namely, that Google Chrome is killing connections to unsecure ws from secure http. To fix that, I can either

 * use `connections.setFactory` at runtime, securing if `path` starts with `ws:`
   * need fixes around missing uses of `LeaderConnection`
   * secure == not by default, ie insecure
 * fix `getWebSocketAddress`
   * break retrocompat
 * add `getWebSocketURL` and use it
   * breaks retrocompat because fixes needed inside most RPC

No solution is nice. WDYT?

Gilthoniel:
>     * use `connections.setFactory` at runtime, securing if `path` starts with `ws:`
>       
>       * need fixes around missing uses of `LeaderConnection`
>       * secure == not by default, ie insecure

No please don't change that. The websocket adapter should remain the same. But somewhat related, you can, and I think that's the best way to go, use [window.isSecureContext](https://developer.mozilla.org/en-US/docs/Web/API/Window/isSecureContext) to detect a browser environment over HTTPS. Like (quickly done):

```javascript
export class WebSocketConnection implements IConnection {
    private url: URL;
    private service: string;
    private timeout: number;

    /**
     * @param addr      Address of the distant peer
     * @param service   Name of the service to reach
     */
    constructor(addr: string, service: string) {
        this.url = new URL(addr);
        // @ts-ignore
        if (global.isSecureContext) {
            this.url.protocol = "wss:";
        }

        this.service = service;
        this.timeout = 30 * 1000; // 30s by default
    }
```

tharvik:
@Gilthoniel so I rewrote the PR, using `window.isSecureContext`, that should do the trick.

external/js/cothority/src/network/connection.ts
```diff
@@ -53,7 +54,12 @@ export class WebSocketConnection implements IConnection {
      * @param service   Name of the service to reach
      */
     constructor(addr: string, service: string) {
-        this.url = addr;
+        const url = new URL(addr, {});
+        if (typeof window !== "undefined" && window.isSecureContext === true) {
```
Gilthoniel: I'm surprised you don't get a reference error here, as it's not defined in a NodeJS environment. Have you run the tests locally ?

tharvik: > I'm surprised you don't get a reference error here, as it's not defined in a NodeJS environment.

Notice the `typeof window` guard to ensure its existence.

> Have you run the tests locally ?

Yup, I did.

Gilthoniel: oh right

PR #2090, opened 2019-10-09
more advanced connection handling

ineiti:
**What this PR does**

This adds a more intelligent connection handling with an active and a reserve
pool of nodes. Depending on the speed and returned connection errors, nodes
pass between the two pools.

---

🙅‍ Friendly checklist:


external/js/cothority/src/network/connection.ts
```diff
@@ -173,35 +178,43 @@ export class RosterWSConnection {
...
+                        if (errors.length >= list.length / 2) {
                             // It's the last connection that also threw an error, so let's quit
                             reject(errors);
-                        }
-                        rotate--;
-                        if (rotate >= 0) {
-                            // Take the oldest connection that hasn't been used yet
-                            this.connectionsPool.push(this.connectionsActive[i]);
-                            this.connectionsActive[i] = this.connectionsPool.shift();
+                            conn = undefined;
```
tharvik: This is never excuted. Maybe also `throw` the list instead of `rejecting` it, as it looks clearer that the control flow is changed.

external/js/cothority/src/network/connection.ts
```diff
@@ -242,3 +250,234 @@ export class LeaderConnection extends WebSocketConnection {
...
+    }
+
+    /**
+     * Creates a new NodeList for one message.
+     * @param service
+     */
+    newList(service: string): NodeList {
+        if (Nodes.parallel > 0) {
+            this.parallel = Nodes.parallel;
+        }
```
tharvik: It's weird to change `parallel` is a seamingly unrelated call. This should have been handled by `set parallel`, no?

external/js/cothority/src/network/connection.ts
```diff
@@ -242,3 +250,234 @@ export class LeaderConnection extends WebSocketConnection {
...
+     * message. It will re-order the nodes to reflect order of arrival. If the
+     * node is more than 10x slower than the fastest node, it will be replaced
+     * with a node from the reserve queue.
+     * @param address node with successful return
+     * @param rang order of arrival
+     * @param ratio delay in answer between first reply and this reply
+     */
+    done(address: string, rang: number, ratio: number) {
+        const index = this.index(address);
+        if (index >= 0) {
```
tharvik: Instead of been a NOP when not found, we can throw an error to show possibly wrong code path.

external/js/cothority/src/network/connection.ts
```diff
@@ -242,3 +250,234 @@ export class LeaderConnection extends WebSocketConnection {
...
+        this.active.concat(this.reserve).forEach((n) => n.setTimeout(t));
+    }
+
+    /**
+     * Replaces the given node from the active queue with the first node from
+     * the reserve queue.
+     * @param index
+     */
+    private replaceActive(index: number) {
+        if (index >= 0) {
```
tharvik: NOP `if index < 0` also.

external/js/cothority/src/network/connection.ts
```diff
@@ -242,3 +250,234 @@ export class LeaderConnection extends WebSocketConnection {
...
+/**
+ * A NodeList is used to interact with the Nodes-class by allowing
+ * the requester to indicate the order of arrival of messages and
+ * which nodes didn't reply correctly.
+ */
+export class NodeList {
+    readonly active: WebSocketConnection[];
+    private reserve: WebSocketConnection[];
+    private readonly start: number;
+    private first: number;
```
tharvik:
```suggestion
    private first: number = 0;
```

.travis.yml
```diff
@@ -92,33 +49,7 @@ jobs:
...
-          branch: master
-        provider: script
-        script: >-
-          cd external/js/kyber && npm ci && npm run link && cd ../../.. &&
-          cd external/js/cothority &&
-          npm ci &&
-          npm link @dedis/kyber &&
-          npm version prerelease --preid=p`date +%y%m.%d%H.%M%S` &&
-          ./publish.sh --tag dev
+        - npm run test:nocover -- --filter=SkipchainRPC
```
tharvik: Hum, maybe you didn't wanted to remove most of travis.

ineiti: yes, that's for testing. For the moment it fails testing, but only on Travis. So I just want to run that one test...

ineiti: This is why it's still in `WIP` ;)

external/js/cothority/src/network/connection.ts
ineiti: There was a (bad) reason for that. I'll think it over again.

external/js/cothority/src/network/connection.ts
ineiti: No, because nodes from a previous connection might receive a message very late and have already been replaced by other nodes.

So it's legit that a node calls the `done`, but that it will not change its position, as it's too late.

external/js/cothority/src/network/connection.ts
ineiti: same reason as for `done`... `gotError(removed_node_address)` will do a `NOP` here, but this is legit.

external/js/cothority/src/network/connection.ts
ineiti: that's not by default in JS?

tharvik: Not in JS at least, it just a name without a type, and I don't think TypeScript does it.

external/js/cothority/spec/skipchain/skipchain-rpc.spec.ts
```diff
@@ -1,4 +1,5 @@
 import { SkipBlock, SkipchainRPC } from "../../src/skipchain";
+import Log from "../../src/log";
```
tharvik: Look unused.

external/js/cothority/src/log.ts
```diff
@@ -7,18 +7,11 @@ const lvlStr = ["E ", "W ", "I ", "!4", "!3", "!2", "!1", "P ", " 1", " 2", " 3"
...
-    get lvl() {
-        return this._lvl;
-    }
-    _lvl: number;
+    lvl: number;
     stackFrameOffset: number = 0;
 
     constructor(lvl: number) {
-        this._lvl = lvl === undefined ? defaultLvl : lvl;
+        this.lvl = lvl === undefined ? defaultLvl : lvl;
```
tharvik: Typewise, `lvl` couldn't be `undefined`, so `defaultLvl` will never be used as value.

external/js/cothority/spec/skipchain/skipchain-rpc.spec.ts
ineiti: s/Look/Looks/

tharvik: `sed'ed`

external/js/cothority/src/network/connection.ts
```diff
@@ -1,7 +1,8 @@
 import { Message, util } from "protobufjs/light";
-import shuffle from "shuffle-array";
+import { sprintf } from "sprintf-js";
```
Gilthoniel: Hmmm I missed when this has been added. Please remove it from this file and from the dependencies (package + @types). There is just no point (or prove me wrong) to have that. You can simply use the \`${var}\` formatting.

external/js/cothority/src/network/connection.ts
```diff
@@ -137,29 +137,31 @@ export class WebSocketConnection implements IConnection {
...
+    nodes: Nodes;
+    private readonly connNbr: number;
+    private msgNbr = 0;
 
     /**
      * @param r         The roster to use
      * @param service   The name of the service to reach
-     * @param parallel how many nodes to contact in parallel
      */
-    constructor(r: Roster, private service: string, parallel: number = 2) {
```
Gilthoniel: aaand that's a breaking change ;)

external/js/cothority/src/network/connection.ts
```diff
@@ -137,29 +137,31 @@ export class WebSocketConnection implements IConnection {
...
  */
 export class RosterWSConnection {
-    private addresses: string[];
-    private connectionsActive: WebSocketConnection[];
-    private connectionsPool: WebSocketConnection[];
+    // debugging variable
+    static totalConnNbr = 0;
+    private static nodes: Map<string, Nodes> = new Map<string, Nodes>();
+    // Can be set after the RosterWSConnection is set up.
+    parallel: number = 3;
```
Gilthoniel: What happens if you set parallel to `-1` ?

ineiti:
I rewrote the parallel thingie to be available form the outside without a class variable.

Started in #2099 to indicate some things that should be changed here...

ineiti:
Also, as far as I know, cothority entered `breaking changes mode`, so stuff like this should be allowed. But probably we'll have to define the set of projects that need to be tested after changes go in.

Or start a `4.0-pre` version tree...

external/js/cothority/src/network/connection.ts
Gilthoniel: @ineiti it's marked as resolved but still there ?

ineiti: what? You want me to actually `git push`?

external/js/cothority/src/skipchain/skipchain-rpc.ts
```diff
@@ -22,14 +22,12 @@ export default class SkipchainRPC {
...
-    private conn: IConnection[];
+    private leader: WebSocketConnection;
 
     constructor(roster: Roster) {
         this.roster = roster;
         this.pool = new RosterWSConnection(roster, SkipchainRPC.serviceName);
-        this.conn = roster.list.map((srvid) => {
-            return new WebSocketConnection(srvid.getWebSocketAddress(), SkipchainRPC.serviceName);
-        });
+        this.leader = new WebSocketConnection(this.roster.list[0].getWebSocketAddress(), SkipchainRPC.serviceName);
```
tharvik: You can use `LeaderConnection`

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -84,17 +84,13 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
      * @param darc          The genesis darc
      * @param blockInterval The interval of block creation in nanoseconds
      */
     static async newByzCoinRPC(roster: Roster, darc: Darc, blockInterval: Long): Promise<ByzCoinRPC> {
-        const rpc = new ByzCoinRPC();
-        rpc.conn = new WebSocketConnection(roster.list[0].getWebSocketAddress(), "ByzCoin");
-        rpc.genesisDarc = darc;
-        rpc.config = new ChainConfig({blockInterval});
-
+        const leader = new WebSocketConnection(roster.list[0].getWebSocketAddress(), "ByzCoin");
```
tharvik: `LeaderConnection` does the same thing, better use it directly.

external/js/cothority/src/network/connection.ts
```diff
@@ -140,21 +145,22 @@ export class WebSocketConnection implements IConnection {
...
+export class RosterWSConnection implements IConnection {
     // debugging variable
     static totalConnNbr = 0;
     private static nodes: Map<string, Nodes> = new Map<string, Nodes>();
-    // Can be set after the RosterWSConnection is set up.
-    parallel: number = 3;
     nodes: Nodes;
     private readonly connNbr: number;
     private msgNbr = 0;
+    private _parallel: number = 3;
```
tharvik: `sed s,_parallel,parallel,g`

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -30,6 +30,7 @@ export const currentVersion = 2;
 const CONFIG_INSTANCE_ID = Buffer.alloc(32, 0);
 
 export default class ByzCoinRPC implements ICounterUpdater {
+    static serviceName = "ByzCoin";
```
tharvik: +readonly :)

external/js/cothority/src/network/connection.ts
```diff
@@ -146,20 +149,22 @@ export class WebSocketConnection {
  * It uses the Nodes class to manage which nodes will be contacted.
  */
 export class RosterWSConnection implements IConnection {
+    // Can be set to override the default parallel value
+    static defaultParallel: number = 3;
```
tharvik: `readonly` then use `setParallel` in `startConodes`

PR #2092, opened 2019-10-09
speed up counter handling in byzcoin

ineiti:
**What this PR does**

Adds a local cache for the counters that is invalidated if a transaction
fails. This speeds up transactions a lot, because after a short set-up
phase, the counters don't have to be updated anymore.

---

🙅‍ Friendly checklist:

### DIFF
--- a/external/js/cothority/src/byzcoin/byzcoin-rpc.ts
+++ b/external/js/cothority/src/byzcoin/byzcoin-rpc.ts
@@ -31,10 +31,6 @@ const CONFIG_INSTANCE_ID = Buffer.alloc(32, 0);
 
 export default class ByzCoinRPC implements ICounterUpdater {
 
-    get genesisID(): InstanceID {
-        return this.genesis.computeHash();
-    }
-
     static staticCounters: Map<string, Map<string, Long>> = new Map<string, Map<string, Long>>();
 
     /**
@@ -123,6 +119,14 @@ export default class ByzCoinRPC implements ICounterUpdater {
     private conn: IConnection;
 
     protected constructor() {
+    }
+
+    get genesisID(): InstanceID {
+        return this.genesis.computeHash();
+    }
+
+    get latest(): SkipBlock {
+        return new SkipBlock(this._latest);
     }
 
     /**
--- a/external/js/cothority/src/byzcoin/byzcoin-rpc.ts
+++ b/external/js/cothority/src/byzcoin/byzcoin-rpc.ts
@@ -31,8 +31,6 @@ const CONFIG_INSTANCE_ID = Buffer.alloc(32, 0);
 
 export default class ByzCoinRPC implements ICounterUpdater {
 
-    static staticCounters: Map<string, Map<string, Long>> = new Map<string, Map<string, Long>>();
-
     /**
      * Helper to create a genesis darc
      * @param signers       Authorized signers
@@ -67,14 +65,14 @@ export default class ByzCoinRPC implements ICounterUpdater {
      * @returns a promise that resolves with the initialized ByzCoin instance
      */
     static async fromByzcoin(roster: Roster, skipchainID: Buffer, waitMatch: number = 0, interval: number = 1000,
-                             latest: SkipBlock = null):
+                             latest?: SkipBlock):
         Promise<ByzCoinRPC> {
         const rpc = new ByzCoinRPC();
         rpc.conn = new RosterWSConnection(roster, "ByzCoin");
 
         const skipchain = new SkipchainRPC(roster);
         rpc.genesis = await skipchain.getSkipBlock(skipchainID);
-        rpc._latest = latest ? latest : rpc.genesis;
+        rpc._latest = latest !== undefined ? latest : rpc.genesis;
 
         const ccProof = await rpc.getProofFromLatest(CONFIG_INSTANCE_ID, waitMatch, interval);
         rpc.config = ChainConfig.fromProof(ccProof);
@@ -112,6 +110,8 @@ export default class ByzCoinRPC implements ICounterUpdater {
 
         return rpc;
     }
+
+    private static staticCounters = new Map<string, Map<string, Long>>();
     private _latest: SkipBlock;
     private genesisDarc: Darc;
     private config: ChainConfig;
@@ -246,7 +246,7 @@ export default class ByzCoinRPC implements ICounterUpdater {
      * @return a promise that resolves with the proof, rejecting otherwise
      */
     async getProofFromLatest(id: Buffer, waitMatch: number = 0, interval: number = 1000): Promise<Proof> {
-        if (!this._latest) {
+        if (this._latest === undefined) {
             throw new Error("no latest block found");
         }
 
--- a/external/js/cothority/src/byzcoin/byzcoin-rpc.ts
+++ b/external/js/cothority/src/byzcoin/byzcoin-rpc.ts
@@ -30,6 +30,13 @@ export const currentVersion = 2;
 const CONFIG_INSTANCE_ID = Buffer.alloc(32, 0);
 
 export default class ByzCoinRPC implements ICounterUpdater {
+
+    get genesisID(): InstanceID {
+        return this.genesis.computeHash();
+    }
+
+    static staticCounters: Map<string, Map<string, Long>> = new Map<string, Map<string, Long>>();
+
     /**
      * Helper to create a genesis darc
      * @param signers       Authorized signers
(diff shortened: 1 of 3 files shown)


external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -30,6 +30,13 @@ export const currentVersion = 2;
 const CONFIG_INSTANCE_ID = Buffer.alloc(32, 0);
 
 export default class ByzCoinRPC implements ICounterUpdater {
+
+    get genesisID(): InstanceID {
```
Gilthoniel: Please keep the getter after the constructor

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -107,17 +116,13 @@ export default class ByzCoinRPC implements ICounterUpdater {
 
         return rpc;
     }
-
+    latest: SkipBlock;
```
Gilthoniel: private ?

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -172,9 +177,21 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
         // The error might be in the response, so we look for it and then reject the promise.
         const resp = await this.conn.send(req, AddTxResponse) as AddTxResponse;
+        // If the transaction has been successful, we update all cached counters. If the transaction
+        // failed, all involved counters are reset and will have to be fetched again.
+        transaction.instructions.forEach((instruction) => {
+            instruction.signerIdentities.forEach((signer, i) => {
+                if (resp.error.length === 0) {
+                    counters.set(signer.toString(), instruction.signerCounter[i]);
+                } else {
+                    counters.delete(signer.toString());
```
Gilthoniel: Does this mean that if a signer counter error occurred, it will remove the counter thus the next try will fetch from a conode ?

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -284,9 +301,49 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
+     * also solves the problem of counters sometimes being out of synch with the chain.
+     * After the call returns, all signers are in ClientTransaction.counters.
+     *
+     * @param rpc should allow updates for counters
+     * @param signers the signers needed
+     */
+    async updateCachedCounters(signers: IIdentity[]): Promise<Long[]> {
+        const counters = this.counters();
+        const newSigners = signers.filter((signer) => {
+            return counters.has(signer.toString()) === false;
```
Gilthoniel: IMHO it's much clearer without the `=== false` and with a `!`. Ignore if you want, not a big deal.

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
ineiti: yes - very simple cache management: if something goes wrong, clear the state and start from scratch.

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
ineiti: added a getter, as I want to read it out, so that I can store it and give it to a new `ByzCoinRPC` for faster startup.

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
ineiti: I can't do that anymore - @tharvik is watching over me and he's making sure that everything is correctly typed!

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -60,21 +63,23 @@ export default class ByzCoinRPC implements ICounterUpdater {
      * @param skipchainID   The genesis block identifier
      * @param waitMatch how many times to wait for a match - useful if its called just after an addTransactionAndWait.
      * @param interval how long to wait between two attempts in waitMatch.
+     * @param latest if given, use this to prove the current state of the blockchain. Needs to be trusted!
      * @returns a promise that resolves with the initialized ByzCoin instance
      */
-    static async fromByzcoin(roster: Roster, skipchainID: Buffer, waitMatch: number = 0, interval: number = 1000):
+    static async fromByzcoin(roster: Roster, skipchainID: Buffer, waitMatch: number = 0, interval: number = 1000,
+                             latest: SkipBlock = null):
```
tharvik:
```suggestion
                             latest?: SkipBlock):
```

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -30,6 +30,9 @@ export const currentVersion = 2;
 const CONFIG_INSTANCE_ID = Buffer.alloc(32, 0);
 
 export default class ByzCoinRPC implements ICounterUpdater {
+
+    static staticCounters: Map<string, Map<string, Long>> = new Map<string, Map<string, Long>>();
```
tharvik: Why having it static? Does the rpc get copied a lot or only referenced?

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -60,21 +63,23 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
+    static async fromByzcoin(roster: Roster, skipchainID: Buffer, waitMatch: number = 0, interval: number = 1000,
+                             latest: SkipBlock = null):
         Promise<ByzCoinRPC> {
         const rpc = new ByzCoinRPC();
         rpc.conn = new RosterWSConnection(roster, "ByzCoin");
 
         const skipchain = new SkipchainRPC(roster);
         rpc.genesis = await skipchain.getSkipBlock(skipchainID);
-        rpc.latest = rpc.genesis;
+        rpc._latest = latest ? latest : rpc.genesis;
```
tharvik:
```suggestion
        rpc._latest = latest !== undefined ? latest : rpc.genesis;
```

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -102,24 +107,28 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
-    protected constructor() {}
+    protected constructor() {
+    }
 
     get genesisID(): InstanceID {
         return this.genesis.computeHash();
     }
 
+    get latest(): SkipBlock {
+        return new SkipBlock(this._latest);
```
tharvik: It a immutable (js style) object, that should be enough.
```suggestion
        return this._latest;
```

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -225,11 +246,11 @@ export default class ByzCoinRPC implements ICounterUpdater {
      * @return a promise that resolves with the proof, rejecting otherwise
      */
     async getProofFromLatest(id: Buffer, waitMatch: number = 0, interval: number = 1000): Promise<Proof> {
-        if (!this.latest) {
+        if (!this._latest) {
```
tharvik:
```suggestion
        if (this._latest === undefined) {
```

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
tharvik: Also, why using `Map<(string,string),Long>` instead of using direclty the `InstanceIDs`?

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
tharvik: I agree with @Gilthoniel on this one, comparison with bool are redundant, but when an object get casted to bool (to test for undef really), it is type-wise wrong.
```suggestion
            return !counters.has(signer.toString());
```

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
ineiti: Static: there are multiple instances of `ByzCoinRPC` in use, at least in the `mobile`, but probably also in the `webapp`.

`Map<InstanceID` - IIRC I did that in the beginning, but it didn't work from time to time. The problem is that `Map.set(Buffer.from("123", 1)).get(Buffer.from("123")` returns `undefined`, as far as I remember. But with strings it works...

See https://runkit.com/ineiti/5d9e0424577a31001a6341d2

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
ineiti: Oh my - I hate getters and setters that don't do anything. 

(deleted 20 lines of rant)

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
tharvik: Ho, okay, it turns out that [JS Map doesn't use structural equality, only reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), so `Object` can't be used as key. To work around that, we can use [immutable.js](https://immutable-js.github.io/immutable-js/) to have better collections, key equality tested using `equals` for Map/List/Seq. which is kinda the standard by default, even if badly used by stdlib.

PR #2094, opened 2019-10-09
Move stackFrameOffset to printcaller

ineiti:
**What this PR does**

Simplifies `log`.

---

🙅‍ Friendly checklist:

### DIFF
--- a/external/js/cothority/src/log.ts
+++ b/external/js/cothority/src/log.ts
@@ -138,10 +138,10 @@ export class Logger {
     catch(e: (Error | string), ...args: any) {
         let errMsg = e;
         if (e instanceof Error) {
-            errMsg = (e as Error).message;
+            errMsg = e.message;
         }
         if (e instanceof Error) {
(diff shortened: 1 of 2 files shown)


external/js/cothority/src/log.ts
```diff
@@ -137,40 +137,23 @@ export class Logger {
 
     catch(e: (Error | string), ...args: any) {
         let errMsg = e;
-        if ((e as Error).message) {
+        if (e instanceof Error) {
             errMsg = (e as Error).message;
```
tharvik: No need to cast now :)
```suggestion
             errMsg = e.message;
```

external/js/cothority/src/log.ts
```diff
@@ -137,40 +137,23 @@ export class Logger {
...
         }
-        if ((e as Error).stack) {
+        if (e instanceof Error) {
             for (let i = 1; i < (e as Error).stack.split("\n").length; i++) {
```
tharvik:
```suggestion
             for (let i = 1; i < e.stack.split("\n").length; i++) {
```

PR #2106, opened 2019-10-15
Fix issue with isSecureContext with local content

### LINKED ISSUE #2104: network.Roster.fromTOML doesn't take Url field into account
I found out that the `network.Roster.fromTOML` function always tries to connect to secure web socket, even if the `Url` field is not set or set to `http://...`.

**How to reproduce**

1) Launch some conodes locally (`./run_nodes.sh`)
2) Set up and use `external/js/cothority/index.html` (run `npm run bundle` before)
3) Try to load the generated `public.toml` file from the step 1, which won't work.

Gilthoniel:
This changes the logic to instead check the current protocol so that
browser context with a location using HTTPS will always use WSS.

Fixes #2104

---

🙅‍ Friendly checklist:

### DIFF
--- a/external/js/cothority/spec/helpers/global-this.ts
+++ b/external/js/cothority/spec/helpers/global-this.ts
@@ -1,4 +1,9 @@
 
+// globalThis in NodeJS appeared since 12.0.0 which means we need to declare the
+// global for the LTS. As it is simply a reference to the global object (_window_
+// for the browser and _global_ for NodeJS), we can assign it.
+// This routine is run only for the tests as the production code should check for
+// existance until we drop version above 12.0.0.
 beforeAll(() => {
     // @ts-ignore
     global.globalThis = global;
--- a/external/js/cothority/spec/helpers/global-this.ts
+++ b/external/js/cothority/spec/helpers/global-this.ts
@@ -1,0 +1,5 @@
+
+beforeAll(() => {
+    // @ts-ignore
+    global.globalThis = global;
+});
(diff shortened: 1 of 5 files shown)


external/js/cothority/src/network/connection.ts
```diff
@@ -61,8 +61,10 @@ export class WebSocketConnection {
      */
     constructor(addr: string, service: string) {
         const url = new URL(addr, {});
-        if (typeof window !== "undefined" && window.isSecureContext === true) {
-            url.set("protocol", "wss");
+        if (typeof globalThis !== "undefined" && typeof globalThis.location !== "undefined") {
```
tharvik: when not running the tests, `globalThis` isn't defined, so, never going to `wss`, even on secure pages?

Gilthoniel: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

tharvik: I learn everyday with MDN.

external/js/cothority/spec/helpers/global-this.ts
```diff
@@ -0,0 +1,5 @@
+
+beforeAll(() => {
+    // @ts-ignore
+    global.globalThis = global;
```
nkcr: Thanks for the fix! In addition I would just explain why this is needed and how it is used.

PR #2115, opened 2019-10-25
Fixes a csadmin test

nkcr:
Commit 79ddf19d2ccb41ffa97a7595e923b5b4aab068a2 included a test that was only compatible with a feature implemented in the odyssey-needs branch.

This PR makes the test pass by adding this missing feature.

---

🙅‍ Friendly checklist:

### DIFF
--- a/calypso/csadmin/clicontracts/write.go
+++ b/calypso/csadmin/clicontracts/write.go
@@ -9,6 +9,7 @@ import (
 	"os"
 
 	"go.dedis.ch/onet/v4/log"
+	"golang.org/x/xerrors"
 
 	"github.com/urfave/cli"
 	"go.dedis.ch/cothority/v4"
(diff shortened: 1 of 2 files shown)


nkcr:
thanks @gnarula for pointing that out.

calypso/csadmin/clicontracts/write.go
```diff
@@ -230,6 +230,19 @@ func WriteGet(c *cli.Context) error {
 		return errors.New("didn't get a write instance: " + err.Error())
 	}
 
+	if c.Bool("export") {
+		_, buf, _, _, err := proof.KeyValue()
+		if err != nil {
+			return errors.New("failed to get value from proof: " + err.Error())
```
gnarula: Can we use `xerrors.Errorf("...: %v", err)` here please to be consistent with #2116

calypso/csadmin/clicontracts/write.go
```diff
@@ -230,6 +230,19 @@ func WriteGet(c *cli.Context) error {
...
+		}
+		reader := bytes.NewReader(buf)
+		_, err = io.Copy(os.Stdout, reader)
+		if err != nil {
+			return errors.New("failed to copy to stdout: " + err.Error())
```
gnarula: Similar to above, but perhaps a `return cothority.ErrorOrNil(err, "failed to copy to stdout")`

nkcr: I am still not familiar with those custom things. Why should I use `cothority.ErrorOrNil` in this case?

PR #2121, opened 2019-10-28
Limit viewchange timeout

ineiti:
The viewchange timeout was not handled correctly in case of very big
numbers. `time.Duration(float64(1e100))*time.Second == 0s`! So the
verification `if duration < 0` was never `true`, and the viewchange
was reset to `0s`!

---

🙅‍ Friendly checklist:

### DIFF
--- a/byzcoin/viewchange/viewchange.go
+++ b/byzcoin/viewchange/viewchange.go
@@ -1,5 +1,5 @@
 // Package viewchange implements the view-change algorithm found in the PBFT
-// paper (OSDI99). The implementation is specific our use-case, meaning that
+// paper (OSDI99). The implementation is specific to our use-case, meaning that
 // our views are represented by skipblocks. However, it is possible to create
 // an interface that abstracts the underlying data-structure of a view.
 //
@@ -29,7 +29,7 @@ import (
 
 // maxTimeout is an upper bound for the view change timeout as it is increasing
 // exponentially.
-const maxTimeout = 5 * time.Minute
+const maxTimeout = 1 * time.Minute
 
 // View assume that the context are the same.
 type View struct {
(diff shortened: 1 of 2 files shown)


byzcoin/viewchange/viewchange.go
```diff
@@ -167,21 +167,25 @@ func (c *Controller) Start(myID network.ServerIdentityID, genesis skipchain.Skip
...
 				// receiving 2*f+1 view-change messages.
-				// Watch out for overflow: When ctr gets to be too high, math.Pow(2, float64(str))
+				// Watch out for overflow: When ctr gets to be too high,
+				// math.Pow(2, float64(ctr))
 				// will be too high, and timeout will end up negative.
-				timeout := time.Duration(math.Pow(2, float64(ctr))) * initialDuration
-				if timeout < 0 || timeout.Seconds() > maxTimeout.Seconds() {
-					timeout = maxTimeout
+				timeout := maxTimeout
+				backoff := float64(ctr - 2*f)
```
Gilthoniel: Can you explain that ?

ineiti: Added comment

Gilthoniel: I actually agreed with the math but my question was about the `float64(ctr -2*f)`

ineiti: this code-path supposes that `ctr > 2*f`. So that it doesn't start with `2**(2*f)*timeout`, which can get high very fast, but that it starts with `2**(1)`.

ineiti: Shoot - mixed up `meta.countOf(ctr)` and `ctr` itself...

PR #2123, opened 2019-10-28
Codacy reports

ineiti:
---

🙅‍ Friendly checklist:


byzcoin/statetrie.go
```diff
@@ -151,7 +151,7 @@ type stateTrie struct {
 func loadStateTrie(db *bbolt.DB, bucket []byte) (*stateTrie, error) {
 	t, err := trie.LoadTrie(trie.NewDiskDB(db, bucket))
 	if err != nil {
-		return nil, xerrors.Errorf("loading trie: %v")
+		return nil, xerrors.Errorf("loading trie")
```
Gilthoniel: The issue is the missing `err` => `: %v", err)`. Same for below.

ineiti: Oupsi...

PR #2130, opened 2019-10-30
Fix message races

### LINKED ISSUE #2129: TestService_DownloadStateRunning fails intermittently
Jenkins reports that `TestService_DownloadStateRunning` fails from time to time.

And there seems also to be a race-condition while doing `TestClose`.

ineiti:
Divers races have been fixed:
- between closing down byzcoin/skipchain and calling propagateForwardLinks
- between resetting byzcoin and calling propagateForwardLinks
- added some `local.WaitDone(time.Second)` to the end of some tests to make sure all messages are passed before closing the services

Plus a proposed change to the ill-named skipchain::Service.WaitBlock

Closes #2129

---

🙅‍ Friendly checklist:

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -2567,7 +2567,6 @@ func (s *Service) TestRestart() error {
 }
 
 func (s *Service) cleanupGoroutines() {
-	log.Lvl1(s.ServerIdentity(), "closing go-routines 1")
 	s.heartbeats.closeAll()
 	s.closeLeaderMonitorChan <- true
 	s.viewChangeMan.closeAll()
@@ -2579,9 +2578,7 @@ func (s *Service) cleanupGoroutines() {
 		delete(s.pollChan, k)
 	}
 	s.pollChanMut.Unlock()
-	log.Lvl1(s.ServerIdentity(), "closing go-routines 2")
 	s.pollChanWG.Wait()
-	log.Lvl1(s.ServerIdentity(), "closing go-routines 3")
 }
 
 func (s *Service) monitorLeaderFailure() {
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -15,6 +15,8 @@ import (
 	"sync"
 	"time"
 
+	uuid "gopkg.in/satori/go.uuid.v1"
+
 	"go.dedis.ch/cothority/v3"
 	"go.dedis.ch/cothority/v3/blscosi/protocol"
 	"go.dedis.ch/cothority/v3/byzcoin/trie"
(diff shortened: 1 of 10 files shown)


skipchain/skipchain.go
```diff
@@ -955,9 +955,24 @@ func (s *Service) ListFollow(list *ListFollow) (*ListFollowReply, error) {
...
+// ChainIsProcessing returns true if the given chain has a block in the
+// blockBuffer.
+func (s *Service) ChainIsProcessing(sid SkipBlockID) bool {
+	return s.blockBuffer.has(sid)
+}
+
 // WaitBlock returns a block by its ID instantly if already stored in the DB
 // or check if the block is inside the buffer. If the block is known, false will
 // be returned because a catch up is not necessary, true otherwise.
+// DEPRECATED
```
nkcr: > To signal that an identifier should not be used, add a paragraph to its doc comment that begins with Deprecated: followed by some information about the deprecation, and a recommendation on what to use instead, if applicable.

I would add some explanations about the deprecation

byzcoin/service.go
```diff
@@ -2569,7 +2579,9 @@ func (s *Service) cleanupGoroutines() {
 		delete(s.pollChan, k)
 	}
 	s.pollChanMut.Unlock()
+	log.Lvl1(s.ServerIdentity(), "closing go-routines 2")
 	s.pollChanWG.Wait()
+	log.Lvl1(s.ServerIdentity(), "closing go-routines 3")
```
tharvik: Debug?

byzcoin/service_test.go
```diff
@@ -2709,8 +2710,8 @@ func (s *ser) testDarcEvolution(t *testing.T, d2 darc.Darc, fail bool) (pr *Proo
 }
 
 func (s *ser) deleteDBs(t *testing.T, index int) {
-	log.Lvl1("Deleting DB of node", index)
 	bc := s.services[index]
+	log.Lvl1("Deleting DB of node", index, bc.ServerIdentity())
```
tharvik:
```suggestion
	log.Lvlf1("%v: deleting DB", bc.ServerIdentity())
```

byzcoin/tx_pipeline.go
```diff
@@ -110,7 +110,11 @@ func (s *defaultTxProcessor) CollectTx() (*collectTxResult, error) {
 		return nil, xerrors.Errorf("reading config: %v", err)
 	}
 
-	_, isNotProcessingBlock := s.skService().WaitBlock(s.scID, nil)
+	if s.skService().ChainIsProcessing(s.scID) {
+		// When a block is processed, wait for the block to be done
+		time.Sleep(bcConfig.BlockInterval / 2)
+		return &collectTxResult{Txs: nil, CommonVersion: CurrentVersion - 1}, nil
```
tharvik: That looks weird, you wait for some time then return the same data you'd have before sleeping.

ineiti: Looking at the call chain, I think I can remove that one. Sending it to Travis for confirmation ;)

PR #2139, opened 2019-11-20
Add functionalities to personhood

ineiti:
During the OpenHouse2019 demo, a couple of new functionalities have been added:
- remove polls
- protect adding of new personhood-gatherings

Both functionalities need to be signed off by the conode-key. The
personhood/phapp binary can do so when given a private.toml file.

This also updates the `external/js/cothority/src/protobuf/models.json` to
include `PollDelete`, so that the cothority-npm can be correctly imported
when using personhood.

---

🙅‍ Friendly checklist:


jeffallen:
FYI, I am currently rebasing this onto v3.

PR #2143, opened 2019-11-21
external/js/kyber: add @types/elliptic for users

tharvik:
**Allow @dedis/{kyber, cothority} to be imported in typescript land without adding extra deps**

This PR moves `@types/elliptic` from dev deps to user deps, thus allowing theses project to use it with typescript direclty, without having to add its own version of the typing lib. It's also removing the now dangling deps of @dedis/cothority.

When building a project using @dedis/cothority or @dedis/kyber, `tsc` yield:
```
node_modules/@dedis/kyber/curve/edwards25519/curve.d.ts:2:23 - error TS7016: Could not find a declaration file for module 'elliptic'. '/home/tharvik/c4dt/service-drynx/library/node_modules/elliptic/lib/elliptic.js' implicitly has an 'any' type.
  Try `npm install @types/elliptic` if it exists or add a new declaration (.d.ts) file containing `declare module 'elliptic';`

2 import { curve } from "elliptic";
                        ~~~~~~~~~~

node_modules/@dedis/kyber/curve/nist/curve.d.ts:1:22 - error TS7016: Could not find a declaration file for module 'elliptic'. '/home/tharvik/c4dt/service-drynx/library/node_modules/elliptic/lib/elliptic.js' implicitly has an 'any' type.
  Try `npm install @types/elliptic` if it exists or add a new declaration (.d.ts) file containing `declare module 'elliptic';`

1 import elliptic from "elliptic";
```

### DIFF
--- a/external/js/cothority/package-lock.json
+++ b/external/js/cothority/package-lock.json
@@ -1061,15 +1061,6 @@
         "@types/node": "*"
       }
     },
-    "@types/elliptic": {
-      "version": "6.4.2",
-      "resolved": "https://registry.npmjs.org/@types/elliptic/-/elliptic-6.4.2.tgz",
-      "integrity": "sha512-aIhX07nKFHKTw8z5mZcsgbXjTD9pOXVbqtcNUbdhd3YhH5pC94+plnSglrPFhbHgeL2dCTobnb4kn6gtf7UC3A==",
-      "dev": true,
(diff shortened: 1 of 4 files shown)


tharvik:
@Gilthoniel can you also make a NPM release for it (and also for @dedis/cothority), please?

PR #2161, opened 2019-12-08
Use new darc-rule only in version 3

ineiti:
There has been a mismatch with the release of the nodes and the spawning of contracts. To keep the chain in a consistent state, the version needs to be updated to version 3, and only then the new darc rule needs to be used.

---

🙅‍ Friendly checklist:

### DIFF
--- a/byzcoin/bcadmin/main.go
+++ b/byzcoin/bcadmin/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/qantik/qrgo"
 	"github.com/urfave/cli"
 	"go.dedis.ch/cothority/v3"
+	_ "go.dedis.ch/cothority/v3/bevm"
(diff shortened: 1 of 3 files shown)


byzcoin/bcadmin/main.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/qantik/qrgo"
 	"github.com/urfave/cli"
 	"go.dedis.ch/cothority/v3"
+	_ "go.dedis.ch/cothority/v3/bevm"
```
Gilthoniel: Why is this in the PR ?

ineiti: Because BEVM is now part of the byzcoin-conodes, and there are BEVM instructions in the chain. So if BEVM is not included in `bcadmin`, the command to verify the chain, `bcadmin db replay`, fails...

PR #2168, opened 2019-12-12
travis: add lint to external/js

tharvik:
* inline single-use anchors
  * we are only building for go 1.12, not testing, is it wanted?
* add lint for external/js

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,18 @@ jobs:
       script:
         - make -C conode verify
         - GO111MODULE=on make test_{fmt,lint}
+    - name: "js > kyber"
+      <<: *language_js
+      script:
+        - cd external/js/kyber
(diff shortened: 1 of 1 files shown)


ineiti:
I think DEDIS' promise on the code is to run it on the last two stable versions, which would be go 1.12 and go 1.13 in December 2019.

ineiti:
> I don't recall how it got to be "build for 1.12 and 1.13, test for 1.13 only", but that's how it was before, so you're not really changing anything here except the linting.

I think I remember it had to do with the testing-time - if you need to test both go 1.12 and 1.13, it takes nearly double the time. Even though travis builds are parallelized, they don't allocate an infinity of tasks to each PR. So more tasks, even if they can be run in parallel, still make it slower.

PR #2169, opened 2019-12-19
Separate personhood-contract and service

ineiti:
For the github.com/c4dt/byzcoin I want to include only the contracts from personhood,
but not the service. This PR puts the contracts in its own directory, so that the
byzcoin-repository can include only that.

Also rewrites the testing of the contracts to not use the service itself, but to
test only the contracts, without the whole byzcoin-service around. This could
also be applied to other contracts...

---

🙅‍ Friendly checklist:


nkcr:
a step towards  https://github.com/dedis/smartcoin ?  :)

ineiti:
As long as we can replay the blocks in the `9cc3` dedis-chain, that sounds like a good idea ;)

The main purpose is to have:
- allow ELCA to do a code-review and then run a node
- allow others to run a node
- regular release of docker-image

PR #2172, opened 2020-01-07
external/kyber: test for BN256 add/mul consistency

tharvik:
PR failing on purpose (for now).
Disclaimer: I don't know much about crypto, I might be completely wrong.

__Multiplication is not consistent with addition on BN256G1 in typescript, ie `s*P != P + P + P + ...`.__ I'm adding a small test showing it, which is really close to the [golang one](https://github.com/dedis/kyber/pull/410). From what I gather, it's the addition which is wrongly implemented.

Now, I don't know how to fix it, so some help from a cryptographer would be nice :)

### DIFF
--- a/external/js/kyber/spec/pairing/point.spec.ts
+++ b/external/js/kyber/spec/pairing/point.spec.ts
@@ -26,6 +26,26 @@ describe("BN256 Point Tests", () => {
         expect(prop).toHold();
     });
 
+    it("should yield the same point by addition and multiplication", () => {
+        const prop = jsc.forall(jsc.uint8, (target) => {
+            const base = new BN256G1Point().base();
+            const scalarUnit = new BN256Scalar().one()
+            const pointUnit = new BN256G1Point().mul(scalarUnit, base)
+
+            const scalarAdder = new BN256Scalar();
+            const pointAdder = new BN256G1Point();
+            for (let i = 0; i < target; i++) {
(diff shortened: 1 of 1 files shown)


external/js/kyber/spec/pairing/point.spec.ts
```diff
@@ -26,6 +26,26 @@ describe("BN256 Point Tests", () => {
...
+        const prop = jsc.forall(jsc.uint8, (target) => {
+            const base = new BN256G1Point().base();
+            const scalarUnit = new BN256Scalar().one()
+            const pointUnit = new BN256G1Point().mul(scalarUnit, base)
+
+            const scalarAdder = new BN256Scalar();
+            const pointAdder = new BN256G1Point();
+            for (let i = 0; i < target; i++) {
+                scalarAdder.add(scalarAdder, scalarUnit)
+                pointAdder.add(pointAdder, pointUnit);
```
Gilthoniel: `pointAdder.add(pointAdder.clone(), pointUnit);` fixes the test so it seems there is something wrong in the addition that modifies the first parameter.

Gilthoniel: I'll have a better look tomorrow.

Gilthoniel:
See #2173 for the fix.

tharvik:
Nice, thanks for investigating; closing this one then.

PR #2173, opened 2020-01-08
KyberJS: fix curve point double stability

Gilthoniel:
This allows a and this to be the same object and still get a correct
result.

Closes #2172 

---

🙅‍ Friendly checklist:


ineiti:
I tried to understand the change you did but I don't see why it's working that way ;) Can you give me a hint, please?

Gilthoniel:
`this.y` was modified and then `a.y` used to compute the z value and as `this` and `a` was the same object, the result was erroneous.

PR #2175, opened 2020-01-08
external/kyber/bn256: fix order for scalar

tharvik:
**The modulo for bn256' scalar in typescript is wrong.**

By looking at the [golang implementation of bn256' scalar](https://github.com/dedis/kyber/blob/master/pairing/bn256/group.go#L69), we see that it uses the order (the number of element of a ring) as modulus. In the typescript, it actually uses `p`, not `group`, which gives that addition is not consistent with multiplication when using negatives.

This PR adds a test triggering the issue, and the fix. It also contain a simplification of multiplication for bn256 (which is unrelated, just easier to understand).

Kudos to @ineiti for helping me debug :)

### DIFF
--- a/external/js/kyber/spec/pairing/point.spec.ts
+++ b/external/js/kyber/spec/pairing/point.spec.ts
@@ -46,6 +46,27 @@ describe("BN256 Point Tests", () => {
         expect(prop).toHold();
     });
 
+    it("should yield the same point by negative addition and multiplication", () => {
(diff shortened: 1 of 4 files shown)


ineiti:
Yes, that was a good run - even though at one moment we thought the go-implementation is wrong and that the private keys used in the byzcoin-testnet are wrong, too... Fortunately it was just the typescript implementation, which has never been used to create private bn256 keys!

PR #2177, opened 2020-01-09
external/kyber/bn256: mod after neg

tharvik:
-1 is not encoded the same when using `neg` rather than `mul`. Adding a small test for it and a fix (yet, I'm no cryptographer, I don't know if it's correct at all).

I'm using `p` here, as throughout the typescript codebase, see dedis/kyber#412 if you find it weird.


ineiti:
Checking in 

https://github.com/dedis/kyber/blob/406d4f502bd17963e8f3ff6982673e0bbe9e588d/group/mod/int.go#L218

This seems to be the correct thing to do.

PR #2180, opened 2020-01-15
Fix LookupSciper problem

jeffallen:
The people.epfl.ch website we are using changed their output format
so adjust to compensate.

### DIFF
--- a/evoting/api_test.go
+++ b/evoting/api_test.go
@@ -1,16 +1,12 @@
 package evoting_test
 
 import (
-	"fmt"
-	"net/http"
-	"net/http/httptest"
 	"testing"
 
 	"go.dedis.ch/onet/v3"
 	"go.dedis.ch/onet/v3/log"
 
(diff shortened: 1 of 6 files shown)


jeffallen:
Please don't commit yet, I realised on the drive home that I meant to make the unit test not depend on external servers, which is a very bad thing to do.

jeffallen:
VPSI says that lookup by SCIPER over HTTPS is going away, and that I need to change this to use LDAP. Will do.

jeffallen:
The build that Github says failed succeeded in Travis, so I don't know what's going on.

ineiti:
> The build that Github says failed succeeded in Travis, so I don't know what's going on.

Did you restart one of the build jobs? Because the `Deploy` only runs if all the other jobs succeed. So if one fails temporarily, and you restart it, you also need to restart `Deploy`. At least I had that problem some time ago...

PR #2182, opened 2020-01-17
GetUpdateChain: send all available updates

### LINKED ISSUE #2181: GetUpdateChain error
The following `else` is not optimal:

https://github.com/dedis/cothority/blob/master/skipchain/skipchain.go#L528

Example given: in the current chain, `conode.c4dt.org` has been removed for some blocks. So it goes through the blocks and finds that the highest link from block 2048 points to a block with a roster where `conode.c4dt.org` is not in anymore. So it stops searching and returns.

Instead, the node should return as many blocks as he can. At least in this case the clients would receive more blocks like that.

Also the `skipchain/Client.GetUpdateChain` doesn't recognize the corner case where a node returns a valid block but where the node is not in the roster. This should not happen.

Also the `external/js/cothority/src/skipchain/SkipchainRPC.getUpdateChain` doesn't work correctly and can fall into an endless loop.

ineiti:
Closes #2181

### DIFF
--- a/external/js/cothority/src/skipchain/skipchain-rpc.ts
+++ b/external/js/cothority/src/skipchain/skipchain-rpc.ts
@@ -1,4 +1,10 @@
-import { IConnection, LeaderConnection, RosterWSConnection, WebSocketConnection } from "../network/connection";
+import Log from "../log";
+import {
+    IConnection,
+    LeaderConnection,
+    RosterWSConnection,
+    WebSocketConnection,
+} from "../network/connection";
 import { Roster } from "../network/proto";
 import {
     GetAllSkipChainIDs,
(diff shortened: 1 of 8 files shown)


external/js/cothority/src/skipchain/skipchain-rpc.ts
```diff
@@ -33,14 +39,15 @@ export default class SkipchainRPC {
...
     private roster: Roster;
-    private conn: IConnection;
+    private conn: RosterWSConnection | IConnection | undefined;
 
-    constructor(nodes: Roster | IConnection) {
+    constructor(nodes: Roster | IConnection | RosterWSConnection) {
         if (nodes instanceof Roster) {
             this.roster = nodes;
             this.conn = new RosterWSConnection(nodes, SkipchainRPC.serviceName);
+        } else if (nodes instanceof RosterWSConnection) {
```
nkcr: looks like this `else if` could be removed

PR #2183, opened 2020-01-22
*: bump go.{mod,sum}

tharvik:
Since https://github.com/btcsuite/btcd/issues/1487, a clean build (without module cache) fails. This PR bumps btcd and others dependencies in go.mod.

This issue was undetected for long as travis and most devs are building with an already populated module cache.


nkcr:
@tharvik can you solve the conflicts before I merge it?

jeffallen:
This was already taken care of in b776c0a1f.

PR #2185, opened 2020-01-29
Add the possibility to rencrypt to a different key from the signing key.

jeffallen:


jeffallen:
No need for this for the moment.

PR #2184, opened 2020-01-23
Adds PaginateBlocks service

nkcr:
This PR implements a new PaginateBlocks service that accepts stream connections.

Is linked with the work of https://github.com/dedis/onet/pull/616, which is part of the "Blockchain visualizer" project.

**Must not be merged**
Split into two different PRs:
- #2192 (Byzcoin service)
- #2193 (js client)

---

New messages:
- PaginateRequest
- PaginateResponse

Update of the cothority/js:
- supports Streaming request
- supports the new PaginateBlocks service

Update the cothority/js websocket connection:
- update the IConnection interface to include a sendStream function

---

🙅‍ Friendly checklist:

- [ ] 0. Code comments are added (or updated) when/where needed and explain the WHY of the code.
- [ ] 1. Design choices, user documentation and any additional doc are added (or updated) in READMEs.
- [ ] 2. Any new behaviour is tested and small units of code that can be are unit tested.
- [ ] 3. Code comments are added on tests to explain what they do.
- [ ] 4. Errors are systematically wrapped with a meaningful message using `xerrors.Errorf` and the `%v` verb.
- [ ] 5. Hard limit of 80 chars is always respected.
- [ ] 6. Changes are backward compatible.
- [ ] 7. Indentation level does not exceed 5, although 4 is already suspicious.
- [ ] 8. Functions, files, and packages are kept to a manageable size and decomposed into smaller units if needed.
- [ ] 9. There are no magic values.

PR #2192, opened 2020-01-30
Adds a PaginateBlocks service to byzcoin

nkcr:
**What this PR does**

This PR add a new PaginateBlocks service that allows one to "page" the blockchain. It has the advantage of using fully bi-directional streams, allowing the client to request an asynchrounous sequence of page, and re-use the opened channel to request more pages.
In a more textual way, the client can say "give me 10 pages that contain each 50 blocks". Then the service will send 10 messages to the client, each containing 50 consecutive blocks.

This PR is related to
- https://github.com/dedis/onet/pull/616 (update of Onet to handle bi-directional streaming)
- #2184 (original PR that has been split to help review process)

---

🙅‍ Friendly checklist:


ineiti:
Following the discussion on the slack channel, shouldn't this go into the skipchain service?

nkcr:
Yes, that would make sense. One reason I see to keep it in the Byzcoin service is to stay consistent with what has already implemented, ie. keeping it with the streaming service.

ineiti:
You could put a `deprecated` for the current streaming service, and put yours in the `skipchain` service.

nkcr:
Ok, let's do that. Should I put it in the `skipchain/skipchain.go` or in a new `skipchain/streaming.go`?

go.mod
```diff
@@ -12,6 +12,8 @@ require (
 	github.com/ethereum/go-ethereum v1.8.27
 	github.com/go-ldap/ldap/v3 v3.1.5
 	github.com/hashicorp/golang-lru v0.5.3 // indirect
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
```
jeffallen: Why?

byzcoin/proto.go
```diff
@@ -301,6 +301,45 @@ type StreamingResponse struct {
...
+	// Identifies the channel the service uses to send back messages. The client
+	// must provide it if he wants to send additional requests using the same
+	// channel.
+	StreamID []byte
+	// Tells if the result contains consecutive blocks in a reversed order.
+	Backward bool
+	// Used to tell the client if an error occured. Any error code not equal to
+	// 0 means that something special happened.
+	ErrorCode uint64
+	// A list of error messages in case something special happended.
```
jeffallen: s/happended/happened/

"Something special" is kind of strange. Anyway, I don't see a place where more than one is returned. What is this for? What am I missing here?

byzcoin/proto.go
```diff
@@ -301,6 +301,45 @@ type StreamingResponse struct {
...
+	PageNumber uint64
+	// Identifies the channel the service uses to send back messages. The client
+	// must provide it if he wants to send additional requests using the same
+	// channel.
+	StreamID []byte
+	// Tells if the result contains consecutive blocks in a reversed order.
+	Backward bool
+	// Used to tell the client if an error occured. Any error code not equal to
+	// 0 means that something special happened.
+	ErrorCode uint64
```
jeffallen: The only two that are used are 0 and 1. Can't the caller know this from len(ErrorText)?

byzcoin/proto.go
nkcr: I actually forgot to make the final update that would use different error codes. Its added now.

One use case is when the user asks for too many blocks than the chain has. In that case, the service returns no blocks (its either all the requested blocks in a page, or none). But the user can check the error code, 5 in this case, and know that the list of error `errorText` contains the index of the last block, which the client can then use to make a new request with the correct number of blocks.

go.mod
nkcr: This one keeps appearing automatically, I'll remove it.

byzcoin/streaming.go
```diff
@@ -105,3 +107,132 @@ func (s *Service) StreamTransactions(msg *StreamingRequest) (chan *StreamingResp
 	}()
 	return outChan, stopChan, nil
 }
+
+// PaginateBlocks return blocks with pagination, ie. N asynchounous requests
+// that contain each K consecutive blocks. If a streamID is given, it uses the
+// same PaginateResponse chan AND the same close chan. The caller is responsible
+// for closing the close chan when the caller wants to close the connection. The
+// close chan should be closed only when no other requests are being processed using
```
Gilthoniel: Is this still true ?

byzcoin/streaming.go
```diff
@@ -105,3 +107,132 @@ func (s *Service) StreamTransactions(msg *StreamingRequest) (chan *StreamingResp
...
+func (s *Service) PaginateBlocks(msg *PaginateRequest) (chan *PaginateResponse, chan bool, error) {
+
+	outChan := make(chan *PaginateResponse)
+	stopChan := make(chan bool)
+
+	go func() {
+
+		if msg.PageSize < 1 {
+			outChan <- &PaginateResponse{
+				ErrorCode: 2,
```
Gilthoniel: Can we have constants for the error codes with meaningful names ?

PR #2196, opened 2020-02-04
external/cothority: WebSocketConnection supports URL

tharvik:
* using `URL` inside `WebSocketConnection` allow for correct/better URL handling
  * previously, using an URL with a pathname already defined (such as adding a slash at the end) fails when used in `WebSocketConnection.send`: `https://conode.dedis.ch/` will send to `https://conode.dedis.ch//...`
  * now it inconditionally sets the pathname, caring solely about the origin
* also adding a structural check for given URL
* also remove `url-parse` and direclty use URL


ineiti:
I have a faint memory of `URL` not working in a given environment - most probably `nativescript`, and that this is why we used `url-parse`. Putting this back into WIP until you confirm that it works in nativescript...

tharvik:
> I have a faint memory of `URL` not working in a given environment - most probably `nativescript`, and that this is why we used `url-parse`. Putting this back into WIP until you confirm that it works in nativescript...

Merging for now and checking that is works on mobile followed in https://github.com/c4dt/omniledger/issues/227

nkcr:
This PR wasn't updated with master, and make the `connection-spec.ts#WebSocketAdapter Tests with sendStream` fail.

PR #2197, opened 2020-02-05
DARC: Add support for Sovrin DID Verification

gnarula:
**What this PR does**

Adds support for Sovrin DID Resolution in Darcs using Indy CLI.
`indy-cli` is expected to be in `$PATH`. The pool configuration file is
assumed to be in `os.UserConfigDir()/indy/genesis`


ineiti:
Just to make sure: is it true that the darc-verification does not call out to the network. 

Because it should always be possible to verify a signature even if there is no network. And old blocks might need to be replayed, so the signature verification should not depend on any services that are available only part-time.

gnarula:
> Just to make sure: is it true that the darc-verification does not call out to the network.
> 

At the moment it does require a network call to resolve a DID to a DID Document (aka determining the public keys associated with a DID). One future improvement can be to cache the DID document.

I should add that using a cached copy comes with the risk of the conodes using obsolete public keys to verify messages which the user has rotated.

> Because it should always be possible to verify a signature even if there is no network. And old blocks might need to be replayed, so the signature verification should not depend on any services that are available only part-time.

The resolution for `did:sov` method relies on the Sovrin blockchain to resolve the DID document and is highly available. Please refer to https://github.com/sovrin-foundation/indyscan/blob/master/daemon/genesis-data.js for an example list of validator nodes to connect to.

I think an edge case would be a user having rotated their keys and the blocks with old signatures being replayed. I recall some discussions about retreiving DIDDoc at timestamp=t which should definitely be possible since DID Docs evolve like DARCs ;) This isn't supported in this PR at the moment because `indy-cli` doesn't expose that functionality.

ineiti:
I might still misunderstand what you're doing here. But from what I understand, this PR breaks one very big, important rule:

- `ClientTransactions` must only depend on the global state and the information available in the block

If you break this rule, various things will happen:
- replay is not possible anymore
- some nodes will fail to confirm the `ClientTransaction`, while others might think it's OK
- the evaluation of the `ClientTransaction` will be delayed because of network problems

This is the reason @jeffallen worked on the https://github.com/dedis/cothority/tree/master/authprox, to map external, centralised, network-accessed resources to something a `ClientTransaction` can trust.

I still might misunderstand what you're doing, but from what I understand now, you'll have to change this to be part of the `authprox` service!

jeffallen:
Thanks for the feedback @ineiti . It is true that as I was helping Gaurav set his direction and plan this work I didn't take into consideration that ByzCoin needs to be self contained.

This work is fundamentally about adding a level of indirection between the Identity that signs a certain ByzCoin instruction and the public key that does the signing. The DID is the long-term name of a given identity, and DID resolution is the point where the DID is changed into the currently active public key. This code is currently structured close to right, in that it uses the DID as the signer's identity, and then does the resolution into a public key at the last possible moment before verifying.

The DID resolution step, in the case of Sovrin DID's, is handled by a permissioned blockchain. So what we are talking about is making a connection between two blockchains. The logical way to do that is for ByzCointo refer to a point in time (the block number) in the Sovrin chain.

So what I propose is to:
* add an indication of which block number a certain DID should be resolved at to the Signer identity
* in this PR or one coming soon to offer a second implementation of interface DIDResolver that can operate entirely in an offline manner, taking as input a genesis transaction (the trust root in the Sovrin world) and a set of blocks stored in a big file.

In this way, online nodes can resolve against the online Sovrin blockchain, and offline verifiers can still check that previous authentication decisions respected the DID holder's policy for which public key to use at the time when they sent in the tx.

jeffallen:
Other comments for reviewers to be aware of:

# How to talk to Sovrin blockchain

calling out to indy-cli is a short-term hack to get us up and running, the focus of this PR needs to be on the flow of tx authorisation data from clients through darc authorisation, and into the ByzCoin ledger.

# Why we are not using authprox

Authprox is for a situation where the transaction signer has a credential but is not able/willing to hold/protect a private key. So instead, the key is stored in a key escrow service for them, and is then used to sign messages on their behalf. In the case of authprox the key escrow is not centralised, but a collective authority that holds shares of a secret and then use their share to contribute a partial signature onto a statement like, "this user proved to me that IdP X recognises them".

In the sovereign identity use case, a ByzCoin transaction originator is holding on to their secret key, and ready to use it to sign. But they want to get the benefits of the existing key management infrastructure provided by other parts of the sovereign identity world.

darc/darc_test.go
```diff
@@ -625,4 +630,128 @@ func TestParseIdentity(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NotNil(t, i.DID)
+	require.Equal(t, strings.Split(in, ":")[2], i.DID.DID)
+	require.Equal(t, "sov", i.DID.Method)
+}
+
+func TestDarc_EvolveToDIDDarc(t *testing.T) {
+	dataPublisher := createDarc(1, "Patient Data Publisher Darc")
+
+	patientIdentity := NewIdentityDID("WRfXPg8dantKVubE3HX8pw", "sov")
```
jeffallen: Flipping these two arguments would make morse sense, I think.

darc/darc_test.go
```diff
@@ -625,4 +630,128 @@ func TestParseIdentity(t *testing.T) {
...
+
+	patientDarc := NewDarc(patientRules, []byte("Patient Darc"))
+
+	signature := "4b61e1e81fb17145d9a560724f16902673a5b9c86cb5bd330d51722fcd66b9311b1821e319e9160062f88eab476a9e3bdc5e770f85b194eec91585fea24d400e"
+	sigBuf := make([]byte, hex.DecodedLen(len([]byte(signature))))
+	hex.Decode(sigBuf, []byte(signature))
+
+	// overriding the global in the test
+	didResolver = &IndyCLIDIDResolver{
+		Path:            "./mock_resolver_outputs/test_resolver",
```
jeffallen: Instead of running a test_resolver from the unit test, teach IndyCLIDIDResolver that if it has an io.Reader set, then it should read from that instead of spawning `indy-cli`. Then you can have the canned results right here in the unit test.

darc/darc_test.go
```diff
@@ -625,4 +630,128 @@ func TestParseIdentity(t *testing.T) {
...
+
+func TestDarc_VerifyWithDIDDarc(t *testing.T) {
+	patientIdentity := NewIdentityDID("CvjH5UcquEEqsz5LrZcfvM", "sov")
+
+	patientRules := InitRules([]Identity{patientIdentity}, []Identity{patientIdentity})
+	patientRules.AddRule("Read", expression.InitOrExpr(patientIdentity.String()))
+
+	patientDarc := NewDarc(patientRules, []byte("Patient Darc"))
+
+	signature := "4b61e1e81fb17145d9a560724f16902673a5b9c86cb5bd330d51722fcd66b9311b1821e319e9160062f88eab476a9e3bdc5e770f85b194eec91585fea24d400e"
```
jeffallen: Include instructions to the next poor sucker how to generate this, if needed.

darc/didresolver.go
```diff
@@ -0,0 +1,172 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"math/rand"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/mr-tron/base58"
```
jeffallen: please use the same base58 already used by aries-framework-go, so as to reduce the dependencies: github.com/btcsuite/btcutil/base58

See also: https://github.com/hyperledger/aries-framework-go/issues/1192

darc/didresolver.go
```diff
@@ -0,0 +1,172 @@
...
+		})
+	}
+	return _createdAt, []did.PublicKey{pk}, svcs, nil
+}
+
+// Resolve resolves a did to a DID document using indy-cli.
+func (r *IndyCLIDIDResolver) Resolve(id string) (*did.Doc, error) {
+	output, err := r.executeCli(id)
+	if strings.Contains(output, "NYM not found") {
+		return nil, errors.New("DID not found in the ledger")
```
jeffallen: Please follow the standard for use of xerror as proposed by @nkcr .

darc/mock_resolver_outputs/test_resolver
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
jeffallen: These files will go away, the content can be inside of the unit tests directly. (less files = more good, within reason)

darc/darc.go
```diff
@@ -919,6 +947,15 @@ func (id Identity) GetPublicBytes() []byte {
 			return nil
 		}
 		return buf
+	case 4:
+		doc, err := didResolver.Resolve(id.DID.DID)
+		if err != nil {
+			return nil
+		}
+		for _, pk := range doc.PublicKey {
```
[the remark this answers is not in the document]
ineiti: @gnarula : this for is a nice solution if you have a map. But as PublicKey is a slice, it's much simpler as jeff suggests.

ineiti:
# AuthProxy usage

With regard to AuthProxy: think of it as a decentralized TownCrier (https://eprint.iacr.org/2016/168.pdf) without the need for SGX. 

Towncrier describes how to create an _Authenticated data feed_ that is verifiable offline. It uses SGX, but AuthProxy can do the same without SGX.

In detail, here is how it could work. I'm sure there are some parts to be optimized. The most important part is that the verification is now offline and doesn't need to contact any outside source.

## Setup AuthProxy - online

First you need to setup AuthProxy to do the DID-lookups for you. You chose a set of nodes that you trust, they do their key-sharing and run.

Everybody can call these AuthProxys to create a signature on a hash of the following data:

- DID itself (the string)
- The DID public key
- Timestamp

The AuthProxy nodes verify this data by looking it up on the Sovrin blockchain, and apply their partial signature if they agree.

The final AuthProxy signature proves that a given DID is linked to a given DID public key at a given time.

## Create a new DID-Identity for DARCs - offline

The DID-Identity holds the following information and does not change over time:

- DID itself (the string)
- AuthProxy public key
- TTL of the AuthProxy signature - how long you trust the AuthProxy signature

This DID-Identity will go into a DARC expression and is stored in ByzCoin to authorize certain tasks.

The consideration of the duration of the TTL needs to include the fact that during the TTL you cannot revoke your key, as ByzCoin will still trust signatures done with the old key. So the TTL must be at least as long as the time it takes for AuthProxy to do the verification, plus the inclusion of the transaction in Byzcoin, so probably 10s. To make it easier for multiple signatures, I think you could go as high as 1h.

## Sign a ClientTransaction using a DID-Identity - online (once per TTL period) / offline

When you want to send a ClientTransaction to ByzCoin, it needs to be signed by the private key of the DID. To make the signature verifiable by ByzCoin, it needs to include the following:

- The signature on the ClientTransaction verifiable using the DID public key
- The DID public key used
- The timestamp of the AuthProxy signature
- The signature from AuthProxy proving that the DID public key and the DID match, verifiable with the AuthProxy public key in the DID-Identity

## Verify a DID-Identity signature - offline

To verify the signature, the method needs to do three things:

- Verify the DID public key is valid for the given timestamp by using the signature from the AuthProxy
- Verify the DID public key is valid by checking the TTL against the timestamp of the block and the timestamp of the AuthProxy
- Verify the signature of the message itself is valid using the DID public key

# Other solutions

I don't know Sovrin in detail, so it might be possible that you can do something similar with claims or ZKPs. But then you'd have to trust the issuer of these claims, which you avoid by using AuthProxy.

gnarula:
Closing this in light of the discussion and the new idea suggested above

PR #2208, opened 2020-02-13
Merges Odyssey needs

nkcr:
**What this PR does**

In order for Odyssey to use Cothority, we need to be able to register custom Darc attribute parsers on a read request. This PR mainly adds a way to call an `AddReadAttrInterpreter` in the Calypso service that will fill a global list of attribute interpreters. Those interpreters are then applied when it comes to check a read request.

It also updates `csadmin` to offer the ability to provide a "projectInstanceID" for a read request that can be optionally used in the context of running a custom attribute interpreter.

---

🙅‍ Friendly checklist:


calypso/service.go
```diff
@@ -69,6 +69,21 @@ const calypsoReshareProto = "calypso_reshare_proto"
...
+// Allows one to register custom MakeAttrInterpreters for the read request
+// verify.
+var readMakeAttrInterpreter = make([]makeAttrInterpreterWrapper, 0)
+
+// makeAttrInterpreterWrapper holds the data needed to register a
+// MakeAttrInterpreter.
+type makeAttrInterpreterWrapper struct {
+	// name is the corresponding name of the custom attribute.
+	name string
+	// interpreter is the fonction producing the interepreter for the given name.
```
jeffallen: function

calypso/csadmin/commands.go
```diff
@@ -241,6 +241,10 @@ var cmds = cli.Commands{
 								Name:  "export, x",
 								Usage: "export the instance id to STDOUT",
 							},
+							cli.StringFlag{
+								Name:  "projectInstID, pid",
+								Usage: "The project instance ID, which contains the metadata for verification (optional). This option is not directly used in the contract, it is only usefull during the verification process in the case you registered a custom makeAttrInterpreter.",
```
jeffallen: useful

go.mod
```diff
@@ -23,7 +23,9 @@ require (
 	github.com/syndtr/goleveldb v1.0.0 // indirect
 	github.com/urfave/cli v1.22.2
 	go.dedis.ch/kyber/v3 v3.0.12
+	go.dedis.ch/kyber/v4 v4.0.0-pre2
```
jeffallen: oops?

PR #2209, opened 2020-02-14
Support for editing elections before they start

jeffallen:
The Open service will edit an election if the election
ID is set in the incoming Election structure and it
has not received any votes yet.

GetBox has been modified to make it possible to dump the
current state of an election.

Added support for this workflow in the evoting-admin tool.


evoting/lib/election_test.go
```diff
@@ -17,3 +17,6 @@ func TestIsCreator(t *testing.T) {
 	assert.True(t, e.IsCreator(0))
 	assert.False(t, e.IsCreator(1))
 }
+
+func TestParse(t *testing.T) {
```
Gilthoniel: Did you forget to write it ?

jeffallen: Decided not to, fixed.

PR #2210, opened 2020-02-17
Adds a short practical summary of ByzCoin

nkcr:
**What this PR does**

*See commit message*


byzcoin/README.md
```diff
@@ -14,6 +14,43 @@ them as seem fit.
...
 use them. It should grow over time as more parts of the system are implemented.
 
+## Short summary of ByzCoin
+
+- ByzCoin is a **distributed ledger**: a database that is
+  distributed across several nodes, and where the authority to add changes
+  to the database is decentralized.
+- To ensure that updates to the database are strictly ordered, ByzCoin
+  uses a **blockchain** called a **Skipchain**: every node has its local copy of the ledger
+  (database). Every change to the ledger is packed into a block that is
```
ineiti: The novelties of skipchains are the following:
- delegation of trust from one block to the next: as the roster is stored in the block, the current nodes delegate trust to the next set of nodes
- finality using forwardlinks: where other blockchains use proof-of-work and longest chain to define the latest state of the system, skipchains never change and don't allow for split views (except twice in the 9cc3 skipchain, due to an error ;)
- short proofs: using only the hash of the genesis block (9cc3...), every node can return a proof that the latest block is valid by sending only the forward-links. In other systems you need to have your own trusted node follow the skipchain so you can be sure that the latest block is valid

and one thing that is not implemented, but describe in `Chainiac`:
- using multiple skipchains, where one holds the definition of the roster, while the other skipchain holds the actual transactions.

byzcoin/README.md
```diff
@@ -14,6 +14,43 @@ them as seem fit.
...
+  cryptographically chained to the previous one, and forward links in
+  the chain represent consensus among the verifiers that these changes
+  belong in the database.
+- In ByzCoin, the data are organised as instances of **Smart
```
ineiti:
```suggestion
- In ByzCoin, the data is organised as instances of **Smart
```

byzcoin/README.md
```diff
@@ -14,6 +14,43 @@ them as seem fit.
...
+  Contracts**, where a smart contract can be seen as a class and an
+  instance an object of this class. In ByzCoin, one can *Spawn* a new
```
ineiti:
```suggestion
  instance as an object of this class. In ByzCoin, one can *Spawn* a new
```

byzcoin/README.md
```diff
@@ -14,6 +14,43 @@ them as seem fit.
...
+  instance of a contract, *Invoke* a command (or method) on an existing
+  instance, or *Delete* an instance. *Spawn*, *Invoke*, and *Delete*
+  represent all the possible actions a user can do to update the ledger.
+- Every request to update the ledger is a transaction made up of one
+  or more instructions. The transactionn is sent to one of the nodes.
```
ineiti:
```suggestion
  or more instructions. The transaction is sent to one of the nodes.
```

byzcoin/README.md
```diff
@@ -14,6 +14,43 @@ them as seem fit.
...
+  All instructions in the transaction must be approved by a quorum of
+  the nodes, or else the entire transaction is refused.
+- As an example, we can have a simple *Project* smart contract that has
```
ineiti: I would not put the example in the `Short summary`, but rather in it's own `## Example`

byzcoin/README.md
```diff
@@ -14,6 +14,43 @@ them as seem fit.
...
+  the status's change is eventually written in a block, allowing anyone to
+  track and witness the evolution of the `state` of the *Project* in the ledger.
+  If the smart contract enforces a rule that `status: pending`
+  must always be followed by `status: active`, then the propsed transaction
+  will be refused by nodes which are honestly implementing the smart contract.
+  Even if a minority of the nodes are dishonestly allowing the transition
+  directly to `state: done`, they will not form a quorum, and the incorrect
+  state change cannot be introduced into the database.
+
 ## Overview
```
ineiti: And we should update the overview. I know of at least one error in the figure, and some things might be outdated a bit.

byzcoin/README.md
nkcr: that's great, I'm adding it to the `skipchain/README`

byzcoin/README.md
nkcr: that's a debate, but it seems there is nothing wrong with this plural form of "data".

byzcoin/README.md
nkcr: We should then provide a more complete example. To keep it simple I prefer if we keep it there.

byzcoin/README.md
nkcr: that would be awesome! Do you want to suggest an update of the overview that we include in this PR?

byzcoin/README.md
ineiti: I was not aware that there is a discussion around this - https://www.theguardian.com/news/datablog/2010/jul/16/data-plural-singular

According to that article you're more of a traditionalist, though ;)

ineiti:
Here you go - only corrected the things I could easily find. Also included @cgrigis latest changes to `Instruction.Delete` which includes `Arguments` now.

[ByzCoin.svg.zip](https://github.com/dedis/cothority/files/4244305/ByzCoin.svg.zip)

ineiti:
Here you go (this replaces the one from my previous comment):

[ByzCoin-2.svg.zip](https://github.com/dedis/cothority/files/4244326/ByzCoin-2.svg.zip)

ineiti:
I wonder whether we should also add a paragraph about DARCs for the access control?

PR #2211, opened 2020-02-17
Observable ByzCoin

ineiti:
Bringing the wonderful world of rxjs to cothority/byzcoin:
- connection:sendString - uses Observable instead of callbacks
- ByzCoinRPC.getNewBlocks - returns a BehaviorSubject
- ByzCoinRPC.proofObservable - returns a BehaviorSubject of a proof

**It breaks API compatibility for the new WebSocketAdapter.sendStream**

If that is a problem, it's quite straightforward to do the following:
- change `sendStream` to `sendStreamObservable`
- add `sendStream` that behaves like before, but calls `sendStreamObservable`

Readmes:
https://rxjs.dev/guide/observable
https://gist.github.com/staltz/868e7e9bc2a7b8c1f754

### DIFF
--- a/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
+++ b/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
@@ -68,7 +68,7 @@ describe("ByzCoinRPC Tests", () => {
             history.push(`${d.version}-${d.description.toString()}`);
         });
         expect(history[0]).toBe("0-initial");
-
+// Evolving the Darc we observe, we expect the history to be updated
         const di = await DarcInstance.fromByzcoin(rpc, darc.getBaseID());
         const newDI = new Darc({
             ...darc.evolve(),
--- a/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
+++ b/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
@@ -61,7 +61,7 @@ describe("ByzCoinRPC Tests", () => {
         const darc = ByzCoinRPC.makeGenesisDarc([SIGNER], roster, "initial");
         const rpc = await ByzCoinRPC.newByzCoinRPC(roster, darc, BLOCK_INTERVAL, cache);
 
-        const instUpdate = await rpc.proofObservable(darc.getBaseID());
+        const instUpdate = await rpc.instanceObservable(darc.getBaseID());
         const history: string[] = [];
         instUpdate.subscribe((inst) => {
             const d = Darc.decode(inst.value);
@@ -93,7 +93,7 @@ describe("ByzCoinRPC Tests", () => {
             await wait100ms();
         }
 // Getting a new proofObservable on the previously updated Darc should return the latest version of the Darc
-        const latestProof = (await rpc.proofObservable(darc.getBaseID())).getValue();
+        const latestProof = (await rpc.instanceObservable(darc.getBaseID())).getValue();
         expect(latestProof.stateChangeBody.version.equals(Long.fromNumber(1)))
             .toBeTruthy();
 
--- a/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
+++ b/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
@@ -92,7 +92,7 @@ describe("ByzCoinRPC Tests", () => {
             expect(history.length).toBe(2);
             await wait100ms();
         }
-
+// Getting a new proofObservable on the previously updated Darc should return the latest version of the Darc
         const latestProof = (await rpc.proofObservable(darc.getBaseID())).getValue();
         expect(latestProof.stateChangeBody.version.equals(Long.fromNumber(1)))
             .toBeTruthy();
--- a/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
+++ b/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
@@ -1,6 +1,10 @@
+import Long from "long";
+import { Log } from "../../src";
 import { ByzCoinRPC } from "../../src/byzcoin";
+import { LocalCache } from "../../src/byzcoin/byzcoin-rpc";
 import DarcInstance from "../../src/byzcoin/contracts/darc-instance";
 import Instance from "../../src/byzcoin/instance";
+import { Darc } from "../../src/darc";
 import { BLOCK_INTERVAL, ROSTER, SIGNER, startConodes } from "../support/conondes";
 
 describe("ByzCoinRPC Tests", () => {
@@ -51,4 +55,52 @@ describe("ByzCoinRPC Tests", () => {
             .toBeRejectedWith(new Error("key not in proof: 010203"));
         await expectAsync(DarcInstance.fromByzcoin(rpc, Buffer.alloc(32, 0))).toBeRejected();
     });
+
+    it("should get updated when an instance is updated", async () => {
+        const cache = new LocalCache();
+        const darc = ByzCoinRPC.makeGenesisDarc([SIGNER], roster, "initial");
+        const rpc = await ByzCoinRPC.newByzCoinRPC(roster, darc, BLOCK_INTERVAL, cache);
+
+        const instUpdate = await rpc.proofObservable(darc.getBaseID());
+        const history: string[] = [];
+        instUpdate.subscribe((inst) => {
+            const d = Darc.decode(inst.value);
+            history.push(`${d.version}-${d.description.toString()}`);
+        });
+        expect(history[0]).toBe("0-initial");
+
+        const di = await DarcInstance.fromByzcoin(rpc, darc.getBaseID());
+        const newDI = new Darc({
+            ...darc.evolve(),
+            description: Buffer.from("new"),
+        });
+        await di.evolveDarcAndWait(newDI, [SIGNER], 10);
+
+        for (let i = 0; i < 5; i++) {
+            if (history.length === 2) {
+                break;
+            }
+            await wait100ms();
+        }
+        expect(history.length).toBe(2);
+        expect(history[1]).toBe("1-new");
+
+        const newDarc = Darc.createBasic([SIGNER], [SIGNER],
+            Buffer.from("darc 2"));
+        await di.spawnDarcAndWait(newDarc, [SIGNER], 2);
+        for (let i = 0; i < 5; i++) {
+            expect(history.length).toBe(2);
+            await wait100ms();
+        }
+
+        const latestProof = (await rpc.proofObservable(darc.getBaseID())).getValue();
+        expect(latestProof.stateChangeBody.version.equals(Long.fromNumber(1)))
+            .toBeTruthy();
+
+        rpc.closeNewBlocks();
+    });
 });
+
+async function wait100ms(): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, 100));
+}
--- a/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
+++ b/external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
@@ -84,7 +84,7 @@ describe("ByzCoinRPC Tests", () => {
         }
         expect(history.length).toBe(2);
         expect(history[1]).toBe("1-new");
-
+// Creating a new Darc, this should not update the history
         const newDarc = Darc.createBasic([SIGNER], [SIGNER],
             Buffer.from("darc 2"));
         await di.spawnDarcAndWait(newDarc, [SIGNER], 2);
--- a/external/js/cothority/spec/byzcoin/stream.spec.ts
+++ b/external/js/cothority/spec/byzcoin/stream.spec.ts
@@ -1,8 +1,7 @@
+import { Log } from "../../src";
 import ByzCoinRPC from "../../src/byzcoin/byzcoin-rpc";
 import { PaginateRequest, PaginateResponse } from "../../src/byzcoin/proto/stream";
-import { WebSocketAdapter } from "../../src/network";
 import { WebSocketConnection } from "../../src/network/connection";
-import { SkipchainRPC } from "../../src/skipchain";
 import { BLOCK_INTERVAL, ROSTER, SIGNER, startConodes } from "../support/conondes";
 
 describe("Stream Tests", () => {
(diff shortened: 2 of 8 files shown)


external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -168,6 +193,52 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
+            dbProof = await this.getProofFromLatest(id);
+        }
+        if (!dbProof.exists(id)) {
+            throw new Error("this instance does not exist");
+        }
+
+        const bsNew = new BehaviorSubject(dbProof);
+        // Set up a pipe from the block to fetch new versions if a new block
+        // is available.
+        (await this.getNewBlocks())
```
nkcr: Please help humans understand this: explaining what each operator do and why you need them would help.

external/js/cothority/src/protobuf/marshaller.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    async getObject<T>(key: string): Promise<T | undefined> {
+        const buf = await this.get(key);
+        if (buf === undefined) {
+            return undefined;
+        }
+        return JSON.parse(buf.toString(), (_, value) => {
+            if (value && typeof value === "object") {
+                if (value.type && value.type === "Buffer") {
```
tharvik: Is `value.type` always existing? I don't see it in the [nodejs/Buffer documentation](https://nodejs.org/api/buffer.html)
If so, why not use it for `Long` also? Else, better use `"type" in value`.

Or if it is wanted to be a falsy value, same idea as before.

external/js/cothority/src/protobuf/marshaller.ts
```diff
@@ -0,0 +1,63 @@
...
+                    return Buffer.from(value);
+                } else if (value.low !== undefined && value.high !== undefined) {
```
tharvik: Same remark as before.

external/js/cothority/src/network/connection.ts
```diff
@@ -191,67 +188,66 @@ export class WebSocketConnection implements IConnection {
...
+    sendStream<T extends Message>(message: Message, reply: typeof Message):
+        Observable<[T, WebSocketAdapter]> {
+        return new Observable((sub) => {
 
-        if (!message.$type) {
-            onError(new Error(`message "${message.constructor.name}" is not registered`));
-            return;
-        }
+            if (!message.$type) {
+                sub.error(new Error(`message "${message.constructor.name}" is not registered`));
```
tharvik: I would move it outside of the subscription, as it's an argument checking error, not a transient one.

external/js/cothority/src/network/connection.ts
```diff
@@ -191,67 +188,66 @@ export class WebSocketConnection implements IConnection {
...
-            } else {
-                onClose(code, reason);
-            }
-        });
+            ws.onError((err: Error) => {
+                clearTimeout(timer);
 
-        ws.onError((err: Error) => {
-            clearTimeout(timer);
+                Log.print("error is", err);
```
tharvik: Debug code?

external/js/cothority/src/network/connection.ts
```diff
@@ -191,67 +188,66 @@ export class WebSocketConnection implements IConnection {
...
 
-        if (!reply.$type) {
-            onError(new Error(`message "${reply}" is not registered`));
-            return;
-        }
+            const path = this.getURL() + "/" + this.service + "/" + message.$type.name.replace(/.*\./, "");
+            Log.lvl4(`Socket: new WebSocket(${path})`);
+            const ws = factory(path);
+            const bytes = Buffer.from(message.$type.encode(message).finish());
+            const timer = setTimeout(() => ws.close(1000, "timeout"), this.timeout);
```
tharvik: It's the same preamble as `send`, consider putting it inside an helper. (it took so much stressfull minutes to try to wrap my head around the fact that my commit about URL was merged in master, but this code was not using it; basic humain mistake of matching `send` and `sendStream` but yeah, code duplication)

external/js/cothority/src/network/connection.ts
```diff
@@ -191,67 +188,66 @@ export class WebSocketConnection implements IConnection {
...
-        const ws = factory(path);
-        const bytes = Buffer.from(message.$type.encode(message).finish());
-        const timer = setTimeout(() => ws.close(1000, "timeout"), this.timeout);
-        ws.onOpen(() => {
-            Log.lvl3("Sending message to", path);
-            ws.send(bytes);
-        });
+            ws.onMessage((data: Buffer) => {
+                clearTimeout(timer);
+                const buf = Buffer.from(data);
```
tharvik: Hum, `data` is already of type `Buffer`, why `Buffer.from` it?

external/js/cothority/src/network/connection.ts
```diff
@@ -191,67 +188,66 @@ export class WebSocketConnection implements IConnection {
...
+
+                try {
+                    const ret = reply.decode(buf) as T;
+                    sub.next([ret, ws]);
+                } catch (err) {
+                    if (err instanceof util.ProtocolError) {
+                        sub.error(err);
+                    } else {
+                        sub.error(
+                            new Error(`Other error: ${err}`),
```
tharvik: That would look weird when arising; why not simply `sub.error` without checking for instance?

external/js/cothority/src/network/connection.ts
```diff
@@ -191,67 +188,66 @@ export class WebSocketConnection implements IConnection {
...
-            try {
-                const ret = reply.decode(buf) as T;
-                onMessage(ret, ws);
-            } catch (err) {
-                if (err instanceof util.ProtocolError) {
-                    onError(err);
+            ws.onClose((code: number, reason: string) => {
+                // nativescript-websocket on iOS doesn't return error-code 1002 in case of error, but sets the 'reason'
+                // to non-null in case of error.
+                if (code !== 1000 || reason) {
```
tharvik:
```suggestion
                if (code !== 1000 || reason !== "") {
```

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
ineiti: Added comments - see also the description of the PR itself.

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -386,11 +457,30 @@ export default class ByzCoinRPC implements ICounterUpdater {
         return reply.actions;
     }
 
+    /**
+     * The returned BehaviorSubject replays all new blocks to all listeners.
+     * The streaming is attached to the first node of the connectionlist.
+     */
+    async getNewBlocks(): Promise<BehaviorSubject<SkipBlock>> {
+        if (this.newBlock === undefined) {
```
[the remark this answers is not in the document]
ineiti: no war here - sometimes you're just right...

external/js/cothority/src/protobuf/marshaller.ts
```diff
@@ -0,0 +1,63 @@
...
+        return this.storage.set(key, value);
+    }
+
+    async getObject<T>(key: string): Promise<T | undefined> {
+        const buf = await this.get(key);
+        if (buf === undefined) {
+            return undefined;
+        }
+        return JSON.parse(buf.toString(), (_, value) => {
+            if (value && typeof value === "object") {
```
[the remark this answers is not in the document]
ineiti: deleted code.

external/js/cothority/src/protobuf/marshaller.ts
ineiti: deleted code - that's what you get when you go too deep in a WIP PR...

external/js/cothority/src/protobuf/marshaller.ts
ineiti: same here...

external/js/cothority/src/network/connection.ts
ineiti: removed

external/js/cothority/src/network/connection.ts
ineiti: I think I remember that one. I'm not sure that all the `nativescript` and other websocket implementations always return a Buffer here. I prefer keeping it here...

external/js/cothority/src/network/connection.ts
ineiti: don't look at old code ;)

external/js/cothority/src/network/connection.ts
ineiti: again, I'm not sure the different websocket implementations actually handle it like that.

external/js/cothority/src/network/connection.ts
ineiti: good idea.

external/js/cothority/src/network/connection.ts
ineiti: In fact it's the original codebase from `send`...

nkcr: yep, we should ask @Gilthoniel

external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
```diff
@@ -51,4 +55,52 @@ describe("ByzCoinRPC Tests", () => {
...
+        const rpc = await ByzCoinRPC.newByzCoinRPC(roster, darc, BLOCK_INTERVAL, cache);
+
+        const instUpdate = await rpc.proofObservable(darc.getBaseID());
+        const history: string[] = [];
+        instUpdate.subscribe((inst) => {
+            const d = Darc.decode(inst.value);
+            history.push(`${d.version}-${d.description.toString()}`);
+        });
+        expect(history[0]).toBe("0-initial");
+
```
nkcr:
```suggestion
// Evolving the Darc we observe, we expect the history to be updated
```

external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
```diff
@@ -51,4 +55,52 @@ describe("ByzCoinRPC Tests", () => {
...
+        for (let i = 0; i < 5; i++) {
+            if (history.length === 2) {
+                break;
+            }
+            await wait100ms();
+        }
+        expect(history.length).toBe(2);
+        expect(history[1]).toBe("1-new");
+
```
nkcr:
```suggestion
// Creating a new Darc, this should not update the history
```

external/js/cothority/spec/byzcoin/byzcoin-rpc.spec.ts
```diff
@@ -51,4 +55,52 @@ describe("ByzCoinRPC Tests", () => {
...
+        const newDarc = Darc.createBasic([SIGNER], [SIGNER],
+            Buffer.from("darc 2"));
+        await di.spawnDarcAndWait(newDarc, [SIGNER], 2);
+        for (let i = 0; i < 5; i++) {
+            expect(history.length).toBe(2);
+            await wait100ms();
+        }
+
```
nkcr:
```suggestion
// Getting a new proofObservable on the previously updated Darc should return the latest version of the Darc
```

external/js/cothority/spec/byzcoin/stream.spec.ts
```diff
@@ -23,16 +22,15 @@ describe("Stream Tests", () => {
 
         const msg = new PaginateRequest({startid: rpc.genesisID, pagesize: 1, numpages: 1,  backward: false});
 
-        const foo = {
-            onClose: (code: number, reason: string) => {
-                expect(code).toEqual(1000);
```
nkcr: Ok so we loose access to the close code and reason. Hope we won't need it.

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -168,6 +188,70 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
+     * It returns a BehaviorSubject, which is a special observable that
+     * keeps the current value.
+     * It also caches all requests, so if there are two requests for the
+     * same id, no new observable will be created.
+     * It connects to `getNewBlocks` to be informed whenever a new block is
+     * created.
+     * @param id of the instance to return
+     * @throws an error if the instance does not exist
+     */
+    async proofObservable(id: InstanceID): Promise<BehaviorSubject<Proof>> {
```
nkcr: Just a reflexion: "proofObservable" is a bit abstract and when you think about it, what the user really wants to observe is an instance, not the proof itself right? Also when I think about a "proof", in my mind that refers to something fixed on the blockchain, and not something that evolves and gets updated. Would changing this to "instanceObservale" make sense?

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
```diff
@@ -168,6 +188,70 @@ export default class ByzCoinRPC implements ICounterUpdater {
...
+        // Set up a pipe from the block to fetch new versions if a new block
+        // arrives.
+        // Start with an observable that emits each new block as it arrives.
+        (await this.getNewBlocks())
+            .pipe(
+                // Make sure only newer blocks than the proof are taken into
+                // account
+                filter((block) => block.index > dbProof.latest.index),
+                // Get a new proof of the instance
+                mergeMap(() => this.getProofFromLatest(id)),
```
nkcr: Instead of calling the `getProofFromLatest`, why not checking directly if the block contains an instruction about the instanceID we observe?

external/js/cothority/spec/byzcoin/stream.spec.ts
ineiti: if the `code` and `reason` of the close is something that points to a non-error, it will resolve to `error`. So we have to make sure that the test for 'non-error' is correct.

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
ineiti: absolutely - naming is hard. It started out as `instanceObservable`, in fact. But then, as it returned a proof, I renamed it to `proofObservable`. And `instanceObservableByReturningProofs` is a bit long ;)

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
ineiti: Yes, I thought about that, but it will not work everytime. If you do an `Instruction` to a `Coin` with the command `transfer`, two instances will change: the one in `InstanceID`, and the one in the argument `destination`. So it would be quite difficult to know which change.

See also #2216

external/js/cothority/spec/byzcoin/stream.spec.ts
nkcr: "if (...) that points to a non-error, it will resolve to `error`". 

why?

external/js/cothority/src/byzcoin/byzcoin-rpc.ts
nkcr: IMO its easier for a user to understand that "we observe an instance, which returns a proof because that's how we reference instances in Byzcoin", rather than "we observe a proof, because that's the result of observing an instance, therefore we observe an instance".

external/js/cothority/spec/byzcoin/stream.spec.ts
ineiti: https://github.com/dedis/cothority/pull/2211/files#diff-5f68f2cd364b6d32e0a39977416c8cdeR189

Issue #2187, opened 2020-01-29
Link Bevm with ByzCoin


ineiti:
Implement something like:
https://cryptpad.c4dt.org/pad/#/2/pad/view/hcUYF0pKjY8fmcxBwjqNnHvmAAGdCmACn6fZfsdT2CA/

Proposed workflow:
- [x] Convert BEVM-Events to Instructions dedis/cothority#2201
- [x] Add `bevm_state` as darc-identity dedis/cothority#2202

cgrigis:
Closes c4dt/byzcoin#7.

PR #2214, opened 2020-02-19
Add bevm darc attr

### LINKED ISSUE #2202: Add bevm as darc attr expression evaluator
To use BEVM contracts in darcs, add a `bevm_state` identity that takes a contractID and a method. The method is a view-method with the parameters `skipblock_index` and `signature`. The `signature` is passed from the `ClientTransaction.Instructions[x].Signature[y]` where x is the corresponding instruction and y is the corresponding darc-identity.

Use-case:
- allow only update of a valueInstance between 8am and 5pm CEST.

Sub-tasks:
- [x] Pass ByzCoin timestamp to EVM
  - [x] For an EVM transaction, timestamp of the ByzCoin block currently being created
  - [x] For an EVM view method, timestamp of latest ByzCoin block OR current time on the executing machine
- [x] Implement `bevm` attribute expression evaluator

cgrigis:
**What this PR does**

This PR adds an `attr:bevm` DARC expression evaluator that performs the verification by delegating to an Ethereum contract deployed in ByzCoin through BEvm. See #2187 and #2202 for more details regarding the motivation and the design.

One notable use case for this attribute is checking against some time information. The EVM allows access to the current block timestamp using the [`now`](https://solidity.readthedocs.io/en/v0.6.3/units-and-global-variables.html#block-and-transaction-properties) special variable. In order to pass this information to the EVM, the `globalState` structure was extended to hold the timestamp and some changes were made in the `creataeNewBlock()` byzcoin method to fill it in before executing instructions (see first commit).

Furthermore, in order to allow contracts using this attribute to pass information to the validation method through arguments, the `Delete` instruction was extended with `Arguments`, making it symmetrical with `Spawn` and `Invoke` (see second commit).

To test the attribute, a simple value contract is used, along with a simple Solidity function that verifies that the new value provided for an update is greater than the current value (not that it is a real use case ;-) ).

Closes #2202 

---

🙅‍ Friendly checklist:

- [X] 0. Code comments are added (or updated) when/where needed and explain the WHY of the code.
- [X] 1. Design choices, user documentation and any additional doc are added (or updated) in READMEs.
- [X] 2. Any new behaviour is tested and small units of code that can be are unit tested.
- [X] 3. Code comments are added on tests to explain what they do.
- [X] 4. Errors are systematically wrapped with a meaningful message using `xerrors.Errorf` and the `%v` verb.
- [X] 5. Hard limit of 80 chars is always respected.
- [X] 6. Changes are backward compatible.
- [X] 7. Indentation level does not exceed 5, although 4 is already suspicious.
- [X] 8. Functions, files, and packages are kept to a manageable size and decomposed into smaller units if needed.
- [X] 9. There are no magic values.


cgrigis:
Closes #2202

byzcoin/service.go
```diff
@@ -2172,17 +2178,22 @@ func (s *Service) addError(tx ClientTransaction, err error) {
...
-	scID skipchain.SkipBlockID) (StateChanges, *stagingStateTrie, error) {
+	scID skipchain.SkipBlockID, timestamp int64) (StateChanges, *stagingStateTrie, error) {
 
 	// Make a new trie for each instruction. If the instruction is
 	// sucessfully implemented and changes applied, then keep it
 	// otherwise dump it.
 	sst = sst.Clone()
+
+	// convert ReadOnlyStateTrie to a GlobalState so that contracts may cast it if they wish
+	roSC := newROSkipChain(s.skService(), scID)
```
nkcr: did you move it there from `executeInstruction` for performance sake?

bevm/bevm_attr.go
```diff
@@ -0,0 +1,267 @@
+package bevm
```
nkcr: I must say: this code is really well formatted, a pleasure to read!! :)

byzcoin/api.go
```diff
@@ -713,6 +713,16 @@ searchLatest:
 	return xerrors.New("didn't get the same blocks from everybody within 10 seconds")
 }
 
+// GetLatestTimestamp returns the timestamp of the latest known block
+func (c *Client) GetLatestTimestamp() (int64, error) {
+	header, err := decodeBlockHeader(c.getLatestKnownBlock())
+	if err != nil {
+		return 0, xerrors.Errorf("Failed to decode skipchain header: %v", err)
```
nkcr:
```suggestion
		return 0, xerrors.Errorf("failed to decode skipchain header: %v", err)
```

byzcoin/service.go
```diff
@@ -2293,17 +2304,13 @@ func (s *Service) GetContractInstance(contractName string, in []byte) (Contract,
...
 		}
 	}()
 
-	// convert ReadOnlyStateTrie to a GlobalState so that contracts may cast it if they wish
-	roSC := newROSkipChain(s.skService(), scID)
-	gs := globalState{st, roSC}
-
 	contents, _, contractID, _, err := gs.GetValues(instr.InstanceID.Slice())
 	if !xerrors.Is(err, errKeyNotSet) && err != nil {
 		err = xerrors.Errorf("Couldn't get contract type of instruction: %v", err)
```
nkcr:
```suggestion
		err = xerrors.Errorf("couldn't get contract type of instruction: %v", err)
```

bevm/bevm_contract_test.go
```diff
@@ -195,11 +195,33 @@ func Test_Time(t *testing.T) {
...
+	// Retrieve stored time
+	storedTime, err := bevmClient.Call(a, timeTestInstance, "getStoredTime")
+	require.Nil(t, err)
+	// Unitialized -- 0
+	assertBigInt0(t, storedTime.(*big.Int))
+
+	// Store current time
+	err = bevmClient.Transaction(txParams.GasLimit, txParams.GasPrice, 0, a,
+		timeTestInstance, "storeCurrentTime")
+	require.Nil(t, err)
```
nkcr:
```suggestion
	require.NoError(t, err)
```
for consistency that would require to update all the other ones in this file. Could be done in another PR.

nkcr: in fact #2213 already updates all the other ones. So you should make sure the new ones actually use `require.NoError`

byzcoin/api.go
```diff
@@ -713,6 +713,16 @@ searchLatest:
 	return xerrors.New("didn't get the same blocks from everybody within 10 seconds")
 }
 
+// GetLatestTimestamp returns the timestamp of the latest known block
+func (c *Client) GetLatestTimestamp() (int64, error) {
+	header, err := decodeBlockHeader(c.getLatestKnownBlock())
```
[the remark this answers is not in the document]
ineiti: Hmm - we discussed this with @cgrigis , and I thought we'd take `time.Now` in the client - but I might have forgotten a good argument against this.

byzcoin/proto.go
```diff
@@ -240,6 +240,8 @@ type Invoke struct {
 type Delete struct {
 	// ContractID represents the kind of contract that is being deleted.
 	ContractID string
+	// Args holds all data necessary to delete the instance.
+	Args Arguments
```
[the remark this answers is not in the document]
ineiti: I originally had them less differentiated than now. But this was one of Bryan's requests to have the structure more clearly showing which type of action you're doing.

ineiti:
Needs to have an entry in a `README.md`

byzcoin/api.go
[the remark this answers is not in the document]
cgrigis: When we discussed it with @ineiti, we mentioned the two possibilities, either taking the current time or the time of the latest block, and to examine how Ethereum did it.
Playing with [Remix](https://remix.ethereum.org/), it seems `now` returns the current machine time when using a JS VM, and the time of the last block when deployed on the Ropsten test chain.
I don't have a preference, so if you think using the current time is better I will update. :)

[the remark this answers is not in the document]
nkcr: I don't have any preference, my concern is about using `c.getLatestKnownBlock()` that uses `c.Latest` that corresponds to `c.Genesis` in the case no other blocks have been fetched by the clients, which can be dangerous. But if you think that's not a problem then let's keep it like that.

byzcoin/service.go
cgrigis: Not really, just because it looked like it could be done earlier, and also it would remove the need to pass the `scID` parameter (which I forgot to do in the commit :grimacing:, coming up).

bevm/bevm_attr.go
cgrigis: Thanks a lot! :blush:

bevm/bevm_contract_test.go
cgrigis: Well spotted! :)
It was my intention to make the new ones correct, but I visibly missed some :unamused: .

byzcoin/api.go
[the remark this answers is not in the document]
cgrigis: You raise a good point. To avoid this, as well as @ineiti's point that ByzCoin has no guarantee that new blocks are created within a specific timeframe, I'll change it to take the machine's current time. It's not the exact same semantic as Ethereum, but it probably does not matter.

PR #2217, opened 2020-02-24
Fixes an unwanted limitation on URLs

nkcr:
Fixes a limitation introduced by #2196

Right now its impossible to host a conode at an adress that contains a path.

**What this PR does**

*See commit message*

### DIFF
--- a/external/js/cothority/src/network/connection.ts
+++ b/external/js/cothority/src/network/connection.ts
@@ -84,6 +84,12 @@ export class WebSocketConnection implements IConnection {
         } else {
             url = addr;
         }
+        // We want any pathname to contain a "/" at the end. This is motivated
+        // by the fact that URL will not allow you to have an empty pathname,
+        // which will always equal to "/" if there isn't any
+        if (url.pathname.slice(-1) !== "/") {
+            url.pathname = url.pathname + "/";
+        }
         if (url.username !== "" || url.password !== "") {
             throw new Error("addr contains authentication, which is not supported");
         }
@@ -153,11 +159,7 @@ export class WebSocketConnection implements IConnection {
 
         return new Observable((sub) => {
             const url = new URL(this.url.href);
-            let prefix = "";
-            if (url.pathname !== "/") {
-                prefix = url.pathname;
-            }
-            url.pathname =  `${prefix}/${this.service}/${message.$type.name.replace(/.*\./, "")}`;
+            url.pathname =  `${url.pathname}${this.service}/${message.$type.name.replace(/.*\./, "")}`;
             Log.lvl4(`Socket: new WebSocket(${url.href})`);
             const ws = factory(url.href);
             const bytes = Buffer.from(message.$type.encode(message).finish());
--- a/external/js/cothority/src/network/connection.ts
+++ b/external/js/cothority/src/network/connection.ts
@@ -110,12 +110,9 @@ export class WebSocketConnection implements IConnection {
 
     /** @inheritdoc */
     getURL(): string {
-        // Retro compatibility: in the case there is no pathname we don't want
-        // the extra "/", thus using url.origin
-        if (this.url.pathname === "/") {
-            return this.url.origin;
-        }
-        return this.url.href;
+        // Retro compatibility: this.url always ends with a slash, but the old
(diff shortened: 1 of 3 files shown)


external/js/cothority/src/network/connection.ts
```diff
@@ -153,7 +153,11 @@ export class WebSocketConnection implements IConnection {
 
         return new Observable((sub) => {
             const url = new URL(this.url.href);
-            url.pathname = `/${this.service}/${message.$type.name.replace(/.*\./, "")}`;
+            let prefix = "";
+            if (url.pathname !== "/") {
+                prefix = url.pathname;
+            }
+            url.pathname =  `${prefix}/${this.service}/${message.$type.name.replace(/.*\./, "")}`;
```
tharvik: Hum, you can use `url.pathname +=` instead.

nkcr: mmm not quite. If there is no pathname, then pathname equals "/", which give `//serviceName...` but its's correct if pathname equals something like "/conode".

tharvik: Ho, you're right, I thought that an URL couldn't contain empty path element

external/js/cothority/src/network/connection.ts
```diff
@@ -104,7 +110,7 @@ export class WebSocketConnection implements IConnection {
 
     /** @inheritdoc */
     getURL(): string {
-        return this.url.origin;
+        return this.url.href;
```
tharvik: It's breaking retrocompat; but if it's wanted, we can also directly return a copy of `this.url` :P

nkcr: gosh, I don't wanna break anything, just restoring the same behaviour as before #2196 ... 😰

tharvik: Yeah, so to have the old behavior would be to explicitly check for a trailing slash and removing it when existing, in getURL and send IMO.

nkcr: yep, that's what I did 
```
        if (this.url.pathname === "/") {
            return this.url.origin;
        }
        return this.url.href;
```

tharvik: Ho, good then, I didn't refresh the page with the latest commit ;)

external/js/cothority/src/network/connection.ts
```diff
@@ -153,7 +164,7 @@ export class WebSocketConnection implements IConnection {
 
         return new Observable((sub) => {
             const url = new URL(this.url.href);
-            url.pathname = `/${this.service}/${message.$type.name.replace(/.*\./, "")}`;
+            url.pathname =  `${url.pathname}${this.service}/${message.$type.name.replace(/.*\./, "")}`;
```
tharvik: Now we can use it as it always ends with a slash :)

```suggestion
            url.pathname += `${this.service}/${message.$type.name.replace(/.*\./, "")}`;
```

external/js/cothority/src/skipchain/skipchain-rpc.ts
```diff
@@ -181,7 +181,7 @@ export default class SkipchainRPC {
             const last = newBlocks[newBlocks.length - 1];
             let isInRoster = false;
             for (const n of last.roster.list) {
-                if (n.getWebSocketAddress() === this.conn.getURL()) {
+                if (new URL(n.getWebSocketAddress()).href === new URL(this.conn.getURL()).href) {
```
tharvik: Why is it needed? The `URL -> string` translation is already done by `getURL` and there should not be retrocompat issue, so, that would be equivalent, no?

nkcr: true, was needed with a previous version.

tharvik:
I've pushed some commits, which should do the trick
 * explicit test for an URL with a pathname
 * found a test polluting the global namespace, fixed it
 * updating `WSConn.getURL` to have the old behavior
Can you review it please @nkcr? For me, it looks good, but yeah, never too sure :)

nkcr:
Ok looks good, thanks for the test. If you approve it I can then finally merge it.

PR #2225, opened 2020-03-17
byzcoin: move C4DT node to 443

tharvik:
Following the recent update of the Roster, bumping the group's config.

Also, bumping the README of bcamin because of the decommissioning of the HTTP server on conde.c4dt.org.


ineiti:
should be on 7770 and 443.

PR #2226, opened 2020-03-17
external/cothority: run tests before publish

tharvik:
Add a `npm test` pass before publishing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,7 @@ jobs:
       script:
         - cd external/js/kyber
         - npm ci
-        - npm run test
+        - npm test
     - name: "js > cothority"
       <<: *language_js
       install: *get_go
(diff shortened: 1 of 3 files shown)


ineiti:
Why is `npm test` in `publish.sh` and not in `travis.yml`?

tharvik:
It's useful when publishing a new stable release of a js library, to ensure that what is packed is correctly passing tests.

ineiti:
This time it's me that I don't like having the tests in a script called `publish.sh`, because it's not what I'd expect it to do...

Can you please move the tests to the appropriate part of `travis.yml`? Or create a `package.json` target `publish` that runs the tests and then runs `publish.sh`.

tharvik:
You're right, that makes more sense; updated.

PR #2227, opened 2020-03-19
Use coinValue in spawner contract

ineiti:
Whenever a spawner contract creates a new coin instance, it fills it with all available coins.
This works fine if you only create one coin in a transaction. But if multiple instructions
with multiple spawners are created in the same transaction, it will fail.

This PR adds a new argument when spawning a coin that allows filling of the coin with only
a partial available amount.

@Gilthoniel there is a _real_ unit tests in this PR - it is possible ;)

### DIFF
--- a/byzcoin/struct.go
+++ b/byzcoin/struct.go
@@ -562,6 +562,20 @@ func (c *Coin) SafeSub(a uint64) error {
 	return xerrors.New("uint64 underflow")
 }
 
+// SafeTransfer takes val from one coin and puts it to another.
+// It checks that the source coin has enough,
+// and that the destination coin will not overflow.
+func (c *Coin) SafeTransfer(to *Coin, val uint64) error {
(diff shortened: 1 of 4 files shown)


nkcr:
> @Gilthoniel there is a real unit tests in this PR - it is possible ;)

`newRstSimul()`. This would have changed my life, if only someone did it in cothority_template or byzcoin/contract to show a good example to a newcomer, I would have used it 😞

ineiti:
> `newRstSimul()`. This would have changed my life, if only someone did it in cothority_template or byzcoin/contract to show a good example to a newcomer, I would have used it 😞

It only came with my latest changes to personhood. I think part of the 'integration testing' was due to the fact that I was not sure how the contracts would need to be used. So it was the only way to be sure that the clients will actually be able to use the contracts.

Now that we have some experience, it makes sense to actually do unit testing... Anyway - that's a good idea for cothority.v4 ;)

PR #2228, opened 2020-03-19
Speed up proofs

ineiti:
In the latest dynacred2, there can be a lot of proofs requested when a new block
comes in. The previous code checked the forward-links for every proof, which
made it very slow.

This PR stores the latest trusted block, and doesn't check the forward-links
if a proof with that latest block is received.

Also changes new darcs with `version=0` to have a random `prevID`, enabling
to store a darc with the same rules and description twice under a different
ID.

Also extends the `Credential.setAttribute` method to allow for strings in the value.

### DIFF
--- a/external/js/cothority/src/byzcoin/proof.ts
+++ b/external/js/cothority/src/byzcoin/proof.ts
@@ -78,6 +78,8 @@ export default class Proof extends Message<Proof> {
     static register() {
         registerMessage("byzcoin.Proof", Proof, InclusionProof, SkipBlock, ForwardLink);
     }
+    // The id of the latest known proof - supposes there is only one ByzCoin
+    private static latestID: InstanceID;
 
     readonly inclusionproof: InclusionProof;
     readonly latest: SkipBlock;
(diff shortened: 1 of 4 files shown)


external/js/cothority/src/byzcoin/proof.ts
```diff
@@ -154,26 +156,32 @@ export default class Proof extends Message<Proof> {
...
-        // Check that all forward-links are correct.
         let prev = links[0].to;
-        for (let i = 1; i < links.length; i++) {
-            const link = links[i];
 
-            const err = link.verifyWithScheme(publics, this.latest.signatureScheme);
-            if (err) {
-                return new Error("invalid forward link signature: " + err.message);
-            }
+        if (Proof.latestID && links[links.length - 1].to.equals(Proof.latestID)) {
```
tharvik:
```suggestion
        if (Proof.latestID !== undefined && links[links.length - 1].to.equals(Proof.latestID)) {
```

external/js/cothority/src/byzcoin/proof.ts
```diff
@@ -78,6 +78,8 @@ export default class Proof extends Message<Proof> {
     static register() {
         registerMessage("byzcoin.Proof", Proof, InclusionProof, SkipBlock, ForwardLink);
     }
+    // The id of the latest known proof - supposes there is only one ByzCoin
+    private static latestID: InstanceID;
```
tharvik:
```suggestion
    private static latestID: InstanceID | undefined;
```

external/js/cothority/src/darc/darc.ts
```diff
@@ -63,7 +64,7 @@ export default class Darc extends Message<Darc> {
         const darc = new Darc({
             baseID: Buffer.from([]),
             description: desc,
-            prevID: createHash("sha256").digest(),
+            prevID: randomBytes(32),
```
tharvik: Why not setting it to undefined?

external/js/cothority/src/personhood/credentials-instance.ts
```diff
@@ -283,7 +281,10 @@ export class CredentialStruct extends Message<CredentialStruct> {
      * @param attribute     Name of the attribute
      * @param value         The value to set
      */
-    setAttribute(credential: string, attribute: string, value?: Buffer) {
+    setAttribute(credential: string, attribute: string, value?: Buffer | string) {
+        if (!(value instanceof Buffer)) {
+            value = Buffer.from(value ? value : "");
```
tharvik:
```suggestion
            value = Buffer.from(value !== undefined ? value : "");
```

Or more readable, have a chain of
```
if (value === undefined)
else if (typeof value === "string")
else
```

tharvik:
Waiting for the lint fixes and then I'll merge it :)

PR #2229, opened 2020-03-19
More exports

ineiti:
Fix some exports, as the `const` construct somehow does not allow to export `types`.

### DIFF
--- a/external/js/kyber/src/curve/nist/point.ts
+++ b/external/js/kyber/src/curve/nist/point.ts
@@ -99,11 +99,13 @@ export default class NistPoint implements Point {
             }
 
             const xRed = x.toRed(this.ref.curve.curve.red);
+            // @ts-ignore
             const aX = xRed.redMul(new BN(this.ref.curve.curve.a));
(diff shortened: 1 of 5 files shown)


external/js/kyber/src/curve/nist/point.ts
```diff
@@ -99,11 +99,13 @@ export default class NistPoint implements Point {
             }
 
             const xRed = x.toRed(this.ref.curve.curve.red);
+            // @ts-ignore
             const aX = xRed.redMul(new BN(this.ref.curve.curve.a));
             // y^2 = x^3 + ax + b
             const y2 = xRed.redSqr()
                 .redMul(xRed)
                 .redAdd(aX)
+                // @ts-ignore
```
tharvik: Do you really need thoses? I've removed it and `npm run build` seems to work.

ineiti: It's mostly if I directly depend on the cothority-sources for development... And my dev-environment is stricter than the cothority-environment.

PR #2234, opened 2020-03-20
roster: update to latest

tharvik:
Bump to latest version of the roster.

### DIFF
--- a/dedis-cothority.toml
+++ b/dedis-cothority.toml
@@ -1,99 +1,97 @@
 [[servers]]
   Address = "tls://conode.dedis.ch:7000"
-  Url = "https://conode.dedis.ch"
   Suite = "Ed25519"
   Public = "ec5c65a3c922d1df32075640e3de606197be24af76059a2ef145501122884bd3"
   Description = "EPFL Cothority-server"
+  URL = "https://conode.dedis.ch"
   [servers.Services]
     [servers.Services.ByzCoin]
       Public = "6f69dc10dbef8f4d80072aa9d1bee191b0f68b137a9d06d006c39fe6667738fa2d3439caf428a1dcb6f4a5bd2ce6ff6f1462ebb1b7374080d95310bc6e1115e105d7ae38f9fed1585094b0cb13dc3a0f3e74daeaa794ca10058e44ef339055510f4d12a7234779f8db2e093dd8a14a03440a7d5a8ef04cac8fd735f20440b589"
       Suite = "bn256.adapter"
-    [servers.Services.PoPServer]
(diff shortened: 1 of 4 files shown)


dedis-cothority.toml
```diff
@@ -1,99 +1,97 @@
 [[servers]]
   Address = "tls://conode.dedis.ch:7000"
-  Url = "https://conode.dedis.ch"
   Suite = "Ed25519"
   Public = "ec5c65a3c922d1df32075640e3de606197be24af76059a2ef145501122884bd3"
   Description = "EPFL Cothority-server"
+  URL = "https://conode.dedis.ch"
```
cgrigis: Is this change of case ("Url" → "URL") intentional? It seems to break `Roster.fromTOML()`...

tharvik: Oups, I overlooked it; it seems that the golang serialisation uses [`URL`](https://github.com/dedis/onet/blob/master/app/config.go#L211) and that the typescript one uses [`Url`](https://github.com/dedis/cothority/blob/master/external/js/cothority/src/network/proto.ts#L59).

Adding a commit fixing it in external/js.

ineiti: Yes, as far as I know, the specs in my head say that both should be accepted. It's unfortunate, but, well, that's what it is...

tharvik: Looks like https://github.com/BurntSushi/toml allow anycase for the keys, even `uRl`...
So, updating the external/js patch, to allow for both casing, with preference to `URL`.

cgrigis: The [TOML spec](https://github.com/toml-lang/toml#spec) seems to be case-sensitive though.

tharvik: Correct.. Maybe we should drop this dep at some point, as upstream is dead since nearly 2 year, and it wrongly implements an old version of TOML.

external/js/cothority/spec/network/proto.spec.ts
```diff
@@ -18,13 +18,16 @@ describe("Network Proto Tests", () => {
...
             Suite = "Ed25519"
-            Url = "ws:127.0.0.1:7010"
+            URL = "ws:127.0.0.1:7010"
         `;
-        const roster = Roster.fromTOML(str);
 
-        expect(roster.length).toBe(2);
-        expect(roster.slice(1, 2).length).toBe(1);
-        expect(roster.aggregate.length).toBeGreaterThan(0);
+        [str, str.replace("URL", "Url")].forEach((toml) => {
```
ineiti: It needs to run with both syntaxes: `URL` and `Url`.

tharvik: Heurm, which it does: once with `str` (containing `URL`), once with `str.replace` (containing `Url`).

PR #2235, opened 2020-03-23
travis: revert to old npm publish

tharvik:
Running tests twice doesn't make much sense; also, I broke deploy on master by not linking kyber in the previous iteration, so, two birds, one stone.


ineiti:
Even though github thinks that travis didn't finish, travis says it did finish successfully...

PR #2236, opened 2020-03-23
Adding ProofsRequest

ineiti:
This PR adds a new service endpoint to ByzCoin:
- ProofsRequest

It takes a slice of InstanceID/Version and returns a proof for every instance
that has a newer version.

Further work could include adding/removing of additional info.


byzcoin/service.go
```diff
@@ -604,6 +604,38 @@ func (s *Service) GetSignerCounters(req *GetSignerCounters) (*GetSignerCountersR
...
+	}
+	if len(sb.ForwardLink) > 0 {
+		return nil, xerrors.New("can only give proofs for latest block")
+	}
+	st, err := s.GetReadOnlyStateTrie(sb.SkipChainID())
+	if err != nil {
+		return nil, xerrors.Errorf("getting state trie: %w", err)
+	}
+	for _, idv := range pr.Instances {
+		_, ver, _, _, err := st.GetValues(idv.ID[:])
```
tharvik: Dropping error.

byzcoin/api_test.go
```diff
@@ -350,6 +350,47 @@ func TestClient_SignerCounterDecoder(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, 1, len(gur.Proofs))
+	require.True(t, gur.Proofs[0].Match(ConfigInstanceID[:]))
+
+	ctx, _ := createConfigTxWithCounter(t, time.Millisecond*100, *s.roster,
+		1e5, s, 1)
+	s.sendTxAndWait(t, ctx, 10)
+
+	req.Flags = 0
+	gur, err = s.service().GetUpdates(req)
```
tharvik: `gur`'s value is unused.

byzcoin/service.go
```diff
@@ -604,6 +604,38 @@ func (s *Service) GetSignerCounters(req *GetSignerCounters) (*GetSignerCountersR
...
+		if ver <= idv.Version &&
+			!(sendVersion0 && ver == 0) {
+			continue
+		}
+		proof, err := st.GetProof(idv.ID[:])
+		if err != nil {
+			log.Warn("Error while looking up proof", err)
```
tharvik: Not failing? Looks like `GetProof` shouldn't fail as it generate a proof for both existence or absence.

byzcoin/service.go
```diff
@@ -604,6 +604,38 @@ func (s *Service) GetSignerCounters(req *GetSignerCounters) (*GetSignerCountersR
 	return &resp, nil
 }
 
+// GetUpdates returns instances that have a newer versions than the ones
+// passed to it.
+func (s *Service) GetUpdates(pr *GetUpdatesRequest) (*GetUpdatesReply, error) {
+	reply := &GetUpdatesReply{}
```
tharvik: I like to delay initialisation until it's really needed (such as before the `for`), I find it more readable as "in the next block, we initialize theses variables".

byzcoin/service.go
```diff
@@ -604,6 +604,38 @@ func (s *Service) GetSignerCounters(req *GetSignerCounters) (*GetSignerCountersR
 	return &resp, nil
 }
 
+// GetUpdates returns instances that have a newer versions than the ones
+// passed to it.
+func (s *Service) GetUpdates(pr *GetUpdatesRequest) (*GetUpdatesReply, error) {
```
tharvik: Please add some newlines to make it clearer.

byzcoin/service_test.go
```diff
@@ -1046,6 +1046,21 @@ func sendTransactionWithCounter(t *testing.T, s *ser, client int, kind string, w
 	return proof, tx.Instructions[0].Hash(), resp, err, err2
 }
 
+func (s *ser) sendInstructions(t *testing.T, wait int,
```
tharvik: Unused.

external/js/cothority/src/byzcoin/proof.ts
```diff
@@ -136,11 +130,11 @@ export default class Proof extends Message<Proof> {
      * Verify that the proof contains a correct chain from the given genesis. Note that
      * this function doesn't verify the first roster of the chain.
      *
-     * @param genesisID The skipchain ID
+     * @param latestID ID of the latest known and trusted block the proof is based upon
      * @returns an error if something is wrong, null otherwise
      * @deprecated use verifyFrom for a complete verification
      */
-    verify(id: InstanceID): Error {
+    verify(latestID: InstanceID): Error {
```
tharvik: While at it, you can return `Error | undefined` (and change the `null` to `undefined` down there).

external/js/cothority/src/byzcoin/proof.ts
```diff
@@ -383,6 +353,7 @@ class InclusionProof extends Message<InclusionProof> {
     leaf: LeafNode;
     empty: EmptyNode;
     nonce: Buffer;
+    protected _state: StateChangeBody;
```
tharvik:
```suggestion
    protected _state: StateChangeBody | undefined;
```

external/js/cothority/src/byzcoin/proof.ts
```diff
@@ -372,6 +329,19 @@ class InclusionProof extends Message<InclusionProof> {
     get value(): Buffer {
         return this.leaf.value;
     }
+
+    /**
+     * Returns the stateChangeBody of this InclusionProof.
+     */
+    get stateChangeBody(): StateChangeBody {
+        if (!this._state) {
```
tharvik:
```suggestion
        if (this._state === undefined) {
```

external/js/cothority/src/byzcoin/proof.ts
tharvik: Hum, you mark it as protected but the field generating it (`leaf.value`) is publicicly modifiable; tag the latter as private?

external/js/cothority/src/byzcoin/proof.ts
```diff
@@ -446,6 +417,53 @@ class InclusionProof extends Message<InclusionProof> {
...
+            }
+
+            return this.key.equals(key);
+        } else if (expectedHash.equals(this.hashEmpty())) {
+            if (_.difference(bits.slice(0, i), this.empty.prefix).length !== 0) {
+                throw new Error("invalid prefix in empty node");
+            }
+
+            return false;
+        }
```
tharvik: Can't you merge both blocks?

external/js/cothority/src/byzcoin/proto/requests.ts
```diff
@@ -224,6 +224,55 @@ export class GetSignerCountersResponse extends Message<GetSignerCountersResponse
...
+/**
+ * Holds an instanceID and it's latest known version.
+ */
+export class IDVersion extends Message<IDVersion> {
+    static register() {
+        registerMessage("IDVersion", IDVersion);
+    }
+
+    readonly id: Buffer;
+    readonly version: Long;
```
tharvik: To be strict, every field set by `Message` can be `undefined` (that's the burden of ProtocolBuffer), so better type it correctly.

external/js/cothority/src/byzcoin/proto/requests.ts
```diff
@@ -224,6 +224,55 @@ export class GetSignerCountersResponse extends Message<GetSignerCountersResponse
...
+
+    static register() {
+        registerMessage("GetUpdatesRequest", GetUpdatesRequest);
+    }
+
+    readonly instances: IDVersion[];
+    readonly flags: Long;
+    readonly latestblockid: Buffer;
+
+    constructor(props?: Properties<GetUpdatesRequest>) {
```
tharvik: Useless, only forward args to superclass.

external/js/cothority/src/byzcoin/proto/requests.ts
```diff
@@ -224,6 +224,55 @@ export class GetSignerCountersResponse extends Message<GetSignerCountersResponse
...
+export class GetUpdatesReply extends Message<GetUpdatesReply> {
+    static register() {
+        registerMessage("GetUpdatesReply", GetUpdatesReply);
+    }
+
+    readonly proofs: InclusionProof[];
+    readonly links: ForwardLink[];
+    readonly latest: SkipBlock;
+
+    constructor(props?: Properties<GetUpdatesReply>) {
```
tharvik: Useless.

external/js/cothority/src/byzcoin/proto/requests.ts
```diff
@@ -224,6 +224,55 @@ export class GetSignerCountersResponse extends Message<GetSignerCountersResponse
...
+
+/**
+ * Requests a set of proofs that will be returned in a single message.
+ * The reply will only hold proofs for instances that have a higher version than the ones given here.
+ */
+export class GetUpdatesRequest extends Message<GetUpdatesRequest> {
+    static readonly sendVersion0 = Long.fromNumber(1);
+
+    static register() {
+        registerMessage("GetUpdatesRequest", GetUpdatesRequest);
```
tharvik: You should define dependencies used by the message.

```suggestion
        registerMessage("GetUpdatesRequest", GetUpdatesRequest, IDVersion);
```

external/js/cothority/src/byzcoin/proto/requests.ts
```diff
@@ -224,6 +224,55 @@ export class GetSignerCountersResponse extends Message<GetSignerCountersResponse
...
+        super(props);
+    }
+}
+
+/**
+ * The requested proofs
+ */
+export class GetUpdatesReply extends Message<GetUpdatesReply> {
+    static register() {
+        registerMessage("GetUpdatesReply", GetUpdatesReply);
```
tharvik:
```suggestion
        registerMessage("GetUpdatesReply", GetUpdatesReply, ForwardLink, InclusionProof, SkipBlock);
```

external/js/cothority/src/byzcoin/proto/requests.ts
```diff
@@ -232,3 +281,6 @@ AddTxRequest.register();
 AddTxResponse.register();
 GetSignerCounters.register();
 GetSignerCountersResponse.register();
+IDVersion.register();
```
tharvik: With correct deps, no need to register it.

.travis.yml
```diff
@@ -62,11 +62,7 @@ jobs:
         - npm run linter
 
     - stage: build
-      name: "go 1.12"
```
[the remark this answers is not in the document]
ineiti: yes, cothority supports the last two versions of go. We might even change this to the last version... will speed up the tests.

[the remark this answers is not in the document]
ineiti: Changed it to try to build using 1.13 and 1.14, but only test 1.14.

external/js/cothority/src/byzcoin/proof.ts
ineiti: I don't like too much changing signatures - hope it doesn't break anything.

external/js/cothority/src/byzcoin/proof.ts
ineiti: changed it to private.

external/js/cothority/src/byzcoin/proof.ts
ineiti: nah, as now it's either undefined or an instance of `StateChangeBody`

external/js/cothority/src/byzcoin/proof.ts
ineiti: that's a copy of the code in go - looks more complicated to merge them than to let it be like that.

external/js/cothority/src/byzcoin/proto/requests.ts
ineiti: Not sure - if the field is required, then protobuf should throw an error if the field is not there.

external/js/cothority/src/byzcoin/proto/requests.ts
ineiti: copy/pasted from other definitions...

byzcoin/service.go
```diff
@@ -604,6 +604,42 @@ func (s *Service) GetSignerCounters(req *GetSignerCounters) (*GetSignerCountersR
...
+	sb := s.db().GetByID(pr.LatestBlockID)
+	if sb == nil {
+		return nil, xerrors.New("cannot find skipblock while getting proof")
+	}
+	if len(sb.ForwardLink) > 0 {
+		return nil, xerrors.New("can only give proofs for latest block")
```
nkcr: why don't you take a skipchainID instead of a skipblockID? Then the service can directly fetch the latest block.

ineiti: That would be an option then. The current implementation never sends the forwardlinks to prove the latest block is valid, as it supposes that the client already has the latest block. Currently the following happens:

- client calls the badly named `StreamingTransactions` and listens for new blocks
- every time a new block is received, the client sends a list of known instances to the node, with the request to send only the updates
- the node checks the versions, and only creates `InclusionProof`s for the instances that have a newer version

The main goal of `GetUpdates` is to use as few as possible of the network. So sending the block when not needed is an advantage.

A possible extension would be the flag `GUFSendProofForLatestBlock` which would fill out the already available `ForwardLinks` and the `Latest`.

ineiti:
This PR passed Travis - it's just github and travis can't communicate right...

PR #2237, opened 2020-03-24
Adding PollDelete and Challenge

ineiti:
For the OpenHouse demo two new functionalities have been added to the personhood
service - one to delete the polls (only for admins) and one to participate
in a challenge.


ineiti:
This has been merged previously in fact...

PR #2239, opened 2020-03-28
Adding deletion and preIDs for Calypso contracts

ineiti:
This PR adds the following new features to ByzCoin:
- deleting instances by default: now the BasicContract includes a default 'Delete'
- added preIDs to Instruction: there is now a default way to calculate the preIDs for a contract
- using preIDs in CalypsoContracts: allowing to choose the pre-image of the IDs of the Write and Read Instances
- fixes small bugs:
  - missing error handling in executeTransaction
  - missing argument strings in calypso-instance.ts

Both the go-code and the js-code have been updated.

### DIFF
--- a/byzcoin/transaction.go
+++ b/byzcoin/transaction.go
@@ -231,6 +231,23 @@ func (instr Instruction) DeriveID(what string) InstanceID {
 	// strict domain separation now.
 }
 
+// DeriveIDArg calculates the id of the new instance for a spawn-instruction.
+// If the 'arg' is not present in the spawn-arguments,
+// it behaves like DeriveID(what).
+// If the instruction is not a spawn-instruction, it returns an error.
+func (instr Instruction) DeriveIDArg(what, arg string) (InstanceID, error) {
+	if instr.GetType() != SpawnType {
+		return NewInstanceID(nil), xerrors.New("can only get ID from spawn")
+	}
+	if instIDBuf := instr.Spawn.Args.Search(arg); instIDBuf != nil {
+		h := sha256.New()
+		h.Write([]byte(instr.Spawn.ContractID))
+		h.Write(instIDBuf)
+		return NewInstanceID(h.Sum(nil)), nil
+	}
+	return instr.DeriveID(what), nil
+}
+
 // Action returns the action that the user wants to do with this
 // instruction.
 func (instr Instruction) Action() string {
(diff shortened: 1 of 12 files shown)


byzcoin/transaction.go
```diff
@@ -231,6 +231,23 @@ func (instr Instruction) DeriveID(what string) InstanceID {
 	// strict domain separation now.
 }
 
+// DeriveIDArg calculates the id of the new instance for a spawn-instruction.
+// If the 'arg' is not present in the spawn-arguments,
+// it behaves like DeriveID(what).
+// If the instruction is not a spawn-instruction, it returns an error.
+func (instr Instruction) DeriveIDArg(what, arg string) (InstanceID, error) {
```
tharvik: Can't we use the `Argument` type direclty (without `Value`)?

```suggestion
func (instr Instruction) DeriveIDArg(what string, arg Argument) (InstanceID, error) {
```

external/js/cothority/src/skipchain/skipchain-rpc.ts
```diff
@@ -180,15 +180,16 @@ export default class SkipchainRPC {
             const last = newBlocks[newBlocks.length - 1];
             let isInRoster = false;
             for (const n of last.roster.list) {
-                if (n.getWebSocketAddress() === this.conn.getURL()) {
+                if (n.getWebSocketAddress().replace(/^ws+:\/\//, "") ===
+                    this.conn.getURL().replace(/^ws+:\/\//, "")) {
```
tharvik: Why is it needed?

ineiti: When running the webapp locally, but connected to the real byzcoin. It's a 'manual testing' usecase, but I still do it a lot...

tharvik: I don't get it, why the URL would be different? `getWebSocketAddress` is used when creating `this.conn` so it should be the same

byzcoin/transaction.go
ineiti: Problem is that the caller does not know where in the `instr.Spawn.Args` the correct argument is. He only knows the name of the argument...

external/js/cothority/src/skipchain/skipchain-rpc.ts
ineiti: if the URL is `https://`, `getWebSocketAddress` will return `wss://`, while the conneted URL returns `ws://`. So it will not match.

byzcoin/transaction.go
tharvik: Yep, and the caller can create a "search "Argument`" as `Argument{Name: "my-name"}`.

external/js/cothority/src/skipchain/skipchain-rpc.ts
tharvik: Ho, right, the weird secure context.

ineiti:
Also, if OK, merge please.

PR #2240, opened 2020-03-30
Moving StateTrie simulator to byzcoin

ineiti:
In order to test also other than personhood contracts in a more unit-testing manner,
move the ReadOnlyStateTrieSimulator to the byzcoin directory.

PR #2241, opened 2020-03-30
Calypso update

ineiti:
Allow the CalypsoWrite contract to update the Data and ExtraData part of the
Write structure.

Depends on #2240

### DIFF
--- a/calypso/contracts.go
+++ b/calypso/contracts.go
@@ -128,14 +128,12 @@ func (c ContractWrite) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruc
 //  - update - it takes a 'data' and/or 'extraData' argument that is used to
 //    update the data and/or extradata part of the write structure.
 func (c *ContractWrite) Invoke(rst byzcoin.ReadOnlyStateTrie,
-	inst byzcoin.Instruction, cin []byzcoin.Coin) (sc []byzcoin.StateChange,
-	cout []byzcoin.Coin, err error) {
-	cin = cout
+	inst byzcoin.Instruction, cin []byzcoin.Coin) ([]byzcoin.StateChange,
+	[]byzcoin.Coin, error) {
 
-	var darcID darc.ID
-	_, _, _, darcID, err = rst.GetValues(inst.InstanceID.Slice())
+	_, _, _, darcID, err := rst.GetValues(inst.InstanceID.Slice())
 	if err != nil {
-		return
+		return nil, nil, err
 	}
 
 	update := false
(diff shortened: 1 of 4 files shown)


calypso/contracts.go
```diff
@@ -124,6 +124,51 @@ func (c ContractWrite) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruc
...
+		if extraData != nil {
+			c.ExtraData = extraData
+			update = true
+		}
+	default:
+		return nil, nil, xerrors.New("only know 'update' command")
+	}
+
+	if !update {
+		return
```
tharvik, line 160: Maybe an error? IMO, `update` without either `data` or `extraData` is usually a bug.

calypso/contracts.go
```diff
@@ -124,6 +124,51 @@ func (c ContractWrite) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruc
 	return
 }
 
+// Invoke supports the following command:
+//  - update - it takes a 'data' and/or 'extraData' argument that is used to
+//    update the data and/or extradata part of the write structure.
+func (c *ContractWrite) Invoke(rst byzcoin.ReadOnlyStateTrie,
+	inst byzcoin.Instruction, cin []byzcoin.Coin) (sc []byzcoin.StateChange,
+	cout []byzcoin.Coin, err error) {
```
tharvik, line 132: Hum, that's probably being pedantic but using theses named return values make the code less readable, as one has to trace back the assignements; and in this case, it's quite easy to return `nil` on error.

external/js/cothority/src/calypso/calypso-instance.ts
```diff
@@ -150,6 +153,28 @@ export class CalypsoWriteInstance extends Instance {
...
+
+    /**
+     * Update sends an update command to an existing CalypsoWriteInstance and changes the data and/or the extraData
+     * field of the Write structure. If neither data, nor extraData is given, it returns without sending a
+     * transaction to byzcoin.
+     * @param u a structure holding the data and/or extraData to be updated.
+     */
+    async update(u: {data?: Buffer; extraData?: Buffer}) {
+        const args: Argument[] = [];
+        if (u.data && u.data.length > 0) {
```
tharvik, line 165: Why checking for the length? How would someone "reset" `data`?

calypso/contracts.go
ineiti: Hmm - still thinking about it. Per se it's a NO-OP, so it's a valid operation. On the other hand, it might point to an error done somewhere in the sending of the transaction...

calypso/contracts.go
ineiti: Re-wrote it without the named returns. But it looks more convoluted to me. Also, all the other `Invoke` do use named returns...

calypso/contracts.go
```diff
@@ -124,6 +124,49 @@ func (c ContractWrite) Spawn(rst byzcoin.ReadOnlyStateTrie, inst byzcoin.Instruc
...
+		return nil, nil, xerrors.New("only know 'update' command")
+	}
+
+	if !update {
+		return nil, cin, nil
+	}
+
+	var ciBuf []byte
+	ciBuf, err = protobuf.Encode(&c.Write)
+	if err == nil {
```
tharvik, line 163: Can you inverse the condition (`err != nil`)? That's the common golang pattern.

calypso/contracts.go
tharvik: Do you? I find it more readable as the `return` clearly indicate what is being returned and it ensure that all fields are set as intended (ie `nil` on error); do you find something specifically convoluted or only the named returns can fix it?

calypso/contracts.go
ineiti: True - now it doesn't make sense anymore. With the named returns it made sense to have it `== nil` there...

PR #2242, opened 2020-03-31
Jenkins failures

ineiti:
Some small improvements to make the tests pass. The viewchange tests are
still flaky - but using the go-test cache, it will eventually pass :(


ineiti:
Merge if OK...

tharvik:
Not waiting for @Gilthoniel then?

PR #2252, opened 2020-04-21
Fixing README of scmgr

### LINKED ISSUE #2251: Problem in scmgr
When I run command "scmgr follow add single SKIPCHAIN_ID 127.0.0.1:7004" or "scmgr skipchain block add -roster public.toml SKIPCHAIN_ID" of the README , I always confront "Error while sending closing type: websocket: close sent" or " couldn't add this block as chain-follower: sending: connection read: websocket: close 1002 (protocol error): unexpected error: processing error: wrong signature of unknown signer" ,Could you please tell me what should I do or which part I have done is wrong.

ineiti:
The scmgr/README.md was out of date with regard to the rest of the code.
Updated the following to work well:
- conode/run_nodes.sh - build public.toml with first node at beginning
- scmgr/app.go - use leader node and store both forward-link and new block

OR-reviewers: first to think it's OK, please merge...
Closes #2251

### DIFF
--- a/scmgr/README.md
+++ b/scmgr/README.md
@@ -31,7 +31,7 @@ you can use with scmgr. In the following examples, we suppose that the
 directory and that you installed `scmgr` using
 
 ```bash
-go get github.com/dedis/cothority/scmgr
+go build ../scmgr
 ```
 
 ## Securing the conode by creating a link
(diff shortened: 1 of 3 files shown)


scmgr/README.md
```diff
@@ -31,7 +31,7 @@ you can use with scmgr. In the following examples, we suppose that the
 directory and that you installed `scmgr` using
 
 ```bash
-go get github.com/dedis/cothority/scmgr
+go build ../scmgr
```
tharvik, line 34:
```suggestion
go build .
```

ineiti: no, because the previous `cd cothority/conode` supposes that the current `wd` is `cothority/conode`. This `wd` is also needed for the other commands.

PR #2253, opened 2020-04-27
Adds a warning about using maps in contracts

nkcr:
I already lost a lot of time on this issue and that just happened to someone else. In particular it produces an error hard to debug since it just says that nodes can't reach a consensus.


tharvik:
Hum, you mean when `protobuf.Encode` for `Value`, right? In that case, modifying `protobuf.Value.MapKeys` to have a determinist order would do the trick IMO

nkcr:
If you think its worth the effort then I suggest we open a new issue and keep this warning in the meantime.

tharvik:
Good idea, followup in https://github.com/dedis/protobuf/issues/72.

PR #2255, opened 2020-04-28
Another fix for the viewchange

ineiti:
This patch fixes the initialization of the viewchange when the node
starts up: instead of calculating 'f' from the genesis-block, it
looks up the latest block and calculates 'f' from this one.

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -2662,6 +2662,8 @@ func (s *Service) monitorLeaderFailure() {
 						LeaderIndex: 1,
 					},
 				}
+				log.Lvlf2("Starting a view-change by putting our own request"+
+					": %+v", req)
 				s.viewChangeMan.addReq(req)
 			}
 		case <-s.closeLeaderMonitorChan:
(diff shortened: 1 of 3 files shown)


byzcoin/service.go
```diff
@@ -2804,8 +2806,10 @@ func (s *Service) startChain(genesisID skipchain.SkipBlockID) error {
 	if err != nil {
 		return xerrors.Errorf("getting initial duration: %v", err)
 	}
-	s.viewChangeMan.add(s.sendViewChangeReq, s.sendNewView, s.isLeader, string(genesisID))
-	s.viewChangeMan.start(s.ServerIdentity().ID, genesisID, initialDur, s.getFaultThreshold(genesisID))
+	s.viewChangeMan.add(s.sendViewChangeReq, s.sendNewView, s.isLeader,
+		string(genesisID))
+	s.viewChangeMan.start(s.ServerIdentity().ID, genesisID, initialDur,
+		s.getFaultThreshold(latest.Hash))
```
ineiti, line 2812: This was the line where it took the `f` from the genesis-block instead of the latest block.

PR #2258, opened 2020-04-30
Add a caching system for the stateTries

### LINKED ISSUE #2257: Verifying of DARCs is very slow
The current verification of DARCs is very slow if there are more than 50 signers or so. Is there a way to speed this up?
The problem is seen in `byzcoin.CheckAuthorization` and in the contract-signature-verification.

ineiti:
The current stateTries have two bottlenecks:
- LoadConfigFromTrie - which is called everytime a LoadDarcFromTrie is called and
decodes the config protobuf structure every time anew
- LoadDarcFromTrie - which is called when evaluating signatures or authorizations

This PR adds a cache for both of these methods, so that calls within the
same block are cached. The test, which is a bit artificial, gives the
following improvements:
- LoadConfigFromTrie - 3x faster authorization check
- LoadDarcFromTrie - 4x faster authorization check

The latter is a bit artificial, because it measures looking up the same darcs
repeatedly, which in practice happens rarely.

The first test, however, is very relevant. In the 9cc3-chain, it is even worse,
as the config contains more nodes, which will make protobuf.Decode even
slower.

Closes #2257


byzcoin/statetrie.go
```diff
@@ -149,13 +155,84 @@ func (t *stagingStateTrie) GetVersion() Version {
...
+		return nil, xerrors.Errorf("reading trie: %v", err)
+	}
+	config, err := tc.loadConfigFromTrie(st)
+	if err != nil {
+		return nil, xerrors.Errorf("reading trie: %v", err)
+	}
+	var ok bool
+	for _, id := range config.DarcContractIDs {
+		if contract == id {
+			ok = true
```
tharvik, line 215: `break` after that

byzcoin/statetrie.go
```diff
@@ -149,13 +155,84 @@ func (t *stagingStateTrie) GetVersion() Version {
...
+		}
+	}
+	if !ok {
+		return nil, xerrors.New("the contract \"" + contract + "\" is not in the set of DARC contracts")
+	}
+	tc.darcs[string(id)], err = darc.NewFromProtobuf(darcBuf)
```
tharvik, line 221: Use a local variable before assigning to map, otherwise, it might get polluted by failing call.

byzcoin/statetrie.go
```diff
@@ -149,13 +155,84 @@ func (t *stagingStateTrie) GetVersion() Version {
...
+		return nil, xerrors.Errorf("reading trie: %v", err)
+	}
+	var ok bool
+	for _, id := range config.DarcContractIDs {
+		if contract == id {
+			ok = true
+		}
+	}
+	if !ok {
+		return nil, xerrors.New("the contract \"" + contract + "\" is not in the set of DARC contracts")
```
tharvik, line 219:
```suggestion
		return nil, fmt.Errorf("the contract \"%v\" is not in the set of DARC contracts", contract)
```

byzcoin/statetrie.go
```diff
@@ -149,13 +155,84 @@ func (t *stagingStateTrie) GetVersion() Version {
...
+	}
+	tc.darcs[string(id)], err = darc.NewFromProtobuf(darcBuf)
+	if err != nil {
+		return nil, xerrors.Errorf("decoding darc: %v", err)
```
tharvik, line 223: Can help with knowing what was wrong in the encoded DARC (which key is missing, what failed exactly, ASO)
```suggestion
		return nil, fmt.Errorf("decoding darc: %w", err)
```

byzcoin/statetrie.go
```diff
@@ -149,13 +155,84 @@ func (t *stagingStateTrie) GetVersion() Version {
...
+}
+
+type trieCache struct {
+	config *ChainConfig
+	darcs  map[string]*darc.Darc
+}
+
+func (tc *trieCache) loadConfigFromTrie(st ReadOnlyStateTrie) (*ChainConfig, error) {
+	if tc.config != nil {
+		return tc.config, nil
```
tharvik, line 173: Depending on how usual it is to retrieve the config, one can move it at initialisation time of the `trieCache`.

byzcoin/statetrie_test.go
```diff
@@ -2,7 +2,18 @@ package byzcoin
...
+	"math/big"
+	"strings"
 	"testing"
+	"time"
+
+	"go.dedis.ch/onet/v3"
+
+	"go.dedis.ch/kyber/v3/util/random"
+
+	"github.com/ethereum/go-ethereum/common/math"
```
tharvik, line 15: Nope

byzcoin/statetrie_test.go
```diff
@@ -2,7 +2,18 @@ package byzcoin
...
+	"go.dedis.ch/onet/v3/log"
```
tharvik, line 16: Pack the related imports (ie, all from dedis)

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
 	require.NoError(t, sst.Commit())
 	require.True(t, bytes.Equal(sst.GetRoot(), newRoot))
 }
+
+// TestDarcRetrieval is used as a benchmark here. It stores a lot of darcs
```
tharvik, line 86: Then, use `testing.B` ;)

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+//   go test -v -cpuprofile cpu.prof -run TestDarcRetrieval
+//   go tool pprof -pdf cpu.prof
+func TestDarcRetrieval(t *testing.T) {
+	darcDepth := 3
+	darcWidth := 10
+	entries := 10000
+
+	s := newSer(t, 1, 10*time.Second)
+	defer s.local.CloseAll()
+	s.local.Check = onet.CheckNone
```
tharvik, line 99: Why dropping goroutine checks?

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+
+	st, err := s.service().getStateTrie(s.genesis.SkipChainID())
+	require.NoError(t, err)
+	require.NotNil(t, st)
```
tharvik, line 103: Drop it, `st != nil` if `err == nil`.

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+	require.NotEqual(t, -1, st.GetIndex())
+
+	log.Lvl1("Creating random entries")
+	value := random.Bits(1024, true, random.New())
+	start := time.Now()
+	var scs []StateChange
+	for e := 1; e <= entries; e++ {
```
tharvik, line 110:
```suggestion
	for e := 0; e < entries; e++ {
```

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+		scs = append(scs, StateChange{
+			StateAction: Create,
+			InstanceID:  random.Bits(256, true, random.New()),
+			ContractID:  ContractDarcID,
+			Value:       value,
+		})
+		if e%100 == 0 {
+			st.StoreAll(scs, 5, CurrentVersion)
+			scs = []StateChange{}
+			log.Print(e, time.Now().Sub(start))
```
tharvik, line 120: Debug? Ha no, it's because it's not a benchmark, move it to one; use `testing.B.N` as `entries`, drop `start` and let go do the magic.

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+	defer s.local.CloseAll()
+	s.local.Check = onet.CheckNone
+
+	st, err := s.service().getStateTrie(s.genesis.SkipChainID())
+	require.NoError(t, err)
+	require.NotNil(t, st)
+	require.NotEqual(t, -1, st.GetIndex())
+
+	log.Lvl1("Creating random entries")
+	value := random.Bits(1024, true, random.New())
```
tharvik, line 107: Why random? It doesn't test much to do that and make it a bit harder to read, just choose a value.

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+	start := time.Now()
+	var scs []StateChange
```
tharvik, line 109: IMO, reallocation will take some time (relative to `entries), better create a slice with the right size already.

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+		})
+		if e%100 == 0 {
+			st.StoreAll(scs, 5, CurrentVersion)
+			scs = []StateChange{}
+			log.Print(e, time.Now().Sub(start))
+			start = time.Now()
+		}
+	}
+
+	log.Lvl1("Creating darcs")
```
tharvik, line 125: As I understood it, it's mostly two unrelated tests, split it into two functions.

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+	counter := big.NewInt(0)
```
tharvik, line 126: No need for BigInt
```suggestion
	var counter uint
```

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+	// Store the tree of darcs in a 2D-array with the root in index 0 and
+	// the leafs in index darcDepth-1
+	darcs := make([][]*darc.Darc, darcDepth)
+	for d := darcDepth - 1; d >= 0; d-- {
+		width := math.Exp(big.NewInt(int64(darcWidth)),
+			big.NewInt(int64(d))).Int64()
+		darcs[d] = make([]*darc.Darc, width)
+		// The leafs just hold a dummy reference to a darc
+		if d == darcDepth-1 {
```
tharvik, line 135: This a "fire once" `if`, move it before the loop.

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+						darcs[d+1][i*darcWidth+id].GetBaseID())
+				}
+				rules := darc.InitRules(ids, ids)
+				darcs[d][i] = darc.NewDarc(rules, []byte(counter.String()))
+			}
+		}
+	}
+
+	scs = make([]StateChange, 0, counter.Int64())
+	for d := range darcs {
```
tharvik, line 159:
```suggestion
	for d, darc := range darcs {
```
And use it instead of `darcs[d]`

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+		for w := range darcs[d] {
```
tharvik, line 160:
```suggestion
		for _, td := range darcs[d] {
```

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+			return nil
+		}
+		id, err := hex.DecodeString(strings.Replace(s, "darc:", "", 1))
+		if err != nil || len(id) != 32 {
+			log.Error("invalid darc id", s, len(id), err)
+			return nil
+		}
+		d, err := st.LoadDarcFromTrie(id)
+		if err != nil {
+			return nil
```
tharvik, line 189: No `log.Error` (not that I approve of this way to mark error but yeah, otherwise, we would need to change `darc.GetDarc` which might be out of scope for this PR (but not of https://github.com/dedis/cothority/issues/2260!))

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+		}
+		return d
+	}
+
+	root := darcs[0][0]
+	rootSign := root.Rules.GetSignExpr()
+	start = time.Now()
```
tharvik, line 196: Another timer, another occasion to have a real benchmark :P

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+	for i := range darcs[darcDepth-1] {
+		id := darc.ID(make([]byte, 32))
+		copy(id, big.NewInt(int64(i+1)).Bytes())
+		darcID := darc.NewIdentityDarc(id)
+		log.Print("Searching id", darcID.String())
+		err = darc.EvalExprDarc(rootSign, getDarcs, true, darcID.String())
```
tharvik, line 202: Either inline in `require.NoError` or make `err` local to the `for`.

byzcoin/statetrie_test.go
```diff
@@ -71,3 +82,125 @@ func TestStateTrie(t *testing.T) {
...
+	}
+
+	root := darcs[0][0]
+	rootSign := root.Rules.GetSignExpr()
+	start = time.Now()
+	for i := range darcs[darcDepth-1] {
+		id := darc.ID(make([]byte, 32))
+		copy(id, big.NewInt(int64(i+1)).Bytes())
+		darcID := darc.NewIdentityDarc(id)
+		log.Print("Searching id", darcID.String())
```
tharvik, line 201: Debug?

tharvik:
The tests are failing also, but login seems to work now :)

byzcoin/statetrie.go
```diff
@@ -149,13 +155,84 @@ func (t *stagingStateTrie) GetVersion() Version {
...
+func (t *stagingStateTrie) LoadDarcFromTrie(id darc.ID) (*darc.Darc, error) {
+	return t.loadDarcFromTrie(t, id)
+}
+
+type trieCache struct {
+	config *ChainConfig
+	darcs  map[string]*darc.Darc
+}
+
+func (tc *trieCache) loadConfigFromTrie(st ReadOnlyStateTrie) (*ChainConfig, error) {
```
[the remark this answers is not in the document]
ineiti: we'll have to have a longer discussion on this. Whenever there is an error plus a value to be returned, I prefer the pointer. That gives a nice `return nil, err`. Else you have to return an unusable structure...

byzcoin/statetrie.go
ineiti: I prefer `%s` for the format here - `%v` is too generic for me ;)

byzcoin/statetrie.go
ineiti: I thought that if you pass `%v` with an `err`, it will correctly create a stack-trace, no?

byzcoin/statetrie.go
ineiti: premature optimisation ;) sometimes the tries are initialized just with the structure, so I'd have to replace all those with a `NewTrie`, too...

byzcoin/statetrie_test.go
ineiti: Thought about it - as bevm already uses it, I thought it's OK to use it here, too...

byzcoin/statetrie_test.go
ineiti: But I'm smarter than go ;) I wanted to be able to change all three of `depth`, `width` and `entries`, which is not possible with a benchmark.

byzcoin/statetrie_test.go
ineiti:
```
	// If the test fails, the profile will be empty.
	// So disable testing of go-routines started by the profiler.
```

byzcoin/statetrie_test.go
ineiti: no, else the `entries` must go until `1001`. The `if e%100` stores the actual entries.

byzcoin/statetrie_test.go
ineiti: irrelevant here. It takes `50ms` for 100 entries.

ineiti: removed the measurement, starting from 0, and allocating.

byzcoin/statetrie_test.go
ineiti: No, the measurement of the random entries was just for me to understand how much faster it is to create a slice and then `st.StoreAll` it all at once.

I wanted to measure:
- how long it takes to verify the darcs
- whether entries in the trie change this

byzcoin/statetrie_test.go
ineiti: yes, for the `math.Exp`

byzcoin/statetrie_test.go
ineiti: but then I need to calculate the `width` and assign the `darcs[d]` twice - once before the loop and once in the loop.

byzcoin/statetrie_test.go
ineiti: you're overwriting a package-name? Sowas...

byzcoin/statetrie_test.go
ineiti: I want the `w` for the logging.

byzcoin/statetrie_test.go
ineiti: as written above, 3d-benchmark...

byzcoin/statetrie_test.go
ineiti: why not re-use it?

byzcoin/statetrie_test.go
ineiti: yes

byzcoin/statetrie_test.go
tharvik: What about something like:
```
func TestDarcRetrievalWithEmptyTrie()
func TestDarcRetrievalWithFilledTrie()
```
both using a common function which is the second part of the second `TestDarcRetrieval`.

byzcoin/statetrie_test.go
tharvik: you can use an helper function if wanted
```golang
getDarcSliceForDepth := func(depth uint) []*darc.Darc
```

byzcoin/statetrie_test.go
tharvik: it's more readable, I don't need to see where this `err` will be used, it's local to the block, not the to the whole function. Faster to read.

byzcoin/statetrie.go
tharvik: Not from what I read on [godoc](https://golang.org/pkg/fmt/#Errorf)

byzcoin/statetrie_test.go
tharvik: okay, that makes sense.

byzcoin/statetrie_test.go
ineiti: doesn't fit my test-case.

byzcoin/statetrie_test.go
ineiti:
```
	// When running with `-cpuprofile`, additional go-routines are added,
	// which should be ignored.
	log.AddUserUninterestingGoroutine("/runtime/cpuprof.go")
```

byzcoin/statetrie_test.go
ineiti: didn't see the `(*big.Int).Exp`

byzcoin/statetrie.go
ineiti: The `Unwrap` method is used to be able to check if an error equals to an existing error. Which is not what we want to do here. So I just follow the DEDIS standard here:

```
4. Errors are systematically wrapped with a meaningful message using xerrors.Errorf and the %v verb.
```

byzcoin/statetrie_test.go
ineiti: That's more complicated and unreadable for me.

Issue #2264, opened 2020-05-07
Simplify view-change


ineiti:
Currently the view-change works like this:

1. the leader stops polling the nodes, which will trigger the heartbeat-watcher
2. all nodes seeing the leader fail send a view-change request to the new leader
3. once the new leader has enough requests, he'll start a BLSCoSi to confirm the view-change
4. if the BLSCoSi is successful, the new leader proposes a new block with the change of the roster

I wonder if `3.` is really necessary or not. Couldn't the verification of the block already include checking that all requests are valid?

Question for @LefKok

LefKok:
- For number 2. What happens if the new leader is faulty too? Do they view-change to the next in line?

- 3 and 4 seem to do the same thing (the new leader tries to see if he has enough support) so I would skip 3.

- What happens if multiple leaders propose a new block? Who do the followers support?

LefKok:
> Couldn't the verification of the block already include checking that all requests are valid?
This sounds dangerous. The leader can have valid requests but still, be too late to the party as a better leader has emerged. The protocol should not only check that the view-change messages are valid but also that the view is not an old view

ineiti:
* 2. - yes, they cycle through the list of the roster, giving each possible leader some time to propose a new block.
* support / checking - good reminder to check that the view-change is not old. If I do this, then for each node there will be only one valid new leader, so it will not have to chose between concurrent blocks

Possible improvement: would it be dangerous if the nodes look at the signature list of the latest forward-link to skip nodes that did not participate in the signature? So if the next node in the roster-list did not participate in the signing of the latest block, this node will be skipped.

ineiti:
Another idea would be to completely remove the signatures, and if the current leader is found to be stuck, the new possible leader just sends out a block where he proposes himself.

All nodes then decide whether they want to sign the new block and thus accepting the new leader.

ineiti:
Closed by #2266

PR #2265, opened 2020-05-07
Sign viewchange only with active nodes

ineiti:
Once the new leader receives confirmation of all nodes that he
can propose a view-change, he will do a blsCoSi to confirm that
the view-change will be happening. See also #2264.

To help BLSCoSi, this PR will only ask nodes who sent a view-change
request to sign it.

Is useless after #2266


ineiti:
Not relevant anymore due to #2266

PR #2266, opened 2020-05-07
Removing the BLS signature from the view-change

ineiti:
As @LefKok validated #2264, this PR removes the BLSCoSi round for the view-change. This makes it faster and simpler.

One more idea would be to avoid the view-change requests completely. But that will break even more stuff. So let's start small...

### DIFF
--- a/byzcoin/viewchange.go
+++ b/byzcoin/viewchange.go
@@ -2,6 +2,8 @@ package byzcoin
 
 import (
 	"bytes"
+	"errors"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


byzcoin/viewchange.go
```diff
@@ -169,38 +170,40 @@ func (s *Service) sendNewView(proof []viewchange.InitReq) {
...
+	sb := s.db().GetByID(proof[0].View.ID)
+	req, err := viewchange.NewNewViewReq(sb.Roster, proof)
 	if err != nil {
-		log.Error("Couldn't encode request:", err)
-		return
-	}
-	if !s.verifyViewChange(req.Hash(), pl) {
-		log.Error("Will not ask for view change I cannot verify")
-		return
+		log.Errorf("couldn't create request: %v", err)
```
tharvik, line 177: Not bailing early?

byzcoin/viewchange/viewchange.go
```diff
@@ -125,7 +127,9 @@ func (c *Controller) Stop() {
 }
 
 // Start begins a blocking process that processes incoming view-requests.
-func (c *Controller) Start(myID network.ServerIdentityID, genesis skipchain.SkipBlockID, initialDuration time.Duration, f int) {
```
tharvik, line 128: Same changes already merged in https://github.com/dedis/cothority/pull/2268 (and some more below)

ineiti: Yes, true. I re-based it on #2268... I'll rebase it on master.

byzcoin/viewchange.go
ineiti: yes, correct. Converted the method to return an error, and handle them correctly.

ineiti:
After presenting my idea, I put it back to `WIP`, as it's possible to make it even simpler...

ineiti:
This time I'm sure I will not go the whole way - removing all the `viewchange` and `heartbeat` seems too much for the moment.
This PR will already make it a little bit simpler. Even simpler is for later...

PR #2282, opened 2020-05-18
Uses published versions of the papers

nkcr:
**What this PR does**

*See commit message*

### DIFF
--- a/bftcosi/README.md
+++ b/bftcosi/README.md
@@ -36,5 +36,5 @@ a tree-structure which reduces the communication cost to O(n).
 
 - [PBFT](http://pmg.csail.mit.edu/papers/osdi99.pdf) describes the original
 PBFT protocol that is limited to 10-15 nodes
-- [ByzCoin](https://arxiv.org/abs/1602.06997) describes the BFTCoSi protocol
-and uses it to enhance bitcoin consensus
+- [ByzCoin](https://www.usenix.org/system/files/conference/usenixsecurity16/sec16_paper_kokoris-kogias.pdf)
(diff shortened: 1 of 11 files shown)


bftcosi/README.md
```diff
@@ -36,5 +36,5 @@ a tree-structure which reduces the communication cost to O(n).
 
 - [PBFT](http://pmg.csail.mit.edu/papers/osdi99.pdf) describes the original
 PBFT protocol that is limited to 10-15 nodes
-- [ByzCoin](https://arxiv.org/abs/1602.06997) describes the BFTCoSi protocol
-and uses it to enhance bitcoin consensus
+- [ByzCoin](https://www.usenix.org/system/files/conference/usenixsecurity16/sec16_paper_kokoris-kogias.pdf)
```
ineiti, line 39: How did you decide that usenix.org has a more reliable linking-system than arxiv.org? I'm a bit skeptic that usenix.org will keep their deep links more stable than the arxiv.org people. What do you think?

nkcr: It's not. I comes from a request of @bford to link published version of papers instead of pre-prints, which have less "value" in term of credibility.

ineiti: OK - makes sense. Do you see any way how we can make sure these links won't get out-of-order?

nkcr: If you really want to monitor that we could add a badge for each link: https://shields.io/category/monitoring, I tried with uptimerobot and it works well. But that seems overkill (though a bit fun). 
But for that matter I doubt that the www.usenix.org will ever expire since it even appears in the first page of the paper.

PR #2285, opened 2020-05-20
Use dedis/onet#640 to filter ByzCoin connections

### LINKED ISSUE #2204: Limit access to the node
Currently the connections to the conode are open to every other conode. This allows a simple amplification attack by requesting a signature from a big roster.

Byzcoin should call dedis/onet#620 every time it receives a new block, and set the current roster to be the only allowed set of nodes to communicate with.

cgrigis:
**What this PR does**

*See commit message*

Depends on dedis/onet#640 and dedis/onet#643 to be merged.

Closes #2204
Closes c4dt/TODO#142

---

🙅‍ Friendly checklist:


skipchain/skipchain.go
```diff
@@ -1962,3 +1969,16 @@ func (s *Service) incrementWorking() error {
 
 	return nil
 }
+
+// SetSkipChainValidPeers is a utility function to assign the set of valid
+// peers for the given skipchain.
+func SetSkipChainValidPeers(ctx *onet.Context, scID SkipBlockID,
+	validPeers []*network.ServerIdentity) {
+	// Compute the PeerSetID as hash(serviceID | skipchainID)
```
ineiti, line 1977: Sorry - missed that in your last PR. My proposition was to have this hash done in `NewPeerSetID`, so that the services don't have to create their own domain-protection, but can rely on `Context` doing the right thing.

skipchain/skipchain.go
```diff
@@ -367,6 +369,11 @@ func (s *Service) StoreSkipBlockInternal(psbd *StoreSkipBlock) (*StoreSkipBlockR
 				"Couldn't get forward signature on block: " + err.Error())
 		}
 
+		// Set the server's set of valid peers for this skipchain with the new
+		// roster.
+		SetSkipChainValidPeers(s.ServiceProcessor.Context, scID,
```
ineiti, line 374: Here you do `SetSkipChainValidPeers` twice: once here, and once in `catchUp`. I propose you keep the one here, and remove the one from `catchUp`.

skipchain/skipchain.go
cgrigis: But the SkipChain ID, being not known by `onet`, must still be provided, isn't it?

skipchain/skipchain.go
cgrigis: I believe I tried something like that at first, but it seemed in the tests that only the first node (the one receiving the request) was updating the list of valid peers, the other ones only doing it much later. My understanding was that they were doing it during the catch-up, hence the idea to have `catchUp()` set the valid peers as well, which (still per my understanding) also covers the `startAllChains()` case.
Is my understanding incorrect?

cgrigis:
> Perhaps leave the skipchain-service as-is, and only do `SetValidPeers` in `byzcoin`? Once in `catchUp`, and once in `startAllChains`?

My understanding was that `startAllChains()` eventually calls `catchUp()`, is that incorrect?

skipchain/skipchain.go
ineiti: yes, but not the serviceID.

skipchain/skipchain.go
ineiti: I think I need holidays, I can't think straight - I mixed `catchUp` with `updateTrieCallback`. 

Did you try to put it `updateTrieCallback` for new blocks and `startAllChains` for startup?

ineiti:
yes, `startAllChains` should call `catchUp`. But then you still have the update in `skipchain` and the startup handling in `byzcoin`, which is strange.

byzcoin/service_test.go
```diff
@@ -1850,6 +1850,70 @@ func TestService_SetConfigRosterReplace(t *testing.T) {
...
+		expectedValidPeers = append(expectedValidPeers, si.ID)
+
+		ctx, _ := createConfigTxWithCounter(t, testInterval, *goodRoster, defaultMaxBlockSize, s, counter)
+		counter++
+		cl := NewClient(s.genesis.SkipChainID(), *goodRoster)
+		resp, err := cl.AddTransactionAndWait(ctx, 10)
+		transactionOK(t, resp, err)
+
+		// Wait until all servers have included the block
+		time.Sleep(testInterval * 2)
```
ineiti, line 1889: Can you replace that with `cl.WaitPropagation(-1)`? It will make sure that all nodes have the same last block present. And as the `AddTransactionAndWait` will not return until at least the contacted node has the new block stored, this will wait the correct time, without having to sleep, which might break if the testing-host is busy with other stuff.

Same below.

And, yes, not all tests use yet `WaitPropagation`. But they should ;)

cgrigis: Excellent, thank you! I was indeed hoping there was a way to do that without the brittle `Sleep`...

skipchain/skipchain.go
cgrigis: Done

jeffallen:
Is this waiting on someone from DEDIS?

ineiti:
> Is this waiting on someone from DEDIS?

No - as it's still in `WIP` I think @cgrigis is finishing it...

calypso/service.go
```diff
@@ -434,6 +449,24 @@ func (s *Service) ReshareLTS(req *ReshareLTS) (*ReshareLTSReply, error) {
...
+	currentRoster := s.storage.Rosters[req.ByzcoinID]
+
+	// Build the new set as the union of the current and new rosters
+	newPeerSet := []*network.ServerIdentity{}
+	if currentRoster != nil {
+		newPeerSet = append(newPeerSet, currentRoster.List...)
+	}
+	newPeerSet = append(newPeerSet, req.NewRoster.List...)
+
+	s.SetValidPeers(s.NewPeerSetID(req.ByzcoinID[:]), newPeerSet)
```
cgrigis, line 465: @ineiti Is that ID a unique identifier for the LTS?

ineiti: TLDR: no, it's not the unique identifier

Longer version:

The LTS stores two things:
- the `ByzCoinID`, which is the hash of the genesis-block, to know how to verify proofs
- a list of `LTSIDs` that have been stored in the byzcoin-instance with the `ByzCoinID`

Every time a client wants to add an LTSID or wants to do a re-encryption, he has to proof he has the right by providing a proof with the `ByzCoinID` as the base.

So the valid peers for Calypso is not the byzcoin-roster, but the roster of the LTS, which can (and are) different. You can find this information in the `s.storage.Rosters` map.

ineiti: And in fact if you use `ByzCoinID` here, it will overwrite the set of peers for different LTSIDs. So you'll have to change for the LTSID.

cgrigis: Summary after discussion: it is actually the correct unique ID, just incorrectly named (it should be named `LTSID`) :-)

cgrigis:
> No - as it's still in `WIP` I think @cgrigis is finishing it...

It should be ready now. :)

calypso/service.go
```diff
@@ -342,6 +353,11 @@ func (s *Service) ReshareLTS(req *ReshareLTS) (*ReshareLTSReply, error) {
 
 		// NOTE: the roster stored in ByzCoin must have myself.
 		tree := roster.GenerateNaryTreeWithRoot(len(roster.List), s.ServerIdentity())
+		if tree == nil {
+			return nil, xerrors.New("failed to generate tree " +
```
ineiti, line 357: As we only support go 1.13 and go 1.14, you can use `errors.New` here, also below, where `fmt.Errorf` can be used instead of `xerrors.Errorf`.

calypso/service.go
```diff
@@ -330,6 +330,17 @@ func (s *Service) ReshareLTS(req *ReshareLTS) (*ReshareLTSReply, error) {
 		return nil, xerrors.Errorf("verifying proof: %v", err)
 	}
 
+	// Iterate through the new roster and update the set of valid peers
+	cl := onet.NewClient(cothority.Suite, ServiceName)
+	var reply updateValidPeersReply
+	for _, srv := range roster.List {
```
ineiti, line 336: Shouldn't this be sent to the union of the old and the new roster?

calypso/service.go
```diff
@@ -711,36 +757,50 @@ func (s *Service) NewProtocol(tn *onet.TreeNodeInstance, conf *onet.GenericConfi
 
 		// Wait for DKG in reshare mode to end
 		go func(id byzcoin.InstanceID) {
+			// TODO: properly propagate errors during execution of DKG protocol
+			// (see dedis/cothority#2320)
 			<-setupDKG.Finished
+
+			setValidPeers := func(r *onet.Roster) {
+				s.SetValidPeers(s.NewPeerSetID(id[:]), r.List)
```
ineiti, line 765: As the argument is always the same, couldn't you just write:
```
setValidPeers := func(){
    s.SetValidPeers(s.NewPeerSetID(id[:]), s.storage.Rosters[id].List)
}
```

calypso/service.go
cgrigis: Since the rest of the code uses `xerrors`, I'd rather use that for conformity. We can open a new PR that updates all errors if needed.

calypso/service.go
cgrigis: I could, but I find it clearer to be explicit at the call point.

calypso/service.go
cgrigis: Good point. It's not entirely clear to me the message exchanges that happen during the reshare.
Doesn't the protocol execution below involve only the nodes from the new roster? (`roster.GenerateNaryTreeWithRoot()`)

calypso/service.go
```diff
@@ -434,6 +450,36 @@ func (s *Service) ReshareLTS(req *ReshareLTS) (*ReshareLTSReply, error) {
...
+	newRoster, ltsID, err := s.getLtsRoster(&req.Proof)
+	if err != nil {
+		return nil, xerrors.Errorf("retrieving roster: %v", err)
+	}
+
+	s.storage.Lock()
+	currentRoster := s.storage.Rosters[ltsID]
+	s.storage.Unlock()
+
+	// Build the new set as the union of the current and new rosters
```
ineiti, line 471: This also stems from my wrong understanding of the calypso re-sharing. We don't need the union here, only the new roster.

calypso/service.go
```diff
@@ -711,36 +757,50 @@ func (s *Service) NewProtocol(tn *onet.TreeNodeInstance, conf *onet.GenericConfi
...
+					setValidPeers(s.storage.Rosters[id])
 					s.storage.Unlock()
 					log.Error("the reshared public point is different")
 					return
 				}
 			}
 			s.storage.Shared[id] = shared
 			s.storage.DKS[id] = dks
+			s.storage.Rosters[id] = roster
+			setValidPeers(roster)
```
ineiti, line 799: My wrong understanding of the calypso re-sharing: the new roster should already be set here...

calypso/service.go
ineiti: Yes, you're right. I was under the impression that you can do a reshare to a new roster that is completely disjoint of the old one. But in fact you need to have a threshold number of old nodes in the new roster, else the secret cannot be reshared. So the new roster includes a threshold of the old roster, but the other nodes from the old roster will not be contacted.

cgrigis: Thank you, that makes it clearer. :)

calypso/service.go
cgrigis: Changed to only use the new roster

calypso/service.go
cgrigis: Removed `setValidPeers()` call

calypso/service.go
```diff
@@ -711,36 +746,49 @@ func (s *Service) NewProtocol(tn *onet.TreeNodeInstance, conf *onet.GenericConfi
 
 		// Wait for DKG in reshare mode to end
 		go func(id byzcoin.InstanceID) {
+			// TODO: properly propagate errors during execution of DKG protocol
+			// (see dedis/cothority#2320)
 			<-setupDKG.Finished
+
+			setValidPeers := func(r *onet.Roster) {
```
ineiti, line 753: I think this should be called in _all_ error-returns under `case calypsoReshareProto`, so also on line 743, 715,707, 704. Does that make sense?

cgrigis: Indeed, that makes sense.
In the same spirit, shouldn't it also be called in all the error cases in `ReshareLTS()`, such as lines 406, 409, 418, etc? (I could not find where protocol execution in general is described, a pointer to the documentation is welcome :) )

ineiti: I think the protocol itself is not described in the code. You'd have to deep-dive into the calypso-paper and follow the references to the rabin/petersen (I'm not even sure which one is used) to understand what's really going on.

Also, we don't have a real use-case for the moment. So any of the following is OK for me:
1. stop here - the happy-path works, and we don't have a real use-case for it for the moment
2. just add the `SetValidPeers` to `NewProtocol` in case of error
3. do the same for `ReshareLTS`

cgrigis: > I think this should be called in _all_ error-returns under `case calypsoReshareProto`, so also on line 743, 715,707, 704. Does that make sense?

Actually, for 704 and 707, I don't yet have the ltsID, so I can't access the previous Roster

cgrigis: Actually, I think I will go with 1), because the more I think about it, the more things should be changed (error in 743 is in a function executed later in `pedersen`, an error in 339 means we have to rollback the changes on the successful nodes, etc.)

jeffallen:
@cgrigis can you please upgrade to Onet v3.2.4 in this PR? It has minor changes, but better to pick it up now than to forget about it until later.

PR #2288, opened 2020-05-25
make sure that a latest block is really the latest

ineiti:
When trying to catch-up, it is possible that a node has a wrong view of the
skipchain. This PR makes sure that all nodes agree on the latest
available block.

### DIFF
--- a/byzcoin/bcadmin/cmd_db.go
+++ b/byzcoin/bcadmin/cmd_db.go
@@ -43,17 +43,22 @@ func dbStatus(c *cli.Context) error {
 
 // dbCatchup uses the live byzcoin chain to update to the latest blocks
 func dbCatchup(c *cli.Context) error {
-	if c.NArg() < 3 {
+	if c.NArg() < 1 {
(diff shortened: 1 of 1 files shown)


byzcoin/bcadmin/cmd_db.go
```diff
@@ -43,17 +43,22 @@ func dbStatus(c *cli.Context) error {
...
+	if c.NArg() < 1 {
 		return xerrors.New("please give the following arguments: " +
-			"conode.db byzCoinID url")
+			"conode.db [byzCoinID [url]]")
 	}
 	fb, err := newFetchBlocks(c)
 	if err != nil {
 		return xerrors.Errorf("couldn't create fetchBlock: %+v", err)
 	}
-	err = fb.addURL(c.Args().Get(2))
```
tharvik, line 54: does nothing when only two args?

ineiti: The two args case is handled in `newFetchBlocks`.

PR #2295, opened 2020-06-03
Lts roster

ineiti:
When creating a new LTS instance on byzcoin, the roster used for the re-encryption can be different from the roster used by byzcoin.

This PR adds the possibility to `csadmin` to indicate a different roster. It also contacts the correct set of nodes for the different operations.

### DIFF
--- a/calypso/csadmin/commands.go
+++ b/calypso/csadmin/commands.go
@@ -118,6 +118,10 @@ var cmds = cli.Commands{
 								Usage:  "the ByzCoin config to use (required)",
 							},
 							cli.StringFlag{
+								Name:  "roster",
+								Usage: "change the roster for the calypso",
+							},
+							cli.StringFlag{
 								Name:  "darc",
 								Usage: "DARC with the right to create an LTS (default is the admin DARC)",
 							},
--- a/calypso/csadmin/commands.go
+++ b/calypso/csadmin/commands.go
@@ -118,8 +118,9 @@ var cmds = cli.Commands{
 								Usage:  "the ByzCoin config to use (required)",
 							},
 							cli.StringFlag{
-								Name:  "roster",
-								Usage: "change the roster for the calypso",
+								Name: "roster",
+								Usage: "the path of a roster file to be used as argument for the spawn. " +
+									"If not provided the config roster is used (optional)",
 							},
 							cli.StringFlag{
 								Name:  "darc",
--- a/calypso/csadmin/main.go
+++ b/calypso/csadmin/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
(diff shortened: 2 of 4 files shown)


calypso/csadmin/main.go
```diff
@@ -133,9 +134,23 @@ func dkgStart(c *cli.Context) error {
...
+	}
+	err = protobuf.Decode(ltsBuf, &ltsInfo)
+	if err != nil {
+		return fmt.Errorf("couldn't decode the Lts-instance: %v", err)
+	}
+
 	reply := &calypso.CreateLTSReply{}
 	oc := onet.NewClient(cothority.Suite, calypso.ServiceName)
-	err = oc.SendProtobuf(cl.Roster.List[0], &calypso.CreateLTS{
+	log.Info("Sending CreateLTS to", ltsInfo.Roster.List[0].URL)
```
nkcr, line 152: This might break someone that uses the `--export` and expects only the public key to be printed.

calypso/csadmin/commands.go
```diff
@@ -117,6 +117,10 @@ var cmds = cli.Commands{
 								EnvVar: "BC",
 								Usage:  "the ByzCoin config to use (required)",
 							},
+							cli.StringFlag{
+								Name:  "roster",
+								Usage: "change the roster for the calypso",
```
nkcr, line 122: for me something like "the path of a roster file to be used as argument for the spawn. If not provided the config roster is used (optional)" would be more understandable.

calypso/csadmin/main.go
ineiti: added an `if !export` in front of the log.Info. Generally speaking, I'm always at a loss to know what's happening when I use the `csadmin`, `bcadmin` commands. So I started adding more `log.Info` to them.

Do you already have a guideline for the dedis/dela, how you can get data out of the CLIs? JSON on stdout? having a `--output=file`? Just curious.

nkcr: no we don't have one yet. For me the best way to get data out of a cli is using stdout, therefore I tend to not print any log.

nkcr:
no worries, but it's not Gaylor, it's Noémien 😄

ineiti:
> no worries, but it's not Gaylor, it's Noémien 😄

Oups - sorry.

calypso/csadmin/main.go
ineiti: true - but as a user I really like to know what the tool is up to. Specifically if it's waiting for 5-10 seconds before finishing.

ineiti: or put the logging-output to stderr and the data to stdout?

nkcr: could be a solution. At the end it should also depend on the command. In some cases its ok to print some status logs when the output is not expected to be used prgramatically, like with a wget.

PR #2296, opened 2020-06-03
Publish @dedis/kyber@3.4.3

ineiti:
Updated the bn.js library to the latest version.
Used npm audit fix.


ineiti:
@jeffallen - I want to update kyber, to be able to update cothority afterwards. Do you publish the npm, or shall I do it?

PR #2297, opened 2020-06-03
Verify timestamp of new block

ineiti:
When a block is proposed, the current check is to make sure the age of the block is between
```- 4 * blockInterval...4 * blockInterval```
In the following scenario the upper age of the block is too small:
- slow transactions - because the timestamp needs to be set _before_ executing the transactions
- subleader retries because of nodes that are down

So this PR changes the verification of the maximum age of the block:
- newer blocks are rejected if their timestamp is _before_ the previous block
- newer blocks are rejected if their timestamp is _after_ the node's time + 4 * blockInterval

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -2089,19 +2089,34 @@ func (s *Service) verifySkipBlock(newID []byte, newSB *skipchain.SkipBlock) bool
 			log.Error("Didn't accept the new roster:", err)
 			return false
 		}
+		previous := s.db().GetByID(newSB.BackLinkIDs[0])
+		if previous != nil {
+			var prevHeader DataHeader
+			err := protobuf.Decode(previous.Data, &prevHeader)
+			if err != nil {
+				log.Errorf("couldn't decode previous header: %v", err)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jeffallen** [comment]
Why? (I've no worries about this change, but it would be helpful for change history to know why, not what.)

File: byzcoin/service.go [hunk 1]
```diff
@@ -2089,19 +2089,33 @@ func (s *Service) verifySkipBlock(newID []byte, newSB *skipchain.SkipBlock) bool
 			log.Error("Didn't accept the new roster:", err)
 			return false
 		}
+		previous := s.db().GetByID(newSB.BackLinkIDs[0])
+		if previous != nil {
+			var prevHeader DataHeader
+			err := protobuf.Decode(previous.Data, &prevHeader)
+			if err != nil {
+				log.Errorf("couldn't decode previous header: %v", err)
```
tharvik, line 2097: `return false`?

ineiti: yes, good catch! Thanks!

PR #2301, opened 2020-06-04
Publish cothority 3.4.6

ineiti:
Using new kyber
Ran npm audit fix

Jumping the 3.4.5 version, which was never published. One of the previous commit already changed the version from 3.4.4 to 3.4.5, but it was not ready. So going directly to 3.4.6.

Depends on #2298 #2299


ineiti:
@jeffallen is it OK if I publish @dedis/cothority 3.4.6 after this PR?

PR #2304, opened 2020-06-08
Fix bcadmin tests

### LINKED ISSUE #2303: `bcadmin` test failure
```
$ cd byzcoin/bcadmin
$ ./test.sh 
Working in ./build
Building bcadmin
Building conode

* Testing testReset
Assert FAIL for 'runBA db resetBlock service_storage/eaffa5527d17385eba2069c6ff2c04819e75a2e590dc7dbb099a018cb9e405da.db 88adc1bed3ce7adf1565a83ded787dbe02586cfa84971ee43dfef3e547099291'

* Testing testDbReplay
Assert OK for 'runBA create public.toml --interval .5s'
Assert FAIL for 'runBA db replay conode.db a010f583d330ec3101c03f21cbbcfaff2b93f5d6f20dc78b81277c52db69be4e'
Assert OK for 'runBA db catchup conode.db a010f583d330ec3101c03f21cbbcfaff2b93f5d6f20dc78b81277c52db69be4e http://localhost:2003'
Assert grepping 'Replaying block at index 0' in 'runBA db replay conode.db a010f583d330ec3101c03f21cbbcfaff2b93f5d6f20dc78b81277c52db69be4e'
Assert OK for 'runBA mint config/bc-a010f583d330ec3101c03f21cbbcfaff2b93f5d6f20dc78b81277c52db69be4e.cfg config/key-ed25519:a5c61882db6519f8134d27741693814d8e90bf389237f20514a0583dffe3b243.cfg a5c61882db6519f8134d27741693814d8e90bf389237f20514a0583dffe3b243 1000'
Assert OK for 'runBA mint config/bc-a010f583d330ec3101c03f21cbbcfaff2b93f5d6f20dc78b81277c52db69be4e.cfg config/key-ed25519:a5c61882db6519f8134d27741693814d8e90bf389237f20514a0583dffe3b243.cfg a5c61882db6519f8134d27741693814d8e90bf389237f20514a0583dffe3b243 1000'
W : api.go:361 (skipchain.(*Client).GetUpdateChainLevel) - Got a possibly outdated block from node tls://http://localhost:2003
Assert grepping 'Replaying block at index 0' in 'runBA db replay conode.db a010f583d330ec3101c03f21cbbcfaff2b93f5d6f20dc78b81277c52db69be4e'
Assert grepping 'index 0' in 'runBA db catchup conode.db a010f583d330ec3101c03f21cbbcfaff2b93f5d6f20dc78b81277c52db69be4e --write --blocks 1'

	FAILED: Didn't find 'index 0' in output of 'runBA db catchup conode.db a010f583d330ec3101c03f21cbbcfaff2b93f5d6f20dc78b81277c52db69be4e --write --blocks 1': 
```

ineiti:
- use runBA0 for test.*Grep calls
- fix error report in byzcoin.service.catchupAll
- correctly initialize fetchBlocks in bcadmin

Closes #2303

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -1351,7 +1351,11 @@ func (s *Service) catchupAll() error {
 		// Get the latest block known by the Cothority.
 		reply, err := cl.GetUpdateChain(sb.Roster, sb.Hash)
 		if err != nil {
-			return xerrors.Errorf("getting chain: %v", err)
+			// Might be that the other nodes are not yet up,
+			// so just continue with the other chains.
+			// Call to s.catchup will probably also fail, so skip it.
+			log.Errorf("couldn't get update chain: %+v", err)
+			continue
 		}
 
 		if len(reply.Update) == 0 {
--- a/byzcoin/bcadmin/cmd_db.go
+++ b/byzcoin/bcadmin/cmd_db.go
@@ -606,6 +606,13 @@ func newFetchBlocks(c *cli.Context) (*fetchBlocks,
 		bi, err = hex.DecodeString(c.Args().Get(1))
 		fb.bcID = &bi
 		fb.genesis = fb.db.GetByID(*fb.bcID)
+		if fb.genesis != nil {
+			latest, err := fb.db.GetLatestByID(fb.genesis.SkipChainID())
+			if err != nil {
+				return nil, fmt.Errorf("couldn't get latest block: %v", err)
+			}
+			fb.roster = latest.Roster
+		}
 	} else {
 		err = fb.getBCID()
 		if err != nil {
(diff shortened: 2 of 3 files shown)


byzcoin/service.go
```diff
@@ -1351,7 +1351,11 @@ func (s *Service) catchupAll() error {
 		// Get the latest block known by the Cothority.
 		reply, err := cl.GetUpdateChain(sb.Roster, sb.Hash)
 		if err != nil {
-			return xerrors.Errorf("getting chain: %v", err)
+			// Might be that the other nodes are not yet up,
+			// so just continue with the other chains.
+			// Call to s.catchup will probably also fail, so skip it.
+			log.Errorf("couldn't get update chain: %+v", err)
+			continue
```
tharvik, line 1358: If it is unable to retrieve any chain, it will still return `nil`, maybe return a specific error?

ineiti: That will fail in the scenario where all nodes are down, and you want to restart them. You will not get the network up and running, except if you manage to have everybody start their node at exactly the same moment. This is why I changed this to simply continue, which will allow for the other nodes to come up. If there is a disagreement of which is the latest block, the catchup-procedure will be restarted later.

ineiti: More precise to your question: it's not about retrieving a chain, but updating it. This code wants to make sure that the node has the latest available block. But in the above scenario, this is just not possible, and will make it so that the network will fail.

Found it in one of the integration tests ;)

PR #2310, opened 2020-06-10
Use crypto-browserify

ineiti:
One of the imports in cothority has been using 'crypto', which
makes it fail when used. Replaced with 'crypto-browserify'.

### DIFF
--- a/external/js/cothority/package.json
+++ b/external/js/cothority/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "@dedis/cothority",
-  "version": "3.4.6",
+  "version": "3.4.7",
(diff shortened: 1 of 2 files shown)


ineiti:
> there is the related c4dt/crypto-ts repo where we were able to remove the need for explicit import of crypto-browserify ([c4dt/crypto-ts#1](https://github.com/c4dt/crypto-ts/pull/1)) as it is already taken care of by any decent web packer such as webpack. why not going the other way around and `s/crypto-browserify/crypto/g`?

Yes, that would also be possible. However, I'm tired of this crypto-war, and prefer just to go with this simple patch to keep it all aligned.

PR #2316, opened 2020-06-12
stateChangeStorage background calculation

ineiti:
- Background calculation of stateChangeStorage - else it takes 90s on startup
- Only calculate the scs once in byzcoin/service.go

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -1130,12 +1130,6 @@ func (s *Service) createNewBlock(scID skipchain.SkipBlockID, r *onet.Roster, tx 
 		return nil, xerrors.Errorf("storing block: %v", err)
 	}
 
-	// State changes are cached only when the block is confirmed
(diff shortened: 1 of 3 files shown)


ineiti:
tests pass, but Travis <-> github miscommunication

byzcoin/service.go
```diff
@@ -1130,12 +1130,6 @@ func (s *Service) createNewBlock(scID skipchain.SkipBlockID, r *onet.Roster, tx
 		return nil, xerrors.Errorf("storing block: %v", err)
 	}
 
-	// State changes are cached only when the block is confirmed
-	err = s.stateChangeStorage.append(scs, ssbReply.Latest)
```
Gilthoniel, line 1134: Is there another call that append the states ?

ineiti: Yes, in `byzcoin.Service.updateTrieCallback`. This is called when the block from `StoreSkipblockInternal` is received by the node.

PR #2318, opened 2020-06-12
When using -short, reduce the number of nodes in testing

ineiti:


ineiti:
Communication failure between travis and github...

PR #2321, opened 2020-06-24
Rollup protocol

ineiti:
This replaces the transaction-gathering loop by a 'create-blocks-as-transactions-arrive'.
Instead of querying every `blockInterval` seconds all nodes for new transactions, this PR has all nodes sending their transactions to the leader whenever they get one.
The leader will immediately start a new block with the first transaction. While the block is created, more transactions can queue up. Once the block is done, all queued transactions will be put in a new block.

- 28/07 - rewrote the tx-pipeline to correctly take transactions into account that arrive just after a version change.
- 29/07 - ready for review. Proposed way of looking at it:
  - The two main entry points are:
    - tx_pipeline.go - no use of looking at the diff, it's nearly all brand new...: https://github.com/dedis/cothority/blob/rollup-protocol/byzcoin/tx_pipeline.go
    - AddTransaction - again, pretty much new: https://github.com/dedis/cothority/blob/rollup-protocol/byzcoin/service.go#L374
      - calls important method https://github.com/dedis/cothority/blob/rollup-protocol/byzcoin/service.go#L2667
  - most of the tests are just to make sure the new system works
  - there are a lot of cleanups, removal of `heartbeat`, `collect_tx`, and others, which are not necessarily interesting...

PLEASE DO NOT MERGE YET!


byzcoin/tx_pipeline.go
```diff
@@ -2,263 +2,230 @@ package byzcoin
...
+
+		// Add as many ClientTransactions as possible to the proposedTransactions
+		// before the block gets too big, then put it in the channel.
+		p.txQueue = currentState.addTransactions(p.processor, p.txQueue)
+		if !currentState.isEmpty() {
+			newBlock <- currentState.copy()
+			currentState.reset()
+		}
+	}
+	p.wg.Wait()
```
tharvik, line 139: so, this function never ends, right? why using the `p.wg` and not a IMO more readable `<-make(chan struct{})`?

byzcoin/viewchange_test.go
```diff
@@ -65,26 +67,36 @@ func testViewChange(t *testing.T, nHosts, nFailures int, interval time.Duration)
...
+	log.Lvl1("creating a first tx to trigger the view-change")
+	tx0, err := createOneClientTx(s.darc.GetBaseID(), dummyContract,
+		NewInstanceID([]byte{1}).Slice(), s.signer)
+	require.NoError(t, err)
+	s.sendTxTo(t, tx0, nFailures)
+
+	sl := s.interval * rw * time.Duration(math.Pow(2, float64(nFailures)))
+	log.Lvl1("Waiting for the propagation to be finished during", sl)
+	time.Sleep(sl)
+	log.Print("waiting")
```
tharvik, line 80: debug? `Print` vs `Lvl?`

byzcoin/service.go
```diff
@@ -1941,26 +1967,31 @@ func loadBlockInfo(st ReadOnlyStateTrie) (time.Duration, int, error) {
...
-			scID:        scID,
-			Service:     s,
-		},
+func (s *Service) startTxPipeline(scID skipchain.SkipBlockID) chan struct{} {
+	latest, err := s.db().GetLatestByID(scID)
+	if err != nil {
+		log.Errorf("Error while searching for %x", scID[:])
+		log.Error("DB is in bad state and cannot find skipchain anymore."+
+			" This function should never be called on a skipchain that does not exist.", err)
+		panic("DB is in bad state and cannot find skipchain anymore.")
```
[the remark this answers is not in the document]
ineiti: yes, that was an old one and always bothered me, too ;)

byzcoin/transaction.go
```diff
@@ -118,6 +117,16 @@ func (ctx *ClientTransaction) SignWith(signers ...darc.Signer) error {
 	return nil
 }
 
+// Clone creates a deep clone of the ClientTransaction - mostly used in the
+// tests.
+func (ctx *ClientTransaction) Clone() ClientTransaction {
+	newCtx := ClientTransaction{}
+	for _, inst := range ctx.Instructions {
+		newCtx.Instructions = append(newCtx.Instructions, inst)
+	}
```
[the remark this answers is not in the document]
ineiti: sorry for not merging your suggestions - it's a very useful format, but I don't like `Commit suggestion`, then pull, then rebase, squash, ... - I'm faster writing it myself again... But keep the `suggested change` format, which is very useful!

byzcoin/tx_pipeline.go
ineiti: if the `stopSignal` channel is closed, the function quits. That's the case when the leader changes. Even though, in most cases, when the leader changes it's because the leader is down...

So this is probably mostly used when the leader exits, another leader takes over, the old leader starts again, starts `txPipeline.start`, catches up the blocks, sees that he's no longer the leader, and closes `stopSignal`.

ineiti:
Thanks, I think I addressed most of the comments. Now I'm tempted to try and switch the chain tomorrow - just before 2 weeks of holidays... Probably not a good idea ;)

ineiti:
OK - let's be crazy. I'll merge it and launch an update of the nodes tomorrow morning (4am UTC).

Issue #2259, opened 2020-04-30
move from xerrors to standard errors/fmt


tharvik:
Now that we have 1.13 as the lowest version, `errors` is a standard package. A quite simple `sed` would do the trick.

ineiti:
In fact we need to do the inverse: only `xerrors.New` and `xerrors.Errorf` allow to have the error-stack included...

tharvik:
FYI, there is also https://github.com/JavierZunzunegui/zerrors which is stable, see [discussion on golang](https://github.com/golang/go/issues/29934#issuecomment-591492106).

nkcr:
I also stumbled upon this one: https://github.com/cockroachdb/errors

ineiti:
Oh my - they grow like the ulcers in my stomach when I work on typescript!

- `xerrors` - works so far, even if I'm not a big fan of how it prints the frame
- `zerrors` - new kid on the block - no changes in the last 6 months, method-names seem to mix `Wrap` with the frame
- `errors` - has it all, but might be overkill - also I don't like the namespace clashing with the go `errors`

Please add your vote:
- @ineiti - xerrors

ineiti:
@tharvik why `errors`? Don't you like the trace feature of `xerrors` to know where it all happened?

tharvik:
> @tharvik why `errors`? Don't you like the trace feature of `xerrors` to know where it all happened?

I meant the `errors` from https://github.com/cockroachdb/errors

But yeah, I prefer the basic golang's `errors`. I think we can quite precisely know where an error happened just by looking at the resulting string. Each time a check for error is done, a context is added using `fmt.Errorf`, which is way more readable than a full stacktrace.

ineiti:
> which is way more readable than a full stacktrace.

Hmm - when debugging, I like having a list of methods and files where the error got passed up to. It helps me to faster dig into where the problem might actually be happening. If I just have the context, a la:

`context 1: context 2: context 3: context 4: actual error`

I first have to search for each context, decide if it's actually a problem, and then move on. If I have
```
file1: method1: line1: context 1
file2: method2: line2: context2
file3: method3: line3: context3
file4: method4: line4: actual error
```
I can directly scan the methods and decide where to jump in...

tharvik:
> `context 1: context 2: context 3: context 4: actual error`
> 
> I first have to search for each context, decide if it's actually a problem, and then move on.

Then the contextes aren't helpful/detailed/relevant enough :P

[...]

PR #2324, opened 2020-07-02
Remove xerrors dependencies

ineiti:
- replaced all xerrors.Errorf with fmt.Errorf
- replaced all xerrors.New with errors.New
- removed most of cothority/error.go
- replaced all %+v in fmt.Errorf with %v
- replaced most %w in fmt.Errorf with %v, only kept constant errors

Most of `cothority/error.go` seems to be redundant to what the `errors` package offers. 
- [ ] @Gilthoniel confirms the above sentence ;)

### DIFF
--- a/bevm/bevm_attr.go
+++ b/bevm/bevm_attr.go
@@ -2,6 +2,7 @@ package bevm
 
 import (
 	"encoding/hex"
+	"fmt"
 	"math/big"
 	"reflect"
 	"regexp"
@@ -14,7 +15,6 @@ import (
 	"go.dedis.ch/cothority/v3/byzcoin"
 	"go.dedis.ch/onet/v3/log"
 	"go.dedis.ch/protobuf"
-	"golang.org/x/xerrors"
 )
 
 // BEvmAttrID identifies the name of the `attr` used in DARC rules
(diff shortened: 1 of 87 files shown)


tharvik:
Following issue https://github.com/dedis/cothority/issues/2259

Gilthoniel:
```
 These functions were incorporated into the standard library's errors package in Go 1.13: - Is - As - Unwrap

Also, Errorf's %w verb was incorporated into fmt.Errorf.

Use this package to get equivalent behavior in all supported Go versions.

No other features of this package were included in Go 1.13, and at present there are no plans to include any of them.
```
From the godoc of the xerrors package.

Yes the functions are supported in Go 1.13 but not all the features. We tried to drop xerrors with Dela but the fact is that you loose the benefit of it when it comes to have nice error stacks.

ineiti:
> Yes the functions are supported in Go 1.13 but not all the features. We tried to drop xerrors with Dela but the fact is that you loose the benefit of it when it comes to have nice error stacks.

Oh dang - so I have to do the inverse in fact: replace all `errors.New` with `xerrors.New` and `fmt.Errorf` with `xerrors.Errorf`

ineiti:
This is getting old - closing

PR #2325, opened 2020-07-02
Cleaning up using staticcheck

ineiti:
To make sure the upcoming rollup-protocol branch is correct, I'm using
staticcheck from https://github.com/dominikh/go-tools .
This pass is over the master branch before the rollup-protocol, so that I
have to consider only the errors shown by the rollup-protocol branch.

Currently, staticcheck complains about two things that I didn't fix here:
- `blscosi` is deprecated - but I don't want to change the consensus-protocol at this stage
- unused methods when using `t.Skip` - so this is kind of right, but still didn't fix

### DIFF
--- a/byzcoin/api.go
+++ b/byzcoin/api.go
@@ -380,7 +380,7 @@ func (c *Client) GetDeferredDataAfter(instrID InstanceID, barrier *skipchain.Ski
 
 	header, err := decodeBlockHeader(c.Latest)
 	if err != nil {
-		return nil, xerrors.Errorf("decoding header: %w")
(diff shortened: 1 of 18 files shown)


ineiti:
Yes, staticchecker is OK. The only false negatives I saw currently is that it treats `t.Skip` as `return` and then complains that methods called only later are not used...

tharvik:
I don't see it as false negative, it warns when there is deadcode and theses `t.Skip` are never-return and not conditional; which can be see as pedantic but also don't allow for deadcode to go unoticed.
btw, aren't theses supposed to be fixed by #2130?

PR #2327, opened 2020-07-08
Handle large numeric return values in view method calls

cgrigis:
**What this PR does**

*See commit message*

These changes make the mechanism symmetrical to how arguments are passed to view methods, transactions or contract deployments: it uses strings to encode numbers with a large range (such as `uint256`), so as to preserve precision. The client can decide how to handle the string, e.g. in JavaScript it can be converted to a `BigInt` or a `Long`.

Closes c4dt/TODO#161

---

🙅‍ Friendly checklist:

### DIFF
--- a/bevm/api.go
+++ b/bevm/api.go
@@ -664,6 +664,74 @@ func decodeEvmArray(abiType string, abi abi.Argument, arg interface{}) (
 	return arr.Interface(), nil
 }
 
+// EncodeEvmResult encodes an EVM call result into JSON.
+// This can be useful for command-line tools or calls serialized over the
+// network.
+func EncodeEvmResult(result interface{}, outputs abi.Arguments) (string, error) {
+	var jsonData []byte
+	var err error
+
+	if len(outputs) == 1 {
+		abiType := outputs[0].Type.String()
+		encodedResult, err := encodeEvmValue(abiType, result)
+		if err != nil {
+			return "", xerrors.Errorf("failed to encode EVM value: %v", err)
+		}
+
+		jsonData, err = json.Marshal(encodedResult)
+	} else {
+		resultArr, ok := result.([]interface{})
+		if !ok {
+			return "", xerrors.Errorf("received result type: %v, expected to "+
+				"be a slice", reflect.TypeOf(result))
+		}
+		if len(resultArr) != len(outputs) {
+			return "", xerrors.Errorf("received slice of length %v, expected "+
+				"to be of length %v", len(resultArr), len(outputs))
+		}
+
+		encodedResult := []interface{}{}
+		for i, output := range outputs {
+			abiType := output.Type.String()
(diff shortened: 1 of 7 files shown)


bevm/api.go
```diff
@@ -664,6 +664,74 @@ func decodeEvmArray(abiType string, abi abi.Argument, arg interface{}) (
 	return arr.Interface(), nil
 }
 
+// EncodeEvmResult encodes an EVM call result into JSON.
+// This can be useful for command-line tools or calls serialized over the
+// network.
+func EncodeEvmResult(result interface{}, outputs abi.Arguments) (string, error) {
```
ineiti, line 670: Just for my readability - chose 0-2 of the following points:
- `s/result/value/`
- re-arrange to have the `result/value` always at 2nd argument (preferred) or 1st argument in `[Ee]ncodeEvmResult`

bevm/api.go
```diff
@@ -664,6 +664,74 @@ func decodeEvmArray(abiType string, abi abi.Argument, arg interface{}) (
...
+		if !ok {
+			return "", xerrors.Errorf("received result type: %v, expected to "+
+				"be a slice", reflect.TypeOf(result))
+		}
+		if len(resultArr) != len(outputs) {
+			return "", xerrors.Errorf("received slice of length %v, expected "+
+				"to be of length %v", len(resultArr), len(outputs))
+		}
+
+		encodedResult := []interface{}{}
```
ineiti, line 693: If you `encodeResult := make([]interface{}, len(outputs))`, you can directly assign it on line 696...

bevm/api.go
```diff
@@ -664,6 +664,74 @@ func decodeEvmArray(abiType string, abi abi.Argument, arg interface{}) (
...
+			}
+
+			encodedResult = append(encodedResult, encodedValue)
+		}
+
+		jsonData, err = json.Marshal(encodedResult)
+	}
+
+	if err != nil {
+		return "", xerrors.Errorf("failed to encode result in JSON: %v", err)
```
ineiti, line 708: good ol' `xerrors` ;)

bevm/api.go
cgrigis: Actually, I chose this naming/order in symmetry with `DecodeEvmArgs()` / `decodeEvmValue()`...
The rationale being that the public functions use "Args" or "Result" indicating the direction of the conversion, and the private lower-level functions use "value".

ineiti: Hmm - but if I read the method here and see `result`, I think that this will hold the result of the method.

Can you at least make the comment more clear that `result` is the input parameter here? The first few times I read the comment I didn't understand it.

cgrigis: I guess different people, different logic ;-)
I renamed functions and variables to use the term "return value" instead of "result" to improve clarity.

ineiti: still confusing to me. Anyway...

PR #2331, opened 2020-07-14
Adding new v2-byzcoin requests

ineiti:
The previous requests and instances were difficult to use for multi-instruction transactions.
Also, they didn't use the new Observable interface given by ByzCoin.

This PR adds a v2/ directory with some soon-to-be stable version of Darc and Coin instances.
Every instance can now be created as a BehaviorSubject that updates automatically whenever
a new block is available.
This allows to program a much more reactive programming of the user-interface: instead of
polling for new values, the UI can subscribe to the BehaviorSubjects and be updated
whenever something changes.

In the long run, other contracts should also find their way into byzcoin/v2.

Depends on #2392

### DIFF
--- a/external/js/cothority/spec/helpers/bctest.ts
+++ b/external/js/cothority/spec/helpers/bctest.ts
@@ -25,8 +25,6 @@ export class BCTest {
     static async singleton(): Promise<BCTest> {
         if (BCTest.bct === undefined) {
             BCTest.bct = await BCTest.init();
-        } else {
-            await new Promise((resolve) => setTimeout(resolve, 1000));
         }
 
         return BCTest.bct;
--- a/external/js/cothority/spec/helpers/bctest.ts
+++ b/external/js/cothority/spec/helpers/bctest.ts
@@ -1,0 +1,78 @@
+import { Log } from "../../src";
+import { ByzCoinRPC, IStorage, LocalCache } from "../../src/byzcoin";
+import { Darc, Rule } from "../../src/darc";
+import { RosterWSConnection } from "../../src/network";
+import { StatusRPC } from "../../src/status";
+import { StatusRequest, StatusResponse } from "../../src/status/proto";
+import { TransactionBuilder } from "../../src/v2/byzcoin";
+import { CoinContract, DarcInst } from "../../src/v2/byzcoin/contracts";
+import { BLOCK_INTERVAL, ROSTER, SIGNER, startConodes, stopConodes } from "../support/conondes";
+
+/**
+ * BCTest allows for using a single ByzCoin instance for multiple tests. It should be called with
+ *
+ *   const bct = await BCTest.singleton()
+ *
+ * in every test where a byzcoin-instance is used. Thereafter the test can use the genesisInst
+ * to create new CoinInstances and DarcInstances.
+ *
+ * Using this class reduces the time to test, as the same ByzCoin instance is used for all tests.
+ * But it also means that the tests need to make sure that the genesis-darc is not made
+ * unusable.
+ */
+export class BCTest {
+
+    static async singleton(): Promise<BCTest> {
+        if (BCTest.bct === undefined) {
+            BCTest.bct = await BCTest.init();
+        } else {
+            await new Promise((resolve) => setTimeout(resolve, 1000));
+        }
+
+        return BCTest.bct;
+    }
+    private static bct: BCTest | undefined;
+
+    private static async init(): Promise<BCTest> {
+        Log.lvl = 1;
+        const roster4 = ROSTER.slice(0, 4);
+
+        let usesDocker = true;
+        try {
+            const ws = new RosterWSConnection(roster4, StatusRPC.serviceName);
+            ws.setParallel(1);
+            await ws.send(new StatusRequest(), StatusResponse);
+            Log.warn("Using already running nodes for test!");
+            usesDocker = false;
+        } catch (e) {
+            await startConodes();
+        }
+
+        const cache = new LocalCache();
+        const genesis = ByzCoinRPC.makeGenesisDarc([SIGNER], roster4, "initial");
+        [CoinContract.ruleFetch, CoinContract.ruleMint, CoinContract.ruleSpawn, CoinContract.ruleStore,
+            CoinContract.ruleTransfer]
+            .forEach((rule) => genesis.addIdentity(rule, SIGNER, Rule.OR));
+        const rpc = await ByzCoinRPC.newByzCoinRPC(roster4, genesis, BLOCK_INTERVAL, cache);
+        rpc.setParallel(1);
+        const tx = new TransactionBuilder(rpc);
+        const genesisInst = await DarcInst.retrieve(rpc, genesis.getBaseID());
+        return new BCTest(cache, genesis, genesisInst, rpc, tx, usesDocker);
+    }
+
+    private constructor(
+        public cache: IStorage,
+        public genesis: Darc,
+        public genesisInst: DarcInst,
+        public rpc: ByzCoinRPC,
+        public tx: TransactionBuilder,
+        public usesDocker: boolean,
+    ) {
+    }
+
+    async shutdown() {
+        if (this.usesDocker) {
+            return stopConodes();
+        }
+    }
+}
--- a/external/js/cothority/src/v2/byzcoin/transactionBuilder.ts
+++ b/external/js/cothority/src/v2/byzcoin/transactionBuilder.ts
@@ -1,0 +1,143 @@
+import { Argument, ByzCoinRPC, ClientTransaction, InstanceID, Instruction } from "../../byzcoin";
+import { AddTxResponse } from "../../byzcoin/proto/requests";
+import ISigner from "../../darc/signer";
+
+/**
+ * TransactionBuilder handles collecting multiple instructions and signing them all
+ * together before sending the transaction to the chain.
+ * There are convenience methods to create spawn, invoke, or delete instructions.
+ * Once all instructions are added, the send method will contact one or more nodes
+ * to submit the transaction.
+ * After a call to the send method, the transaction is ready for new instructions.
+ *
+ * If any of the instructions in this transaction fails, all other instructions will
+ * fail, too.
+ */
+export class TransactionBuilder {
+    private instructions: Instruction[] = [];
+
+    constructor(protected bc: ByzCoinRPC) {
+    }
+
+    /**
+     * Signs all instructions and sends them to the nodes.
+     * The `instructions` field is only emptied if the transaction has been accepted successfully.
+     * If the transaction fails, it can be retried.
+     *
+     * @param signers one set of signers per instruction. If there is only one set for multiple
+     * instructions, always the same set of signers will be used.
+     * @param wait if 0, doesn't wait for inclusion. If > 0, waits for inclusion for this many blockIntervals.
+     */
+    async send(signers: ISigner[][], wait = 0): Promise<[ClientTransaction, AddTxResponse]> {
+        const ctx = ClientTransaction.make(this.bc.getProtocolVersion(), ...this.instructions);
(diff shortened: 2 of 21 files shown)


external/js/cothority/tsconfig.spec.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    // Specify the root directory of input files. Use to control the output directory structure with --outDir.
+    "rootDir": "./",
+    "baseUrl": "./"
```
tharvik, line 6: how much adding the `strict` flag makes eveything explose?

external/js/cothority/src/byzcoin/v2/darcsBS.ts
```diff
@@ -0,0 +1,268 @@
...
+     *
+     * @param tx where the instruction will be appended to
+     * @param idSign the IIdentity or the InstanceID of the DARC that is delegated
+     * @param idEvolve per default the same as idSign, but can be given as a different argument
+     * @return the new DARC as it will appear on ByzCoin if the transaction is accepted
+     */
+    setSignEvolve(tx: TransactionBuilder, idSign: IIdentity | InstanceID, idEvolve = idSign): Darc {
+        const rules = this.getValue().rules.clone();
+        rules.setRule(Darc.ruleSign, toIId(idSign));
+        if (idEvolve) {
```
tharvik, line 120: is always != undefined

external/js/cothority/src/byzcoin/v2/darcsBS.ts
```diff
@@ -0,0 +1,268 @@
...
+     * added to the signature rule
+     * @param idEvolve IIdentity or DARC-ID to be added to the evolve rule. Per default the same as the idSign.
+     * @return the new DARC as it will appear on ByzCoin if the transaction is accepted
+     */
+    addSignEvolve(tx: TransactionBuilder, idSign: IIdentity | InstanceID | undefined, idEvolve = idSign): Darc {
+        const rules = this.getValue().rules.clone();
+        if (idSign) {
+            rules.appendToRule(Darc.ruleSign, toIId(idSign), Rule.OR);
+        }
+        if (idEvolve) {
```
tharvik, line 140: is always != undefined

external/js/cothority/tsconfig.spec.json
ineiti: Didn't try that - but that's for another day ;) My goal here is not to revamp the whole of dedis/cothority, but to add a new layer (and hope the bottom layers will disappear someday ;)

external/js/cothority/tsconfig.spec.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    // Specify the root directory of input files. Use to control the output directory structure with --outDir.
```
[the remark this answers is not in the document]
ineiti: Ah, that's why I hate json...

external/js/cothority/src/byzcoin/v2/darcsBS.ts
ineiti: no, you can pass `undefined` in the method.

external/js/cothority/src/byzcoin/v2/darcsBS.ts
ineiti: no, you can pass `undefined` in the method.

external/js/cothority/src/byzcoin/v2/darcsBS.ts
tharvik: if you pass `undefined`, it will get assigned `idSign`, which typewise can't be `undefined`

ineiti:
* v1 / v2 -> for the moment I'd like to keep v1, so for making it clearer, change `byzcoin/v2` to `v2/byzcoin`, and add eventual other services to it, like `v2/skipchain`

external/js/cothority/src/byzcoin/v2/darcsBS.ts
ineiti: OK - need to test - I was sure it works...

ineiti:
I also prefer not to use the following:
```
export namespace Coin {
    export class BehaviorSubject {}
    export class Instance {}
    export namespace Contract {}
}
```

Because of the collision of the names `Coin`, `BehaviorSubject`, and `Instance`. So I build things with the following structure:

```
export class CoinBS{}
export class CoinInst{}
export namespace Contract{}
```

Again, `CoinInstance` is already taken in v1, and to avoid unnecessary suffering of the users, I don't want to use the same name.

external/js/cothority/src/byzcoin/v2/darcsBS.ts
ineiti: Makes sense - tests you are right.

Even though I'd like to see `ts` behave so that I can give `undefined` there ;)

external/js/cothority/src/byzcoin/v2/darcsBS.ts
ineiti: yes, you can pass, but as @tharvik writes, ts decides that `undefined` === no argument...

external/js/cothority/src/byzcoin/v2/darcsBS.ts
tharvik: Haa, that's ECMA standard (and classic JS way to do it); if you want to retain the two behavior, you can spit the function into `setSign` and `setEvolve`

tharvik:
> Again, `CoinInstance` is already taken in v1, and to avoid unnecessary suffering of the users, I don't want to use the same name.

So you expect that people mixes v1 & v2? Then, is it really a new version or more of an extension (so in something like `byzcoin-bs`)?

> Another set I see is:
> 
> * Always the latest data with: `CoinInst extends BehaviorSubject<CoinVersion>`

That's weird that a stream of data is _an_ instance; `CoinStream` maybe or `CoinUpdates`?
I like to play with `Struct & Structs`, the first being an element, the second being a stream/list/container of the first.

> * One version of the data: `CoinVersion extends Coin`

`CoinUpdate`?

external/js/cothority/src/byzcoin/v2/darcsBS.ts
ineiti: I'm trying to take a `Map`, or an `Object` now, with `action: expression` syntax, so that it's easy callable.

The problem with `setSign` and `setEvolve` is that the darc-version would change twice, and then it gets difficult keeping track of the intermediate version...

Currently the system just supposes that you only update the darc once per transaction. If you update it twice, it will be refused, as the versions will not match.

ineiti:
> So you expect that people mixes v1 & v2? Then, is it really a new version or more of an extension (so in something like `byzcoin-bs`)?

Well it all is in the `@dedis/cothority` package, so if two classes have the same name, but come from a different path, confusion is guaranteed.

> That's weird that a stream of data is _an_ instance; `CoinStream` maybe or `CoinUpdates`?

That's the ByzCoin terminology: an instance points to data, and the contract defines how this data can be changed.

> I like to play with `Struct & Structs`, the first being an element, the second being a stream/list/container of the first.

OK, mixing it all well:
- `CoinStruct` - containing `Coin` and `readonly inst byzcoin.Instance`
- `CoinStream` / `CoinInst` - `BehaviorSubject<CoinStruct>`
- `CoinContract` - constants and command-generators

So the `CoinInst` would be in reference to the byzcoin-instance, which by definition is changeable.

external/js/cothority/src/byzcoin/v2/darcsBS.ts
tharvik: Hum, I'm thinking of removing it and pushing people to use `evolve` directly then, but that might be a bit cumbersome.

The easy way would be to have `idEvolve?` as argument then you can do the undefined check yourself, that seems easier.

tharvik:
> > So you expect that people mixes v1 & v2? Then, is it really a new version or more of an extension (so in something like `byzcoin-bs`)?
> 
> Well it all is in the `@dedis/cothority` package, so if two classes have the same name, but come from a different path, confusion is guaranteed.

Right, that makes it confusing for some people..

> * `CoinStruct` - containing `Coin` and `readonly inst byzcoin.Instance`
> * `CoinStream` / `CoinInst` - `BehaviorSubject<CoinStruct>`
> * `CoinContract` - constants and command-generators
> 
> So the `CoinInst` would be in reference to the byzcoin-instance, which by definition is changeable.

That looks more consistent, more readable; good for you?

ineiti:
yes, starts to look OK for me. However, now it complains about the following:

```
export namespace CoinContract {
```
with
```
ERROR: src/v2/byzcoin/contracts/coinInst.ts:108:1 - 'namespace' and 'module' are disallowed
```

Perhaps I'll make it a class with only `static` fields and methods anyway?

tharvik:
Hum, it seems that when using `module`, you can't use `namespace`, you should put any namespace into a new module (ie a new file) and export that you want, which is semantically the same

ineiti:
Seems to be one of these eternal PRs...

external/js/cothority/src/v2/byzcoin/transactionBuilder.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * returns a useful and readable representation of all instructions in this transaction.
+     */
+    toString(): string {
+        return this.instructions.map((inst, i) => {
+            const t = ["Spawn", "Invoke", "Delete"][inst.type];
+            let cid: string;
+            let args: Argument[];
+            switch (inst.type) {
```
tharvik, line 125: Consider updating return type of `Instruction.get type` & remove `Instruction.type*` to smth like `type InstructionType = "spawn" | "invoke" | "delete"`

tharvik (approved):

external/js/cothority/spec/helpers/bctest.ts
```diff
@@ -0,0 +1,78 @@
...
+ * But it also means that the tests need to make sure that the genesis-darc is not made
+ * unusable.
+ */
+export class BCTest {
+
+    static async singleton(): Promise<BCTest> {
+        if (BCTest.bct === undefined) {
+            BCTest.bct = await BCTest.init();
+        } else {
+            await new Promise((resolve) => setTimeout(resolve, 1000));
```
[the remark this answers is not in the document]
ineiti: Added a comment.

external/js/cothority/src/v2/byzcoin/transactionBuilder.ts
tharvik: > but that doesn't seem to be correct TS

Yep, as `Instruction.typeSpawn` is a value of type `number`, not the type `1` (it's kinda close but there is two different DSL, one for values as in JS and one for types for TS).
In the `0 | 1 | 2` case, it creates a type which matches only the `number` value 0, 1 or 2.
In the `"spawn" | "invoke" | "delete"`, it creates a type which matches only the `string` value "spawn", "invoke" or "delete".

> So I kept the numbers - does it look better?

Having the numbers is safer, and using discrete `string` would make it more readable. Having the serialization (ie the representation as `number`) separated from the coder-readable values does allow for more safety and less code-context to keep around. 
Anyway, it's kinda nitpicking, feel free to disregard.

external/js/cothority/spec/helpers/bctest.ts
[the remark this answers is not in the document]
tharvik: Hum, as the test are stopping at the first failed assert, dropping this line would show a test not cleaning correctly or even missing some relevant check. Maybe it was added during developpement and isn't needed now?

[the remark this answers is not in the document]
ineiti: OK, ran it 78 times until travis-timeout without the sleep...

tharvik (approved):

PR #2332, opened 2020-07-15
Use js ethereum libs & improve View Method calls

cgrigis:
**What this PR does**

Rework to use JS libraries instead of methods of the BEvm service to perform Ethereum operations.

* Remove no longer needed service methods.
* Change `PerformCall()` service method to use packed arguments instead, simplifying the code and providing support for large integers (using [BN](https://github.com/indutny/bn.js/)).
* Simplify frontend code by using JS libraries wherever possible.
* Change `PerformCall()` service method to use the local conode's ReadOnlyStateTrie to back the EVM state DB, instead of a ByzCoin client.

---

🙅‍ Friendly checklist:


ineiti:
Great work! I couldn't follow everything, so I just make some naming comments:
- the `service.ts::BEvmService::performCall` should have `view` in it, to correctly separate it with a 'real' call that changes the state
- I would like to see the `service.ts::BEvmClient` either in the classical format and naming scheme of `byzcoin/contracts/coin-instance.ts`, or in the upcoming `v2` format in #2331 
- if you follow the `classical` naming, please follow the same naming as in `calypso`:
  - `service.ts` -> `bevm-rpc.ts`
  - `client.ts` -> `bevm-instance.ts`

And please be my guest to comment #2331 to make sure that the `v2` naming scheme is OK!

cgrigis:
Thank you for the comments!
I've rename things according to the "classical" format, please let me know if I forgot anything.

external/js/cothority/src/bevm/bevm-instance.ts
```diff
@@ -238,27 +205,27 @@ export class BEvmClient extends Instance {
...
+            throw new Error(`mismatch contract name: ${inst.contractID} vs ${BEvmInstance.contractID}`);
         }
     }
 
     /**
      * Set the BEvm service to use
      */
-    setBEvmService(bevmService: BEvmService) {
-        this.bevmService = bevmService;
+    setBEvmRPC(bevmRPC: BEvmRPC) {
```
ineiti, line 227: if we suppose that each node has the bevm-service, is this still needed?

external/js/cothority/src/bevm/bevm-rpc.ts
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @param byzcoinId         ByzCoin ID
+     * @param bevmInstanceId    BEvm instance ID
+     * @param accountAddress    Address of the EVM account
+     * @param contractAddress   Address of the smart contract
+     * @param callData          ABI-packed call arguments
+     *
+     * @return Result of the view method execution
+     */
+    async performViewMethodCall(byzcoinId: Buffer,
```
ineiti, line 43: This would be clearer to me, but if you have a good reasoning for the current name, keep it... s/performViewMethodCall/callViewMethod/ ?

external/js/cothority/src/bevm/bevm-instance.ts
cgrigis: True, we probably don't need to set this from the outside.

Removed `setBEvmRPC()` and initialize the service in `BEvmInstance` constructor.

cgrigis:
> One last question: as you're making some changes to the contract, can you please make sure that `bcadmin db replay` still works? You'll find some instructions and where to get the `cached.db` here: https://github.com/dedis/cothority/tree/master/byzcoin/bcadmin

Working on it... :)

cgrigis:
So I went into `byzcoin/bcadmin`, `go build`, downloaded `cached.db`, then:
```
$ ./bcadmin db catchup cached.db 9cc36071ccb902a1de7e0d21a2c176d73894b1cf88ae4cc2ba4c95cd76f474f3 https://conode.dedis.ch/
(...)
$ ./bcadmin db replay cached.db 9cc36071ccb902a1de7e0d21a2c176d73894b1cf88ae4cc2ba4c95cd76f474f3
(...)
Replaying block at index 67842
Got correct block from 2020-07-21 17:05:44 +0200 CEST with 1 txs, out of which 1 txs got accepted. Tps: 0.0
Successfully checked and replayed all blocks.
```

I am pretty sure this increased the ambient temperature in my room by a couple ⁰C :-)

external/js/cothority/src/bevm/bevm-rpc.ts
ineiti: In fact that's a good argument ;) I like when the go-services and the methods that call it have the same name. It's a manual grpc...

ineiti:
> I am pretty sure this increased the ambient temperature in my room by a couple °C :-)

Shall we make such an article, too?
https://digiconomist.net/bitcoin-energy-consumption/

external/js/cothority/src/bevm/bevm-rpc.ts
ineiti: Starting to see why grpc is better - it does this automatically what I ask you to do manually...

Here we have:
- bevm/api.go - CallEVM
- bevm/api.go - Client.PerformCall
- bevm/proto.go - CallRequest
- bevm/service.go - Service.PerformCall
- cothority/src/bevm/bevm-rpc.ts - BEvmRPC.callViewMethod

Compare this to something like calypso:
- calypso/api.go - Client.Authorize
- calypso/proto.go - Authorize
- calypso/service.go - Service.Authorize
- cothority/src/calypso/calypso-rpc.ts - OnChainSecretRPC.authorize

So if you have to trace the call across ts/go, I find the latter much easier to read... Unfortunately, if we change the `proto.go` definition, the popcoin app will stop working :( So if we decide to change the name here, we should inform them beforehand.

cgrigis:
> Shall we make such an article, too?
> https://digiconomist.net/bitcoin-energy-consumption/

Sure, I'll put it on your TODO list ;-)

external/js/cothority/src/bevm/bevm-rpc.ts
cgrigis: Yes, that's one of the advantages of an RPC framework ;-)

I would say that `CallEVM()` is not related, as it is a function called by several places in BEvm, and just happens to be called by the service implementation of `PerformCall()`.

To me, the important link is indeed between the TS  service client and the Go service implementation function.
The name in `api.go` might not always be related, for example we could have a Client call that is implemented using two or more service calls (to various services even). Actually, I realize that `PerformCall()` in `api.go` is only used by the test, as a Go client will normally use `Call()` (symmetric to `BEvmInstance.call()`); maybe I should just move it to the test?

For the message name itself in `proto.go`, I agree, but there is a pair of messages (request and response). How about `PerformCallRequest` and `PerformCallResponse`? (btw, I see byzcoin code does not seem to follow the naming there, e.g. `AddTransaction()` <-> `AddTxRequest`)

By the way, the `proto.go` is already changed in the PR (removed methods, and `CallRequest`/`CallResponse` members changed). There will anyway be breaking changes for the Popcoin app as the TS API changes as well. But for now, the Popcoin app still refers to the "Stainless" conode for the service... so that should continue to work, right?

ineiti: 1. `CallEVM` - true - that should not be on the list

2. `PerformCall` and `Call` - let's discuss this in the meeting this afternoon

3. Yes, and this is wrong :( Btw I stumble regularly to find the appropriate method for this call, now I know why...

4. Yes, at least for a grace-period. I'd ask them to move to the new API, and if you want you can do the same for your stainless-demo before you update the stainless conode

cgrigis: Renamed as discussed during the meeting.

PR #2338, opened 2020-07-20
js/kyber: mv bn.js to deps

tharvik:
As js/kyber exposes some BN types, it needs to have its type definition. Having it in devs deps doesn't publihs it alongside it. Moving it to normal deps should help.

### DIFF
--- a/external/js/kyber/package-lock.json
+++ b/external/js/kyber/package-lock.json
@@ -5218,7 +5218,7 @@
         },
         "resolve-from": {
           "version": "4.0.0",
-          "resolved": false,
+          "resolved": "",
           "integrity": "sha512-pb/MYmXstAkysRFx8piNI1tGFNQIFA3vkE3Gq4EuA1dF6gHp/+vgZqsCGJapvy8N3Q+4o7FwvquPJcnZ7RYy4g==",
           "dev": true
         },
--- a/external/js/kyber/package.json
+++ b/external/js/kyber/package.json
@@ -33,13 +33,13 @@
   "homepage": "https://github.com/dedis/cothority",
   "dependencies": {
     "@stablelib/blake2xs": "^0.10.4",
+    "@types/bn.js": "^4.11.6",
     "bn.js": "^5.1.2",
     "crypto-browserify": "^3.12.0",
     "elliptic": "^6.4.1",
     "hash.js": "^1.1.3"
   },
   "devDependencies": {
-    "@types/bn.js": "^4.11.6",
     "@types/elliptic": "^6.4.1",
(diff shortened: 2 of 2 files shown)


ineiti:
Can you explain where a problem occurs because of this? I thought that all `@types/` are only dev-dependencies... And all of non-kyber-defined names are mangled by `webpack` anyway, no?

tharvik:
I was bumping deps of service-drynx and it failed because of a missing `@types/bn.js`, which I'm not using direclty), kyber was.
there's some weird stuff happening with the releases of kyber, I don't quite get why there is a need for a tarball containing both a `bundle.min.js` (generated by `webpack`) and other JS files.
anyway, `./constants.d.ts` actually reference `bn.js` so it didn't erase it as wanted.

ineiti:
As far as I understand (but this might be a good pair-programming-hacking-session), both kyber and cothority npms contain code for the browser (`bundle.min.js`) and for usage as a node-module (`index.js`). But I might be wrong, thus the usefulness of a pphs...

And `bn.js` and `@types/bn.js` are two different things, no? So it's normal that `constants.d.ts` references `bn.js`, which is in the dependencies, but not `@types/bn.js`...

tharvik:
> As far as I understand (but this might be a good pair-programming-hacking-session), both kyber and cothority npms contain code for the browser (`bundle.min.js`) and for usage as a node-module (`index.js`). But I might be wrong, thus the usefulness of a pphs...

Good idea, let's try to sort it out one of theses days :)

> And `bn.js` and `@types/bn.js` are two different things, no? So it's normal that `constants.d.ts` references `bn.js`, which is in the dependencies, but not `@types/bn.js`...

Correct, the first one is the implementation, the second, the type definition. However, in your example, it's the type definition of kyber (in `*.d.ts`) which reference other type definition, so it actually need `@types/bn.js`, not `bn.js`. It seems that the implementation in `bundle.min.js` already contains `bn.js`.

> What is the error you get?

see https://travis-ci.com/github/c4dt/service-drynx/builds/176303381
```
node_modules/@dedis/kyber/constants.d.ts:2:16 - error TS7016: Could not find a declaration file for module 'bn.js'. '/home/travis/build/c4dt/service-drynx/library/node_modules/bn.js/lib/bn.js' implicitly has an 'any' type.
Try `npm install @types/bn.js` if it exists or add a new declaration (.d.ts) file containing `declare module 'bn.js';`

import BN from "bn.js";
```

ineiti:
I wouldn't say it makes sense, this being typescript and all. But I see where it's going wrong now, thanks!

PR #2339, opened 2020-07-20
js/kyber: mv @types/elliptic to deps

tharvik:
Same issue and fix than #2338


ineiti:
What I don't understand is why you're the first one to experience this? Did I simply import these types in my projects?

tharvik:
I think that most project does add theses to dev deps, without finding the source of the issue. It is also don't trigger when linking the sources directly, due to the "bundle & unbundle" organisation. And finally, it depends on which file one import from kyber. So, quite hard to track down and why, and an easy fix with `npm i @types/elliptic`.

PR #2340, opened 2020-07-21
Update documentation

### LINKED ISSUE #2337: Better describe release for npms
Currently the npm release description omits the creation of a tag, whereas the https://github.com/dedis/cothority/blob/master/PUBLISH.md mentions usage of tags and deployment for the backend. This should be changed to:

- backend: refer to continuous deployment using https://github.com/c4dt/byzcoin
- frontend: add `git tag` to the description
- frontend: no API breaking changes with patch-levels or new minor versions, except 'experimental' APIs
- frontend: for removal of API or breaking changes, new major version
- frontend: add disclaimer to use `index.ts` instead of files

ineiti:
Closes #2337

@jeffallen - are you OK with us dropping binary releases of the conodes? Or are you aware of users != byzcoin-nodes that depend on the binary releases?

### DIFF
--- a/PUBLISH.md
+++ b/PUBLISH.md
@@ -15,7 +15,7 @@ This pins the library to the latest available development version.
 
 ## Releasing a new version for npm @dedis/*
 
-The npm-packages should follow semantic versioning.
+The npm-packages should follow [semantic versioning](https://semver.org).
 As we don't plan to have any big changes, the major-version is probably stuck
  at `3`, except if we go the linux path and reach 3.9, a 4.0 could follow
  , but more to avoid 3.10 than because anything else - or not.
@@ -35,9 +35,10 @@ As the cothority depends on the kyber package, it's currently a bit akward to
 3. update the kyber-dependency in `cothority/package.json` and the cothority
 -version, which has to be the same as the kyber-version, push to master
 4. publish the new cothority-npm using `cothority/publish.sh`
-5. tag the latest commit using 
+5. use an annotated tag on the latest commit with - this should be signed
+ with your gpg-key 
 ```
-git tag v3.4.6
+git tag -s v3.4.6
 git push origin v3.4.6
 ```
 
--- a/PUBLISH.md
+++ b/PUBLISH.md
@@ -1,33 +1,56 @@
-# Releasing a new version
+# Version management
 
-To be able to do `docker push` you need to have your Docker Hub account added to the dedis organisation.
-Here are the steps we use to make a conode binary release:
+For the latest version of the conodes, please use 
+https://github.com/c4dt/byzcoin, which provides a continuous release of valid
+ conodes.
+
+For the `@dedis/kyber` and `@dedis/cothority` npm libraries, we provide a
+ development and a somewhat stable build.
+The development build is always up-to-date with the latest commit and can be
+ used with a reference to `@dedis/cothority@dev`:
+```bash
+npm i -s @dedis/cothority@dev
 ```
-git checkout master
-git diff -> be sure you have no outstanding changes
-git log
-```
-review to be sure this is where the tag should be
+This pins the library to the latest available development version.
+
+## Releasing a new version for npm @dedis/*
+
+The npm-packages should follow semantic versioning.
+As we don't plan to have any big changes, the major-version is probably stuck
+ at `3`, except if we go the linux path and reach 3.9, a 4.0 could follow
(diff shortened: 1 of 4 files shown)


PUBLISH.md
```diff
@@ -1,33 +1,56 @@
...
-Here are the steps we use to make a conode binary release:
-```
-git checkout master
-git diff -> be sure you have no outstanding changes
-git log
-```
-review to be sure this is where the tag should be
+For the latest version of the conodes, please use 
+https://github.com/c4dt/byzcoin, which provides a continuous release of valid
+ conodes.
```
cgrigis, line 5: Commenting here even though that belongs to `https://github.com/c4dt/byzcoin`:

For the version string associated to the ByzCoin release, I would suggest to have, in addition to the tag and date, the git commit. This would allow to pinpoint precisely to the code that is running. Something similar to what `git describe` produces:
```$ git describe --tags
v3.4.5-85-g3ce8cc8a6
```
What do you think?

PUBLISH.md
```diff
@@ -1,33 +1,56 @@
...
+The development build is always up-to-date with the latest commit and can be
+ used with a reference to `@dedis/cothority@dev`:
+```bash
+npm i -s @dedis/cothority@dev
+```
+This pins the library to the latest available development version.
+
+## Releasing a new version for npm @dedis/*
+
+The npm-packages should follow semantic versioning.
```
cgrigis, line 18: Could provide a link to https://semver.org/, in case people are unfamiliar with it.

PUBLISH.md
```diff
@@ -1,33 +1,56 @@
...
+As the cothority depends on the kyber package, it's currently a bit akward to
+ update the package-version:
+1. update the kyber-version in `kyber/package.json` and push to master
+2. publish the new kyber-npm using `kyber/publish.sh`
+3. update the kyber-dependency in `cothority/package.json` and the cothority
+-version, which has to be the same as the kyber-version, push to master
+4. publish the new cothority-npm using `cothority/publish.sh`
+5. tag the latest commit using 
 ```
 git tag v3.4.6
```
cgrigis, line 40: I would make it an "annotated" tag, possibly even signed.
From the `git tag` documentation:
```
Tag objects (created with -a, -s, or -u) are called "annotated" tags; they contain a creation date, the tagger name
and e-mail, a tagging message, and an optional GnuPG signature. Whereas a "lightweight" tag is simply a name
for an object (usually a commit object).

Annotated tags are meant for release while lightweight tags are meant for private or temporary object labels.
For this reason, some git commands for naming objects (like git describe) will ignore lightweight tags by default.
```

external/js/cothority/README.md
```diff
@@ -4,20 +4,36 @@ This library offers methods to talk to a cothority node. At this point, it
...
-Check out the example in index.html for a browser-based usage
+Check out the example in index.html for a browser-based usage.
+For typescript projects, it is easier to directly import from the subfolders:
 
-# Documentation
+```typescript
+import {DarcInstance} from "@dedis/cothority/byzcoin/contracts"
+```
+
+Avoid importing from the files in the subfolders, as these are not guaranteed
```
cgrigis, line 21: Suggestion:

Do not import directly from _files_ within the subfolders, as they can be moved or renamed at any time, while the index files will always be correct.

PUBLISH.md
ineiti: Sounds good - can you please open an issue at c4dt/byzcoin? You can assign it to me. There might be some more discussion ;)

cgrigis: [Done](c4dt/byzcoin#28)

PUBLISH.md
```diff
@@ -35,9 +35,10 @@ As the cothority depends on the kyber package, it's currently a bit akward to
 3. update the kyber-dependency in `cothority/package.json` and the cothority
 -version, which has to be the same as the kyber-version, push to master
 4. publish the new cothority-npm using `cothority/publish.sh`
-5. tag the latest commit using 
+5. use an annotated tag on the latest commit with - this should be signed
+ with your gpg-key 
 ```
-git tag v3.4.6
+git tag -a v3.4.6
```
cgrigis, line 41:
```suggestion
git tag -s v3.4.6
```

cgrigis: `-a` is unsigned...

cgrigis:
Not yet merging, due to the open question to @jeffallen

PR #2342, opened 2020-07-23
Backporting some of the rollup work

ineiti:
This are different things I found when working on rollup:
- typo in byzcoin/RosterChange.md
- byzcoin/Service.AddTransaction need for checking if the node is closing (less test-failures)
- byzcoin/Service.catchupAll iterating over all nodes to get the latest chain
- byzcoin/*_test.go simplifications and more stability for some tests
- messaging/propagate.go use a start propagation instead of a tree propagation, as the latter cannot handle failures

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -1360,7 +1360,7 @@ func (s *Service) catchupAll() error {
 		for i, node := range sb.Roster.List {
 			cl.UseNode(i)
 			replyTmp, err := cl.GetUpdateChain(sb.Roster, sb.Hash)
-			if err != nil || replyTmp == nil {
+			if err != nil {
 				log.Warn("couldn't get update from", node)
 				continue
 			}
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -379,13 +379,20 @@ func (s *Service) prepareTxResponse(req *AddTxRequest, tx *TxResult) (*AddTxResp
 // error value to find out if an error has occured. The caller must also check
 // AddTxResponse.Error even if the error return value is nil.
 func (s *Service) AddTransaction(req *AddTxRequest) (*AddTxResponse, error) {
+	s.closedMutex.Lock()
+	if s.closed {
+		s.closedMutex.Unlock()
+		return nil, xerrors.New("node is closed")
+	}
+	s.closedMutex.Unlock()
+
 	if len(req.Transaction.Instructions) == 0 {
 		return nil, xerrors.New("no transactions to add")
 	}
 
 	gen := s.db().GetByID(req.SkipchainID)
 	if gen == nil || gen.Index != 0 {
-		return nil, xerrors.New("skipchain ID is does not exist")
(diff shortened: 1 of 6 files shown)


byzcoin/service.go
```diff
@@ -1349,20 +1356,34 @@ func (s *Service) catchupAll() error {
 
 		cl := skipchain.NewClient()
 		// Get the latest block known by the Cothority.
-		reply, err := cl.GetUpdateChain(sb.Roster, sb.Hash)
-		if err != nil {
+		var reply *skipchain.GetUpdateChainReply
+		for i, node := range sb.Roster.List {
+			cl.UseNode(i)
+			replyTmp, err := cl.GetUpdateChain(sb.Roster, sb.Hash)
+			if err != nil || replyTmp == nil {
```
tharvik, line 1363:
```suggestion
			if err != nil {
```

messaging/propagate.go
```diff
@@ -201,21 +201,18 @@ func (p *Propagate) Dispatch() error {
...
+			log.Lvl3(p.ServerIdentity(), "Sending to children", p.Children())
+			if errs = p.SendToChildrenInParallel(&msg.PropagateSendData); len(errs) != 0 {
+				var errsStr []string
+				for _, e := range errs {
+					errsStr = append(errsStr, e.Error())
+				}
+				if len(errs) > p.allowedFailures {
+					return errors.New(strings.Join(errsStr, "\n"))
 				}
+				log.Lvl2("Error while sending to children:", errsStr)
```
tharvik, line 215: Logging after potentially returning?

messaging/propagate.go
```diff
@@ -201,21 +201,18 @@ func (p *Propagate) Dispatch() error {
...
-						return errors.New(strings.Join(errsStr, "\n"))
-					}
-					log.Lvl2("Error while sending to children:", errsStr)
+			}
+			log.Lvl3(p.ServerIdentity(), "Sending to children", p.Children())
+			if errs = p.SendToChildrenInParallel(&msg.PropagateSendData); len(errs) != 0 {
+				var errsStr []string
+				for _, e := range errs {
+					errsStr = append(errsStr, e.Error())
+				}
```
tharvik, line 211:
```suggestion
				errsStr := make([]string, len(errs))
				for i := range errsStr {
					errsStr[i] = e.Error()
				}
```

messaging/propagate.go
```diff
@@ -201,21 +201,18 @@ func (p *Propagate) Dispatch() error {
...
+				if len(errs) > p.allowedFailures {
```
tharvik, line 212: You can move the check earlier, when checking for `len(errs) != 0`; but that depends on what you want to log.

ineiti: Yes, that whole thing is quite crappy anyway: it was supposed to be a fast propagation protocol in a tree-like structure. But then it failed when an intermediate node was down, so I flattened the tree... Now it has lots of stuff that is not really needed anymore.

But, yes, I'd like to log any error I get in there.

PR #2364, opened 2020-08-19
Fixes addJSON

### LINKED ISSUE #2363: @dedis/cothority - Multiple calls to "addJSON" panics
Performing two calls to `addJSON` in `cothority/src/protobuf/index.ts` results in the following error:

```
Error: Root is not a member of Root
```

The following updated test from `cothority/spec/protobuf/index.spec.ts` reproduces the error:

```js
import { Message } from "protobufjs/light";
import { addJSON, registerMessage } from "../../src/protobuf";

class MessageTest extends Message<MessageTest> {
    readonly testField: string;
}

class MessageTest2 extends Message<MessageTest2> {
    readonly testField2: string;
}

describe("Protobuf Tests", () => {
    it("should add the new model to the root", () => {
        const json = {
            nested: {
                TestMessage: {
                    fields: {
                        testField: { type: "string", id: 1 },
                    },
                },
            },
        };

        const json2 = {
            nested: {
                TestMessage2: {
                    fields: {
                        testField2: { type: "string", id: 1 },
                    },
                },
            },
        };

        addJSON(json);
        registerMessage("TestMessage", MessageTest);

        const buf = MessageTest.encode(new MessageTest({ testField: "abc" })).finish();
        const msg = MessageTest.decode(buf);
        expect(msg.testField).toBe("abc");

        // The following panics:
        addJSON(json2);
        registerMessage("TestMessage2", MessageTest2);

        const buf2 = MessageTest2.encode(new MessageTest2({ testField2: "abc" })).finish();
        const msg2 = MessageTest2.decode(buf2);
        expect(msg2.testField2).toBe("abc");
    });
});
```

nkcr:
**What this PR does**

*See commit message*

Fixes #2363


ineiti:
Will the following still work?

https://github.com/c4dt/service-drynx/blob/c6ce72ab88677473f2e087627ad76585e8b1e88a/library/src/client.ts#L17

nkcr:
I tried the fix against the skipchain-explorer, which loads two consecutive models.json, and that works.
https://github.com/dedis/skipchain-explorer/blob/master/src/proto.js

tharvik (approved):

PR #2368, opened 2020-08-21
Some cleanup

ineiti:
Things I saw while working on the Jenkins-patch.
Should not change anything in the processing, mostly cleanup and niceifying of logging-output

### DIFF
--- a/blscosi/protocol/protocol.go
+++ b/blscosi/protocol/protocol.go
@@ -6,8 +6,8 @@
 package protocol
 
 import (
-	"errors"
 	"fmt"
+	"golang.org/x/xerrors"
 	"math"
 	"sync"
 	"time"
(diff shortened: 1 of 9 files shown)


blscosi/protocol/protocol.go
```diff
@@ -133,7 +135,7 @@ func (p *BlsCosi) SetNbrSubTree(nbr int) error {
 	var err error
 	p.subTrees, err = NewBlsProtocolTree(p.Tree(), nbr)
 	if err != nil {
-		return fmt.Errorf("error in tree generation: %s", err.Error())
+		return xerrors.Errorf("error in tree generation: %s", err.Error())
```
tharvik, line 138:
```suggestion
		return xerrors.Errorf("error in tree generation: %v", err)
```

blscosi/protocol/sub_protocol.go
```diff
@@ -218,7 +214,8 @@ func (p *SubBlsCosi) dispatchRoot() error {
 	case <-time.After(p.Timeout):
 		// It might be only the subleader then we send a notification
 		// to let the parent protocol take actions
-		log.Warn(p.ServerIdentity(), "timed out while waiting for subleader response")
+		log.Warn(p.ServerIdentity(),
+			"timed out while waiting for subleader response")
```
tharvik, line 218: I don't get why your editor is always changing the max line width but not consistently, there is a bunch of small and irrelevant such patches. Is there a way to avoid that?

ineiti: There still is a 80-character limit in the code-guidelines, but my editor only imposes it if I edit the line.

We could add this to the `pre-commit` list of things to do. But then I'll propose a 100-character limit...

tharvik: Haa, that's why! Yep, a `pre-commit` is the way to go.

tharvik (approved):

PR #2369, opened 2020-08-24
Skipchain test

ineiti:
In fact the propagation protocol has been broken in the case of
one single node propagating to himself...

### DIFF
--- a/messaging/propagate.go
+++ b/messaging/propagate.go
@@ -200,7 +200,7 @@ func (p *Propagate) Dispatch() error {
 					}
 				}
 			}
-			if !p.IsRoot() {
+			if !p.IsRoot() || len(p.Tree().List()) == 1 {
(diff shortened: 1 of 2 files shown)


messaging/propagate.go
```diff
@@ -200,7 +200,7 @@ func (p *Propagate) Dispatch() error {
 					}
 				}
 			}
-			if !p.IsRoot() {
+			if !p.IsRoot() || len(p.Tree().List()) == 1 {
```
tharvik, line 203:
```suggestion
			if !p.IsRoot() || p.Tree().Size() == 1 {
```

tharvik (commented):
Added a small beautification

tharvik (dismissed):

ineiti:
If the tests pass, please squash and merge...

tharvik (approved):

PR #2371, opened 2020-08-24
insist on node to fetch blocks from

### LINKED ISSUE #2357: Level0 forward-link missing in 52947
from @jeffallen :
```
So I wanted to traverse the entire production skipchain offline to check kyber #434 
against it, and I found that replay stops at block 52944 with this error:
Warning for block 52944: bad forward-link 1/3: to-field points to non-existing block
Successfully checked and replayed all blocks.
4:17 PM
(that’s a lie, there are over 80k blocks!) (edited) 
4:21 PM
gaylor.bosson Isn't it only a warning meaning it used a different link to move forward ? 
4:28 PM
jeff.allen No, it’s failing to walk the chain past 52947 because it has no forward links. 
Corrupted. :disappointed: (edited) 
4:48 PM
jeff.allen Well, I’ve got to go now, but I plan to find a solution for this, I’d like to see if 
it can be lower-cost than the other things we’ve done in the past, i.e. “bcadmin db 
fetch”, where we go fetch one single set of forward links in order to fix a db.
11:13 PM
jeff.allen So I’ve currently got dedis.nella.org offline to repair it. Using “db replay” 
I found that it didn’t have correct level 0 forward links out to the end of the chain, 
and so I used “db catchup”, which seems to have worked ok, though the majority 
of the blocks I downloaded allegedly came from “dedis.nella.org”, which was 
impossible, because my server was down. 
So something is not right in “db replay” logging, still looking for what.
11:14 PM
I am now running the replay again, hopefully it will go out to the end of the chain this 
time. (edited) 
12:53 AM
jeff.allen It is repaired and back up. It looks like tls://conode.c4dt.org:7780 is down 
right now. Also, I got some feedback from the catchup that gasser.blue might 
have missing 0 level links as well.
```

Todo:
- [x] check the logs in `bcadmin` to see why it says it got the blocks from itself
- [x] "db catchup" did not terminate on it's own; it got to the end of the chain and then got into a loop of asking other servers for confirmation

ineiti:
Closes #2357

### DIFF
--- a/byzcoin/bcadmin/cmd_db.go
+++ b/byzcoin/bcadmin/cmd_db.go
@@ -73,7 +73,7 @@ func dbCatchup(c *cli.Context) error {
 	}
 	log.Info("Last index in local db", lastIndex)
 
-	fb.index = 0
+	fb.setNode(0)
 	for {
 		sb, err := fb.gbMulti(latestID)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


tharvik:
I don't see a fix for

>  "db catchup" did not terminate on it's own; it got to the end of the chain and then got into a loop of asking other servers for confirmation

How was it done?

ineiti:
Couldn't reproduce it.

tharvik (approved):

Issue #2080, opened 2019-10-02
Add committed block to viewchange


ineiti:
Discussing this with @LefKok today, I though this issue already exists, but I can't find it anymore. So here you go. When a new block is created, the following happens:

1. the leader proposes a block
2. the followers sign the `propose` and send their signatures to the leader
3. the followers sign the `commit` and send their signatures to the leader
4. the leader sends the block to the followers

If, for any reason, there is a viewchange between 3. and 4., the block is silently dropped, and the chain is effectively forked, as the new leader will create a different block at the same height.

To avoid this, the view-change must include an additional step:
- when asking for a view-change, the nodes must not only indicate their latest block, but also the latest block for which they sent a `commit` signature
- if there is a pending block, then the new leader must collect all `commit` signatures and propagate that new block

For an example where this happened, see #2079

Gilthoniel:
I'm not sure it works for any situation. If the collective signature is only almost accepted, meaning one less signature would have been refused then without the previous leader signature, the block won't be created and thus preventing from moving forward.

ineiti:
> I'm not sure it works for any situation. If the collective signature is only almost accepted, meaning one less signature would have been refused then without the previous leader signature, the block won't be created and thus preventing from moving forward.

Correct - added one step to the issue itself.

Gilthoniel:
I recently noticed that the PBFT implementation using CoSi is not implemented correctly as it is effectively doing two prepare phases. If that was the case, the commit phase would provide the proof that enough nodes have committed the block (n+1). The chain would be approved only if the commit signature exists for the block n+1 so if the leader fails between 3 and 4, the commit signature couldn't exist thus making the proposed block (n+1) not committed. Participants should then accept a view change and continue from the latest fully committed block (n).

For the case where the leader is malicious and keep the commit signature for himself, it is important that a node only signs a commit once for a given leader and only accepts to sign a view change otherwise. In the worst case, a leader could create a valid block but the chain could not continue.

Gilthoniel:
@jeffallen FYI

PR #2380, opened 2020-09-25
More documentation on block creation

ineiti:


byzcoin/README.md
```diff
@@ -104,6 +104,98 @@ The design of the view-change is similar to the view-change protocol in PBFT
...
+1. A client creates a `ClientTransaction` and sends it to one or more nodes.
+ Sending it to more than one node increases the probability that at least
+ one node correctly forwards the `ClientTransaction`.
+
+2. The nodes send the `ClientTransaction` to the leader.
+
+2.a If the leader is not responding, they send the `ClientTransaction` to
+ the other nodes, indicating that the node failed to send the transaction.
+ Every node that fails to send such a transaction to the leader will start
+ requesting a viewchange.
```
pierluca, line 120: previously view change was written as 2 words, sometimes hyphened (line 91, 98). It would be nice to be consistent so that it's easier to navigate the file.

byzcoin/README.md
```diff
@@ -104,6 +104,98 @@ The design of the view-change is similar to the view-change protocol in PBFT
...
+
+3. The leader verifies it's a new `ClientTransaction`, and then puts the
+ `ClientTransction` in a queue. As the client might have sent the
+ `ClientTransaction` to multiple nodes, the leader might receive them more
+ than once, so he has to make sure that only unique `ClientTransaction`s are
```
pierluca, line 125: I'm quite sure the leader is an "it" :)

byzcoin/README.md
```diff
@@ -104,6 +104,98 @@ The design of the view-change is similar to the view-change protocol in PBFT
...
+ possible, assuring that the resulting proposed block is smaller than the
+ `MaxBlockSize`.
+
+6. The leader creates a proposed block and fills the header with the
+ backward-links, the timestamp, and the other information of the skipchain.
+
+7. Then the leader executes every `ClientTransaction` in order, updating the
+ temporary state of the blockchain after every execution. Valid transactions
+ are marked as `Accepted = true` and update the temporary state, while failing
+ transactions are included with `Accepted = false` and don't update the
```
pierluca, line 140: Is "included" the right word here? I assume that if they fail, they never become part of the skip chain , or do they?

edit: I understand that they are from reading later on.
I would avoid creating the ambiguity here though and simply say: "while failing transactions are marked as `Accepted = false`"

byzcoin/README.md
```diff
@@ -104,6 +104,98 @@ The design of the view-change is similar to the view-change protocol in PBFT
...
+9. Then the leader sends the proposed block to all followers. Now every follower
+ verifies that the `ClientTransaction`s actually execute and produce the same
+ `Accepted` flags as given by the leader. The followers also verify that the
+ temporary Merkle tree root hash is the same as given by the leader.
+
+10. Every follower that is OK with the proposed block signs the forward-link
+ from the previous block to the proposed block and sends it to the leader. The
+ signature is quite complicated and described in the ByzCoin paper. It is
+ done over various rounds with a prepare and a commit phase. One long-running
+ bug is the fact that a view-change after a successful prepare phase does not
```
pierluca, line 160: just for my understanding, this bug just delays the addition of a new block with all transactions, right?
I mean, it's inefficient but no data loss is involved, correct?

pierluca (commented):
Minor remarks, but thank you very much for the very clear documentation !

byzcoin/README.md
ineiti: In https://arxiv.org/pdf/1602.06997.pdf it's mostly `view-change`, and only some `view change`, so I'm taking the one with the hyphen.

byzcoin/README.md
ineiti: it/he/she...

byzcoin/README.md
ineiti: Rewrote it - is it clearer now?

byzcoin/README.md
ineiti: See https://github.com/dedis/cothority/issues/2080:
- all the transactions are lost (which is not a problem per se, as the clients cannot rely on the transactions being included in the first place)
- there might be a fork if some of the nodes did include the new block. Up to now we had the failing leader include the block, which made it refuse future blocks

ineiti:
Did the changes - should be OK now.

pierluca (approved):
LGTM

PR #2381, opened 2020-09-25
Adding GetInstanceVersion calls

ineiti:
The calls to recover different instance versions from byzcoin have not been implemented
in the @dedis/cothority library.
This PR adds all the necessary calls to recover older versions of the instances.

### DIFF
--- a/external/js/cothority/package.json
+++ b/external/js/cothority/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "@dedis/cothority",
-  "version": "3.5.3",
+  "version": "3.5.4",
   "description": "A typescript implementation of the cothority",
   "main": "index.js",
   "browser": "bundle.min.js",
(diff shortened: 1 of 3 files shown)


external/js/cothority/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "@dedis/cothority",
-  "version": "3.5.3",
+  "version": "3.5.4",
```
tharvik, line 4: It wont' create a `@dedis/cothority@latest` on NPM, only a prerelease on `@dev`, so `npm i @dedis/cothority` will install version 3.5.2, the current `@latest`.
You can add a [deploy on tags as I did in angular-components](https://github.com/c4dt/angular-components/blob/32772155d3051d80b63edced42e4b54f69fe85d0/.travis.yml#L16-L20) so that it published to `@latest` when tagging a commit.
While I'm at it, I think that you discussed it with @cgrigis some weeks ago so he might want to add to this comment.

ineiti: Yes, there are still some manual operations needed:

https://github.com/dedis/cothority/blob/master/PUBLISH.md

tharvik (approved):

Issue #2367, opened 2020-08-21
Spuriously failing tests


ineiti:
List tests here that are failing from time to time.

Jenkins:
- only from time to time
  - Test_SpawnAndDelete: https://jenkins.cothority.net/job/cothority_master/5949/consoleFull
  - TestAttrBevm: https://jenkins.cothority.net/job/cothority_master/5956/console
  - blscosi/simulation/TestSimulation: https://jenkins.cothority.net/job/cothority_master/5973/consoleFull

Travis:
- `ftcosi.TestSimulation`

PR #2383, opened 2020-09-28
testing travis-failing test

ineiti:
This PR renames the previous `ser` structure for byzcoin testing to `BCTest` and adds a nice builder and runner, so that it can also be run from other byzcoin subdirectories.
The goal is to have a correct shutdown of the failing tests in #2367 .
It is quite a big PR as it needs a lot of refactoring. There are no changes in the byzcoin-code (except cleanups), only in the tests.

The entrypoint for code review is `byzcoin/bctest.go`. It is part of the main-package, which is not ideal, but I'd like it to be used also in other packages. Any idea how to make that happen is welcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,118 +29,13 @@ _:
 dist: trusty
 
 stages:
-  - lint
-  - build
   - test
-  - deploy
 
 jobs:
   include:
-    - stage: lint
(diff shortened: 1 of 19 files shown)


ineiti:
Closed in favour of #2384

PR #2384, opened 2020-10-02
Extract BCTest

ineiti:
This PR makes a nicer BCTest structure that can be used in the different tests.
Its main purpose was to have a correct 'CloseAll' method that waits for all
activity to stop.
But finally it also cleans up a lot of trailing cruft in different tests.

The goal is to have a BCTest structure that can also be used by other modules.
For this reason BCTest is exported in the main package.
For any idea how to make this cleaner, comments are welcome :)

Ran the tests from #2387 in a loop here:
https://travis-ci.com/github/dedis/cothority/jobs/394084093
https://travis-ci.com/github/dedis/cothority/builds/187765748

Closes #2387

PR #2385, opened 2020-10-02
Small fixes from test-branch

### LINKED ISSUE #2386: Spuriously failing test
`byzcoin.contracts.TestDeferred_DefaultExpireBlockIdx` - race detected in `s.working` between `byzcoin.Service.TestClose` and `byzcoin.AddTransaction`

ineiti:
Moving the small fixes to the main code outside of #2384

Closes #2386

### DIFF
--- a/byzcoin/struct.go
+++ b/byzcoin/struct.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"go.dedis.ch/onet/v3/log"
 	"sort"
 	"strings"
 	"sync"
@@ -21,6 +22,12 @@ const defaultMaxSize = (1 << 31) - 1 // maximum 32-bit int
 const versionLength = 64 / 8         // bytes
 const prefixLength = 32              // bytes
 const cleanThreshold = 0.8
+
+// number of blocks in the queue before the service doesn't notify new
+// blocks anymore.
+// This is per-queue, so if one queue doesn't collect its blocks,
+// the other queues continue normally.
+const notificationQueueLenght = 8
 
 var bucketStateChangeStorage = []byte("statechangestorage")
 var errLengthInstanceID = xerrors.New("InstanceID must have 32 bytes")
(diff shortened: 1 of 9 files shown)


byzcoin/struct.go
```diff
@@ -639,6 +639,14 @@ func (bc *bcNotifications) unregisterForBlocks(ch waitChannel) {
 
 	for i, listener := range bc.blockListeners {
 		if listener == ch {
+		emptyChannel:
```
tharvik, line 642: It doesn't really drain a channel, only the available messages. If the channels are supposed to be closed at some point, you can use `for range ch {}`, if not, then, that's harder, there isn't a correct way to do it (it comes down to the Turing halting problem). Maybe add a `close` before breaking to catch channel misuse?

ineiti: True - there might be a block coming in just now. So I'll just close it here and handle that case where it got stuck before:

https://github.com/dedis/cothority/blob/8eec9f1265ba03060878d92f2d88e7dac421de6a/byzcoin/struct.go#L621

ineiti: In fact, it might be enough. The `bc.Lock` _nearly_ assures that no new messages get in the channel. 

The _nearly_ being because of this:
```
// Pseudocode
bc.Lock()
go func(){
  bc.blockListeners[0] <- something
}
bc.UnLock()
```

So there is a small time-window between the `bc.Lock` and the execution of the go-routine, where a new message might come in. But as the main purpose of the drain is to stop the go-routine to be stuck, and the `bc.blockListeners` is of length 1, it would have to race twice the `bc.Lock` and `go func()` before the go-routine gets stuck anyway.

But if you have a cleaner solution, I'm all for it...

ineiti: I could make the wait-channel of length 10 in line 630, and then replace the go-routine in line 620 with a `select` and print a warning if the channel overflows - would that be better?

tharvik: Hum, I'm more versed in the second option, that allow for most of the gouroutine to hopefully finish normally.

ineiti: OK, added a queue that will throw away blocks if they are not fetched in time.

tharvik (approved):

PR #2390, opened 2020-10-06
Use golang-docker to create docker-test

ineiti:
Under MacOSX, the docker-test image doesn't contain the bevm contract, which makes it
impossible to run all tests in external/js/cothority.
This PR uses golang-docker to create the docker-test image, so that also under MacOSX
all tests from external/js/cothority can be run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,6 @@ proto:
 	./proto.sh
 	make -C external
 
-# 	docker run -t -v $(PWD):/cothority neilotoole/xcgo \
-
 docker:
 	docker run -t -v $(PWD):/cothority golang:1.15-buster \
 		bash -c "cd /cothority; go build -o external/docker/conode ./conode"
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@
 
 EXCLUDE_LINT := should be.*UI
 
+TEST_IMAGE_NAME = dedis/conode-test
+
 Coding/bin/Makefile.base:
 	git clone https://github.com/dedis/Coding
 include Coding/bin/Makefile.base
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -25,9 +27,13 @@ proto:
 	./proto.sh
 	make -C external
 
+# 	docker run -t -v $(PWD):/cothority neilotoole/xcgo \
```
tharvik, line 30: Dangling comment?

external/docker/Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM dedis/conode:dev
+FROM golang:1.15-buster
+RUN apt update && apt install -y procps ca-certificates && apt clean
+WORKDIR /root/
+
+EXPOSE 7770 7771
```
tharvik, line 5: It isn't in sync with what `CMD` starts, there is more openned ports.

tharvik (commented):
It might be more readable to use a docker builder image à la `conode/Dockerfile`.

ineiti:
I didn't find a way to build a docker-image with a `Dockerfile` while 'mounting' the local filesystem. Only `COPY` is available in the `Dockerfile`, which might be quite heavy...

tharvik:
> I didn't find a way to build a docker-image with a `Dockerfile` while 'mounting' the local filesystem. Only `COPY` is available in the `Dockerfile`, which might be quite heavy...

Yep, using a local builder might create big temporary images, but if that's an issue, you can use the `.dockerignore` files.
I've recently done a PR on spindle using bother a builder image and ignoring some files: https://github.com/ldsec/spindle/pull/18

Makefile
ineiti: fixed

external/docker/Dockerfile
ineiti: fixed

ineiti:
Is there no way to copy only the "*.go" files to the docker-builder?

tharvik:
> Is there no way to copy only the "*.go" files to the docker-builder?

Try something like the following `.dockerignore`
```
*
!*.go
```

ineiti:
Anyway - thinking this through, to make it nice, I think a lot more would need to be changed. Just adding `Dockerfile` and `.dockerignore` to the root to get two lines out of the `Makefile` doesn't make it nicer.

So if you're OK, I'll leave it like this with the `docker-go` build command in the `Makefile`.

tharvik (approved):

PR #2391, opened 2020-10-08
Add DID to Darcs

jeffallen:
Add DID to Darcs, update proto files. Add support for DIDs into Cothority JS.

Also: Updated protoc version.


ineiti:
Wow - wonderful! I'll have a look at it next week!

Just scanning through the files: I only see the definition of the DID-identity, but not the implementation itself. Is this on purpose?

gnarula:
> Just scanning through the files: I only see the definition of the DID-identity, but not the implementation itself. Is this on purpose?

Yes, the purpose is to have enough to be able to parse it on the skipchain explorer and reserve a spot in the proto files

ineiti:
> Yes, the purpose is to have enough to be able to parse it on the skipchain explorer and reserve a spot in the proto files

Cool - any chance that you also add the rest of the logic to the master branch? I just want to make sure that there is some commitment to finish the implementation, so that we don't have (yet more) unfinished code in the master branch...

gnarula:
> > Yes, the purpose is to have enough to be able to parse it on the skipchain explorer and reserve a spot in the proto files
> 
> Cool - any chance that you also add the rest of the logic to the master branch? I just want to make sure that there is some commitment to finish the implementation, so that we don't have (yet more) unfinished code in the master branch...

The rest of the code is relies on https://github.com/hyperledger/indy-vdr at the moment to resolve a DID I'm afraid. As far as commitment goes, there's some work done to verify Indy proof states and build the merkle trie in another branch but it's still a WIP.

PS: https://github.com/dedis/skipchain-explorer/tree/decode-did depends on changes here and it's required for a project demo on 20th Oct.

nkcr (approved):

ineiti:
The dependency with indy is not a problem per se. BEVM depends on ethereum... Or do you need to have an indy-node running for it to work?

Theoretically I would prefer to have this in a fork until it can join with the full implementation in the cothority. But if it's difficult ot keep a separate fork just for the demo, let's put it in. Except if you never plan to port the rest of the code to the cothority, then I'd prefer to have it in a fork/branch...

One last request though: can you somewhere describe a bit what the future plans are for the did-identity? somewhere like `darc/did.go`?

nkcr (approved):

PR #2393, opened 2020-10-19
Use byzcoin.BCTest

ineiti:
As part of the slow migration, this repalces the bevm.bctTest structure
with the byzcoin.BCTest one. Hope it makes the error go away...


cgrigis (approved):
Thank you!

PR #2394, opened 2020-10-20
Replace URL with url-parse

ineiti:
As the dedis/cothority is not only used in the browser, but also in node
and nativescript, URL cannot be used reliably.

This PR replaces URL with the url-parse implementation.

### DIFF
--- a/external/js/cothority/src/network/websocket.ts
+++ b/external/js/cothority/src/network/websocket.ts
@@ -1,5 +1,6 @@
 import { Message } from "protobufjs";
 import { Observable } from "rxjs";
+import URL from "url-parse";
 import Log from "../log";
 import { IConnection } from "./nodes";
 import { Roster } from "./proto";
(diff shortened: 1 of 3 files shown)


external/js/cothority/src/network/websocket.ts
```diff
@@ -39,18 +40,18 @@ export class WebSocketConnection implements IConnection {
...
         // which will always equal to "/" if there isn't any
         if (url.pathname.slice(-1) !== "/") {
-            url.pathname = url.pathname + "/";
+            url.set("pathname", url.pathname + "/");
         }
         if (url.username !== "" || url.password !== "") {
             throw new Error("addr contains authentication, which is not supported");
         }
-        if (url.search !== "" || url.hash !== "") {
+        if (url.hash !== "") {
```
tharvik, line 48:
```suggestion
        if (url.query !== "" || url.hash !== "") {
```

tharvik (changes requested):

tharvik:
Hum, that's weird, it looks like the types definition for `url-parse` aren't correct. When using `new URL(..., true)`, `query` is indeed an object, but when `new URL(..., false)`, it is a `string` (which is the default behavior btw).. You can cast it to `unknown` before comparing, that should do it.

external/js/cothority/src/network/websocket.ts
```diff
@@ -39,18 +40,18 @@ export class WebSocketConnection implements IConnection {
...
         // which will always equal to "/" if there isn't any
         if (url.pathname.slice(-1) !== "/") {
-            url.pathname = url.pathname + "/";
+            url.set("pathname", url.pathname + "/");
         }
         if (url.username !== "" || url.password !== "") {
             throw new Error("addr contains authentication, which is not supported");
         }
-        if (url.search !== "" || url.hash !== "") {
+        if (url.query !== undefined || url.hash !== "") {
```
tharvik, line 48: The object is never `undefined`. The most precise test would be to check for the returned object size, which does works for both `string` and `Object`.

```suggestion
        if (Object.entries(url.query).length === 0 || url.hash !== "") {
```

ineiti:
Should work now - the condition was `=== 0` instead of `> 0`

tharvik (approved):

PR #2398, opened 2020-11-03
Convert websocket error-codes to text

### LINKED ISSUE #2396: Better error-report in websocket.ts
The current `websocket.ts` `onClose` method does not a good job if the underlying websocket library only returns a `code` without a `reason`.

Suggested fix:
- do a `switch` with all websocket close codes and return the string if no `reason` is given.

https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent

ineiti:
Took https://tools.ietf.org/html/rfc6455 to convert the error-codes of the
websocket to text.

Closes #2396

### DIFF
--- a/external/js/cothority/src/network/websocket.ts
+++ b/external/js/cothority/src/network/websocket.ts
@@ -151,7 +151,7 @@ export class WebSocketConnection implements IConnection {
                             err = `Endpoint terminated the connection due to a protocol error.`;
                             break;
                         case 1003:
-                            err = `Endpoint is terminated the connection
+                            err = `Endpoint terminated the connection
                             because it has received a type of data it cannot accept (e.g., an
                             endpoint that understands only text data MAY send this if it
                             receives a binary message).`;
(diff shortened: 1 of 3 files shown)


external/js/cothority/src/network/websocket.ts
```diff
@@ -141,7 +141,68 @@ export class WebSocketConnection implements IConnection {
...
                 if (code !== 1000 || (reason && reason !== "")) {
-                    sub.error(new Error(reason));
+                    let err = "Unknown close error";
+                    switch (code) {
+                        case 1001:
+                            err = `Endpoint is "going away": server going down or a browser
+                            having navigated away from the page.`;
+                            break;
+                        case 1002:
+                            err = `Endpoint is terminated the connection due to a protocol error.`;
```
nkcr, line 151:
```suggestion
                            err = `Endpoint terminated the connection due to a protocol error.`;
```

external/js/cothority/src/network/websocket.ts
```diff
@@ -141,7 +141,68 @@ export class WebSocketConnection implements IConnection {
...
+                            break;
+                        case 1003:
+                            err = `Endpoint is terminated the connection
```
nkcr, line 154:
```suggestion
                            err = `Endpoint terminated the connection
```

nkcr (approved):

nkcr:
Now that I see those new messages in my apps, I think we should've included the error code as the first part of the message. Like `1001: Endpoint is "going away": server going down or a browser having navigated away from the page.`. That would allow us to display custom error messages if needed.

ineiti:
Feel free to add a PR ;)

PR #2399, opened 2020-11-03
Improve db checking

ineiti:
This PR improves a bit the output of the 'bcadmin db check' command by
removing superfluous messages and differentiating between warnings and
errors.

### DIFF
--- a/byzcoin/bcadmin/cmd_db.go
+++ b/byzcoin/bcadmin/cmd_db.go
@@ -530,7 +530,7 @@ func dbCheck(c *cli.Context) error {
 			}
 			errStrFl := fmt.Sprintf("%s forwardLink at height %d", errStr, i)
 
-			if i > sb.GetForwardLen() || sb.ForwardLink[i].IsEmpty() {
+			if i >= sb.GetForwardLen() || sb.ForwardLink[i].IsEmpty() {
 				log.Warnf("%s missing: should point to %d", errStrFl,
 					index)
 				continue
--- a/byzcoin/bcadmin/cmd_db.go
+++ b/byzcoin/bcadmin/cmd_db.go
@@ -481,12 +481,17 @@ func dbCheck(c *cli.Context) error {
 		if blocks%process == 0 {
 			log.Infof("Processed %d blocks so far", blocks)
 		}
+
+		// Basic sanity check of the block header
 		errStr := fmt.Sprintf("found block %d with", sb.Index)
 		if !sb.Hash.Equal(sb.CalculateHash()) {
 			log.Errorf("%s wrong hash: %x instead"+
 				" of %x", errStr, sb.Hash, sb.CalculateHash())
 		}
 
+		// Check backward-links for all blocks except the genesis-block,
+		// as its backward-link only serves as random input to the hash of
+		// the genesis block.
 		if sb.Index > 0 {
 			if !sb.SkipChainID().Equal(*fb.bcID) {
 				log.Errorf(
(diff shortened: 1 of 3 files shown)


byzcoin/bcadmin/cmd_db.go
```diff
@@ -510,6 +514,23 @@ func dbCheck(c *cli.Context) error {
...
+
+		indexes := sb.GetFLIndexes()
+		for i, index := range indexes {
+			if index <= last.Index {
+				if i >= len(sb.ForwardLink) {
+					maxLink := len(indexes) - 1
+					for ; indexes[maxLink] > last.Index; maxLink-- {
+					}
+					log.Warnf("%s only %d out of %d forward-links",
+						errStr, i, maxLink+1)
```
nkcr, line 526: `i`  is the index of the missing forward link, and `maxLink+1` the index of the highest possible forward link pointing to a block. Therefore, the message should rather be:

```suggestion
						errStr, len(sb.ForwardLink), maxLink+1)
```

byzcoin/bcadmin/cmd_db.go
```diff
@@ -525,12 +546,22 @@ func dbCheck(c *cli.Context) error {
...
-				sb.Roster.ServicePublics(skipchain.ServiceName),
-				sb.SignatureScheme)
-			if err != nil {
-				log.Errorf("%s fails signature verification: %+v",
-					errStrFl, err)
+			if len(next.BackLinkIDs) <= i {
+				log.Errorf("%s points to a block with too few backlinks", errStrFl)
+				continue
+			}
+			if !next.BackLinkIDs[i].Equal(sb.Hash) {
```
nkcr, line 553: too late for that, but this code doesn't breath. Each block of `if` is like a different paragraph of an essay. It needs blanks lines to help the reader understand and rhythm the story.

byzcoin/bcadmin/cmd_db.go
```diff
@@ -510,6 +514,23 @@ func dbCheck(c *cli.Context) error {
...
 			}
 		}
+
+		indexes := sb.GetFLIndexes()
+		for i, index := range indexes {
+			if index <= last.Index {
+				if i >= len(sb.ForwardLink) {
+					maxLink := len(indexes) - 1
+					for ; indexes[maxLink] > last.Index; maxLink-- {
+					}
```
nkcr, line 524:
```suggestion

          // Get the index of the highest possible forwardlink based on the last
          // block.
					for ; indexes[maxLink] > last.Index; maxLink-- {
					}
```

byzcoin/bcadmin/cmd_db.go
ineiti: Nice catch. In fact the forward-links can be even worse than what I wrote here:
Given a block that should have 4 forward-links {0,1,2,3}, it can have only {0,2}, so missing two.

That whole `for`-loop there is bad - I inserted it in the `for`-loop below, I hope that's more clear.

byzcoin/bcadmin/cmd_db.go
ineiti: I have a personal preference of putting similar `if` blocks together, else it get's too 'spacy'. But you're right, it needed some more empty lines... Is this better?

byzcoin/bcadmin/cmd_db.go
ineiti: I removed that one and inserted it in the `for`-loop below.

ineiti:
@nkcr - finally had some time to look at this. Thanks for the comments.

byzcoin/bcadmin/cmd_db.go
```diff
@@ -496,25 +507,36 @@ func dbCheck(c *cli.Context) error {
...
+		// blocks they point to.
+		// If enabled, also verify the signature of the forward-links.
+		indexes := sb.GetFLIndexes()
+		for i, index := range indexes {
+			if index > last.Index {
+				break
 			}
 			errStrFl := fmt.Sprintf("%s forwardLink at height %d", errStr, i)
+
+			if i > sb.GetForwardLen() || sb.ForwardLink[i].IsEmpty() {
```
nkcr, line 533:
```suggestion
			if i >= sb.GetForwardLen() || sb.ForwardLink[i].IsEmpty() {
```

In the case there is one FL in the block instead of 2 for example, in the second iteration, `1 > 1` is false, then you do sb.ForwardLink[1], which will panic.

ineiti: Thanks! I can't believe no test failed for this. Besides not believing I still do those off-by-1-errors :(

nkcr (approved):

PR #2401, opened 2020-11-06
@cothority/byzcoin meaningful instruction print

nkcr:
**What this PR does**

This PR implements a `instruction.getArranged()` function on @cothority/byzcoin that returns a general-purpose JSON representation of an instruction. This representation is contract aware in the sense that arguments should have a meaningful string representation, based on each different contract.

The JSON has the following structure:

```
     *  {
     *      status: 0 | 1,   // Non-zero means an error occured
     *      type: "spawn|invoke|delete|UNKNOWN",
     *      contract: "config|darc|...",
     *      args: [
     *          {
     *              "name": "<NAME>",
     *              "value": "<VALUE>",    // Arg value or its summary
     *              "full": "<FULL VALUE>" // Optional. useful if the argument
     *                                     // have a very long representation.
     *          }
     *          ...
     *      ]
     *  }
```

This functionality is much needed for the Columbus project (blockchain browser). I suggest we include that as an experimental feature until the end of that project (January 2021), where at that point we will have a definitive structure.

Feel free to discuss, I'm open to discussion.

@ineiti's requests:
- [ ] Use an abstract class that the Beautifiers have to implement. Perhaps something like this would work: https://stackoverflow.com/questions/43723313/typescript-abstract-class-static-method-not-enforced
- [x] Give longer names to the classes, this will also avoid any name-clashing:
    - `Config` -> `ConfigBeautify`
    - `Darc` -> `DarcBeautify`
- [x] The tests should pass.
- [x] Squash the commits to have 1 or 2 commits.
- [x] add a `beautifier/index.ts` and import only this. That way you're free to re-arrange any files in there and the user can treat each sub-directory as a package without having to care about the files

### DIFF
--- a/external/js/cothority/src/byzcoin/beautifier/config.ts
+++ b/external/js/cothority/src/byzcoin/beautifier/config.ts
@@ -36,6 +36,9 @@ export class ConfigBeautifier {
                 case "darc_contracts":
                     res.push({name: "darc_contracts", value: arg.value.toString("hex")});
                     break;
+                default:
+                    res.push({name: arg.name, value: "unspecified", full: arg.value.toString("hex")});
+                    break;
             }
         });
 
@@ -52,6 +55,9 @@ export class ConfigBeautifier {
                     // const config = ChainConfig.decode(arg.value);
                     // res.push({name: "darc", value: "chain config", full: config.toString()});
                     res.push({name: "darc", value: "chain config"});
+                    break;
+                default:
+                    res.push({name: arg.name, value: "unspecified", full: arg.value.toString("hex")});
                     break;
             }
         });
--- a/external/js/cothority/src/byzcoin/beautifier/config.ts
+++ b/external/js/cothority/src/byzcoin/beautifier/config.ts
@@ -1,0 +1,61 @@
+import { Darc } from "../../darc";
+import { Roster } from "../../network";
+import { Argument } from "../index";
+import { ChainConfig } from "../index";
+import { IBeautifyArgument } from "./utils";
+// tslint:disable-next-line
+const varint = require("varint");
+
+/**
+ * Arrange arguments for a config contract, ie. provide a meaningful
+ * representation of its arguments.
+ */
+export class ConfigBeautifier {
+    static Spawn(args: Argument[]): IBeautifyArgument[] {
+        const res = Array<IBeautifyArgument>();
+
+        args.forEach((arg) => {
+            switch (arg.name) {
+                case "darc":
+                    const darc = Darc.decode(arg.value);
+                    res.push({name: "darc", value: darc.description.toString(), full: darc.toString()});
+                    break;
+                case "block_interval":
+                    res.push({name: "block_interval", value: `${varint.decode(arg.value, 0) / 1e6} ms`});
+                    break;
+                case "max_block_size":
+                    res.push({name: "max_block_size", value: `${varint.decode(arg.value, 0)} bytes`});
+                    break;
+                case "roster":
+                    const r = Roster.decode(arg.value);
+                    res.push({name: "roster", value: r.id.toString("hex"), full: r.toTOML()});
+                    break;
+                case "trie_nonce":
+                    res.push({name: "trie_nonce", value: `${varint.decode(arg.value, 0)}`});
+                    break;
+                case "darc_contracts":
+                    res.push({name: "darc_contracts", value: arg.value.toString("hex")});
+                    break;
+            }
+        });
+
+        return res;
+    }
+
+    static Invoke(args: Argument[]): IBeautifyArgument[] {
+        const res = Array<IBeautifyArgument>();
+
+        args.forEach((arg) => {
+            switch (arg.name) {
+                case "config":
+                    // Calypso tests won't pass if we use ChainConfig.
+                    // const config = ChainConfig.decode(arg.value);
+                    // res.push({name: "darc", value: "chain config", full: config.toString()});
+                    res.push({name: "darc", value: "chain config"});
+                    break;
+            }
+        });
+
+        return res;
+    }
+}
--- a/external/js/cothority/src/byzcoin/beautifier/utils.ts
+++ b/external/js/cothority/src/byzcoin/beautifier/utils.ts
@@ -2,7 +2,7 @@ import { Argument } from "../index";
 
 export interface IBeautifierSchema {
     status: 0 | 1;
-    type: "spawn" | "invoke" | "delete" | "UNKNOWN";
+    type: "spawn" | "invoke" | "delete";
     contract: string;
     args: IBeautifyArgument[];
 }
--- a/external/js/cothority/src/byzcoin/beautifier/utils.ts
+++ b/external/js/cothority/src/byzcoin/beautifier/utils.ts
@@ -1,0 +1,14 @@
+import { Argument } from "../index";
+
+export interface IBeautifierSchema {
+    status: 0 | 1;
+    type: "spawn" | "invoke" | "delete" | "UNKNOWN";
+    contract: string;
+    args: IBeautifyArgument[];
+}
+
+export interface IBeautifyArgument {
+    name: string;
+    value: string;
+    full?: string;
+}
--- a/external/js/cothority/src/byzcoin/client-transaction.ts
+++ b/external/js/cothority/src/byzcoin/client-transaction.ts
@@ -5,6 +5,7 @@ import { Message, Properties } from "protobufjs/light";
 import IdentityWrapper, { IIdentity } from "../darc/identity-wrapper";
 import Signer from "../darc/signer";
 import { EMPTY_BUFFER, registerMessage } from "../protobuf";
+import { ConfigBeautifier, DarcBeautifier, DefaultBeautifier, IBeautifierSchema } from "./beautifier/index";
 import Instance, { InstanceID } from "./instance";
 
 export interface ICounterUpdater {
@@ -122,9 +123,6 @@ type InstructionType = 0 | 1 | 2;
  * An instruction represents one action
  */
 export class Instruction extends Message<Instruction> {
-    static readonly typeSpawn = 0;
-    static readonly typeInvoke = 1;
-    static readonly typeDelete = 2;
 
     /**
      * Get the type of the instruction
@@ -142,6 +140,9 @@ export class Instruction extends Message<Instruction> {
         }
         throw new Error("instruction without type");
     }
+    static readonly typeSpawn = 0;
+    static readonly typeInvoke = 1;
+    static readonly typeDelete = 2;
 
     /**
      * @see README#Message classes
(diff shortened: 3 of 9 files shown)


external/js/cothority/src/byzcoin/client-transaction.ts
```diff
@@ -364,6 +366,73 @@ export class Instruction extends Message<Instruction> {
...
+     *              "value": "<VALUE>",    // Arg value or its summary
+     *              "full": "<FULL VALUE>" // Optional. useful if the argument
+     *                                     // have a very long representation.
+     *          }
+     *          ...
+     *      ]
+     *  }
+     * 
+     */
+    getArranged(): { [k: string]: any } {
```
ineiti, line 391: The return type should be something more restrictive. From a quick look, I see that all returns contain the `contract` and `args` field.

external/js/cothority/src/byzcoin/arrange/config.ts
```diff
@@ -0,0 +1,57 @@
+import {Argument} from "../index";
+import {Darc as d} from "../../darc";
```
ineiti, line 2: I would not shorten the imports here - it's perhaps a little bit faster to write, but so much harder to read.

external/js/cothority/src/byzcoin/arrange/utils.ts
```diff
@@ -0,0 +1,32 @@
+
+/**
+ * Utility function to arrange arguments on contracts
+ */
+export class Utils {
+
+    // transform a protobuf encoded number to a number
```
ineiti, line 7: I don't think it's a protobuf, but rather a `VarInt`. Also, I would directly import the library from here:

https://github.com/chrisdickinson/varint/blob/master/decode.js

Which keeps the range check that is removed in this copy/paste...

external/js/cothority/src/byzcoin/client-transaction.ts
```diff
@@ -364,6 +366,73 @@ export class Instruction extends Message<Instruction> {
...
+                        res.args = Darc.Invoke(this.invoke.args)
+                        break;
+                }
+
+                break;
+
+            case Instruction.typeDelete:
+                res.type = "delete";
+                break;
+        }
```
ineiti, line 432: Missing `default` case: there should be either a "unknown class", or even an error thrown. Else somebody using the method might be surprised that it doesn't work...

external/js/cothority/src/byzcoin/config.ts
```diff
@@ -46,6 +46,19 @@ export default class ChainConfig extends Message<ChainConfig> {
             },
         });
     }
+
+    /**
+     * Return a string representation of a config
+     */
+    toString(): string {
```
ineiti, line 53: Couldn't that use directly the new pretty-printing?

ineiti:
Very nice. The following parts need to be improved, however:
- programming
  - use interfaces for the classes
  - use interfaces for the data structures
  - add `default` clauses in the `case`s
  - find a better name than `arrange` - pretty-printer, toJSON, beautifier, printDetail, toHTML, ...
- functionality
  - add interpretation of the instance itself
  - add link suggestions to the structures

If it's not to be included anytime soon, you can `Convert to draft`.

external/js/cothority/src/byzcoin/arrange/config.ts
nkcr: I agree. Just followed an ongoing practice in the actual code though.

external/js/cothority/src/byzcoin/config.ts
nkcr: I would rather not. pretty-printing is about printing a transaction, toString() is about printing an instance.

nkcr:
Thank you for the review. I addressed most of the comments:

> use interfaces for the classes

I can't, as all the methods are static

> functionality: ...

I'd like not to implement too much functionality at first, if that's ok I'd like to include this first iteration and add more on top of that when the use cases come along the way.

external/js/cothority/src/byzcoin/beautifier/config.ts
```diff
@@ -0,0 +1,59 @@
+import { Darc } from "../../darc";
+import { Roster } from "../../network";
+import { Argument } from "../index";
+import { ChainConfig } from "../index";
+import { IArgsSchema } from "./utils";
+// tslint:disable-next-line
+const varint = require("varint");
```
ineiti, line 7: I never know - will any of the following work?
```
import * as varint from "varint";
import varint from "varint";
```
???

external/js/cothority/src/byzcoin/beautifier/config.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Arrange arguments for a config contract, ie. provide a meaningful
+ * representation of its arguments.
+ */
+export class Config {
+    static Spawn(args: Argument[]): IArgsSchema {
+        const res = Array<{name: string, value: string, full?: string}>();
```
ineiti, line 15: The `{name...full}` should be an interface, as it's used in many parts.

external/js/cothority/src/byzcoin/beautifier/utils.ts
```diff
@@ -0,0 +1,11 @@
+
+export interface IJSONSchema {
+    status: 0 | 1;
+    type: "spawn" | "invoke" | "delete" | "UNKNOWN";
+    contract: string;
+    args: IArgsSchema;
+}
+
+export interface IArgsSchema {
+    [index: number]: {name: string, value: string, full?: string};
```
ineiti, line 10: What's the advantage of this against
```
export interface IArgument {
  name: string;
  value: string;
  full?: string;
}
```
And then use it like:
```
static Invoke(args: Argument[]): IArgument[]
```
?

external/js/cothority/src/byzcoin/client-transaction.ts
```diff
@@ -320,6 +323,88 @@ export class Instruction extends Message<Instruction> {
...
+    /**
+     * Return a general-purpose json representation of an instruction that is
+     * contracts aware, ie. knows how to render specific contracts arguments. In
+     * the case a problem occurs (for example the requested contract does not
+     * implement a beautify function), the status property is set to non-zero.
+     *
+     * The JSON has the following structure:
+     *
+     *  {
+     *      status: 0 | 1
```
ineiti, line 335: What is the status here? Shouldn't it just throw an error in case of error?

external/js/cothority/src/byzcoin/client-transaction.ts
```diff
@@ -320,6 +323,88 @@ export class Instruction extends Message<Instruction> {
...
+     *      type: "spawn|invoke|delete|UNKNOWN",
```
ineiti, line 336: Why `UNKNOWN`? Is this even a possibility?

external/js/cothority/src/byzcoin/config.ts
ineiti: I suppose you'll need to print instances, too. At least that's what I'm doing in my `ol-explorer`, and it's handy - I had to undo a deletion, again ;)

ineiti (changes requested):
I put the list of things I think are still needed in the PR message, so you can simply tick the ones you did. Of course feel free to argument any or all of the points ;)

external/js/cothority/src/byzcoin/beautifier/config.ts
nkcr: > I never know - will any of the following work?
unfortunately no : /

external/js/cothority/src/byzcoin/client-transaction.ts
nkcr: For the moment non-zero means there is an error, and we only use 1 but we might add additional error code.
I find it more convenient to use than throwing an error.

external/js/cothority/src/byzcoin/client-transaction.ts
nkcr: In theory it should never happen

external/js/cothority/src/byzcoin/config.ts
nkcr: Sure, something we can still add

external/js/cothority/src/byzcoin/beautifier/utils.ts
ineiti: Now that I read my own comment I think that `Argument` and `IArgument` are too close. Would `IArgumentDetail` or something be better?

external/js/cothority/src/byzcoin/client-transaction.ts
ineiti: Hmm - thinking out loud (or whatever applies when writing text):

Is it an error if `beautify` cannot return a nice representation?

Perhaps one solution would be to include a default representation which only sets the type-string and iterates over all arguments and sets the `value` to `key:value.toString('hex')`. Then I think the status would make sense: 
- `0` means only the default representation is returned
- `1` means it has been interpreted specifically
- `2` means it also has links (or whatever)

external/js/cothority/src/byzcoin/client-transaction.ts
ineiti: Then remove it. If the `Instruction` is neither of the three, then it makes sense to `throw new Error`, as this is something that should not happen.

nkcr:
@ineiti tests are passing, comments addressed. The proposed trick to use an interface doesn't work.
The reason why Calypso tests are not passing is still mysterious to me, but I don't have time to investigate more. I made it pass by not using `ChainConfig` in `config.ts`.

external/js/cothority/src/byzcoin/client-transaction.ts
ineiti: Now it's cleaned up, so it's OK. But in general we should try to break code culture only when it goes into 'best practices'. And using `status` instead of `throw new Error` is not a best practice.

ineiti (changes requested):
The switches in the DarcBeautifier and ConfigBeautifier are missing a `default` entry.

nkcr:
@ineiti defaults added.

external/js/cothority/src/byzcoin/beautifier/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { Argument } from "../index";
+
+export interface IBeautifierSchema {
+    status: 0 | 1;
+    type: "spawn" | "invoke" | "delete" | "UNKNOWN";
```
ineiti, line 5:
```suggestion
    type: "spawn" | "invoke" | "delete";
```

ineiti:
Just remove the `UNKNOWN`, and I think we're good.

ineiti (approved):

ineiti:
If you think this is ready, you can move it to R4M. If you think you'll modify it a lot during development of Columbus (which I suppose will happen), you can also teach the student to use a branch of cothority for his development.

PR #2402, opened 2020-11-09
Feature: connectivity check

gnarula:
This PR adds a new protocol that checks connectivity with all other nodes in the roster. The results of the check are cached by default and controlled using the `CONNECTIVITY_TTL` environment variable.

Still a WIP. Need to remove `time.Sleep` in the tests

---

🙅‍ Friendly checklist:

### DIFF
--- a/connectivity/api.go
+++ b/connectivity/api.go
@@ -1,0 +1,27 @@
+package connectivity
+
+import (
+	"go.dedis.ch/cothority/v3"
+	"go.dedis.ch/onet/v3"
+	"go.dedis.ch/onet/v3/network"
+	"golang.org/x/xerrors"
+)
+
+type Client struct {
+	*onet.Client
+}
+
+func NewClient() *Client {
+	return &Client{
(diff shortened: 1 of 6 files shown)


ineiti:
How is this different from https://github.com/dedis/cothority/blob/11cfd645218140693d889d7cb0d8951a82809444/status/service/status.go#L60 ?

gnarula:
I should've grepped or asked you before :) Closing this one and I'll rather propose changes in 'status' if required.

ineiti:
Yes, and it's difficult to grep through it if you don't know. I was also not sure whether this has actually been implemented or not ;) But one line in #engineers with a mention is enough and I'll try to chime in. Or open an issue and mention @ineiti should also work.

Btw: one thing I'm sure the status-thingie doesn't do is to check opening a connection from both ends. E.g.: if you start from a node that is in your local network, it will be able to start connections with all other nodes and they will be able to reply. But if the other nodes will try to contact yours, it will fail.

PR #2404, opened 2020-11-13
Cleaning up mutexes in byzcoin.Service

ineiti:
The use of mutexes in byzcoin.Service got somehow out of hand.
This PR cleans up the use of 'Service.close' and 'Service.working' by
adding two new structures that do all the locking.
It also contains minimal changes to the tests to make them pass.

For review: can you please look at the two structures `runSingleWG` and `tasksWG` in `byzcoin/struct.go` and tell me if they look OK? It simplifies a lot the locking in the `byzcoin.Service`. I hope the tests are relevant ;)
I tried to think of how to merge the two

### DIFF
--- a/byzcoin/service.go
+++ b/byzcoin/service.go
@@ -112,8 +112,8 @@ type Service struct {
 	*onet.ServiceProcessor
 	// stateTries contains a reference to all the tries that the service is
 	// responsible for, one for each skipchain.
-	stateTries     map[string]*stateTrie
-	stateTriesLock sync.Mutex
+	stateTries      map[string]*stateTrie
+	stateTriesMutex sync.Mutex
 	// We need to store the state changes for keeping track
 	// of the history of an instance
 	stateChangeStorage *stateChangeStorage
@@ -141,18 +141,15 @@ type Service struct {
 
 	stateChangeCache stateChangeCache
 
-	closed        bool
-	closedMutex   sync.Mutex
-	working       sync.WaitGroup
+	tasks         tasksWG
 	viewChangeMan viewChangeManager
 
 	streamingMan streamingManager
 
-	updateTrieLock        sync.Mutex
-	catchingLock          sync.Mutex
-	catchingUp            bool
-	catchingUpHistory     map[string]time.Time
-	catchingUpHistoryLock sync.Mutex
+	updateTrieMutex        sync.Mutex
+	catchingUpWG           runSingleWG
+	catchingUpHistory      map[string]time.Time
+	catchingUpHistoryMutex sync.Mutex
 
 	downloadState downloadState
 
@@ -162,8 +159,8 @@ type Service struct {
 
 	// defaultVersion is the new version to use for new
 	// ByzCoin chains.
-	defaultVersion     Version
-	defaultVersionLock sync.Mutex
+	defaultVersion      Version
+	defaultVersionMutex sync.Mutex
 }
 
 type downloadState struct {
@@ -197,9 +194,9 @@ type bcStorage struct {
 // GetProtocolVersion returns the version of the Byzcoin protocol for the current
 // conode.
 func (s *Service) GetProtocolVersion() Version {
-	s.defaultVersionLock.Lock()
+	s.defaultVersionMutex.Lock()
 	v := s.defaultVersion
-	s.defaultVersionLock.Unlock()
+	s.defaultVersionMutex.Unlock()
 	return v
 }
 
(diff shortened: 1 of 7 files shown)


byzcoin/service.go
```diff
@@ -2023,21 +1982,14 @@ func (s *Service) startTxPipeline(scID skipchain.SkipBlockID) chan struct{} {
...
-			return
-		}
-
-		s.working.Add(1)
-		defer s.working.Done()
-		s.closedMutex.Unlock()
-		pipeline.start(&initialState, stopChan)
-	}()
+	if s.tasks.Add(1) {
+		s.stopTxPipelineWG.Add(1)
```
tharvik, line 1986: You don't wrap it in an `if`, is it wanted?

byzcoin/struct.go
```diff
@@ -740,3 +741,75 @@ const (
 	// instances. If not present, missing instances are ignored.
 	GUFSendMissingProofs
 )
+
+// runSingleWG allows to ensure a single execution of a method.
+type runSingleWG struct {
+	sync.WaitGroup
```
tharvik, line 747: I don't think you need to bundle a `sync.WaitGroup` as `running` is already doing the whole work.

byzcoin/struct.go
```diff
@@ -740,3 +741,75 @@ const (
...
+	}
+	wwg.running = false
+	return true
+}
+
+func (wwg *tasksWG) Resume() {
+	wwg.Lock()
+	defer wwg.Unlock()
+	if wwg.running {
+		panic("cannot open a closed wwg")
```
tharvik, line 806:
```suggestion
		panic("cannot resume a running tasksWG")
```

byzcoin/struct.go
```diff
@@ -740,3 +741,75 @@ const (
...
+}
+
+// Check if currently running
+func (cg *runSingleWG) Running() bool {
+	return atomic.LoadInt32(&cg.running) > 0
+}
+
+// tasksWG is a special workGroup that can be paused or resumed.
+type tasksWG struct {
+	sync.Mutex
```
tharvik, line 777: No need to expose the `sync.Mutex` methods, otherwise, someone might do `Lock` the `tasksWG` from outside.

byzcoin/struct.go
```diff
@@ -740,3 +741,75 @@ const (
...
+	cg.WaitGroup.Done()
+}
+
+// Check if currently running
+func (cg *runSingleWG) Running() bool {
+	return atomic.LoadInt32(&cg.running) > 0
+}
+
+// tasksWG is a special workGroup that can be paused or resumed.
+type tasksWG struct {
```
tharvik, line 776: You do not implement `Done` so `tasksWG` can be still running even if all expected children calls `Done`.

tharvik (changes requested):

byzcoin/struct.go
ineiti: That's in a later PR where I need to `runSingleWG.Wait()` to be sure the catchup is done.

byzcoin/struct.go
ineiti: I was hesitating, but because the structure is package-only anyway, so even if I write `m sync.Mutex`, the package itself can still access the mutex.
Well, it's one step more to do a stupid thing, so I added the `m sync.Mutex`...

byzcoin/struct.go
ineiti: OK - probably I need to rename `running` here. It is normal for the structure to have `running = true`, but no tasks are waiting.

I can rename `(r|R)unning` to `(t|T)asksAllowed` - would that make it clearer? Other proposition for the name?

ineiti: renamed and added a comment to the structure. Also lowercased all methods on the structure, except for `Add`, as this one overwrites the one from `sync.WaitGroup`.

byzcoin/struct.go
tharvik: Okay, makes sense then. So I'd recommend to avoid exposing `sync.WaitGroup` direclty now, as you define `done` which doesn't override `sync.WaitGroup.Done`, one can go negative by running the latter before the former. And one can `sync.WaitGroup.Add` anything.

byzcoin/struct.go
tharvik: The more steps needed to do something stupid, the less likely it will happen :) If you want to push it further, you can create a new sub-package containing only theses primitives

But I don't see the `m sync.Mutex` commit, so..

ineiti: branch mixup - removed all embedding and replaced with references. IIRC, Christian also prefers those.

byzcoin/struct.go
ineiti: Btw, a negative WG-counter panics: https://play.golang.org/p/H-jQb6HAgs4

tharvik (approved):

PR #2405, opened 2020-11-17
Updates @cothority/byzcoin import

nkcr:
In the beautifier I couldn't use ChainConfig for some strange reasons because it would fail the Calypso tests. This situation is fixed by changing the  ClientTransaction imports.


ineiti:
Nice catch! Any idea why that happened? I looked at it but couldn't see why this helps...

nkcr:
No idea, this is some kind of sorcery to me 🧙‍♂️.
I got this intuition because there were an error from calypso-rpc.ts saying that `ClientTransaction` were undefined.

nkcr:
Be aware that this change removes the default export of `ClientTransaction`. This change goes toward only using `index` for imports, which is something that already brought small incompatibility in the past when Linus made the change in some parts, without too much trouble (but a lot of frustration..).

tharvik (changes requested):
You actually don't need to import from `[...]/byzcoin/index` but simply from `[...]/byzcoin`, as the `index.ts` is already exporting what's needed.

ineiti:
Waiting for cleanup of `/index` imports

ineiti (commented):
Like @tharvik - remove the `/index` from the imports.

external/js/cothority/src/byzcoin/beautifier/config.ts
```diff
@@ -1,7 +1,7 @@
+import { Argument } from "../";
+import { ChainConfig } from "../";
```
ineiti, line 2:
```suggestion
import { Argument, ChainConfig } from "../";
```

tharvik (changes requested):
Can also you remove the dangling slash at the end of imports path?

tharvik (approved):

PR #2407, opened 2020-11-19
(js/cothority) Fixes bug in beautify

nkcr:
**What this PR does**

*See commit message*

### DIFF
--- a/external/js/cothority/src/byzcoin/client-transaction.ts
+++ b/external/js/cothority/src/byzcoin/client-transaction.ts
@@ -357,14 +357,14 @@ export class Instruction extends Message<Instruction> {
                 switch (this.spawn.contractID) {
                     case "config":
                         res.contract = "config";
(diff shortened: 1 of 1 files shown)


ineiti (approved):
I really suggest you show your student how to work on this library locally. It will be much faster to work with if you don't have to go through a complete PR/merge session for every change...

nkcr:
> I really suggest you show your student how to work on this library locally. It will be much faster to work with if you don't have to go through a complete PR/merge session for every change...

True, I might do that. Would be better if the instruction to do that (the one that works, with `npm run link`) hadn't been removed from the README 👀 .

ineiti:
They didn't work for me, so I erronously supposed they don't work for others, neither... Here is the `old way`:

https://github.com/dedis/cothority/blob/0f895e0de01dc8d8f5eacf5a810ce7ab6354e42d/external/js/cothority/README.md

PR #2408, opened 2020-11-20
Fixes timeout in Websocket

nkcr:
Clearing the timeout doesn't set it back. This would make the websocket connection have a connection error (coming from Onet I guess..?) instead of the expected timeout.

Also added a `npm run watch` to automatically compile files on changes.

Fixes https://github.com/dedis/columbus-united/issues/40

### DIFF
--- a/external/js/cothority/package.json
+++ b/external/js/cothority/package.json
@@ -12,6 +12,7 @@
   "types": "index.d.ts",
   "scripts": {
     "build": "npm run protobuf && tsc",
+    "watch": "tsc --watch",
     "linter": "tslint 'src/{,**/}*.ts' 'spec/{,**/}*.ts' && prettier --check *.html",
     "linter:fix": "tslint --fix 'src/{,**/}*.ts' 'spec/{,**/}*.ts' && prettier --write *.html",
     "test": "npm run linter && npm run cover",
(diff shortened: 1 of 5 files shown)


external/js/cothority/package.json
```diff
@@ -12,6 +12,7 @@
   "types": "index.d.ts",
   "scripts": {
     "build": "npm run protobuf && tsc",
+    "watch": "tsc --watch",
```
tharvik, line 15: FYI, you can use `npm run build -- --watch` to obtain the same expected behavior.

tharvik (commented):
Travis is failing, can you see what it is about?

nkcr:
> Travis is failing, can you see what it is about?

Yes, I'm on it.

external/js/cothority/package.json
nkcr: Ok, good to know !
I think it doesn't hurt having an explicit command, it can be of a great help for non-initiated javascript makers (like me ^^).

tharvik: I agree, the small the barrier to entry, the more people can participate :)

ineiti (approved):

ineiti:
It's still in WIP, so I'll not merge it...

PR #2411, opened 2020-11-25
status: allow exporting ConnectivityCheck results over HTTP

gnarula:
Adds a command that initiates ConnectivityCheck every `interval`
and serves the results in the form of prometheus metrics.

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -33,7 +33,7 @@ const prometheusTemplate = `
 # TYPE voting_conodes_status gauge
 voting_conodes_status{} {{ .Connectivity }}
 
-# HELP voting_conodes_status_timestamp timestamp in milliseconds since epoch denoting last probe time
+# HELP voting_conodes_status_timestamp timestamp in seconds since epoch denoting last probe time
 # TYPE voting_conodes_status_timestamp counter
 voting_conodes_status_timestamp{} {{ .LastCheckedAt }}
 
--- a/status/status.go
+++ b/status/status.go
@@ -99,7 +99,9 @@ func main() {
 			Subcommands: []cli.Command{
 				{
 					Name:  "serve",
-					Usage: "exposes the connectivity check results on HTTP. Assumes -findFaulty",
+					Usage: "exposes the connectivity check results on HTTP. Assumes -findFaulty\n   " +
+						"Note: The HTTP server does not rate limit connections. Administrators are advised\n   " +
+						"to rate-limit connections or firewall the HTTP port themselves.",
 					Flags: []cli.Flag{
 						cli.StringFlag{
 							Name:  "format, fo",
--- a/status/status.go
+++ b/status/status.go
@@ -98,7 +98,7 @@ func main() {
 			},
 			Subcommands: []cli.Command{
 				{
-					Name:  "serve",
+					Name: "serve",
 					Usage: "exposes the connectivity check results on HTTP. Assumes -findFaulty\n   " +
 						"Note: The HTTP server does not rate limit connections. Administrators are advised\n   " +
 						"to rate-limit connections or firewall the HTTP port themselves.",
--- a/status/status.go
+++ b/status/status.go
@@ -6,12 +6,18 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
+	"os/signal"
 	"sort"
 	"strings"
+	"sync"
+	"syscall"
+	"text/template"
 	"time"
 
 	cli "github.com/urfave/cli"
@@ -21,6 +27,21 @@ import (
 	"go.dedis.ch/onet/v3/log"
 	"go.dedis.ch/onet/v3/network"
 )
+
+const prometheusTemplate = `
+# HELP voting_conodes_status voting global conodes status: X is the number of conodes; X >= 0.66 OK, 0 < X < 0.66 KO
+# TYPE voting_conodes_status gauge
+voting_conodes_status{} {{.Connectivity}}
+# HELP voting_conode_status voting conode status: 1=OK, 0=KO
+# TYPE voting_conode_status gauge
+{{- range $conode, $status := .Matrix -}}
+{{ if eq $conode $.Self }}
+voting_conode_status{conode="{{ $conode}}", critical="true"} {{$status}}
+{{ else }}
+voting_conode_status{conode="{{ $conode}}", critical="false"} {{$status}}
+{{ end }}
+{{- end -}}
+`
 
 func main() {
 	app := cli.NewApp()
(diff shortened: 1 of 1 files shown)


ineiti (approved):
I think I like that approach very much: do I correctly understand that you put the charge of fetching the status and sending it to prometheus on the `status` binary?

In our current systems, this would work best as a one-shot thing: call `status serve` once every 5 minutes. But YMMV...

pierluca:
> I think I like that approach very much: do I correctly understand that you put the charge of fetching the status and sending it to prometheus on the `status` binary?
> 
> In our current systems, this would work best as a one-shot thing: call `status serve` once every 5 minutes. But YMMV...

This is not how prometheus works. Prometheus works by scraping various systems and letting them expose their status.
I.e. Prometheus just needs a few addresses to scrape, but every service exposes metrics that are relevant for itself.
As a matter of fact, "sending to prometheus" is considered an anti-pattern and only recommended for ephemereal / batch jobs ( https://prometheus.io/docs/practices/pushing/ ) 

As such, this starts a webserver that exposes the metrics on a webpage.

status/status.go
```diff
@@ -22,6 +28,21 @@ import (
...
+# TYPE voting_conodes_status gauge
+voting_conodes_status{} {{.Connectivity}}
+# HELP voting_conode_status voting conode status: 1=OK, 0=KO
+# TYPE voting_conode_status gauge
+{{- range $conode, $status := .Matrix -}}
+{{ if eq $conode $.Self }}
+voting_conode_status{conode="{{ $conode}}", critical="true"} {{$status}}
+{{ else }}
+voting_conode_status{conode="{{ $conode}}", critical="false"} {{$status}}
+{{ end }}
```
pierluca, line 42: Even if it was not requested by EPFL, I would add a field to expose the timestamp at which we last checked connectivity.

It is a typical Prometheus pattern and allows one to detect if the connectivity check stopped for some reason, and we keep serving the same old vlaues.

status/status.go
```diff
@@ -244,3 +294,153 @@ func printJSON(all []se) {
...
+
+				resp, err := status.NewClient().CheckConnectivity(
+					s.si.GetPrivate(),
+					s.list,
+					s.timeout,
+					true,
+				)
+
+				if err != nil {
+					log.Errorf("error checking connectivity: %s", err)
```
pierluca, line 416: For debugging purposes, I'm wondering:
should we output this as a comment on the prometheus page?
It might help DEDIS debug problems without having access to the VPSI server.

pierluca (commented):
LGTM, minor remarks to improve our debugging capabilities

status/status.go
```diff
@@ -244,3 +294,153 @@ func printJSON(all []se) {
...
+		list:     list,
+		si:       si,
+		quit:     make(chan struct{}),
+	}
+
+	http.HandleFunc(endpoint, s.serveHandler)
+	httpServer := &http.Server{Addr: fmt.Sprintf(":%d", port)}
+
+	done := s.probe()
+	go func() {
```
nkcr, line 364: There might be a simpler way to run the http server, I've often seen that `ListenAndServe` is called blocking. You can then have something like that, which has slightly less channels:

```go
	signal.Notify(quit, os.Interrupt)

	go func() {
		<-quit
		log.Info("Server is shutting down...")

		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
		defer cancel()

		if err := server.Shutdown(ctx); err != nil {
			log.Fatalf("Could not gracefully shutdown the server: %v\n", err)
		}
		close(done)
	}()

	log.Infof("Starting HTTP server on %d\n", port)

	err := h.server.ListenAndServe()
	if err != nil && err != http.ErrServerClosed {
		log.Fatalf("Could not listen on %s: %v", h.listenAddr, err)
	}

	<-done
	log.Info("Server stopped")
```

and then you can use `done` in the `probe()`.

gnarula: > and then you can use done in the probe().

wouldn't that lead to a race condition on which goroutine is scheduled to recognise close on done first? Let's say the main goroutine is scheduled first and accepts the channel close. I think it would just log "Server stopped" and exit, thereby leaking the probe goroutine.

nkcr: Once a channel is closed, every subsequent call to it will pass:
https://play.golang.org/p/_N4rtwZ7HVB

gnarula: But the other goroutine might not finish at the same time as main: https://play.golang.org/p/G3TlAARPIcQ and https://play.golang.org/p/sXpLqESWHS2

nkcr: Oh, I see what you mean now. That shouldn't be a problem, as long as the goroutine doesn't have to do some cleanup, you can exit a program with active goroutines: https://stackoverflow.com/questions/25518531/ok-to-exit-program-with-active-goroutine

But this is a detail, I'm perfectly fine with keeping the current implementation.

ineiti:
> Prometheus works by scraping various systems and letting them expose their status.

Hmm - true. I was still thinking about Grafana. You'll have to explain to me sometimes why a logger that calls the service to be logged is a good idea.

Getting back to my previous idea - shouldn't this be implemented using https://github.com/dedis/onet/blob/992a708c6c664b744ec67d8ea6f9b0181c16d166/processor.go#L192 ? Then you wouldn't even have to start yet another process, but could let the nodes serve Prometheus. You would have to add a rate-limitation, but other than that, I would prefer that...

pierluca:
> > Prometheus works by scraping various systems and letting them expose their status.
> 
> Hmm - true. I was still thinking about Grafana. You'll have to explain to me sometimes why a logger that calls the service to be logged is a good idea.
> 
> Getting back to my previous idea - shouldn't this be implemented using https://github.com/dedis/onet/blob/992a708c6c664b744ec67d8ea6f9b0181c16d166/processor.go#L192 ? Then you wouldn't even have to start yet another process, but could let the nodes serve Prometheus. You would have to add a rate-limitation, but other than that, I would prefer that...

Grafana just does visualisation. I assume you're thinking about Graphite.
Prometheus is explicitly NOT about logging, and "why scraping"... I could talk about for hours :) 
Happy to have that conversation around coffee :)

gnarula:
> Getting back to my previous idea - shouldn't this be implemented using https://github.com/dedis/onet/blob/992a708c6c664b744ec67d8ea6f9b0181c16d166/processor.go#L192 ? Then you wouldn't even have to start yet another process, but could let the nodes serve Prometheus. You would have to add a rate-limitation, but other than that, I would prefer that...

I think the limitation there is the REST Handler only supports JSON responses but Prometheus expects a text format in the response.

Re: rate-limitation, I think the metrics would only be consumed by the conode internally and I'd assume the HTTP port would be firewalled to allow connections only from trusted sources (i.e. the prometheus server).

pierluca:
There's no need for rate limiting : Prometheus metrics are never meant to be "computed" on HTTP request, merely displayed.
We're following this pattern : we check the connectivity regularly, display it on request.
The cost of displaying the metric, i.e. a template with some data, is minimal.
Worst case we can cache the metric page for optimisation purposes, but that's an unlikely scenario.

pierluca (approved):
LGTM. Awesome :)

nkcr (approved):

PR #2413, opened 2020-11-27
Adds an optional acceptForeigner flag

### LINKED ISSUE #2409: js/byzcoin - use getLatestBlock from a proxy node ..?
Actually it is not possible for a foreign node (like a proxy one) to use `getLatestBlock`, because of this check:
https://github.com/dedis/cothority/blob/master/external/js/cothority/src/skipchain/skipchain-rpc.ts#L189

Why so ? Is it necessary ?

nkcr:
*See commit message*

Closes #2409

### DIFF
--- a/external/js/cothority/src/skipchain/skipchain-rpc.ts
+++ b/external/js/cothority/src/skipchain/skipchain-rpc.ts
@@ -149,7 +149,7 @@ export default class SkipchainRPC {
      * @param verify    Verify the integrity of the chain when true
      * @returns a promise that resolves with the list of blocks
      */
-    async getUpdateChain(latestID: Buffer, verify = true, acceptForeigner = false): Promise<SkipBlock[]> {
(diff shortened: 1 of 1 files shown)


external/js/cothority/src/skipchain/skipchain-rpc.ts
```diff
@@ -149,7 +149,7 @@ export default class SkipchainRPC {
      * @param verify    Verify the integrity of the chain when true
      * @returns a promise that resolves with the list of blocks
      */
-    async getUpdateChain(latestID: Buffer, verify = true): Promise<SkipBlock[]> {
+    async getUpdateChain(latestID: Buffer, verify = true, acceptForeigner = false): Promise<SkipBlock[]> {
```
ineiti, line 152: Here and in `getLatestBlock`:
`s/acceptForeigner/acceptProxy/`

we accept foreigners always, regardless of this setting ;)

ineiti (changes requested):

ineiti (approved):

PR #2414, opened 2020-11-27
Adds beautifier on the Coin + Credential contracts

nkcr:
**What this PR does**

*See commit message*

### DIFF
--- a/external/js/cothority/src/byzcoin/beautifier/coin.ts
+++ b/external/js/cothority/src/byzcoin/beautifier/coin.ts
@@ -1,0 +1,69 @@
+import Long from "long";
+import { Argument } from "..";
+import { Darc } from "../../darc";
+import { IBeautifyArgument } from "./utils";
+
+/**
+ * Arrange arguments for a Coin contract, ie. provide a meaningful
+ * representation of its arguments.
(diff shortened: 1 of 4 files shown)


external/js/cothority/src/byzcoin/beautifier/coin.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Arrange arguments for a Coin contract, ie. provide a meaningful
+ * representation of its arguments.
+ */
+export class CoinBeautifier {
+    static Spawn(args: Argument[]): IBeautifyArgument[] {
+        const res = Array<IBeautifyArgument>();
+
+        args.forEach((arg) => {
+            switch (arg.name) {
+                case "public":
```
ineiti, line 16: If you want to go the dark way, you can have:
```
case "public":
case "coinID":
case "type":
  res.push...
  break;
```

nkcr: I prefer when things are explicit (truly in this case sparing a few lines just makes the code harder to read). I don't feel like going dark today 😇

ineiti (approved):

PR #2420, opened 2020-12-04
Add release workflow

gnarula:
Allows creating a release with built tarballs on pushing a tag


ineiti (approved):

ineiti:
I was wondering how you avoid having a release for every push. Wonderful `on: push: tag`!

ineiti (approved):

ineiti (changes requested):
Thanks for the release action. Can you please add two more things to this PR?
- Update the `Versioning and Roadmap` section in cothority/README.md
- Update the cothority/PUBLISH.md file with instructions how and when to create new releases

To not create confusion between c4dt/byzcoin and the cothority releases, it might be worth thinking to reduce the cothority/conode.go to only include the things needed for evoting...

ineiti:
This PR should be rebased on master for the correct tests.

gnarula:
It turns out the tests depend on the other services. The way around it would be to decouple the tests from conode/conode.go which is well beyond the scope of this PR imho.

I'm thinking about undoing the last commit and work on the de-coupling of tests in another PR. What do you think @ineiti ?

ineiti:
One simple thing you can do: add a file `conode/full.go` with a compilation flag. Then you can compile the binary twice: once with the flag for the tests, once without the flag for the release.

gnarula:
thanks! SonarCloud complains but it's a false negative because of full.go

ineiti:
Coveralls is right ;) My idea was to have a `full.go` with only the imports. Then you can remove the `// +build !test` from `conode.go`...

```
// +build test

// Importing all necessary packages for tests
package main

import (
	_ "go.dedis.ch/cothority/v3/authprox"
	_ "go.dedis.ch/cothority/v3/byzcoin"
	_ "go.dedis.ch/cothority/v3/byzcoin/contracts"
	_ "go.dedis.ch/cothority/v3/calypso"
	_ "go.dedis.ch/cothority/v3/eventlog"
	_ "go.dedis.ch/cothority/v3/personhood"
	_ "go.dedis.ch/cothority/v3/skipchain"
)
```

PUBLISH.md
```diff
@@ -13,6 +16,19 @@ npm i -s @dedis/cothority@dev
...
 
+## Releasing a new version for e-voting conode
+
+Tag the release using
+
+```bash
+git tag -s vX.Y.Z -m "Cothority Release vX.Y.Z"
+```
+
+and push it to master with `git push origin vX.Y.Z`. A tag push would trigger
```
ineiti, line 27:
```suggestion
and push it to master with `git push origin vX.Y.Z`. A tag push triggers
```

PUBLISH.md
```diff
@@ -13,6 +16,19 @@ npm i -s @dedis/cothority@dev
...
+the `release` action workflow with new release binaries available at
```
ineiti, line 28:
```suggestion
the `release` action workflow which releases new binaries in
```

ineiti (changes requested):
Last corrections, then it's OK.

ineiti (approved):

PR #2421, opened 2020-12-04
Put test in lint_build

ineiti:
Using on:workflow_run puts the result of the workflow in the master-branch,
which makes it impossible to use the results of test.yaml in the requirement of the PR.

This PR creates a monster-workflow lint_build_test to avoid that problem.


.github/workflows/lint_build_test.yaml
```diff
@@ -0,0 +1,164 @@
...
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Installing golint
+        run: GO111MODULE=off go get -u golang.org/x/lint/golint
+
+      - name: Running pre-commit
+        uses: pre-commit/action@v2.0.0
+
+  # The barrier job is used to avoid writing a lot of needs
+  barrier:
```
tharvik, line 85:
```suggestion
  lint-and-build:
```

But it's not really correct, as `golang` doesn't build. Consider splitting it into a `lint` and another `build` barrier-job. Also rename the "javascript" & "golang" in "lint-javascript" & "lint-golang".

.github/workflows/lint_build_test.yaml
```diff
@@ -0,0 +1,164 @@
...
+        with:
+          go-version: ${{ matrix.golang-version }}
+
+      - name: Installing golint
+        run: GO111MODULE=off go get -u golang.org/x/lint/golint
+
+      - name: Testing proto-files
+        run: |
+          make test_proto
+          go version
```
tharvik, line 58:
```suggestion
        run: make test_proto
```

.github/workflows/lint_build_test.yaml
```diff
@@ -0,0 +1,164 @@
...
+  # The barrier job is used to avoid writing a lot of needs
+  barrier:
+    runs-on: ubuntu-latest
+    needs:
+      - javascript
+      - golang
+      - pre-commit
+    steps:
+      - name: Done linting and building
+        run: echo Linting and Building done
```
tharvik, line 93: Do you really need a `steps` block? It would be nice to avoid spawning a VM for a simple echo.

.github/workflows/lint_build_test.yaml
ineiti: Hmm - currently the `javascript` target lints and builds the code. So _should_ the `golang` target. I'll push a version with that proposition.

.github/workflows/lint_build_test.yaml
ineiti: done

.github/workflows/lint_build_test.yaml
ineiti: yep, needed: 

Screenshot 2020-12-08 at 14 42 22

tharvik:
Again on this `barrier` job, did you consider using `needs: [javascript, golang, pre-commit]` inplace of `needs: lint-build`?

ineiti:
> Again on this `barrier` job, did you consider using `needs: [javascript, golang, pre-commit]` inplace of `needs: lint-build`?

Yes, I did, and I thought it's nicer with a barrier - it represents better the difference between the `lint-build` that contains 3 jobs, and the `test` job... But if you convince @cgrigis it's cleaner with a `needs: [...]`, I'm OK with it...

cgrigis:
> Yes, I did, and I thought it's nicer with a barrier - it represents better the difference between the `lint-build` that contains 3 jobs, and the `test` job... But if you convince @cgrigis it's cleaner with a `needs: [...]`, I'm OK with it...

I lean towards `needs: lint-build`, as it is used 4 times and thus avoids repetition and eases maintenance.
It is unfortunate that this requires a whole job though, it would be much simpler to allow a variable, but that does not appear possible. :-(   What's one more VM spawn anyway? ;-)

tharvik (approved):

PR #2423, opened 2020-12-08
Adding SkipchainDB.GetProofFromIndex

ineiti:
Previously the GetByID was only available as a service-endpoint, but as it is
a handy method in general, this PR moves it to the SkipchainDB as `GetFullProof`,
which can do much more.

Also contains a fix to SkipBlock.pathForIndex

### DIFF
--- a/skipchain/struct.go
+++ b/skipchain/struct.go
@@ -605,6 +605,32 @@ func (sbs Proof) VerifyFromID(id SkipBlockID) error {
 	return sbs.verifyChain()
 }
 
+// GetForwardLinks creates a slice of ForwardLinks of the proof,
+// starting with a forwardlink pointing out of nil to the genesis block to
+// give the roster.
+func (sbs Proof) GetForwardLinks() (links []*ForwardLink, err error) {
+	if len(sbs) == 0 {
+		return nil, xerrors.New("cannot create links for empty proof")
+	}
+	links = make([]*ForwardLink, len(sbs))
+	links[0] = &ForwardLink{
+		To:        sbs[0].Hash,
+		NewRoster: sbs[0].Roster,
+	}
+
+	logBH := math.Log(float64(sbs[0].BaseHeight))
+	for i, sb := range sbs[:len(sbs)-1] {
+		logDist := math.Log(float64(sbs[i+1].Index - sb.Index))
+		// Using math.Round here to ignore rounding errors in math.Log
+		height := int(math.Round(logDist / logBH))
+		if len(sb.ForwardLink) <= height || sb.ForwardLink[height].IsEmpty() {
+			return nil, xerrors.New("missing forward-link in proof")
+		}
+		links[i+1] = sb.ForwardLink[height]
+	}
+	return
+}
+
 func (sbs Proof) verifyChain() error {
 	for i, sb := range sbs {
 		if !sb.CalculateHash().Equal(sb.Hash) {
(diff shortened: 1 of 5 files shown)


skipchain/struct.go
```diff
@@ -1148,112 +1151,105 @@ func (db *SkipBlockDB) GetFuzzy(id string) (*SkipBlock, error) {
...
 		}
+	}
+	return nil, nil, xerrors.New("didn't find any appropriate block")
+}
 
-		return nil
-	})
+// GetProof returns the shortest chain from the genesis to the latest block
+// using the highest forward-links available in the local db
+func (db *SkipBlockDB) GetProof(sid SkipBlockID) (sbs []*SkipBlock,
```
tharvik, line 1239: As it is pretty much the same as `GetFullProof`, maybe it can be remove?

tharvik:
And there is a Code Smell for the naming convention of test case. btw, I finally found my "golang standard" argument, in [effective go](https://golang.org/doc/effective_go.html#mixed-caps).

ineiti:
OK, bikeshedding for `TestSkipBlockDB_GetFullProof`:
- 1st google match for `golang test naming` cites hashicorp to do the same: https://ieftimov.com/post/testing-in-go-naming-conventions/
- golang proposal to impose naming on tests doesn't reach a consensus: https://github.com/golang/go/issues/33688

So my take here is: _follow the sheeps_. dedis/cothority has `TestStruct_Method` in many places already, so just accept it.

skipchain/struct.go
ineiti: I'm hesitating - it is public, so theoretically it could be used elsewhere, and I didn't want to break it. But looking at the usage, it's only in the `skipchain` package, so it _should_ be fine. I can mark it `Deprecated` (even though we never _really_ removed anything, did we?)

tharvik:
> So my take here is: _follow the sheeps_. dedis/cothority has `TestStruct_Method` in many places already, so just accept it.

Okay, then let's keep it as is, but you'll need to fix SonarCloud to avoid the next code smells :P

skipchain/struct.go
tharvik: > I'm hesitating - it is public, so theoretically it could be used elsewhere, and I didn't want to break it. But looking at the usage, it's only in the `skipchain` package, so it _should_ be fine.

As you wish, I'm always in favor of reducing the number of exposed function, especially if it serves the same purpose.

> I can mark it `Deprecated` (even though we never _really_ removed anything, did we?)

Yeah, to be nice with semver, we should do a new major release where we remove the deprecated functions, but that might never happen.

cgrigis:
Agreed, whatever the accepted rules are, the tools should be adjusted accordingly to remove false positives. :)

ineiti:
Opened #2425 so we can go forward with this.

ineiti:
OK - changed the names and signatures of the `GetProof`* methods and added some tests.

skipchain/struct.go
ineiti: Renamed the methods to make more sense:
- GetProofFromIndex - takes a skipchain-ID and an index
- GetProof - for compatibility, deprecated, not used anymore in the code
- GetProofForLatest - calls `GetProofFromIndex` with magical value `-1` - replaces previous `GetProof`

And the return values of all these methods is now the same. The `Proof` structure got a `GetForwardLinks` method.

tharvik (approved):

PR #2429, opened 2021-02-24
Adds a threshold rule for DARC expressions.

nkcr:
**What this PR does**

This PR adds a new expression in DARC to support a threshold of identities. It comes in the following form:

`threshold<n/d, id1, id2, ...>`

### DIFF
--- a/darc/darc.go
+++ b/darc/darc.go
@@ -1242,8 +1242,14 @@ func evalThreshold(visited map[string]bool, getDarc GetDarc,
 		}
 
 		// In case of a DARC, we need to check the _sign of that DARC and ensure
-		// we don't fall in a loop.
+		// we don't fall in a loop. A wrong evaluation will continue the loop
+		// and not include the DARC in the valid identities.
 		if strings.HasPrefix(entry, "darc") {
+
+			_, found := validIds[entry]
+			if acceptDarc && found {
+				continue
+			}
 
 			if _, ok := visited[entry]; ok {
 				return xerrors.Errorf("cycle detected")
@@ -1278,7 +1284,7 @@ func evalThreshold(visited map[string]bool, getDarc GetDarc,
 		}
 	}
 
-	// we floor the number of identities to 1, so we avoid 0/0.
+	// we clamp the number of identities to at least 1, so we avoid 0/0.
 	if len(uniqIds) == 0 {
 		uniqIds["fake"] = struct{}{}
 	}
--- a/darc/darc.go
+++ b/darc/darc.go
@@ -1194,7 +1194,8 @@ func parseIDEvmContract(in string) (Identity, error) {
 func evalThreshold(visited map[string]bool, getDarc GetDarc,
 	attrFuncs AttrInterpreters, acceptDarc bool, s string, ids []string) error {
 
-	// A provided threshold of 0/0 is undefined,
+	// A provided threshold of 0/0 is wrong, but will always make the validation
+	//   pass,
 	// A provided threshold of 1/0 will always make the validation fail,
 	// A provided threshold of 0/N will always make the validation pass.
 
--- a/darc/darc.go
+++ b/darc/darc.go
@@ -1194,7 +1194,11 @@ func parseIDEvmContract(in string) (Identity, error) {
 func evalThreshold(visited map[string]bool, getDarc GetDarc,
 	attrFuncs AttrInterpreters, acceptDarc bool, s string, ids []string) error {
 
-	// we have something of form 'threshold<1/2,darc:aa,ed25519:bb,...>'
+	// A provided threshold of 0/0 is undefined,
+	// A provided threshold of 1/0 will always make the validation fail,
+	// A provided threshold of 0/N will always make the validation pass.
+
+	// s is of form 'threshold<1/2,darc:aa,ed25519:bb,id...>'
 
 	s = strings.TrimPrefix(s, "threshold<")
 	s = strings.TrimSuffix(s, ">")
@@ -1207,26 +1211,25 @@ func evalThreshold(visited map[string]bool, getDarc GetDarc,
 
 	fractionStr := entries[0]
 
-	ok := regexp.MustCompile(`\d+/\d+`).MatchString(fractionStr)
-	if !ok {
+	r := regexp.MustCompile(`(\d+)/(\d+)`)
+	matches := r.FindStringSubmatch(fractionStr)
+	// One match for the entire expression, one for the Num, and one for Denum.
+	if len(matches) != 3 {
 		return xerrors.Errorf("the threshold fraction is incorrect: %s", fractionStr)
 	}
 
-	fractionElem := strings.Split(fractionStr, "/")
-	numerator, err := strconv.Atoi(fractionElem[0])
+	numerator, err := strconv.Atoi(matches[1])
 	if err != nil {
 		return xerrors.Errorf("failed to convert numerator: %v", err)
 	}
 
-	denominator, err := strconv.Atoi(fractionElem[1])
+	denominator, err := strconv.Atoi(matches[2])
 	if err != nil {
 		return xerrors.Errorf("failed to convert denominator: %v", err)
 	}
 
-	threshold := float64(numerator) / float64(denominator)
-
-	uniqIds := map[string]struct{}{}
-	validIds := map[string]struct{}{}
+	uniqIds := make(map[string]struct{}, len(entries)-1)
+	validIds := make(map[string]struct{}, len(entries)-1)
 
 	for _, entry := range entries[1:] {
 		uniqIds[entry] = struct{}{}
@@ -1241,10 +1244,6 @@ func evalThreshold(visited map[string]bool, getDarc GetDarc,
 		// In case of a DARC, we need to check the _sign of that DARC and ensure
 		// we don't fall in a loop.
 		if strings.HasPrefix(entry, "darc") {
-			_, found := validIds[entry]
-			if found && acceptDarc {
-				continue
-			}
 
 			if _, ok := visited[entry]; ok {
 				return xerrors.Errorf("cycle detected")
@@ -1279,10 +1278,15 @@ func evalThreshold(visited map[string]bool, getDarc GetDarc,
 		}
 	}
 
-	computedFraction := float64(len(validIds)) / float64(len(uniqIds))
-	if computedFraction < threshold {
-		return xerrors.Errorf("computed fraction is lower than threshold: %f < %f",
-			computedFraction, threshold)
+	// we floor the number of identities to 1, so we avoid 0/0.
+	if len(uniqIds) == 0 {
+		uniqIds["fake"] = struct{}{}
+	}
+
+	// a1/a2 < b1/b2 <=> a1*b2 < a2*b1
+	if denominator*len(validIds) < numerator*len(uniqIds) {
+		return xerrors.Errorf("computed fraction is lower than threshold: "+
+			"%d/%d < %d/%d", len(validIds), len(uniqIds), numerator, denominator)
 	}
 
 	return nil
--- a/darc/darc.go
+++ b/darc/darc.go
@@ -42,6 +42,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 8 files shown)


darc/darc.go
```diff
@@ -1180,6 +1191,103 @@ func parseIDEvmContract(in string) (Identity, error) {
...
+			if err != nil {
+				continue
+			}
+
+			validIds[entry] = struct{}{}
+		}
+	}
+
+	computedFraction := float64(len(validIds)) / float64(len(uniqIds))
+	if computedFraction < threshold {
```
cgrigis, line 1283: Just a thought: in order to avoid floats (and possible rounding issues), you can keep the comparison in ints using:

```
 a       c
---  <  ---  <=>  a * d  <  b * c
 b       d
```

therefore coming to:

```
if len(validIDs) * denominator < len(uniqIds) * numerator {
...
```

darc/darc.go
```diff
@@ -1180,6 +1191,103 @@ func parseIDEvmContract(in string) (Identity, error) {
...
+	}
+
+	fractionStr := entries[0]
+
+	ok := regexp.MustCompile(`\d+/\d+`).MatchString(fractionStr)
+	if !ok {
+		return xerrors.Errorf("the threshold fraction is incorrect: %s", fractionStr)
+	}
+
+	fractionElem := strings.Split(fractionStr, "/")
```
tharvik, line 1215: You can simplify it by using a single regex, which would be safer IMO. smth like `threshold<(\d+)/(\d+)(,((darc:[[:xdigit:]]+)|(ed25519:[[:xdigit:]]+)))+>`

darc/darc.go
```diff
@@ -1180,6 +1191,103 @@ func parseIDEvmContract(in string) (Identity, error) {
...
+
+	denominator, err := strconv.Atoi(fractionElem[1])
+	if err != nil {
+		return xerrors.Errorf("failed to convert denominator: %v", err)
+	}
+
+	threshold := float64(numerator) / float64(denominator)
+
+	uniqIds := map[string]struct{}{}
+	validIds := map[string]struct{}{}
```
tharvik, line 1229: To avoid relocation, you can `make` it with the expected capacity of `len(entries)-1`.

tharvik (changes requested):
I found some erronous test cases
 * `threshold<x/0,...>` yields a `NaN/Inf` threshold which is accepted
 * `threshold<x/y>` (no ids) yields a `NaN/Inf` actual threshold which is accepted

You should check that `numerator <= denominator` and that none are zero.

Also, you need to check that the denominator is equal to the number of ids.
You might even want to remove the denominator entirely as it is given by the number of valid ID, that's up to you.

darc/darc.go
```diff
@@ -1180,6 +1191,103 @@ func parseIDEvmContract(in string) (Identity, error) {
...
+		}
+
+		// In case of a DARC, we need to check the _sign of that DARC and ensure
+		// we don't fall in a loop.
+		if strings.HasPrefix(entry, "darc") {
+			_, found := validIds[entry]
+			if found && acceptDarc {
+				continue
+			}
+
```
cgrigis, line 1248: At this point, either `found == false` or `acceptDarc == false` (or both).
If `found == true` and `acceptDarc == false`, the following recursive validation of DARCs that, if invalid, shortcuts `validIds[entry] = struct{}{}` does not matter because we already have `validIds[entry] == struct{}{}` since `found == true`.

Am I missing something?

darc/darc.go
nkcr: > I found some erronous test cases
> 
>     * `threshold<x/0,...>` yields a `NaN/Inf` threshold which is accepted
> 
>     * `threshold<x/y>` (no ids) yields a `NaN/Inf` actual threshold which is accepted

I didn't want to force users to behave correctly. If one wants to use a threshold that doesn't make sense, it is his responsibility, as long as it doesn't make the system panic.

If we want to go that path I would do the following:

A provided threshold of **0/0** is undefined,  
A provided threshold of **1/0** will always make the validation fail,  
A provided threshold of **0/N** will always make the validation pass.

Plus, we always assume that there is at least one identity in the threshold function, by flooring the computed denominator to 1.

> Also, you need to check that the denominator is equal to the number of ids.

That's not the idea. I should be able to put a threshold saying "whatever identities there are, I want a threshold of 50%", using 1/2, or 50/100, or whatever fraction.

> You should check that `numerator <= denominator` and that none are zero.

Again, we should let users do stupid things. If the denominator is greater than the numerator, that means the threshold will never be reached. Which is fine.

nkcr:
> I found some erronous test cases
> 
>     * `threshold<x/0,...>` yields a `NaN/Inf` threshold which is accepted
> 
>     * `threshold<x/y>` (no ids) yields a `NaN/Inf` actual threshold which is accepted

I didn't want to force users to behave correctly. If one wants to use a threshold that doesn't make sense, it is his responsibility, as long as it doesn't make the system panic.

If we want to go that path I would do the following:

A provided threshold of **0/0** is undefined,  
A provided threshold of **1/0** will always make the validation fail,  
A provided threshold of **0/N** will always make the validation pass.

Plus, we always assume that there is at least one identity in the threshold function, by flooring the computed denominator to 1.

> Also, you need to check that the denominator is equal to the number of ids.

That's not the idea. I should be able to put a threshold saying "whatever identities there are, I want a threshold of 50%", using 1/2, or 50/100, or whatever fraction.

> You should check that `numerator <= denominator` and that none are zero.

Again, we should let users do stupid things. If the denominator is greater than the numerator, that means the threshold will never be reached. Which is fine.

darc/darc.go
nkcr: To be honest I also couldn't understand the logic there. I borrowed the DARC loop check code from above. I was about to remove that piece of code since it seems useless, but I was too afraid I missed something 😅.

cgrigis: Yes, I figured it had been borrowed from `evalExprDarc()` :smile: 
But, if I understand correctly, in the above function, the same behavior does not happen, because the "decision" is directly a `return`... Don't know if I am very clear, do you see what I mean?

darc/darc.go
nkcr: Very nice! 😀

darc/darc.go
nkcr: Oh, I see, so yes we should remove that check.

darc/darc.go
tharvik: Hum, you're right, maybe use a simpler `threshold<(\d+)/(\d+)(,[^,]+)+>` at least? It allows to extract the numerator & denominator, and each id.

darc/darc.go
```diff
@@ -1180,6 +1191,107 @@ func parseIDEvmContract(in string) (Identity, error) {
...
+		// we don't fall in a loop.
+		if strings.HasPrefix(entry, "darc") {
+
+			if _, ok := visited[entry]; ok {
+				return xerrors.Errorf("cycle detected")
+			}
+
+			// we make a copy so that diamond delegation will work,
+			// see TestDarc_DelegationDiamond
+			newVisited := make(map[string]bool)
```
tharvik, line 1254: You can also avoid reallocation here, by using `make` with `len(visited)+1`

tharvik (approved):

darc/darc.go
```diff
@@ -1180,6 +1191,107 @@ func parseIDEvmContract(in string) (Identity, error) {
...
+		}
+	}
+
+	// we floor the number of identities to 1, so we avoid 0/0.
+	if len(uniqIds) == 0 {
+		uniqIds["fake"] = struct{}{}
+	}
+
+	// a1/a2 < b1/b2 <=> a1*b2 < a2*b1
+	if denominator*len(validIds) < numerator*len(uniqIds) {
```
cgrigis, line 1287: :heart:

darc/darc.go
```diff
@@ -1279,10 +1278,15 @@ func evalThreshold(visited map[string]bool, getDarc GetDarc,
 		}
 	}
 
-	computedFraction := float64(len(validIds)) / float64(len(uniqIds))
-	if computedFraction < threshold {
-		return xerrors.Errorf("computed fraction is lower than threshold: %f < %f",
-			computedFraction, threshold)
+	// we floor the number of identities to 1, so we avoid 0/0.
```
cgrigis, line 1281: nitpicking :smile: 
I am not sure "floor" is the proper term here, as it usually means "rounding to the largest integer smaller than"... maybe "we force ... to be at least 1"? Like I said, nitpicking :laughing:

darc/darc.go
cgrigis: Actually I did not mean to remove the check itself (I know I wasn't very clear, sorry).

Per my understanding, the whole point of the `if strings.HasPrefix(entry, "darc") {` test branch is, in case the entry is a DARC, to validate that DARC (possibly recursively, depending on `acceptDarc`). In the original `evalExprDarc()`, the end result of the validation is a boolean in the closure. In your case, it is the inclusion or not of the entry in the `validIds` set. Since you already include the entry in the `validIds` [in the first loop](https://github.com/dedis/cothority/pull/2429/files#diff-3a3df335267a4b1574fe6cb5c8fcace447bb2d612b042e6191ed9e1edcfd8ec2R1237-R1242), even if the recursive validation fails, the entry still remains included. The fix probably involves, like in `evalExprDarc()`, using a boolean variable in the first loop, which would be negated if the validation fails, and the inclusion takes place at the end only if that variable is `true`.

Does this make sense?

Again, this is per my understanding, so it could be totally off. :smile:

cgrigis (changes requested):

darc/darc.go
tharvik: (single-sided) clamping then?

cgrigis: Kotlin has the `coerceAtLeast()`, `coerceAtMost()` and `coerceIn()` [functions](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/#functions) for this purpose...

darc/darc.go
nkcr: Not working: https://play.golang.org/p/oPV3LrRjNRG

ineiti: @tharvik I think the test for 0 is good, but the denominator does not need to be the same as the number of entries. I also started with the idea of only giving the number of signatures needed, so getting rid of the denominator. But then you need to update the number every time you add/remove a signature.

So I prefer @nkcr 's solution where you can give a fraction that will stay the same. Even though, to be perfect, it should allow to do things like `(- 1, 2 / 3)` to use it in a BFT context where you need `2f+1` signatures... But that's definitely out of scope for this PR.

ineiti:
> If we want to go that path I would do the following:
> 
> A provided threshold of **0/0** is undefined,
> A provided threshold of **1/0** will always make the validation fail,
> A provided threshold of **0/N** will always make the validation pass.

Having undefined behaviour is not what you want in a consensus system... So I propose the partially wrong interpretation (with a nice comment saying that it's partially wrong):
- if denominator == 0 -> fail
- if numerator == 0 -> pass

ineiti (changes requested):
My request with regard to more testing was along these lines:

https://github.com/dedis/cothority/pull/2431

If you're OK with it, merge the PR in yours.

darc/darc.go
nkcr: > Since you already include the entry in the validIds in the first loop, even if the recursive validation fails, the entry still remains included.

Not necessarily. It get included in the first loop only if there is a signed identities of type `darc:...`, but we normally include the identities of the _signed rule.

ineiti:
Test-java passes - nice ;)

ineiti (approved):

darc/darc.go
cgrigis: > Not necessarily. It get included in the first loop only if there is a signed identities of type `darc:...`, but we normally include the identities of the _signed rule.
> 
> So at the end the original if is right, I believe. It is a condition the avoid checking the DARC if it is accepted as an identity and present in the list of signed identities.

Right. And if it it not accepted as an identity but is present in the list of signed identities?

cgrigis:
Oops! Didn't mean to delete your comment @tharvik :disappointed:  How do I undelete?

cgrigis (approved):

PR #2433, opened 2021-03-10
Send view method call to whole roster

cgrigis:
**What this PR does**

This PR increases the reliability of BEvm view methods called from the frontend by sending the request to all the members of the roster. Currently, it was sent only to the first node in the roster, which seems to be unreliable as connection errors appear regularly when using the Stainless demo.
To ensure view methods are executed with a state that is current, the latest block ID is tracked during transaction calls and sent for view methods: using it, the node determines whether it needs to wait for an updated state before executing the call.

---

🙅‍ Friendly checklist:


cgrigis:
SonarCloud appears to get confused about what constitutes _new code_. I suspect it might be because my PR comes from the C4DT repo, which has a different master branch. Anyone knows how to handle this?

ineiti:
Yes, good point. IIRC, I had similar problems and just ended up ignoring Sonarcloud... Do you have write access to the dedis/cothority, so that you can do a PR from here?

cgrigis:
I tried a PR from the dedis repo (see #2434) but it seems to have the same outcome, so that is not the reason...

cgrigis:
Closed in favor of #2434.

ineiti:
> It's annoying that we can no longer make clean pull requests from C4DT's fork though...

I agree - if you want to fight SonarCloud over it, feel free. I mostly PR from the repo itself anyway.

cgrigis:
Was there a reason to keep the master branch in the C4DT fork not synchronized with upstream?

ineiti:
None - only lazyness. I wanted to have a cleaner development by using separate repositories. But I don't feel very strongly about this. And if SonarCloud has trouble with that, I'm OK with removing the c4dt/cothority and just keep the dedis/cothority.

cgrigis:
I meant it more in the sense that we can then synchronize master and see if that fixes it on the next PR.

ineiti:
That is included in the `fight with SonarCloud` ;)

cgrigis:
Challenge accepted :)

PR #2434, opened 2021-03-11
Send view method call to whole roster

cgrigis:
**What this PR does**

This PR increases the reliability of BEvm view methods called from the frontend by sending the request to all the members of the roster. Currently, it was sent only to the first node in the roster, which seems to be unreliable as connection errors appear regularly when using the Stainless demo.
To ensure view methods are executed with a state that is current, the latest block ID is tracked during transaction calls and sent for view methods: using it, the node determines whether it needs to wait for an updated state before executing the call.

---

🙅‍ Friendly checklist:


bevm/service.go
```diff
@@ -36,6 +36,41 @@ func init() {
...
+		currentIndex := rst.GetIndex()
+
+		log.Lvlf2("[%v] rst index: %v >= %v?", maxTries-triesLeft,
+			currentIndex, minBlockIndex)
+
+		if currentIndex >= minBlockIndex {
+			break
+		}
+
+		triesLeft--
```
ineiti, line 61: Why is this not in the `for` loop above? Then you could return on success in line 58, and return the error if it goes out of the for-loop.

ineiti (changes requested):
LGTM - just a syntax go-for question to be resolved.

bevm/service.go
cgrigis: I just found it clearer this way.

ineiti (approved):

PR #2435, opened 2021-03-12
Moved qrgo repo

ineiti:
As qantik/qrgo disappeared, a copy of it has been put up at c4dt/qrgo.


ineiti:
Let's ignore the code smell of `main.go` where I just changed the import path.

tharvik (approved):

PR #2436, opened 2021-03-15
js/cothority: rm url-parse

tharvik:
So, `url-parse` came back to bit my hand, first because `@types/url-parse` was missing yet the `URL` type was exposed, then because I wasn't able to *compile* some of its deps.. It was initially added because webpack doesn't deal well with "subdir imports" and that some platforms doesn't come with the same definition of `URL` (Node, NS, ...). I don't think that the Node target is needed anymore (correct me @nkcr) and we just removed the support for mobile/NS in omniledger. Let's remove (once again) `url-parse`.

I was thinking of finally fixing this webpack issue as a fallback plan, but if this PR is merged, I'll wait until an unfixable issue comes along. (Disclaimer: to fix webpack, I'll probably have to reorganise the whole import hierarchy)

### DIFF
--- a/external/js/cothority/package-lock.json
+++ b/external/js/cothority/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@dedis/cothority",
-  "version": "3.6.2",
+  "version": "3.6.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


ineiti (approved):

nkcr:
> I don't think that the Node target is needed anymore (correct me @nkcr) 

No, nothing I'm aware of. I though that mobile support was needed for the PoP app, did it change ?

ineiti:
@nkcr - Which PoP app? Then one I wrote is not supported anymore and has been removed from the repo. The one from Haoquian I didn't hear anything about it for quite some time, do you have news about that one?

nkcr:
I was thinking about your PoP app. For the rest I don't know. I'm a bit surprised because two years ago, back when I started, the javascript library was only used for mobile and I struggled using it on a browser. Now we are totally shifting, I guess times change 🤷‍♂️

PR #2437, opened 2021-03-17
Fixes the pagination service

nkcr:
**What this PR does**

This PR updates the pagination service that is currently not implemented correctly. The pagination should allow clients to submit multiple requests to the same instance of a service, which is why it must use the same channels for each instance. This wasn't the case and it was also causing some goroutine leaks since the stopChan would be called only once, but a new stopChan was created each time.

### DIFF
--- a/byzcoin/streaming.go
+++ b/byzcoin/streaming.go
@@ -120,57 +120,123 @@ func (s *Service) StreamTransactions(msg *StreamingRequest) (chan *StreamingResp
 	return outChan, stopChan, nil
 }
 
-// PaginateBlocks returns blocks with pagination, ie. N asynchronous requests
-// that contain each K consecutive block. The caller is responsible for closing
-// the close chan when the caller wants to close the connection.
-func (s *Service) PaginateBlocks(msg *PaginateRequest) (chan *PaginateResponse, chan bool, error) {
+// getPaginateService returns a paginate service. We need to use the same output
+// channel across client calls, and one closing channel. Which is why we use a
+// "statefull" service.
+func (s *Service) getPaginateService() func(msg *PaginateRequest) (chan *PaginateResponse, chan bool, error) {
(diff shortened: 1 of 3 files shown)


nkcr:
Any of @ineiti @tharvik @cgrigis can review and merge.

byzcoin/streaming.go
```diff
@@ -120,57 +120,123 @@ func (s *Service) StreamTransactions(msg *StreamingRequest) (chan *StreamingResp
 	return outChan, stopChan, nil
 }
 
-// PaginateBlocks returns blocks with pagination, ie. N asynchronous requests
-// that contain each K consecutive block. The caller is responsible for closing
-// the close chan when the caller wants to close the connection.
-func (s *Service) PaginateBlocks(msg *PaginateRequest) (chan *PaginateResponse, chan bool, error) {
+// getPaginateService returns a paginate service. We need to use the same output
+// channel across client calls, and one closing channel. Which is why we use a
+// "statefull" service.
```
ineiti, line 125:
```suggestion
// "stateful" service.
```

ineiti (changes requested):
There are several things I don't understand:
- why do you register the return result of `getPaginateService` to `RegisterStreamingHandlers`? From what I see, `getPaginateService` is called only once. So you don't need to use it as a callback here
- why should every call to `PaginateRequest` return the same channel? What happens if two clients connect at the same time and want to access two paginateservices?

Perhaps we should keep the meeting tomorrow, so that you can explain it to me ;)

ineiti:
> Discussion is welcome

OK, I re-sent the invite ;) I also invited @tharvik , as he's been working on this last month or so. I hope his memory is fresher than mine with regard to the streaming service...

tharvik:
Closed in favor of dedis/onet#674

PR #2439, opened 2021-03-19
Bypros proxy service

nkcr:
**What this PR does**

This PR add a proxy service that can listen on blocks, catch up, and return results of arbitrary read-only SQL queries.

Needs https://github.com/dedis/onet/pull/674 for tests to pass.


bypros/api.go
```diff
@@ -0,0 +1,213 @@
...
+// Client defines a proxy client
+//
+// - implements OverlayClient
+type Client struct {
+	OverlayClient
+}
+
+// Follow sends a request to start following a node, ie. listening to new blocks
+// and updating the database accordingly.
+func (c *Client) Follow(host, target *network.ServerIdentity,
```
ineiti, line 49: Shouldn't this take a full roster?

bypros/api.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, xerrors.Errorf("failed to send query request: %v", err)
+	}
+
+	return resp.Result, nil
+}
+
+// CatchUP sends a request to catch up from a particular block. A catch up will
+// update the database from the specified block until the end of the chain.
+// Every specifies the interval at which updates are sent.
+func (c *Client) CatchUP(ctx context.Context, host, target *network.ServerIdentity,
```
ineiti, line 100: I thought this would happen automagically. Once the proxy starts, it connects to the `StreamTransactions` byzcoin endpoint and then catches up.

bypros/api.go
```diff
@@ -0,0 +1,213 @@
...
+				close(outChan)
+				return
+			}
+		}
+	}()
+
+	return outChan, nil
+}
+
+// onetOverlay provides an overlay implementation based on onet and gorilla.
```
ineiti, line 174: Is this for easier testing?

bypros/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+	c, _, err := websocket.DefaultDialer.Dial(apiURL, nil)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to dial %s: %v", apiURL, err)
+	}
+
+	streamReq := byzcoin.StreamingRequest{
+		ID: req.ScID,
+	}
+
+	buf, err := protobuf.Encode(&streamReq)
```
ineiti, line 57: Why don't you use `byzcoin.Client.StreamTransactions` here?

bypros/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+	select {
+	case s.stopFollow <- struct{}{}:
+	default:
+	}
+}
+
+// CatchUP updates the database from a given block until the end of the chain.
+// It uses the streaming service to periodically send back the catch up state to
+// the client. This is appropriate since a catch up can be quite long.
+func (s *Service) CatchUP(req *CatchUpMsg) (chan *CatchUpResponse, chan bool, error) {
```
ineiti, line 154: As written above: why is that a call that has to be done manually by the client? For a proxy service, I would think that it tries to handle an up-to-date state all the time.

bypros/test/docker-entrypoint-initdb.d/schema.sql
```diff
@@ -0,0 +1,394 @@
+--
```
ineiti, line 1: Should there be a
```
echo schema.sql binary >> .gitattributes
```
here?

byzcoin/streaming.go
```diff
@@ -129,7 +129,6 @@ func (s *Service) PaginateBlocks(msg *PaginateRequest) (chan *PaginateResponse,
 	stopChan := make(chan bool)
 
 	go func() {
```
ineiti, line 131: Shouldn't there be a `defer close(outChan)` in this go-routine? So the `return`s would inform the caller that something went wrong.

Hmm - in the other PR I supposed that the `outChan` never closes... What is the good thing to do here?

ineiti (commented):
Just looked at it from a high level. I think I don't understand yet what your suppositions are with regard to the interface between the proxy and byzcoin. I believe it could be made simpler, see the questions in the comments. Depending on the answers you give, I might very well be wrong ;)

bypros/api.go
nkcr: no. The service allows you to follow a specific node.

bypros/api.go
nkcr: That was initially the case. It came out that splitting it into a separate service call makes the service much more manageable, from the developer and client's side. You can start following a node, then if you know that you *might* not be in sync, you can run a catchup in the background. You can also send a catchup request at some point if you just want to be sure that you are in sync.

bypros/api.go
nkcr: no, it is for proper decoupling. But it comes out that this makes life easier for testing ;)

bypros/proxy.go
nkcr: 1) To have a connection that stays alive beyond the hard-coded 5 minutes of the cothority client  
2) To remove a source of uncertainty

bypros/proxy.go
nkcr: The proxy service gives the tools for a user to keep an updated proxy: you can follow a node, and perform a catch up if wanted. Those are two different operations that one might want to do separately: I want just to catch up at a given point in time, or I want just to follow from that point in time, for example.

Making a service call doing one specific action also makes it much more manageable, as previously said in another comment. And combining two service calls is also really easy to do if wanted, it should just not be part of the service.

bypros/test/docker-entrypoint-initdb.d/schema.sql
nkcr: no, the schema is plain text

byzcoin/streaming.go
nkcr: It is the user's responsibility to check for any errors in a paginate response, and close the connection or send a new message. The connection is closed after 5 minutes of inactivity anyway.

bypros/api.go
ineiti: That doesn't really make sense. A ByzCoin chain is not defined by the nodes, but rather by the genesis-ID and all the forward-links. Nodes can come and go, they might be down. So if you have this `Follow` with only one node, and then only follow that node, the proxy might be quite unstable.

bypros/api.go
ineiti: As I wrote above, following one node is not what you should do. To follow the chain, you need to follow the roster. The cothority-js does this by sending out a request to all nodes, and then keeps only the two fastest nodes for further request. If one of these nodes fails, it is put at the end of the list and the next node in the list is put in its place.

This allows for nodes to fail, and the system still to be manageable. Also, nodes might disappear altogether. If the client has to start catchup and give the node name, you put quite a big burden on him.

bypros/proxy.go
ineiti: 1. If you add a new parameter to `onet.Options`, you might also change this
  - Having connections time out from time to time is not so bad - as long as you reconnect
  - e.g., the dynacred-library happily reconnects the streaming listener - might be even every 30 seconds...
2. Why uncertiainty?

bypros/proxy.go
ineiti: Hmm - still not convinced - but it's your code. Perhaps bump it off some of the engineers in your team to see if they also think that way about a proxy.

I would at least remove the notion of `node` and replace it with a notion of `byzcoin`, which includes a `roster` that might change. As discussed on #conode lately, even if the nodes are currently very static, that is not the 'natural' behaviour of a blockchain...

bypros/test/docker-entrypoint-initdb.d/schema.sql
ineiti: It's just to remove it from the git-diff. Makes it more readable, specifically if it's machine-generated.

bypros/proxy.go
nkcr: I understand you view. I was implementing the proxy with another mindset: a user chose to follow a node it trusts, and the proxy service doesn't reflect the chain, but the content of a node, like this is the case of Columbus. That's a subtle difference, and I guess that's the essence of our discussion. 

In our case, we will probably run a proxy service following our DEDIS node, which we have full control of, and that we don't plan to have occasionally off. We run it as a production service: if it is down then there is a problem that must be fixed, and we don't expect our proxy to automagically choose another node at random. We fix our production conode, and make our proxy follow again + catch up.

In the case a user really wants to "follow the chain" nothing prevents it to use a roster, do the work of checking for the best node, and rotate in case a node is down. Again I think this is not something the service should do.

What do you think @gnarula @pierluca ?

pierluca: I haven't yet had the pleasure to go through the PR, so here's my (limited) understanding and my take on it:
In principle, my understanding of this proxy is that it's designed to be deployed in a "sidecar pattern", where we're creating a separate, isolated process to index information contained in the ledger and query it later. As it points to the information contained in the ledger (as seen by the local node), we can always query the ledger itself to verify the information we obtained through the proxy.
Furthermore, as a sidecar service, it is expected that its lifecycle would be tightly coupled with that of the node it is deployed next to. If the node is down (or performing in a "degraded" mode), the sidecar service might be unable to process queries, or provide out-of-date results.
If my understanding is correct, I have no philosophical objections to the approach.

As I don't particularly want to remain ignorant and I'd love to have a more informed opinion, I've added myself to the reviewers and will go over this ASAP :-)

ineiti: From a blockchain point of view, following a specific node doesn't make sense: you should never trust one node. The whole idea of a blockchain is to trust 2/3 of the nodes to do the right thing. So if you trust just one node, you remove one of the advantages of it being part of a blockchain.

If you really just want to trust one node, then you don't need a proxy. You just put a firewall on the websocket-port of your node and only allow specific applications to access that.

Saying `a user chose to follow a node it trusts` is not something you should ever do in a blockchain. It's already bad if you have to trust a proxy, but now you have to trust the proxy _and_ the node. Having forward-links makes this a little bit less awkward than in a Proof-of-Work system, as the failing node can only stop you, but still...

byzcoin/streaming.go
ineiti: We should probably better define what the interface of the streaming service should actually be...

bypros/proxy.go
pierluca: @ineiti I believe the point here is that we aren't using the proxy for trust. The blockchain is still the source of trust, but a client can use the proxy to make queries. Whatever the query returns would still need to be validated with the blockchain, but that's not the proxy's responsibility.

Furthermore, because this is thought of as a sidecar pattern, I think trusting the node and the proxy is fundamentally the same.  Both are software written by the same people, and you can expect them to be operated by the same entity on the same hardware, and to be in a "degraded" or flat-out "failing" state at the same time (due to the nature of the deployment)

Finally, I disagree that "you don't need a proxy", in so far as the proxy offers functionality (i.e. "query-ability") that the blockchain does not natively offer. 

I'm under the impression that what you're describing as an alternative is not a sidecar service, but a full blown server, connected to the roster and allowing clients to perform queries on it. If I understand correctly, this effectively would be the same as trusting "itself" as a node, which is ultimately no different to the sidecar pattern. A network outage would still cause the services to be provided in a "degraded" way and it would return outdated data.

Am I missing something here?

ineiti: Oh my - this got out of hand. Long time I didn't have a long conversation on github. I always think it's a bad medium to discuss in that kind of depth.

From a technical point of view, I still reject the PR for the following reasons:
1. depending on one node instead of the roster
2. update management by the client instead of the service
3. patching onet's 5-minute rule by re-implementing the websocket-client

I see three possibilities:
A) as the cothority-code is reaching end-of-life (C4DT will probably stop supporting it in April 2022), I close my eyes and we force-merge
B) rewrite this as a library / separate app instead of a cothority-service
C) discuss over zoom why I think the above three points need to be fixed

bypros/storage/sqlstore/README.md
```diff
@@ -0,0 +1,3 @@
+# SQL store
+
+You can find the SQL schema in ../test/docker-entrypoint-initdb.d/
```
pierluca: That's probably not the best place to keep the schema :)

bypros/storage/sqlstore/mod.go
```diff
@@ -0,0 +1,230 @@
...
+// GetBlock implements storage.Storage. It returns -1 if the block is not found,
+// otherwise its primary key.
+func (s SQL) GetBlock(blockHash []byte) (int, error) {
+	var blockID int
+
+	query := `SELECT block_id 
+	FROM cothority.block 
+	WHERE hash=$1`
+
+	err := s.db.QueryRow(query, blockHash).Scan(&blockID)
```
pierluca, line 96: why wouldn't this use dbRo ?

bypros/storage/sqlstore/mod.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return -1, xerrors.Errorf("failed to begin transaction: %v", err)
+	}
+
+	storeTx := storeTx{
+		sqlTx: sqlTx,
+		block: block,
+	}
+
+	blockID, err := storeTx.store()
```
pierluca, line 127: this API seems very odd to me, as opposed to having `store(block, sqlTx)` 
I really don't see how `storeTx` is not just a function call disguised as an object.

bypros/test/docker-entrypoint-initdb.d/schema.sql
```diff
@@ -0,0 +1,395 @@
...
```
pierluca, line 1: This is a dump and it should probably be a clean SQL schema instead, with higher expressivity and clearly indicating which fields are primary and foreign keys, and which are the indices, directly when defining the table.

bypros/test/integration_test.go
```diff
@@ -0,0 +1,488 @@
...
+	"github.com/docker/go-connections/nat"
+	"github.com/stretchr/testify/require"
+	"go.dedis.ch/cothority/v3/bypros"
+	"go.dedis.ch/cothority/v3/bypros/storage/sqlstore"
+	"go.dedis.ch/cothority/v3/byzcoin"
+	"go.dedis.ch/cothority/v3/byzcoin/contracts"
+)
+
+func TestIntegration_Simple(t *testing.T) {
+	os.Setenv("PROXY_DB_URL", "postgres://bypros:docker@localhost:4567/bypros")
```
pierluca: These test functions are huge and there's no way I can spot a bug in something this big. 

I'm not sure how to approach this though, as I imagine it's not necessarily an easy refactoring?

byzcoin/streaming.go
```diff
@@ -251,13 +184,68 @@ func (p *paginateService) handleMsg(msg *PaginateRequest) {
...
-		}
+		// Waiting for the streaming connection to stop. This signal comes
+		// from onet, which sets it when the client closes the connection.
+		// <-stopChan
+	}()
 
-		p.outChan <- response
-	}
+	go func() {
+		<-stopChan
```
pierluca, line 246: I'm a bit confused, who's closing stopChan?

pierluca (changes requested):
Overalls the PR is quite huge and I can't say that I've reviewed it 100% thoroughly, but overalls I feel that:
1) the general approach seems sound to me
2) some parts could benefit from better readability / maintainability (smaller functions, cleaner SQL, etc.), especially in the tests

bypros/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+Bypros can listen on a conode and save the chain in a relational database. It
+offers a read-only service call that can execute a provided SQL query.
+
+Apart from listening and executing a query, there is also a service call to
+catch up from a given block to the end of the chain.
+
+## Some technical details
+
+### Run postgres in a docker
```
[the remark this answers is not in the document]
ineiti: As this DB is only used for caching and doesn't hold any data that cannot be reproduced from the chain, I think this should be OK.

bypros/api.go
nkcr: not really feasible, that would require to know the node we want to follow at compile time, ie. when the service is included.

bypros/proxy.go
nkcr: 2: onet has its own way of handling ws connections, which was supposedly raising some errors and making the tests (more) unstable. I came to a point where I didn't have time to fight it anymore.

bypros/api.go
ineiti: Why do you need this on compile-time? here is what the ByzCoin service does on startup:
- load the config from the db
- tries to catchup from other nodes in the background
- starts operations

So when the bypros service receives a call to `Follow`, it can store the ID and the ServerIdentity in the DB offered from onet using `Context.Save`. When the bypros service is started the next time, it can use `Context.Load` to see if something is in the DB. If there is an ID and a ServerIdentity stored, it can call the catchup mechanism.

bypros/proxy.go
ineiti: So it will be A): the current code is between `welded side-car` (where the service bypros and byzcoin run on the same conode, so `ServerIdentity` is equal to the conode's ServerIdentity) and `standalone` app, where a full `Roster` would make most sense to use.

Once 3) is taken care of, I'm happy to merge.

nkcr: Regarding 3) we can't spent more time to fix/update onet. We already spent quite some time on it and we definitely need to move forward with this PR, as it is on the critical path of a project that needs to move forward.

bypros/test/integration_test.go
nkcr: We could refactor the test functions to use multiple sub-functions, but I'm afraid we would loose in readability. IMO integration tests should offer a single "thread of events" that one can follow to see how the system work in an "end-to-end" fashion.

bypros/storage/sqlstore/schema/schema.sql
```diff
@@ -0,0 +1,152 @@
...
+
+SET default_tablespace = '';
+SET default_table_access_method = heap;
+
+--
+-- Migration / Versioning
+--
+
+CREATE TABLE cothority.version (
+    version_id integer
```
pierluca, line 24: no need for version_id here, right?
we can just use the database_version as primary key (there can't be 2 identical DB versions)

bypros/storage/sqlstore/schema/schema.sql
```diff
@@ -0,0 +1,152 @@
...
+        PRIMARY KEY
+        GENERATED ALWAYS AS IDENTITY,
+
+    database_version integer NOT NULL,
```
pierluca, line 28: technically, schema_version 😅

pierluca (commented):
Just had a very quick look at the SQL file.
Some comments we discussed orally, the rest is here.

bypros/storage/sqlstore/schema/schema.sql
pierluca: forgot: you might also want to include the very first row, saying ( version: 0, timestamp: now() )

bypros/storage/sqlstore/schema/schema.sql
nkcr: ok. Should it be auto-increment, or do we let the user choose the version ?

pierluca: I'd say no auto-increment, as version management is probably best left to humans :)
the timestamp could be auto-generated though, now that I think of it :)

bypros/storage/sqlstore/schema/schema.sql
```diff
@@ -0,0 +1,150 @@
...
+        REFERENCES cothority.transaction(transaction_id)
+        ON UPDATE CASCADE ON DELETE CASCADE,
+
+    type_id integer NOT NULL
+        REFERENCES cothority."instructionType"(type_id)
+        ON UPDATE CASCADE ON DELETE CASCADE,
+
+    action character varying NOT NULL,
+    instance_iid bytea NOT NULL,
+    contract_iid bytea NOT NULL,
```
pierluca, line 97: does the same logic as the hash apply here (i.e. it'd be reasonable to transform them into `char(n)`? I'm not 100% sure, but I think so.

I'd also consider adding an index on them, to make the retrieval reasonably fast.

bypros/storage/sqlstore/schema/schema.sql
```diff
@@ -0,0 +1,150 @@
...
+CREATE TABLE cothority.instruction (
+    instruction_id integer
+        PRIMARY KEY
+        GENERATED ALWAYS AS IDENTITY,
+
+    transaction_id integer NOT NULL
+        REFERENCES cothority.transaction(transaction_id)
+        ON UPDATE CASCADE ON DELETE CASCADE,
+
+    type_id integer NOT NULL
```
pierluca, line 91: My understanding is that this could be enum-valued.

pierluca (approved):
As discussed with @ineiti and @nkcr, we can merge this now to unblock the MedChain project.

A clear path for subsequent improvements has been defined.
1. Database schema versioning table is in place, to enable ease of migration
2. This review contains a few comments to address some (perceived) shortcomings of the SQL, which can be dealt with now or subsequently.
3. As discussed with @ineiti , we will start a subsequent PR to use ONet instead of a direct web socket around mid-April. @ineiti will support us in debugging the issues are arising there.

Thanks to all those involved for their insights :)

ineiti:
Please make sure that the java test passes. Currently it fails in the service setup:

```
F 21/26/03 17:37:51.314593079: service.go:354 (v3.newServiceManager)    - Trying to instantiate service ByzcoinProxy: creating service:
    go.dedis.ch/onet/v3.(*serviceFactory).start
        /go/pkg/mod/go.dedis.ch/onet/v3@v3.2.7/service.go:295
  - failed to get storage:
    go.dedis.ch/cothority/v3/bypros.newProxyService
        /cothority/bypros/service.go:50
  - failed to get DB URL: did you set PROXY_DB_URL ?:
    go.dedis.ch/cothority/v3/bypros/storage/sqlstore.NewSQL
        /cothority/bypros/storage/sqlstore/mod.go:45
```

PR #2447, opened 2021-04-19
Bypros minor updates

nkcr:
**What this PR does**

This PR adds "stability" updates to Bypros. It also fixes a bug in getting the Websocket address.


bypros/proxy.go
```diff
@@ -148,6 +148,9 @@ func (s *Service) listenBlocks(conn *websocket.Conn) {
 				// that can happen when we close the connection
 				log.Warnf("failed to read request: %v", err)
 			}
+
+			log.LLvl1("stops listening on blocks")
```
ineiti, line 152:
```suggestion
			log.Lvl1("stops listening on blocks")
```

Is this really a level 1?

For debugging prints, please use `log.Print`. That is the unofficial `log.Debug`...

bypros/proxy.go
```diff
@@ -208,6 +211,11 @@ func (s *Service) doCatchUP(outChan catchUpOut, stopChan chan bool, req *CatchUp
 
 	browseDone := make(chan struct{})
 
+	if req.UpdateEvery < 1 {
+		log.Warn("got invalid 'updateEvery' value, setting it to 1")
```
ineiti, line 215: Hmm - I would even return an error here. Except if you have some stable code out there that doesn't support this value yet. But if `updateEvery < 1` is invalid, that's an error...

bypros/proxy.go
```diff
@@ -303,7 +311,7 @@ func (s *Service) Unfollow(req *Unfollow) (*EmptyReply, error) {
 func (s *Service) Query(req *Query) (*QueryReply, error) {
 	res, err := s.storage.Query(req.Query)
 	if err != nil {
-		return nil, xerrors.Errorf("failed to query: %v", err)
+		res = []byte(xerrors.Errorf("ERROR: failed to query: %v", err).Error())
```
ineiti, line 314: can't you catch it if it returns an error?

bypros/proxy.go
ineiti: It's official now:

https://github.com/dedis/cothority/pull/2447/checks?check_run_id=2380240597#step:6:114

;)

bypros/proxy.go
nkcr: when using the JavaScript library you will only get something like "the connection closed abnormally", but in this case I need the error message, which can indicate a wrong SQL syntax.

bypros/proxy.go
nkcr: > Is this really a level 1?

Yes, this log is intended to be there.

bypros/proxy.go
ineiti: OK, thanks - I thought that has been fixed in the meantime :(

ineiti (approved):

PR #2448, opened 2021-04-19
Updated npm packages

ineiti:
Run `npm audit fix` on `external/js/{cothority,kyber}`


tharvik (approved):
perfect, I can fix https://github.com/c4dt/service-drynx/pull/29 without arrière pensée :)

PR #2449, opened 2021-04-19
Uses the onet client instead of pure websocket

nkcr:
**What this PR does**

This PR removes the "pure" websocket use in Bypros and relies on the Onet client.
@ineiti discussion is open.

### DIFF
--- a/bypros/browse/mod.go
+++ b/bypros/browse/mod.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"go.dedis.ch/cothority/v3/skipchain"
+	"go.dedis.ch/onet/v3/network"
 )
 
 // Handler defines the type of function called on each block
@@ -12,7 +13,7 @@ type Handler func(block *skipchain.SkipBlock) error
 // Service defines the primitive of a browsing service.
 type Service interface {
 	// GetBrowser returns a new browsing actor
-	GetBrowser(handler Handler, id skipchain.SkipBlockID, target string) Actor
+	GetBrowser(handler Handler, id skipchain.SkipBlockID, target *network.ServerIdentity) Actor
 }
 
 // Actor defines the primitives of a browsing actor. An actor is created with a
--- a/bypros/browse/paginate/mod.go
+++ b/bypros/browse/paginate/mod.go
@@ -3,14 +3,14 @@ package paginate
 import (
 	"context"
 	"strconv"
-	"time"
 
-	"github.com/gorilla/websocket"
+	"go.dedis.ch/cothority/v3"
 	"go.dedis.ch/cothority/v3/bypros/browse"
 	"go.dedis.ch/cothority/v3/byzcoin"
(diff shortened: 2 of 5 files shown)


nkcr:
@ineiti this is in response to the commitment we made for the Byzcoin proxy PR to be accepted. This PR makes Bypros rely on Onet, but there are two missing features from Onet that would need to be implemented: `Ping`, and `ReadMessageKeep`. 
As those two features don't seem that big to implement, I can invest some time to implement them, provided we are on the same page and it doesn't end up with an eternal philosophical debate 🙃. So, what's your opinion ?

ineiti:
But I like philosophical discussions, more specifically after some beers...

If you have time, that would be great. If you give a signature and a 'contract' for the `ReadMessageKeep`, we can get philosophical on those.

For the `Ping`, would it be enough to fetch the `/ok` path from the websocket? Or is the ping related to an open websocket?

nkcr:
> If you have time, that would be great. If you give a signature and a 'contract' for the ReadMessageKeep, we can get philosophical on those.

Would have the same signature as `ReadMessage`

> For the Ping, would it be enough to fetch the /ok path from the websocket? Or is the ping related to an open websocket?

I was thinking of the "native" Ping control message (https://tools.ietf.org/html/rfc6455#section-5.5.2), on an open ws connection.

ineiti:
> Would have the same signature as `ReadMessage`

So why not add an `if c.IsKeep` or something to `ReadMessage`?

> I was thinking of the "native" Ping control message (https://tools.ietf.org/html/rfc6455#section-5.5.2), on an open ws connection.

OK, makes sense.

nkcr:
> So why not add an if c.IsKeep or something to ReadMessage?

Could be a possibility. Personally I dislike having this king a attribute that shadows the behavior of a function. That makes thing unclear and prone to error.

ineiti:
> Personally I dislike having this king a attribute that shadows the behavior of a function.

That's why you can have a clean slate with `Dela`... Here I propose to use the flag, as it's already there. Else it's confusing that there is a flag _and_ two different methods, when one would be enough.

nkcr:
> Here I propose to use the flag, as it's already there.

This flag is used for another purpose, and is private. So you want to have an `c.keep` and `c.IsKeep` ?
I'm still not convinced of your solution. Plus my solution allows to add additional parameters if needed.

ineiti:
I don't understand: you say that `ReadMessageKeep` has the same signature as `ReadMessage`, but from the name of the method I suppose it's used for connections that have the `keep` flag.

Even though I just see now that `ReadMessage` is defined on the structure `StreamingConn`. So now I need to understand how that is even different from `ReadMessageKeep`...

nkcr:
We are doing that in the wrong order: please review first https://github.com/dedis/onet/pull/677 then we can come back to this one.

ineiti (approved):
Looks much nicer to me. I didn't check that the functionality is still the same, so I rely on the tests...

PR #2451, opened 2021-04-27
Update prometheus template based on feedback from VPSI

gnarula:
**What this PR does**
Updates the prometheus template in `status` based on feedback from VPSI.

Changes made:
* Remove newline at the beginning of the template
* Remove newlines between conode metrics
* Add conode description as a tag (In context of evoting, the node's description is set as the school which runs the conode - eg `IC`, `SV`...)

---

🙅‍ Friendly checklist:

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -28,8 +28,7 @@ import (
 	"go.dedis.ch/onet/v3/network"
 )
 
-const prometheusTemplate = `
-# HELP voting_conodes_status voting global conodes status: X is the number of conodes; X >= 0.66 OK, 0 < X < 0.66 KO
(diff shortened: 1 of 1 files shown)


nkcr (approved):

ineiti (commented):
Doesn't that break the current status.dedis.ch? Or does status.dedis.ch ignore `Matrix`?

gnarula:
> Doesn't that break the current status.dedis.ch? Or does status.dedis.ch ignore `Matrix`?

I reckon status.dedis.ch sends a `status.Request` message to the conode and gets a `status.Response` back. The prometheus template isn't a part of the flow - it invokes `status.CheckConnectivity` instead

ineiti:
So for the workflow, it just needs to be put into the `Cothority` project and in the `Ready4Merge` pipeline, then we can put this in main...

PR #2452, opened 2021-04-28
Move master branch to main branch

ineiti:
There are some `Master` references in the code, specifically in the `evoting` section, that I don't know if it's correct to replace them... @gnarula , any idea?

### DIFF
--- a/PUBLISH.md
+++ b/PUBLISH.md
@@ -24,7 +24,7 @@ Tag the release using
 git tag -s vX.Y.Z -m "Cothority Release vX.Y.Z"
 ```
 
-and push it to master with `git push origin vX.Y.Z`. A tag push triggers
+and push it to `main` with `git push origin vX.Y.Z`. A tag push triggers
 the `release` action workflow which releases new binaries in
 `https://github.com/dedis/cothority/releases/tag/vX.Y.Z`. Please make sure to
 add the release description by editing that page.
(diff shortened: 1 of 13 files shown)


gnarula:
I believe references to `Master` are for the skipchain that tracks all the other election skipchains. `s/Master/Main` and `s/master/main` shouldn't be okay. Is it fine if I add a commit to this branch?

ineiti:
> Is it fine if I add a commit to this branch?

Yes, of course. I was not sure whether I'd break anything, so I left it as-is. But if we're replacing the branch-name `master`, it makes also sense to replace the variable names called `master`...

gnarula:
Digging into this, I see we'd require changes in `dedis/epfl-evoting` recreating the proto file in that repo.

I think we're too close to actual elections to be making changes and suggesting VPSI to use it for this cycle. Having said that I do have to make a release with #2451 though.

I don't want to block you on this PR but I'd suggest going ahead with the current changes. I can work on the evoting related changes in a separate PR in the next sprint (so sometime between the 10th-21st) if that's okay?

gnarula (approved):

tharvik (approved):

ineiti:
> I don't want to block you on this PR so I'd suggest going ahead with the current changes.

I thought of simply accepting that you still use `Master` for the evoting, but to move all the rest to `main`. Once you're ready to change the evoting code from `master` to `main`, this can be included later.

Added #2454 

On an unrelated note:  former master of India, `prince Philipp`, just died, might be a good occasion to move to `main` now ;)

PR #2453, opened 2021-04-28
Bypros update

nkcr:
**What this PR does**

This PR updates the part where we listen to blocks, so that when the connection unexpectedly closes, it tries to reconnect.

### DIFF
--- a/bypros/proxy.go
+++ b/bypros/proxy.go
@@ -39,6 +39,7 @@ func (s *Service) Follow(req *Follow) (*EmptyReply, error) {
 	}
 
 	if !s.scID.Equal(req.ScID) {
+		s.follow <- struct{}{}
 		return nil, xerrors.Errorf("wrong skipchain ID: expected '%x', got '%x'", s.scID, req.ScID)
 	}
 
@@ -47,6 +48,8 @@ func (s *Service) Follow(req *Follow) (*EmptyReply, error) {
 
 	conn, err := subscribe(req)
 	if err != nil {
+		s.follow <- struct{}{}
+		s.following = false
 		return nil, xerrors.Errorf("failed to subscribe: %v", err)
 	}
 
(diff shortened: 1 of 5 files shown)


bypros/proxy_test.go
```diff
@@ -246,13 +246,18 @@ func TestProxyCatchUp_Wrong_Skipchain(t *testing.T) {
...
 
-	_, stop, err := s.CatchUP(&CatchUpMsg{
+	out, stop, err := s.CatchUP(&CatchUpMsg{
 		ScID:        bct.Genesis.SkipChainID(),
 		Target:      bct.Roster.Get(0),
 		FromBlock:   bct.Genesis.SkipChainID(),
 		UpdateEvery: 9,
 	})
 
+	go func() {
```
ineiti, line 256: Can't you initialize the channel with a length of 1 so you can get rid of this go-routine?

ineiti (changes requested):

ineiti (approved):

ineiti:
OK for me to merge, except if you want @gnarula 's input first...

bypros/proxy.go
```diff
@@ -40,15 +43,22 @@ func (s *Service) Follow(req *Follow) (*EmptyReply, error) {
 	}
 
 	if !s.scID.Equal(req.ScID) {
-		return nil, xerrors.Errorf("wrong skipchain ID: expected '%x', got '%x'", s.scID, req.ScID)
+		s.follow <- struct{}{}
```
gnarula, line 46: I don't quite get why we're sending a message on this channel in case of an error? Is it to unblock `s.Follow()` in case of a retry in `listenBlocks`?

nkcr: It is to allow for a new call to `s.Follow()` by the client.

gnarula: Ah yes of course!

gnarula (approved):

bypros/proxy.go
nkcr: just to add more precision: this is actually a bug fix, not directly related to the reconnect stuff.

PR #2464, opened 2021-08-24
Better update of chains

ineiti:
The following happened in August 2021:
* Some nodes did not have the whole state
* When resetting the db, the node tried to get this invalid state

Now the update tries to get the state from one node after another until it finds
a correct state

This depends on dedis/onet#678 and will need a bump in the onet dependency once 678 is merged


byzcoinx/byzcoinx.go
```diff
@@ -184,7 +186,8 @@ func (bft *ByzCoinX) Dispatch() error {
 	select {
 	case commitSig = <-commitProto.FinalSignature:
 		log.Lvl2(bft.ServerIdentity(), "Finished commit phase")
-	case <-time.After(bft.Timeout / time.Duration(2) * time.Duration(bft.SubleaderFailures+1)):
+	case <-time.After(bft.Timeout / time.Duration(3) * time.Duration(bft.
```
calctopian, line 189: I’ve seen some exciting changes that you plan to commit:

      case <-time.After(bft.Timeout / time.Duration(3) * time.Duration(bft.SubleaderFailures+1)):

In practice, they will lower the timeout from the current 15/(2*3)=2,5 secs to 15/(3*3)=1,66 secs, [with bft.Timeout defined as](https://github.com/dedis/cothority/blob/main/skipchain/struct.go#L36):
`var defaultPropagateTimeout = 15 * time.Second`
and [bft.SubleaderFailures defined as](https://github.com/dedis/cothority/blob/main/blscosi/protocol/protocol.go#L35):
`	const defaultSubleaderFailures = 2`

Fine-tuning timeouts is more of an art than a science, but still critical to achieve a performant consensus protocol: a ballpark estimate suggests that the timeout can be further lowered.

Assume that the RTT delay on a Internet setting is [around 200 ms with a standard deviation of 20 ms](https://inrg.soe.ucsc.edu/wp-content/uploads/2017/08/q2swinet2017.pdf), then the timeout keeping the consensus protocol stable with a probability distribution function of the RTT delays equal to 90% can be guesstimated to be >175 ms (i.e., an order of magnitude lower than the current timeout).

What's your take on this interesting topic?

ineiti: Thanks for your comment. I just realized that I changed the wrong part of the quotient! 

What happened was that the I was forcing a new version to the nodes, but two nodes didn't update. And these two nodes happened to be the two chosen subleaders. And the system only had enough timeout for one subleader change. This made that the system didn't create new blocks as desired. So I wanted the leader to wait enough for two subleader changes. But that would've been in the numerator and not the denominator.

With regard to the fine-tuning of the timeouts, yes, that's kind of black magic. One thing the blscosi protocol does is to go on as soon as it has enough signatures. So it only waits for the timeout if it doesn't have 2f+1 signatures.

Another thing to take into account is that the timeout doesn't only depend on the RTT. The total time is:

```
RTT + block_transmission_time + block_verification_time
```

While RTT is quite low, the other two can easily add another 500ms each, for a lot of transactions it might even take up to 1 second. So the timeout is probably too low if the chain has heavy load! I played around with this once to find out what the optimal tree-structure is, depending on the block_tx and block_verify time. For some values of these parameters, a shallow tree is better, for others, a deep tree is advantageous.

PR #2470, opened 2021-11-18
Refine byzcoin

ineiti:
This is a collection of patches for byzcoin and skipchain:
- solve a race condition where the bftForwardLinkLevel0Ack has been called before bftForwardLinkLevel0 is done
- allow storing of new SkipBlocks in existing chains, even if a Client has been registered
- changing the byzcoin.Client.GetInstance method to also do the protobuf-decoding
- adding a byzcoin.Client.UpdateNodes in case some of the nodes didn't catch up

This has been tested in production and behaves nicer. Of course there are still some bugs left...

### DIFF
--- a/byzcoin/api.go
+++ b/byzcoin/api.go
@@ -33,6 +33,8 @@ type Client struct {
 	Genesis *skipchain.SkipBlock
 	// Latest keeps track of the most recent known block for the client.
 	Latest *skipchain.SkipBlock
+	// nodes that are most probably useful and alive
+	nodes []*network.ServerIdentity
 	// Keeps the server identities that replied first to a DownloadState request
 	noncesSI map[uint64]*network.ServerIdentity
 	// Used for SendProtobufParallel. If it is nil, default values will be used.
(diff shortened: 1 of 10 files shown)


byzcoin/api.go
```diff
@@ -85,6 +87,70 @@ func (c *Client) getLatestKnownBlock() *skipchain.SkipBlock {
...
+	var nodeList []nodeBlock
+	for i := 0; i < len(c.Roster.List); i++ {
+		select {
+		case n := <-nodes:
+			if c.Latest == nil || n.sb.Index > c.Latest.Index {
+				c.Latest = n.sb
+				c.Roster = *n.sb.Roster
+			}
+			nodeList = append(nodeList, n)
+		case <-time.After(time.Second * 2):
```
tharvik, line 128: The loop will timeout iff there is no new `nodeBlock` for two seconds, so it might takes as long as `2s * len(c.Roster.List)`; wanted?

tharvik (approved):

byzcoin/api.go
ineiti: I'm never sure what the `break` does here - won't it break from the `for`-loop?

ineiti: After discussion, yes, if the answers trickle in, that is OK.

calctopian:
About changing "Roster.List" to "GetNodes()":
- "Roster.List" is used over all the repository, although it doesn't always refer to ByzCoin.Client
- There are instances where it does refer to ByzCoin.Client, for example in "eventlog/api.go":
209: 	if err := c.c.SendProtobuf(c.ByzCoin.Roster.List[0], req, reply); err != nil {
267: 	blocks, err := c.sc.GetUpdateChainLevel(&c.ByzCoin.Roster, id, 0, -1)
**These must also be changed to GetNodes()**
- I haven't reviewed all the codebase, this is just a cursory review: I think there are more places where this must be changed.

ineiti:
@calctopian - thanks for your comment. The change I did here is really just relevant to the `byzcoin.Client`. The other `Roster.List` are very often needed as-is and should _not_ be replaced by the `nodes` here. The one in line 209 refers to the leader, and should stay like this. Perhaps the one in line 267 should be changed. Are you using the `eventlog` service? In my mind it is marked as _deprecated_.

The goal is for a client to know which nodes they can connect to without problems. This is needed because some of the nodes might be faulty. Of course this doesn't help against byzantine nodes. But currently we're more concerned by too slow nodes than anything else :)

The best place for this to be would be probably in the `cothority/skipchain/api.go`, and then re-use it by the `cothority/byzcoin/api.go`.

calctopian:
No, I wasn't using the `eventlog` service, just finding instances where `byzcoin.Client` was being used.

OK, I fully agree with you that the priority is to connect to the fastest nodes :)

PR #2471, opened 2021-11-18
Separate signup node

ineiti:
As the conode.c4dt.org is on the switch-node and doesn't have an SMTP server
that allows sending of emails without authentication, the signup-service is
changed to interact with a byzcoin on a remote node.


personhood/email.go
```diff
@@ -143,3 +145,11 @@ func (ec *emailConfig) tooManyEmails(clockHours int64) bool {
 	ec.Emails++
 	return false
 }
+
+func (ec *emailConfig) getClient() (*byzcoin.Client, error) {
```
tharvik, line 149: it can't fail
```suggestion
func (ec *emailConfig) getClient() byzcoin.Client
```

tharvik (approved):

PR #2478, opened 2021-12-09
Fix multiple recoveries

ineiti:
The evolve-instruction was badly calculated, and thus only one recovery was
possible.

Unfortunately all accounts with a recovery cannot be recovered :(

### DIFF
--- a/personhood/service_test.go
+++ b/personhood/service_test.go
@@ -180,12 +180,24 @@ func TestService_EmailRecover(t *testing.T) {
 	require.Equal(t, uint64(2000), newUserCoin.Value)
 	require.Equal(t, contracts.SpawnerCoin, newUserCoin.Name)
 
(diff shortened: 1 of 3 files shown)


tharvik (approved):

tharvik:
> Unfortunately all accounts with a recovery cannot be recovered :(

Maybe reset it to the correct instructions when someone logins?

ineiti:
> > Unfortunately all accounts with a recovery cannot be recovered :(
> 
> Arf, too bad; maybe reset it to the correct instructions when someone logins?

Yes, I thought to do so. But as it's only 5 people with 3 of them in the GymInf class, plus Laurent from SPRING, I just sent out an email with manual instructions to remove and re-add the email recovery option. Else I'm sure that I'll add a new error while fixing it :)

PR #2480, opened 2021-12-14
Updates dependencies with vulnerabilities

nkcr:
Updated:

- go-ethereum from v1.8.27 to v1.10.13
- containerd from v1.4.4 to v1.5.8

Then `go mod tidy`


ineiti:
The go-ethereum cannot be updated, as far as I remember. This is because in the new go-ethereum, there is no clear separation anymore between the database holding the contracts data and the chain. So we cannot use it.

@cgrigis can you confirm that?

cgrigis:
As I remember looking at it at the time, 1.9.x introduced several changes in the data structures that we plug into with ByzCoin, and the modifications needed to adapt to it were not trivial. I have not looked at the latest evolution in go-ethereum since then, but I guess that will be a significant amount of work.

nkcr:
Ok, I was aiming at "just a quick security update" thinking that backward compatibility was there for those libraries. Let's forget about it then.

PR #2481, opened 2021-12-14
Fix subleader handling

ineiti:
This PR does the following:
- increase the number of subleaders to sqrt(roster.len - 1), according
  to the OmniLedger paper
- put all calls to Send in the blscosi protocol in a go-routine, so that
  a blocking subleader doesn't stop the protocol

### DIFF
--- a/blscosi/protocol/protocol.go
+++ b/blscosi/protocol/protocol.go
@@ -99,13 +99,14 @@ func DefaultThreshold(n int) int {
 }
 
 // DefaultSubLeaders returns the number of sub-leaders, which is the
-// cube-root of the number of nodes.
+// square root of the number of nodes - 1.
 func DefaultSubLeaders(nodes int) int {
-	// As `math.Pow` calculates `8 ** (1/3) < 2`,
+	if nodes == 1 {
+		return 1
+	}
+	// As `math.Pow` calculates `9 ** (1/2) < 3`,
 	// we add 0.0001 for the rounding error.
-	// This works for up to 57 ** 3 = 185'193, which should be enough
-	// nodes.
-	return int(math.Pow(float64(nodes), 1.0/3.0) + 0.0001)
+	return int(math.Pow(float64(nodes-1), 1.0/2.0) + 0.0001)
 }
 
 // NewBlsCosi method is used to define the blscosi protocol.
--- a/blscosi/protocol/protocol.go
+++ b/blscosi/protocol/protocol.go
@@ -104,9 +104,7 @@ func DefaultSubLeaders(nodes int) int {
 	if nodes == 1 {
 		return 1
 	}
-	// As `math.Pow` calculates `9 ** (1/2) < 3`,
-	// we add 0.0001 for the rounding error.
-	return int(math.Pow(float64(nodes-1), 1.0/2.0) + 0.0001)
+	return int(math.Pow(float64(nodes-1), 1.0/2.0))
 }
 
 // NewBlsCosi method is used to define the blscosi protocol.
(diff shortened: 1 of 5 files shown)


blscosi/protocol/protocol.go
```diff
@@ -99,13 +99,14 @@ func DefaultThreshold(n int) int {
...
 
 // DefaultSubLeaders returns the number of sub-leaders, which is the
-// cube-root of the number of nodes.
+// square root of the number of nodes - 1.
 func DefaultSubLeaders(nodes int) int {
-	// As `math.Pow` calculates `8 ** (1/3) < 2`,
+	if nodes == 1 {
+		return 1
+	}
+	// As `math.Pow` calculates `9 ** (1/2) < 3`,
```
tharvik, line 107: `math.Pow(9.0, 1.0/2.0) < 3.0 == false` for me

blscosi/protocol/sub_protocol.go
```diff
@@ -445,3 +447,16 @@ func (p *SubBlsCosi) checkIntegrity() error {
...
+func (p *SubBlsCosi) sendToChildren(msg interface{}) {
+	for _, c := range p.Children() {
+		go func(tn *onet.TreeNode) {
+			err := p.SendTo(tn, msg)
+			if err != nil {
+				log.Warnf("Error while sending to children: %v", err)
+			}
+		}(c)
+	}
+}
```
tharvik, line 462: You know I really don't like dropping errors (I even feel that this `func` comment was just for me :smile: )

What about smth like? That gives you a stream of error, closing it when all childs are done, so you can both know the end of the broadcast and if any error happened :)
```suggestion
func (p *SubBlsCosi) sendToChildren(msg interface{}) <-chan error {
	ret := make(chan error)

	childsRets := make(chan error)
	go func() {
		for range p.Children() {
			if err := <-childsRets; err != nil {
				ret <- xerrors.Errorf("send to children: %v", err)
			}
		}
		close(ret)
	}()

	for _, c := range p.Children() {
		go func(tn *onet.TreeNode) {
			childsRets <- p.SendTo(tn, msg)
		}(c)
	}

	return ret
}
```

blscosi/protocol/sub_protocol.go
```diff
@@ -190,36 +190,43 @@ func (p *SubBlsCosi) dispatchRoot() error {
...
-		if reply.Equal(p.Root().Children[0]) {
-			// Transfer the response to the parent protocol
-			p.subResponse <- reply
+	for {
+		select {
+		case err := <-subLeaderActive:
+			if err != nil {
+				p.subleaderNotResponding <- true
+				log.Warnf("Couldn't contact subleader: %v", err)
+				return nil
```
tharvik, line 211: feels like it should return the `err`

blscosi/protocol/protocol.go
ineiti: Ah, that makes sense in fact. I just copied the lines from the previous `math.Pow(8.0, 1.0/3.0)` which is smaller than `2`. But as you can represent 1/2 in binary, but not 1/3, that's probably the cause. So I remove that.

blscosi/protocol/sub_protocol.go
ineiti: I thought that if `Dispatch` returns true, the protocol aborts. But it seems to continue, so that's better.

tharvik (approved):

ineiti:
Forgot one file - in the propagation part, it also used the `SendToChildrenInParallel`, so I replaced that one, too.

The `SendToChildrenInParallel` is the place where things go down, as it will block if one of the connections cannot be set up. Which could be better handled with the method you proposed.

ineiti:
Here is the issue for the `SendToChildrenInParallel`: dedis/onet#682

PR #2484, opened 2021-12-21
Darc ECDSA 

pm-duokey:
**What this PR does**

*See commit message*
This PR

---

🙅‍ Friendly checklist:

### DIFF
--- a/darc/darc.go
+++ b/darc/darc.go
@@ -33,6 +33,7 @@ package darc
 import (
 	"bytes"
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/sha256"
 	"crypto/sha512"
 	"crypto/x509"
@@ -773,6 +774,8 @@ func (s Signer) Type() int {
 		return 3
 	case s.EvmContract != nil:
 		return 4
+	case s.ECDSA != nil:
+		return 5
 	default:
 		return -1
 	}
@@ -790,6 +793,8 @@ func (s Signer) Identity() Identity {
 		return NewIdentityProxy(s.Proxy)
 	case 4:
 		return NewIdentityEvmContract(s.EvmContract)
+	case 5:
+		return NewIdentityECDSA(s.ECDSA.PublicKey)
 	default:
 		return Identity{}
 	}
@@ -863,6 +868,8 @@ func (id Identity) Type() int {
 		return 3
 	case id.EvmContract != nil:
 		return 4
+	case id.ECDSA != nil:
+		return 5
 	}
 	return -1
 }
@@ -937,6 +944,8 @@ func (id Identity) Verify(msg, sig []byte) error {
 		return id.Proxy.Verify(msg, sig)
 	case 4:
 		return id.EvmContract.Verify(msg, sig)
+	case 5:
+		return id.ECDSA.Verify(msg, sig)
 	default:
 		return errors.New("unknown identity")
 	}
@@ -964,6 +973,10 @@ func (id Identity) GetPublicBytes() []byte {
 		return buf
 	case 4:
 		return id.EvmContract.Address[:]
+	case 5:
+		buf := elliptic.Marshal(id.ECDSA.PublicKey.Curve, id.ECDSA.PublicKey.X, id.ECDSA.PublicKey.Y)
+		//TODO: add error check here?
+		return buf
 	default:
 		return nil
 	}
@@ -1010,6 +1023,25 @@ func NewIdentityX509EC(public []byte) Identity {
 			Public: public,
 		},
 	}
+}
+
+// NewIdentityECDSA creates a new ECDSA identity struct given a public key
+func NewIdentityECDSA(publicKey ecdsa.PublicKey) Identity {
+	return Identity{
+		ECDSA: &IdentityECDSA{
+			PublicKey: publicKey,
+		},
+	}
+}
+
+//TODO make calls to tsm  available
+func (ide IdentityECDSA) Verify(msg []byte, sig []byte) error {
+	hashMsg := sha256.Sum256(msg)
+	valid := ecdsa.VerifyASN1(&ide.PublicKey, hashMsg[:], sig)
+	if !valid {
+		return errors.New("Signature failed to verify")
+	}
+	return nil
 }
 
 // NewIdentityProxy creates a new OpenID Connect identity struct.
@@ -1120,6 +1152,8 @@ func ParseIdentity(in string) (Identity, error) {
 		return parseIDProxy(fields[1])
 	case "evm_contract":
 		return parseIDEvmContract(fields[1])
+	case "secp256k1":
+		return parseIDECDSA(fields[1])
 	default:
 		return Identity{}, fmt.Errorf("unknown identity type %v", fields[0])
 	}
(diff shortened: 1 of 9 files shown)


darc/darc.go
```diff
@@ -964,6 +973,10 @@ func (id Identity) GetPublicBytes() []byte {
 		return buf
 	case 4:
 		return id.EvmContract.Address[:]
+	case 5:
+		buf := elliptic.Marshal(id.ECDSA.PublicKey.Curve, id.ECDSA.PublicKey.X, id.ECDSA.PublicKey.Y)
+		//TODO: add error check here?
```
ineiti, line 978: What kind of error check can you do here? The `GetPublicBytes` doesn't return an error, so all you could do is `panic`.

darc/darc.go
```diff
@@ -1012,6 +1025,25 @@ func NewIdentityX509EC(public []byte) Identity {
...
+// NewIdentityECDSA creates a new ECDSA identity struct given a public key
+func NewIdentityECDSA(publicKey ecdsa.PublicKey) Identity {
+	return Identity{
+		ECDSA: &IdentityECDSA{
+			PublicKey: publicKey,
+		},
+	}
+}
+
+//TODO make calls to tsm  available
```
ineiti, line 1037: Please resolve all TODOs.

From what I understand, the `Verify` method will not call to the TSM, no?

darc/darc.go
```diff
@@ -1142,6 +1176,25 @@ func parseIDX509ec(in string) (Identity, error) {
 	return Identity{X509EC: &IdentityX509EC{Public: id}}, nil
 }
 
+//necessary function, needs to be refactored only supports elliptic.P256 curve
+//needs to be tested Unmarshal might not work
```
ineiti, line 1180:
```suggestion
// Tries to convert a string into a sec256k1 point.
// TODO: needs to be tested Unmarshal might not work
```

darc/darc.go
```diff
@@ -790,6 +793,8 @@ func (s Signer) Identity() Identity {
 		return NewIdentityProxy(s.Proxy)
 	case 4:
 		return NewIdentityEvmContract(s.EvmContract)
+	case 5:
+		return NewIdentityECDSA(s.ECDSA.PublicKey)
```
ineiti, line 797: Here and everywhere else: we should call the `ECDSA` identity `MPCECDSA` to indicate we're doing something special..

darc/darc.go
```diff
@@ -1142,6 +1176,25 @@ func parseIDX509ec(in string) (Identity, error) {
 	return Identity{X509EC: &IdentityX509EC{Public: id}}, nil
 }
 
+//necessary function, needs to be refactored only supports elliptic.P256 curve
+//needs to be tested Unmarshal might not work
+func parseIDECDSA(in string) (Identity, error) {
+	id := make([]byte, hex.DecodedLen(len(in)))
+	_, err := hex.Decode(id, []byte(in))
```
ineiti, line 1183:
```suggestion
    buf, err := hex.DecodeString(in)
    if err != nil {
        return xerrors.Errorf("couldn't parse hex-string: %v", err)
    }
```

darc/darc.go
```diff
@@ -1142,6 +1176,25 @@ func parseIDX509ec(in string) (Identity, error) {
...
+
+	x, y := elliptic.Unmarshal(elliptic.P256(), id)
```
ineiti, line 1185:
```suggestion
	x, y := elliptic.Unmarshal(elliptic.P256(), buf)
```

darc/darc.go
```diff
@@ -1142,6 +1176,25 @@ func parseIDX509ec(in string) (Identity, error) {
...
+
+	pubkey := ecdsa.PublicKey{
+		Curve: elliptic.P256(),
+		X:     x,
+		Y:     y,
+	}
+	if err != nil {
+		return Identity{}, err
+	}
```
ineiti, line 1194:
```suggestion
```

Treat the error as close as possible to the source. Else it is very confusing.

darc/darc.go
```diff
@@ -1447,6 +1500,22 @@ func (kcs SignerX509EC) Sign(msg []byte) ([]byte, error) {
 	return nil, errors.New("not yet implemented")
 }
 
+// new signer creates a signer only with a public key used to verify signatures
```
ineiti, line 1503:
```suggestion
// NewSignerECDSA only takes a public key as the MPC is needed to sign data.
```

darc/darc.go
```diff
@@ -1447,6 +1500,22 @@ func (kcs SignerX509EC) Sign(msg []byte) ([]byte, error) {
...
+func NewSignerECDSA(public ecdsa.PublicKey) Signer {
+	if public.X == nil {
+		return Signer{}
+	}
+	return Signer{ECDSA: &SignerECDSA{
+		PublicKey: public,
+	}}
+}
+
+//TODO
```
ineiti, line 1513: One way to implement sign would be to give the necessary configuration when calling `NewSignerECDSA`, and then having `Sign` call the MPC backend. But that would introduce a dependency on the MPC code that we probably don't want to have in here.

```suggestion
// Sign cannot be implemented, as only the MPC can sign a message.
```

darc/darc_test.go
```diff
@@ -750,17 +754,25 @@ func TestParseIdentity(t *testing.T) {
...
 // Test any identity
-func testIdentity(t *testing.T, sig Signer) {
-	msg := []byte("something secret")
-	signed, err := sig.Sign(msg)
-	require.NoError(t, err)
+func testIdentity(t *testing.T, id Identity) {
+	msg := []byte(`Hello World`)
+
+	//Signature from code example go-tsm-sdk corresponding to ecdsa public key example
+	signed, _ := hex.DecodeString("304402204f0b20a44efacec7b0514683233a79552026fe80e468078f6fed6cfe3f3e8a0402201eb12db7f6fe0828cafe8b0a032a37ff377b342799cfe77cfbac40c8ec1fa9e8")
```
ineiti, line 761:
```suggestion
	signed, err := hex.DecodeString("304402204f0b20a44efacec7b0514683233a79552026fe80e468078f6fed6cfe3f3e8a0402201eb12db7f6fe0828cafe8b0a032a37ff377b342799cfe77cfbac40c8ec1fa9e8")
	require.NoError(t, err)
```

ALWAYS do error checking! You will save yourself a lot of pain!

darc/darc_test.go
```diff
@@ -750,17 +754,25 @@ func TestParseIdentity(t *testing.T) {
...
 	require.NoError(t, id.Verify(msg, signed))
 	require.Error(t, id.Verify([]byte("wrong message"), signed))
 }
 
 // Test the different identities available - currently only Ed25519.
 func TestIdentities(t *testing.T) {
-	testIdentity(t, NewSignerEd25519(nil, nil))
+	//Ecdsa public key example
+	var x, _ = new(big.Int).SetString("25613385885653880697990944418179706546134037329992108968315147853972798913688", 10)
+	var y, _ = new(big.Int).SetString("74946767262888349555270609195205284686604880870734462312238891495596941025713", 10)
```
ineiti, line 771: Error checking needed.

ineiti (changes requested):
Thanks for this PR. I think the most difficult bit is the naming :) You called it `ECDSA`, but I think it would be more appropriate to call it something like `MPCECDSA`, or `TSMECDSA` or such.

ineiti:
You can also use the `Commit Suggestions` in the comments to directly add the suggestions I made.

PR #2485, opened 2022-02-11
Add the TSM identity to Darc

ineiti:
The Trusted Secure Module is used by DuoKey to sign messages using a threshold signature
scheme. This PR adds support for Darcs to verify those messages. It is used for the
Master Thesis protject of Pilar Marxer.


darc/darc.go
```diff
@@ -964,6 +982,9 @@ func (id Identity) GetPublicBytes() []byte {
 		return buf
 	case 4:
 		return id.EvmContract.Address[:]
+	case 6:
+		buf := id.TSM.PublicKey
+		return buf
```
tharvik, line 987:
```suggestion
		return id.TSM.PublicKey
```

tharvik (approved):
some cosmetics changes, otherwise LGTM

darc/darc.go
```diff
@@ -863,26 +875,18 @@ func (id Identity) Type() int {
...
-		return true
-	case id.X509EC != nil:
-		return true
-	case id.Proxy != nil:
-		return true
-	case id.EvmContract != nil:
-		return true
-	}
-	return false
+	return id.Type() != 0
```
cgrigis, line 889: Does this mean identities other than Darc are always Primary?

darc/darc.go
```diff
@@ -937,6 +953,8 @@ func (id Identity) Verify(msg, sig []byte) error {
 		return id.Proxy.Verify(msg, sig)
 	case 4:
 		return id.EvmContract.Verify(msg, sig)
+	case 6:
+		return id.TSM.Verify(msg, sig)
```
cgrigis, line 957: Isn't `case 5:` missing?

darc/darc.go
cgrigis, line 987: Isn't `case 5:` missing?

darc/darc.go
```diff
@@ -1120,6 +1187,8 @@ func ParseIdentity(in string) (Identity, error) {
 		return parseIDProxy(fields[1])
 	case "evm_contract":
 		return parseIDEvmContract(fields[1])
+	case "tsm":
+		return parseIDTSM(fields[1])
 	default:
```
cgrigis, line 1192: Nothing for DID?

cgrigis (approved):
A few comments, I'll let you judge if they are relevant.

darc/darc.go
ineiti: Yes - this is used to know which identities have counters. It should probably be renamed, but here I just thought it's useless to have this long `switch/case`, when a simple test would do...

darc/darc.go
ineiti: The DID identity didn't get implemented completely, so it doesn't have a verification.

darc/darc.go
ineiti: Yes, this is probably missing. I think the DID hasn't been completely implemented...

PR #2488, opened 2022-02-25
Change the encoding of the IdentityTSM

ineiti:
The first PR to implement the IdentityTSM used a (Un)MarshalBinary that was not compatible
with the protobuf-encoding we were doing.

This PR contains also some tests used to discover this discrepancy.

Because there is no TSM-identity stored in binary form, this should not break anything.

### DIFF
--- a/external/js/cothority/tsconfig.json
+++ b/external/js/cothority/tsconfig.json
@@ -14,7 +14,17 @@
     "typeRoots": [
       "../types",
       "./node_modules/@types"
-    ]
+    ],
+    "paths": {
+      "@dedis/kyber": [
+        "node_modules/@dedis/kyber/dist",
+        "node_modules/@dedis/kyber"
+      ],
+      "@dedis/kyber/*": [
+        "node_modules/@dedis/kyber/dist/*",
+        "node_modules/@dedis/kyber/*"
+      ]
+    }
   },
   "include": [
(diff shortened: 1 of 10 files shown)


external/js/cothority/tsconfig.json
```diff
@@ -14,7 +14,17 @@
...
+    "paths": {
+      "@dedis/kyber": [
+        "node_modules/@dedis/kyber/dist",
+        "node_modules/@dedis/kyber"
+      ],
+      "@dedis/kyber/*": [
+        "node_modules/@dedis/kyber/dist/*",
+        "node_modules/@dedis/kyber/*"
+      ]
+    }
```
tharvik, line 27: debug? should you use `npm run link` in kyber to do so?

tharvik (approved):

external/js/cothority/tsconfig.json
ineiti: It's for debugging, but I thought to leave it there. The problem with `npm link` is that it points to the root-directory of the package. And the compiled code is in the `dist` directory. So this searches first in the `dist` directory, and only then in the actual package.

The first is used with `npm link`, the second with a 'normal' `npm ci` install.

It is the most clean way to use `npm link` with typescript, IMO.

PR #2495, opened 2022-06-02
Add evoting protobuf

ineiti:
Moves the evoting structures to `proto.go`, to have the protobuf-definitions in @dedis/cothority. Also adds the structures in the typescript-implementation.

Also adds a `scan` command to `scmgr` so that one can check skipchains.


tharvik (approved):
My last PR review for cothority :')

